PR #3, opened 2018-10-20
Added initial template for openshift deployment

quarckster:
It's supposed to be used with https://github.com/RedHatInsights/e2e-deploy


app/consumers/compliance_reports_consumer.rb
```diff
@@ -3,7 +3,7 @@
 # Receives messages from the Kafka topic, converts them into jobs
 # for processing
 class ComplianceReportsConsumer < Racecar::Consumer
-  subscribes_to 'compliance'
+  subscribes_to 'platform.upload.compliance'
```
dLobatog: I will make this pick up an env variable or something.

PR #9, opened 2018-11-05
Fix Rubocop and Travis config

dLobatog:


dLobatog:
Unfortunately  FFI::NotFoundError: Function 'oscap_source_new_from_file' not found in [libopenscap.so.8] keeps on showing up as the OpenSCAP lib in ubuntu may not have that function

PR #22, opened 2018-11-29
Updated openshift templates

quarckster:


openshift/templates/compliance-backend-bc.yaml
```diff
@@ -0,0 +1,90 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: builder-insights-compliance-backend
+
+parameters:
+- name: NAMESPACE
```
dLobatog: 2 keys in the same level, no indentation, is this correct?

dLobatog: There are several instances of this in this 'parameters'  key

openshift/templates/compliance-backend.yml
```diff
@@ -0,0 +1,149 @@
...
+          - name: POSTGRESQL_SHARED_BUFFERS
+            value: "${POSTGRESQL_SHARED_BUFFERS}"
+          - name: RAILS_ENV
+            value: "${RAILS_ENV}"
+          resources:
+            limits:
+              memory: "${MEMORY_LIMIT}"
+
+parameters:
+- name: NAMESPACE
```
dLobatog: ditto

openshift/templates/compliance-consumer.yaml
```diff
@@ -0,0 +1,138 @@
...
+          - name: RAILS_ENV
+            value: "${RAILS_ENV}"
+          - name: KAFKAMQ
+            value: "${KAFKAMQ}"
+          resources:
+            limits:
+              memory: "${MEMORY_LIMIT}"
+
+parameters:
+- name: NAMESPACE
```
dLobatog: ditto

openshift/templates/compliance-db.yaml
```diff
@@ -0,0 +1,125 @@
...
+          - name: POSTGRESQL_MAX_CONNECTIONS
+            value: "${POSTGRESQL_MAX_CONNECTIONS}"
+          - name: POSTGRESQL_SHARED_BUFFERS
+            value: "${POSTGRESQL_SHARED_BUFFERS}"
+          resources:
+            limits:
+              memory: "${MEMORY_POSTGRESQL_LIMIT}"
+
+parameters:
+- name: NAMESPACE
```
dLobatog: ditto

openshift/templates/compliance-backend-bc.yaml
quarckster: it's required by `ocdeployer` https://github.com/RedHatInsights/e2e-deploy#how-do-template-variables-work.:
> By default, the 'NAMESPACE' variable is passed to all templates, this corresponds to the project name.
It just needs to be defined in parameters.

quarckster: The indentation is correct. The template is deployed successfully.

PR #41, opened 2019-01-15
Only save profiles which have test results

dLobatog:


dLobatog:
We have all related tests in https://github.com/RedHatInsights/compliance-backend/blob/master/test/services/xccdf_report_parser_test.rb - the reason why is that the code was only modularized after. I think it'd make sense to include the report with multiple profiles in the tests indeed.

PR #45, opened 2019-01-22
Add docker-compose for devel w/ postgres

akofink:
~I'd like some help with the following issues:~

- ~Racecar can't communicate with kafka running in insights-upload for some reason. I'm not sure if we need to define a network or something.~
- ~Right now there are no volume mounts for the database, as I was hitting selinux issues when mounting with `-v /tmp/db:/var/lib/postgresql/data:Z`. This means you get a fresh DB every time you bring the services down and back up.~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ node_modules/
 config/settings.local.yml
 config/settings/*.local.yml
 config/environments/*.local.yml
+config/database.yml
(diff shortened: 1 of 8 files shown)


akofink:
TODO before merge: Update the Readme

akofink:
Yes, just for development. I'd be open to converting this to more Red Hat ways of containers (podman, kompose, minishift?), but I'd need guidance there.

akofink:
Updated the readme

PR #46, opened 2019-01-24
Slight update of Jenkins configuration

quarckster:


Jenkinsfile
```diff
@@ -14,52 +14,41 @@ node {
...
-                // If Gemfiles or Jenknis slave's Dockerfile changed we need to rebuild the jenkins slave image
-                stage("Rebuild the image") {
-                    openshiftBuild(
-                        bldCfg: "jenkins-slave-base-centos7-ruby25-openscap",
-                        namespace: "jenkins",
-                    )
-                }
+        if (currentBuild.currentResult == "SUCCESS" &&
+            env.BRANCH_NAME == "master" &&
+            ("Gemfile.lock" in changedFiles || "Gemfile" in changedFiles || "openshift/Jenkins/Dockerfile" in changedFiles)) {
```
akofink: I'm not sure we should be committing the Gemfile.lock, but for now, I guess this is fine.

quarckster: We need to rebuild jenkins slave image to update gem cache

PR #49, opened 2019-01-28
Prometheus instrumentation

dLobatog:
Do NOT merge until we actually have a pod running `prometheus_exporter` to which we can connect.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,3 +1,5 @@
 platform_kafka_topic: 'platform.upload.compliance'
 platform_kafka_validation_topic: 'platform.upload.validation'
 host_inventory_url: http://insights-inventory.platform-ci.svc.cluster.local:8080
+prometheus_exporter_host: compliance-prometheus-exporter.compliance-ci.svc
+prometheus_exporter_port: 9394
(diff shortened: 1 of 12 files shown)


config/settings.yml
```diff
@@ -1,3 +1,5 @@
 platform_kafka_topic: 'platform.upload.compliance'
 platform_kafka_validation_topic: 'platform.upload.validation'
 host_inventory_url: http://insights-inventory.platform-ci.svc.cluster.local:8080
+prometheus_exporter_host: compliance-prometheus-exporter.compliance-ci.svc
```
quarckster: Please do not hardcode the configs, they should be substituted by environment variables when it's being deployed.

Dockerfile.prometheus
```diff
@@ -0,0 +1,5 @@
+from centos/ruby-25-centos7
+workdir /app
+add . .
+add entrypoint.prometheus.sh /app/
+expose 9394
```
akofink: This really isn't necessary. The following works great in the docker-compose (there are other changes not shown - I need to submit a PR to update the docker-compose.yml, but here is the [Dockerfile](https://github.com/akofink/compliance-backend/blob/prometheus-with-docker/Dockerfile) and [docker-compose.yml](https://github.com/akofink/compliance-backend/blob/prometheus-with-docker/docker-compose.yml) for now):

```yml
  prometheus:
    restart: on-failure
    volumes:
      - .:/app:z
    entrypoint: bash -c 'bundle exec prometheus_exporter -c lib/graphql_collector.rb'
```

PR #50, opened 2019-01-28
Remove tempfile after parsing report

dLobatog:
Meant to fix https://projects.engineering.redhat.com/browse/RHICOMPL-39

PR #52, opened 2019-01-29
Fixes RHICOMPL-37: Handle empty/StringIO uploads

akofink:
I'm not sure how this ever worked :)

### DIFF
--- a/app/services/safe_downloader.rb
+++ b/app/services/safe_downloader.rb
@@ -48,6 +48,7 @@ class SafeDownloader
     def patch_if_less_than_10k(downloaded_file, tempfile)
       return downloaded_file unless downloaded_file.is_a?(StringIO)
 
+      tempfile = Tempfile.new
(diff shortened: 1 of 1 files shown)


quarckster:
@akofink could please add tests for that?

app/services/safe_downloader.rb
```diff
@@ -48,6 +48,7 @@ def encode_url(url)
     def patch_if_less_than_10k(downloaded_file, tempfile)
       return downloaded_file unless downloaded_file.is_a?(StringIO)
 
+      tempfile = Tempfile.new
```
quarckster: Don't you receive `tempfile` as an argument in this method, do you?

akofink: Wow, yeah - let me redo this PR.

akofink:
This is fixed by https://github.com/RedHatInsights/compliance-backend/pull/50

PR #66, opened 2019-02-21
Notify exceptions on the Slack channel

dLobatog:


Gemfile
```diff
@@ -50,6 +50,9 @@ gem 'graphiql-rails'
 # Pundit
 gem 'pundit'
 
+gem 'exception_notification'
+gem 'slack-notifier'
```
akofink: We want to run this in production and all other environments?

dLobatog: According to this https://github.com/smartinez87/exception_notification#rails-1 - it should be installed in all environments, I already have a setting in the config to only send notifications on production. In any case it won't send any notifications if you don't set the webhook.

PR #70, opened 2019-02-28
Send report through ActionJob

dLobatog:
This allows us to have Sidekiq workers without sharing any volumes. The
report is send as a string over to Redis, serialized as an ActiveJob
job.

### DIFF
--- a/app/jobs/parse_report_job.rb
+++ b/app/jobs/parse_report_job.rb
@@ -6,7 +6,7 @@ class ParseReportJob < ApplicationJob
     parser = XCCDFReportParser.new(file, account, b64_identity)
     parser.save_rule_results
   ensure
-    File.delete(file)
+    File.delete(parser.report_path) if File.exist? parser.report_path
   end
 
   rescue_from(OpenSCAP::OpenSCAPError) do |e|
--- a/test/controllers/systems_controller_test.rb
+++ b/test/controllers/systems_controller_test.rb
@@ -16,10 +16,11 @@ class SystemsControllerTest < ActionDispatch::IntegrationTest
 
   test 'index accepts search' do
     scope = mock('scope')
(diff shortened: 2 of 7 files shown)


app/jobs/parse_report_job.rb
```diff
@@ -6,7 +6,7 @@ def perform(file, account, b64_identity)
     parser = XCCDFReportParser.new(file, account, b64_identity)
     parser.save_rule_results
   ensure
-    File.delete(file)
+    File.delete(parser.report_path) if File.exist? parser.report_path
```
dLobatog: This shouldn't really get called as `Tempfile` destroys the file as soon as it's not reclaimed by the context, at garbage collection time, which is why I'm ensuring the file exists before removing it

test/controllers/systems_controller_test.rb
```diff
@@ -16,10 +16,11 @@ class SystemsControllerTest < ActionDispatch::IntegrationTest
 
   test 'index accepts search' do
     scope = mock('scope')
-    SystemsController.any_instance.expects(:policy_scope).with(Host).returns(scope)
+    SystemsController.any_instance.expects(:policy_scope).with(Host)
+                     .returns(scope)
     scope.expects(:search_for).with('foo=bar')
 
-    get systems_url, params: {search: 'foo=bar'}
+    get systems_url, params: { search: 'foo=bar' }
```
akofink: I'm really not a fan of this kind of thing in PRs, unless you put it in a separate commit. It's just noise in the review. If you'd like to enforce some of these style things like line length, etc., let's update the rubocop config! I'm fine to keep the changes in this one PR, though.

dLobatog: Tests didn't pass if I didn't add this apparently :disappointed:

PR #79, opened 2019-03-06
Modify public REST API to get profiles for a host

akofink:
@dLobatog WIP for feedback: This endpoint is required for the client. ~Should we be using subscription-manager ID or host fqdn (or allow either)? There's also an `insights_id` supplied by the canonical facts, but I'm not sure if we're storing that anywhere?~ it seems we don't store sub-man ID or insights_id anywhere, so hostname is the only identifier for hosts. Please correct me if that's not the case.

~(Also WIP for tests)~

### DIFF
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -4,15 +4,31 @@
 class ProfilesController < ApplicationController
   def index
     render json: ProfileSerializer.new(
-      policy_scope(Profile.includes(:rules, :hosts))
+      policy_scope(index_relation)
       .paginate(page: params[:page], per_page: params[:per_page])
       .sort_by(&:score)
     )
+  end
+
+  def index_relation
+    relation = Profile.includes(:rules, :hosts)
+    if profile_index_params[:hostname]
+      relation = relation
+                 .where(hosts: { name: profile_index_params[:hostname] })
+                 .includes(:profile_hosts)
+    end
+    relation
   end
 
   def show
     profile = Profile.friendly.find(params[:id])
     authorize profile
     render json: ProfileSerializer.new(profile)
+  end
+
+  private
+
+  def profile_index_params
+    params.permit(:hostname)
   end
 end
--- a/test/controllers/profiles_controller_test.rb
+++ b/test/controllers/profiles_controller_test.rb
@@ -3,4 +3,40 @@
 require 'test_helper'
 
 class ProfilesControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    ProfilesController.any_instance.stubs(:authenticate_user)
+    @relation = mock('relation')
+  end
+
+  test 'index lists all profiles' do
+    @relation.expects(:paginate).returns(@relation)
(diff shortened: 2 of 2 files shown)


akofink:
Another thing that seems obvious but may be incorrect: we don't need to make a request to the systems API to see if the system exists if we just request this profiles API. Systems registered with old insights would not be in our compliance systems response anyway.

app/controllers/profiles_controller.rb
```diff
@@ -4,15 +4,27 @@
...
+      policy_scope(index_relation)
       .paginate(page: params[:page], per_page: params[:per_page])
       .sort_by(&:score)
     )
   end
 
+  def index_relation
+    relation = Profile.includes(:rules, :hosts)
+    profile_index_params[:hosts].each do |host|
+      relation.where(hosts: {name: host[:name]})
```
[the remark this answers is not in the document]
akofink: Nevermind - this doesn't work because we're not supposed to pass a body with a GET request :(

app/controllers/profiles_controller.rb
```diff
@@ -4,15 +4,27 @@
...
-      policy_scope(Profile.includes(:rules, :hosts))
+      policy_scope(index_relation)
       .paginate(page: params[:page], per_page: params[:per_page])
       .sort_by(&:score)
     )
   end
 
+  def index_relation
+    relation = Profile.includes(:rules, :hosts)
+    unless profile_index_params[:hostname].nil?
```
dLobatog: nitpick

```ruby
return (relation = Profile.includes(:rules, :hosts) if profile_index_params[:hostname].blank?
relation.where(hosts: {name: profile_index_params[:hostname]})
```

akofink: I have two issues with this:

1. It's not extensible. In the future if we decide we want the ability to filter on other columns, we'd have to modify this code rather than adding code. I certainly wanted to write it like this, but see the next point:

```rb
relation.where(hosts: {name: profile_index_params[:hostname]}) unless profile_index_params[:hostname].nil?
```

2. It's not as readable (and it goes over the 80 character line limit - I think rubocop is enforcing that default)

app/controllers/profiles_controller.rb
```diff
@@ -4,15 +4,28 @@
...
+      policy_scope(index_relation)
       .paginate(page: params[:page], per_page: params[:per_page])
       .sort_by(&:score)
     )
   end
 
+  def index_relation
+    relation = Profile.includes(:rules, :hosts)
+    relation.where(hosts: { name: profile_index_params[:hostname] }) \
+      unless profile_index_params[:hostname].nil?
```
akofink: @dLobatog how about this? It's kind of a compromise. :man_shrugging:

akofink:
Updated with tests. Rubocop doesn't like multiline trailing conditionals, so I had to use the traditional `if`. My previous points about extensibility and readability still hold.

akofink:
This PR is out of date. We should use scoped search for this.

PR #81, opened 2019-03-15
Openshift imagestream scanning MVP 

dLobatog:
This was just meant to work during the demo - you can see some of the changes here to the models, data structure, etc.. A few other things should be done before we enable this, including limiting who's got access to it (policies), testing, etc..


app/jobs/scan_image_job.rb
```diff
@@ -0,0 +1,47 @@
...
+    )
+  end
+
+  def download_image(imagestream)
+    Docker.authenticate!(
+      'username' => imagestream.openshift_connection.username,
+      'password' => imagestream.openshift_connection.token,
+      'serveraddress' => @registry_url
+    )
+    Docker::Image.create(
```
akofink: So... is this creating a new image in the customer's openshift? Or is this local to our pod?

app/controllers/imagestreams_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+class ImagestreamsController < ApplicationController
+  def create
+    openshift_connection = OpenshiftConnection.new(openshift_connection_params)
+    openshift_connection.save
+    imagestream = Imagestream.new(
+      imagestream_params.merge(openshift_connection: openshift_connection)
+    )
+    if imagestream.save
+      params['policy'].keys.each do |profile|
+        ScanImageJob.perform_later(imagestream, profile, identity_header)
```
akofink: I'm skeptical here because, if the job fails, the entire contents of the arguments passed to the job are logged. There's some effort to be able to filter these args, but so far, it seems like [it's not yet in Rails](https://github.com/rails/rails/pull/34438).

We have a similar problem with the parse report job - the entire text contents of a report are passed in, and when the job fails for whatever reason, all 5M characters of the string are printed to sidekiq logs. I also think this has implications for how ActiveJobs are using memory while waiting to execute (holding onto all that report text until it finishes, rather than (i.e.) just passing the filename). I learned today that openshift pods only scale for CPU, not memory usage, so we should opt for the low-memory, cpu-intensive implementations.

akofink: Here's a failure example: https://paste.fedoraproject.org/paste/jM8A4SuU3BwWlm1M7D9-Ew

Gemfile
```diff
@@ -55,6 +55,12 @@ gem 'slack-notifier'
 # Sidekiq
 gem 'sidekiq'
 
+gem 'tty-command'
```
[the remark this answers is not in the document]
dLobatog: They do, perhaps this is overkill for this PR. I used this library aiming to easily check stderr, and display progress in the future.

app/models/openshift_connection.rb
```diff
@@ -0,0 +1,14 @@
...
+  belongs_to :account
+  has_many :imagestreams
+
+  validates :api_url, presence: true
+  validates :registry_api_url, presence: true
+  validates :username, presence: true
+  validates :token, presence: true
+  validates_associated :account
+
+  attr_encrypted :token, key: Base64.decode64(Settings.openshift_tokens_secret), encode: true, encode_iv: true, encode_salt: true
```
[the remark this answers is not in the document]
dLobatog: Updated

app/jobs/scan_image_job.rb
dLobatog: This is local. Perhaps the library naming is weird, but if you track `journalctl -lf -u docker` while you run this, you will see it just downloads the image.

app/controllers/imagestreams_controller.rb
dLobatog: Just passing the filename is not an option for us as we cannot share storage between pods on Openshift (volumes are only available for 1 pod at time ). We can investigate other options for the ParseReportJob, in this case, we are just sending the serialized `Imagestream` object, a string 'profile', usually very short, and the header. 

The problematic part here is that the entire thing is logged if it fails, right? I don't think having a 5M argument is that much of an issue. We can try to write our own monkeypatch for ActiveJobs middleware to avoid this logging.

akofink: I see two problems:

1. The log being filled with a bunch of information that's not useful to us and hides relevant logs
1. The user's identity header being logged

I experimented a bit with monkeypatching the ActiveJob, but I wasn't able to get it working completely. It seems like it's possible, though.

PR #87, opened 2019-03-20
Import rules and rule results in bulk

dLobatog:
** WIP **

This brings the parsing and importing of rules and rule_results down to only a few queries:

1 SELECT  * FROM hosts WHERE id == report host 
1 SELECT * FROM rules WHERE rules.ref_id IN (list of ref ids from report)  
1 INSERT INTO RULES (all rules)
1 INSERT INTO RULE_RESULTS (all rule results)

In practice this brings down the parsing to around 1s for a normal report, which should hopefully lower our sidekiq load.

### DIFF
--- a/app/services/concerns/xccdf_report/rules.rb
+++ b/app/services/concerns/xccdf_report/rules.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module XCCDFReport
+  # Methods related to parsing rules
+  module Rules
+    extend ActiveSupport::Concern
+
+    included do
+      def rule_ids
+        test_result.rr.keys
+      end
+
+      def rule_objects
+        return @rule_objects if @rule_objects.present?
+
+        @rule_objects ||= @benchmark.items.select do |_, v|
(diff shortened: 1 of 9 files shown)


app/services/concerns/xccdf_report/rules.rb
```diff
@@ -0,0 +1,53 @@
...
+      def rules_already_saved
+        return @rules_already_saved if @rules_already_saved.present?
+
+        @rules_already_saved = Rule.where(ref_id: rule_objects.map(&:id))
+                                   .includes(:profiles)
+      end
+
+      def add_profiles_to_old_rules(rules, new_profiles)
+        rules.each do |rule|
+          new_profiles.each do |profile|
```
akofink: Anywhere we have an `ActiveRecord::Relation`, we should probably use `find_each`

dLobatog:
@akofink Thanks for the review, I'll address the comments as soon as I can - regarding the first upload, there's an API request to inventory (create host, then parse json, then save it in our db), but 10s sounds like too much...

PR #93, opened 2019-03-28
IPP 12 - Support operationId and PATH_PREFIX

dLobatog:


config/routes.rb
```diff
@@ -1,17 +1,17 @@
...
+  scope "#{ENV['PATH_PREFIX']}/#{ENV['APP_NAME']}" do
     resources :profiles, only: [:index, :show]
     resources :rule_results, only: [:index]
     resources :systems, only: [:index]
     resources :rules, only: [:index, :show]
-    mount Rswag::Api::Engine => '/api-docs'
-    mount Rswag::Ui::Engine => '/api-docs'
-    mount ActionCable.server => '/cable'
+    mount Rswag::Api::Engine => '/'
+    mount Rswag::Ui::Engine => '/'
```
akofink: Could you give some context here? Our docs are moving to `/api/compliance/` from `/api/compliance/api-docs/`?

openshift/templates/compliance-backend.yml
```diff
@@ -96,15 +98,17 @@ objects:
             value: "${POSTGRESQL_SHARED_BUFFERS}"
           - name: RAILS_ENV
             value: "${RAILS_ENV}"
+          - name: PATH_PREFIX
+            value: "${PATH_PREFIX}"
           livenessProbe:
             httpGet:
-              path: /r/insights/platform/compliance/api-docs/index.html
+              path: /api/compliance/api-docs/index.html
```
akofink: Couldn't this also be using the env variables?

openshift/templates/compliance-backend.yml
```diff
@@ -96,15 +98,17 @@ objects:
...
               port: 8080
             initialDelaySeconds: 120
             timeoutSeconds: 1
           readinessProbe:
             httpGet:
-              path: /r/insights/platform/compliance/api-docs/index.html
+              path: /api/compliance/api-docs/index.html
```
akofink: Same here

spec/swagger_helper.rb
```diff
@@ -9,12 +9,12 @@
     'v1/swagger.json' => {
```
akofink: Is it still at `swagger.json`, or now should be `openapi.json`?

swagger/v1/openapi.json
```diff
@@ -1,12 +1,12 @@
...
   "info": {
-    "title": "Insights Compliance API V1",
+    "title": "Cloud Services for RHEL Compliance API V1",
     "version": "v1",
-    "description": "This is the API for Insights Compliance. You can find out more about Red Hat Insights at [https://access.redhat.com/products/red-hat-insights/](https://access.redhat.com/products/red-hat-insights/)"
+    "description": "This is the API for Cloud Services for RHEL Compliance. You can find out more about Red Hat Cloud Services for RHEL at [https://cloud.redhat.com/](https://cloud.redhat.com/)"
   },
   "paths": {
-    "/r/insights/platform/compliance/profiles": {
+    "/api/compliance/profiles": {
```
akofink: Any way to use env variables here?

config/routes.rb
dLobatog: Correct, if you visit the root of the API - it displays the Swagger docs.

openshift/templates/compliance-backend.yml
dLobatog: I'm not sure how, these YAML files are just uploaded to Openshift.

spec/swagger_helper.rb
dLobatog: It should be openapi.json as per https://platform-docs.cloud.paas.upshift.redhat.com/rest/requirements.html

swagger/v1/openapi.json
dLobatog: This file is auto-generated from the specs, the ENV variables are taken into account directly when running `rake rswag:specs:swaggerize`

spec/swagger_helper.rb
akofink: So.. this needs updating, was my point :)

PR #94, opened 2019-03-28
IPP 12 - Change per-page/page to limit/offset and add links 

dLobatog:
Depends on https://github.com/RedHatInsights/compliance-backend/pull/93

### DIFF
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -3,10 +3,10 @@
 # API for Profiles
 class ProfilesController < ApplicationController
   def index
+    profiles = scope_search.sort_by(&:score)
     render json: ProfileSerializer.new(
-      policy_scope(Profile.includes(:rules, :hosts))
-      .paginate(page: params[:page], per_page: params[:per_page])
-      .sort_by(&:score)
+      profiles,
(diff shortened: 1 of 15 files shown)


app/controllers/rules_controller.rb
```diff
@@ -3,7 +3,12 @@
 # REST API for Rules
 class RulesController < ApplicationController
   def index
-    render json: RuleSerializer.new(policy_scope(Rule).to_a)
+    rules = policy_scope(Rule)
+            .paginate(page: pagination_offset, per_page: pagination_limit)
+    render json: RuleSerializer.new(
+      rules,
+      metadata(total: rules.count, resource: RuleResult)
```
akofink: `resource: Rule` ?

app/controllers/profiles_controller.rb
```diff
@@ -15,4 +15,10 @@ def show
     authorize profile
     render json: ProfileSerializer.new(profile)
   end
+
+  private
+
+  def resource
```
akofink: I'm confused where we're using this private method.

dLobatog: In `search.rb`, we use it to know which model to scope_search for

PR #95, opened 2019-03-28
IPP 12 - Add Rule and RuleResult to Swagger

dLobatog:
Depends on : 
https://github.com/RedHatInsights/compliance-backend/pull/94/files
https://github.com/RedHatInsights/compliance-backend/pull/93/files

### DIFF
--- a/spec/swagger_helper.rb
+++ b/spec/swagger_helper.rb
@@ -81,6 +81,56 @@ RSpec.configure do |config|
               example: 'xccdf_org.ssgproject.content_profile_standard'
             }
           }
+        },
+        rule_result: {
+          type: 'object',
+          required: %w[result],
+          properties: {
+            result: {
+              type: 'string',
+              example: 'passed'
+            }
+          }
+        },
+        rule: {
+          type: 'object',
+          required: %w[title ref_id],
+          properties: {
+            title: {
+              type: 'string',
+              example: 'Record Access Events to Audit Log directory'
+            },
+            ref_id: {
+              type: 'string',
+              example: 'xccdf_org.ssgproject.content_rule_directory_access_'\
+              'var_log_audit'
+            },
+            severity: {
+              type: 'string',
+              example: 'Low'
+            },
+            description: {
+              type: 'string',
+              example: 'The audit system should collect access events to read '\
+              'audit log directory.\nThe following audit rule will assure '\
+              'that access to audit log directory are\ncollected.\n-a '\
+              'always,exit -F dir=/var/log/audit/ -F perm=r -F auid>=1000'\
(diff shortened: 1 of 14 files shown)


app/controllers/profiles_controller.rb
```diff
@@ -18,4 +15,10 @@ def show
     authorize profile
     render json: ProfileSerializer.new(profile)
   end
+
+  private
+
+  def resource
+    Profile
```
akofink: I'm a bit confused. Are you using this method? And why not put these changes in the "IPP 12 - Change per-page/page to limit/offset and add links" commit where you add the Metadata and Pagination modules?

spec/swagger_helper.rb
```diff
@@ -81,6 +81,56 @@
...
+            ref_id: {
+              type: 'string',
+              example: 'xccdf_org.ssgproject.content_rule_directory_access_'\
+              'var_log_audit'
+            },
+            severity: {
+              type: 'string',
+              example: 'Low'
+            },
+            description: {
```
akofink: Where is this stuff coming from? This file isn't autogenerated, is it?

swagger/v1/openapi.json
```diff
@@ -167,6 +167,257 @@
         }
       }
     },
+    "/api/compliance/rule_results": {
```
akofink: This _is_ autogenerated, right? Do we have instructions in the readme for how to generate the openapi docs?

test/controllers/systems_controller_test.rb
```diff
@@ -5,20 +5,17 @@
 class SystemsControllerTest < ActionDispatch::IntegrationTest
   setup do
     SystemsController.any_instance.expects(:authenticate_user)
+    SystemsController.any_instance.expects(:policy_scope).with(Host)
```
akofink: This isn't going to work here once we add tests for `show`. You can put it in a method, though, if you want it more DRY.

spec/swagger_helper.rb
dLobatog: It's not autogenerated, this is the OpenAPI definition I wrote

swagger/v1/openapi.json
dLobatog: Correct: I use this: https://github.com/domaindrivendev/rswag `rake rswag:specs:swaggerize`

test/controllers/systems_controller_test.rb
dLobatog: Done, I've kept it separate for now

swagger/v1/openapi.json
akofink: Do you think it'd be a good idea to mention that somewhere in the Readme?

dLobatog:
https://github.com/RedHatInsights/compliance-backend/pull/97

PR #98, opened 2019-03-29
Basic API to remove hosts

dLobatog:
Call the API using `DELETE /systems/host-id-in-the-inventory`. It only allows to remove hosts within your account scope.


app/controllers/systems_controller.rb
```diff
@@ -18,6 +18,13 @@ def index
     end
   end
 
+  def destroy
+    host = Host.find(params[:id])
```
akofink: Strong params? We can add in a future PR.

PR #101, opened 2019-04-03
Fixes RHICOMPL-134: Delete the profile if it has no hosts

akofink:
https://projects.engineering.redhat.com/browse/RHICOMPL-134


akofink:
The previous test failure on this PR wasn't related to the change.

akofink:
@dLobatog this one should be good to go

akofink:
Updated with a test

PR #103, opened 2019-04-05
CSV responses identical to UI table data

dLobatog:
This lacks a few things:
 * Tests
 * On the ProfileDetails table, the CSV exported contains attributes for all profiles, instead of a CSV tailored for that profile. e.g: compliance_score takes into account all profiles, same for rules_failed..


dLobatog:
Irrelevant after https://github.com/RedHatInsights/compliance-frontend/commit/f2747e7c073b08bbe6227f15c21ad75336f3511a

PR #110, opened 2019-04-15
Import ProfileRule in bulk

dLobatog:
https://gitlab.com/gitlab-org/sidekiq-reliable-fetch/tree/master

### DIFF
--- a/app/services/concerns/xccdf_report/rules.rb
+++ b/app/services/concerns/xccdf_report/rules.rb
@@ -31,13 +31,13 @@ module XCCDFReport
         preexisting_profiles = ProfileRule.select(:profile_id)
                                           .where(rule_id: rules.pluck(:id))
                                           .pluck(:profile_id).uniq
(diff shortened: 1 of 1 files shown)


app/services/concerns/xccdf_report/rules.rb
```diff
@@ -31,13 +31,13 @@ def add_profiles_to_old_rules(rules, new_profiles)
         preexisting_profiles = ProfileRule.select(:profile_id)
                                           .where(rule_id: rules.pluck(:id))
                                           .pluck(:profile_id).uniq
-        rules.find_each do |rule|
+        ProfileRule.import(rules.find_each do |rule|
           new_profiles.each do |profile|
             unless preexisting_profiles.include?(profile.id)
               ProfileRule.create(rule_id: rule.id, profile_id: profile.id)
```
akofink: Should this be `new` instead of `create`?

PR #116, opened 2019-04-22
Test for match_account in policies

dLobatog:


app/policies/application_policy.rb
```diff
@@ -54,6 +54,8 @@ def initialize(user, scope)
     end
 
     def only_matching_account
+      return scope.where('1=0') if user.account_id.blank?
```
dLobatog: I noticed that when `profile.account_id` is nil, and `user.account_id` is nil too (this should never happen), users without any account can see all profiles.

I'm returning a scope `1=0` (false in SQL) and not `[]` because the code calling this method is sometimes calling methods only callable upon a scope, like `search_for`.

PR #117, opened 2019-04-23
Use time from xccdf report, not created_at for RuleResult

akofink:
- Adds the start/end time to RuleResult
- Parses start/end time from TestResult xccdf report
- Updates graphql queries to use end_time as the date of last scan

Issues:
There is currently no time zone info in the report. It appears to be the local time zone of the machine on which it was run. Perhaps we could get this from metadata? For now though, I think this is a huge improvement. You can upload a report generated yesterday, and that time difference is reflected in the DB and UI.

### DIFF
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -33,7 +33,7 @@ class XCCDFReportParser
     begin
       @test_result = ::OpenSCAP::Xccdf::TestResult.new(source)
     rescue ::OpenSCAP::OpenSCAPError => e
-      Sidekiq.logger.error('Error: ', e)
+      Sidekiq.logger.error("Error: #{e}")
     end
   end
 
@@ -61,11 +61,11 @@ class XCCDFReportParser
   end
 
   def start_time
-    @start_time ||= DateTime.parse(test_result_node['start-time'])
+    @start_time ||= DateTime.parse(test_result_node['start-time']).in_time_zone
   end
 
   def end_time
-    @end_time ||= DateTime.parse(test_result_node['end-time'])
+    @end_time ||= DateTime.parse(test_result_node['end-time']).in_time_zone
   end
 
   def rule_results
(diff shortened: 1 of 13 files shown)


app/services/xccdf_report_parser.rb
```diff
@@ -33,7 +33,7 @@ def test_result
     begin
       @test_result = ::OpenSCAP::Xccdf::TestResult.new(source)
     rescue ::OpenSCAP::OpenSCAPError => e
-      Sidekiq.logger.error('Error: ', e)
+      Sidekiq.logger.error("Error: #{e}")
```
akofink: By pushing for more coverage, I've covered this error with tests and realized we were using this method incorrectly. Doh!

test/test_helper.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 require 'simplecov'
-SimpleCov.start 'rails'
+SimpleCov.start 'rails' do
+  add_filter 'app/models/schema.rb' # don't test the schema! https://github.com/rmosolgo/graphql-ruby/blob/master/guides/schema/testing.md#dont-test-the-schema
+  add_filter 'app/serializers' # these just contain the description of the attributes of the API responses, no code (https://github.com/Netflix/fast_jsonapi)
+  add_filter 'lib/graphql_collector.rb' # empty class that has to be there for prometheus integration
```
akofink: We're using the codecov formatter for simplecov for codecov.io. This is the way to ignore files with plain old simplecov. To see accurate coverage locally (if you don't have the CODECOV_TOKEN), simplecov uses the HTMLFormatter, and so we have to use the simplecov filters to ignore files rather than codecov's config. This shouldn't affect the codecov formatter - it should continue to ignore these files like it used to.

akofink:
Somehow, codecov isn't getting this latest report :man_shrugging: but here it is https://codecov.io/gh/akofink/compliance-backend/commit/3c906a0b970ac11e880fcc5e6cd48dcf7f4c877b We're less than 0.5% away from 80%!

akofink:
client canonical facts does not contain timezone information, but we could conceivably add it in the future

PR #123, opened 2019-04-29
Move Report parsing to XML and Nokogiri exclusively

dLobatog:
This allows the app to work without ruby-openscap, which may have a memory leak according to the Satellite team (who's avoiding it and is not experiencing leaks since then)


akofink:
We need to remove all references to `OpenSCAP` like [here](https://github.com/RedHatInsights/compliance-backend/blob/master/app/jobs/parse_report_job.rb#L11)

PR #124, opened 2019-04-29
Remove double slash from inventory URI

akofink:
Before: `https://ci.cloud.paas.upshift.redhat.com//api/inventory/v1/hosts`

After: `https://ci.cloud.paas.upshift.redhat.com/api/inventory/v1/hosts`

The one with two slashes returns a 404.


dLobatog:
Thanks!

PR #125, opened 2019-05-01
Use delete instead of delete! in case \n isn't found

akofink:
delete! returns nil if the substring isn't found. delete always returns
the string, regardless of if the substring is found and deleted

### DIFF
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -11,7 +11,10 @@ class Rule < ApplicationRecord
   has_many :rule_results, dependent: :delete_all
   has_many :hosts, through: :rule_results, source: :host
 
+  validates :title, presence: true
   validates :ref_id, uniqueness: true, presence: true
+  validates :description, presence: true
+  validates :severity, presence: true
   validates_associated :profile_rules
   validates_associated :rule_results
 
--- a/app/services/concerns/xccdf_report/rules.rb
+++ b/app/services/concerns/xccdf_report/rules.rb
@@ -19,11 +19,11 @@ class RuleOscapObject
   end
 
   def description
-    @description ||= @rule_xml.at_css('description').text.delete!("\n")
+    @description ||= @rule_xml.at_css('description').text.delete("\n")
   end
 
   def rationale
-    @rationale ||= @rule_xml.at_css('rationale').children.text.delete!("\n")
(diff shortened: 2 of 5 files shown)


app/models/rule.rb
```diff
@@ -11,7 +11,10 @@ class Rule < ApplicationRecord
   has_many :rule_results, dependent: :delete_all
   has_many :hosts, through: :rule_results, source: :host
 
+  validates :title, presence: true
   validates :ref_id, uniqueness: true, presence: true
+  validates :description, presence: true
+  validates :severity, presence: true
```
akofink: Let's validate here if we're going to fail at the graphql layer with nil attributes.

app/services/concerns/xccdf_report/rules.rb
```diff
@@ -76,7 +76,7 @@ def new_rules
       def save_rules
         new_profiles = Profile.where(ref_id: profiles.keys)
         add_profiles_to_old_rules(rules_already_saved, new_profiles)
-        Rule.import(
+        Rule.import!(
```
akofink: This bang makes the whole import fail if a validation fails for any record.

app/services/concerns/xccdf_report/rules.rb
```diff
@@ -19,11 +19,11 @@ def title
   end
 
   def description
-    @description ||= @rule_xml.at_css('description').text.delete!("\n")
+    @description ||= @rule_xml.at_css('description').text.delete("\n")
```
beav: Is this code supposed to just remove any `\n` in the text and return the result?

app/services/concerns/xccdf_report/rules.rb
beav: i'm not very familiar w/ this code, where is the error caught?

app/services/concerns/xccdf_report/rules.rb
akofink: Yes, exactly

akofink:
Updated with test fixes

app/services/concerns/xccdf_report/rules.rb
akofink: It's not caught. It's raised, and sidekiq (the job worker) handles it by re-enqueuing the job

PR #128, opened 2019-05-03
Allow pagination in profiles graphql

akofink:


app/models/schema.rb
```diff
@@ -189,6 +189,9 @@
   field :allProfiles do
     type types[ProfileType]
     description 'All profiles visible by the user'
+    argument :search, types.String, 'Search query'
+    argument :per_page, types.Int, 'Page'
+    argument :page, types.Int, 'Per page'
     resolve lambda { |_obj, _args, ctx|
       Pundit.policy_scope(ctx[:current_user], Profile).includes(:hosts)
```
dLobatog: This line would also need to accept the arguments to paginate

akofink:
Updated

akofink:
@dLobatog remind me again - what's this PR for? There must be some frontend work that needs to happen as well.

akofink:
Closing this, as it's stale, and I don't think we need it anymore anyway.

PR #133, opened 2019-05-09
Add health check endpoint to the kafka consumer

akofink:
Towards https://projects.engineering.redhat.com/browse/RHICOMPL-56

This is one possible solution. Openshift allows us to either supply an HTTP endpoint or a command to run in the pod. I played around with the ideas:

- Using `ss` to look for established connections; would need to install ss as a dependency
- Creating a `compliance.ping` kafka topic and using the already-installed ruby-kafka to produce a message and wait for a response; this isn't good IMO because it also tests the readiness/liveness of the kafka service, not just our consumer
- Running a tiny HTTP endpoint on an unpublished, internal port as part of the consumer; requires an additional gem+dependencies (This PR implements this option)

I'm totally open to other ideas. The insights-upload consumer does it this way as well (one k8s service for a consumer _and_ an HTTP endpoint), but they use HTTP for more than just the health check, of course.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,7 @@ gem 'docker-api'
 gem 'faraday'
 gem 'friendly_id', '~> 5.2.4'
 gem 'scoped_search'
+gem 'sinatra'
 gem 'will_paginate'
 gem 'prometheus_exporter'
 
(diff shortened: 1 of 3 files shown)


akofink:
Do we want to just close this? Or merge it?

akofink:
Closing due to inactivity

PR #136, opened 2019-05-10
Check metadata for insights_id for host ID

akofink:


akofink:
@dLobatog I'm stuck here - it seems like if we use the `pgcrypto` uuid as IDs, there's no way to manually set the Host id on create or update:

```rb
[3] pry(main)> h=Host.new(name: 'foo')
=> #<Host:0x0000561392835f08 id: nil, name: "foo", created_at: nil, updated_at: nil, account_id: nil>
[4] pry(main)> h=Host.new(id: 'foo', name: 'foo')
=> #<Host:0x0000561392878740 id: nil, name: "foo", created_at: nil, updated_at: nil, account_id: nil>
[5] pry(main)> h=Host.new(name: 'foo') { |h| h.id = 'foo' }
=> #<Host:0x00005613928c6698 id: nil, name: "foo", created_at: nil, updated_at: nil, account_id: nil>
[6] pry(main)> h.id = 'foo'
=> "foo"
[7] pry(main)> h
=> #<Host:0x00005613928c6698 id: nil, name: "foo", created_at: nil, updated_at: nil, account_id: nil>
```

Any ideas? This is crucial to this PR because we need to explicitly set the Host id to the insights_id in the metadata sent from insights-client.

app/services/xccdf_report_parser.rb
```diff
@@ -31,6 +32,7 @@ def inventory_api
 
   def save_host
     @host = Host.find_or_initialize_by(
+      id: @metadata&.dig('insights_id'),
```
dLobatog: I wonder what should we do in the case the insights_id or the metadata are not available. Perhaps we should just refuse to use the message and fail the validation?

test/services/xccdf_report_parser_test.rb
```diff
@@ -73,8 +83,8 @@ class XCCDFReportParserTest < ActiveSupport::TestCase
     should 'return the host object even if it already existed' do
       HostInventoryAPI.any_instance
                       .stubs(:host_already_in_inventory)
-                      .returns(true)
-      Host.create(name: @report_parser.report_host, account: accounts(:test))
+                      .returns('id' => '1')
+      Host.create(id: '1', name: @report_parser.report_host, account: accounts(:test))
```
dLobatog: These have to be real UUIDs

app/services/xccdf_report_parser.rb
dLobatog: A few lines below, we call `inventory_api.sync`. Is that necessary at all at this point? We are already taking in the insights_id coming from the inventory

app/consumers/compliance_reports_consumer.rb
```diff
@@ -6,40 +6,38 @@ class ComplianceReportsConsumer < ApplicationConsumer
   subscribes_to Settings.platform_kafka_topic
 
   def process(message)
-    @value = JSON.parse(message.value)
+    @msg_value = JSON.parse(message.value)
```
dLobatog: Any reason for this variable rename?

akofink: `value` has got to be one of the worst variable names there is (after `foo`, `bar`, etc.), especially as an instance variable. I think this adds to the readability of the code by adding a bit more context. I'm certainly open to alternate names, but not `value`. :)

app/services/xccdf_report_parser.rb
akofink: I think we can fail the validation, but I'm still a bit worried the metadata isn't making it to the parser in prod or CI. We need to double check before we push this and potentially break all the compliance uploads.

What should we do if `@metadata['fqdn']` doesn't equal `report_xml.search('target').text` ? Should we even check?

akofink: What if the host doesn't exist in the inventory? We need to verify that it is created without the `inventory_api.sync` call.

PR #137, opened 2019-05-12
Add various Sidekiq metrics to Prometheus

dLobatog:


akofink:
This breaks when there is no access to prometheus. I've made a PR to prometheus_exporter to fix the issue https://github.com/discourse/prometheus_exporter/pull/72

PR #139, opened 2019-05-14
Run initializer only when running Sidekiq

dLobatog:


config/initializers/sidekiq.rb
```diff
@@ -1,41 +1,43 @@
-Sidekiq.configure_server do |config|
-  config.redis = {
-    url: "redis://#{Settings.redis_url}",
-    network_timeout: 5 # Default is 1 second, let's be more lenient
-  }
-  Sidekiq::ReliableFetch.setup_reliable_fetch!(config)
-  config.server_middleware do |chain|
-    require 'prometheus_exporter/instrumentation'
-    chain.add PrometheusExporter::Instrumentation::Sidekiq
+if $0.include?('sidekiq')
```
dLobatog: This is the only change I made, which checks for the process that's running. Maybe there are better ways to do this.

PR #140, opened 2019-05-14
Update Dockerfile for more incremental builds

akofink:
add prometheus port and sidekiq env variables in docker-compose

The `MALLOC_ARENA_MAX` env variable helps keep the dev environment resources low.

The exposed prometheus port allows looking at the local exporter at `http://localhost:9394/metrics`.

The Dockerfile changes enable the ability to only install gems (not packages) when building the image with updated Gemfile[.lock].

This PR does not affect production (only docker-compose development).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,10 @@ from ruby:2.5
 
 workdir /app
 
-add . .
-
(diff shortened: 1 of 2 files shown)


akofink:
I'm the only one that is affected by this PR, so merging!

PR #144, opened 2019-05-14
Add rails to the sidekiq initializer

akofink:
This is to (potentially) fix the horizontal pod autoscaler issue that metrics cannot be found.


akofink:
Merging without review to see if this fixes CI

akofink:
@dLobatog here are the messages we were getting in CI before this PR:

image

Do you know where the metrics come from? Prometheus? Just trying to figure out why the sidekiq initializer affects the Rails deployment.

PR #146, opened 2019-05-16
Parse rule identifiers and references from report XML

akofink:
Towards Aha! COMP-E-99

WIP for:

- More tests

Preliminary performance impact (tested locally with docker-compose):

Before:

```
ParseReportJob JID-1e96bb7b4f72c413dfd91b50 INFO: start
ParseReportJob JID-1e96bb7b4f72c413dfd91b50 INFO: done: 8.186 sec
ParseReportJob JID-72b589705281a3ea232e6fdd INFO: start
ParseReportJob JID-72b589705281a3ea232e6fdd INFO: done: 2.302 sec
ParseReportJob JID-6f71131a8caf4540eee9e718 INFO: start
ParseReportJob JID-6f71131a8caf4540eee9e718 INFO: done: 2.493 sec
ParseReportJob JID-fa8f9b509a16410a4185b4c5 INFO: start
ParseReportJob JID-fa8f9b509a16410a4185b4c5 INFO: done: 2.118 sec
ParseReportJob JID-4a0d0780019476f8974c2e81 INFO: start
ParseReportJob JID-4a0d0780019476f8974c2e81 INFO: done: 2.476 sec
```

After:

```
ParseReportJob JID-1a814121e1c07d9981c1b04f INFO: start
ParseReportJob JID-1a814121e1c07d9981c1b04f INFO: done: 24.544 sec
ParseReportJob JID-9a450dcf2da318efc26c891e INFO: start
ParseReportJob JID-9a450dcf2da318efc26c891e INFO: done: 2.281 sec
ParseReportJob JID-187cbfe45e7b203a1056060a INFO: start
ParseReportJob JID-187cbfe45e7b203a1056060a INFO: done: 2.491 sec
ParseReportJob JID-1aabc589c2fe8cb806e25446 INFO: start
ParseReportJob JID-1aabc589c2fe8cb806e25446 INFO: done: 3.987 sec
```

It appears new imports which create Rule objects will take longer (which is expected), but subsequent imports remain the same as before this change.

### DIFF
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -10,6 +10,11 @@ class Rule < ApplicationRecord
   has_many :profiles, through: :profile_rules, source: :profile
   has_many :rule_results, dependent: :delete_all
   has_many :hosts, through: :rule_results, source: :host
+  has_many :rule_references_rules
+  has_many :rule_references, through: :rule_references_rules
+  alias references rule_references
+  has_one :rule_identifier
+  alias identifier rule_identifier
 
   validates :title, presence: true
   validates :ref_id, uniqueness: true, presence: true
@@ -24,7 +29,6 @@ class Rule < ApplicationRecord
     self.rationale = oscap_rule.rationale
     self.description = oscap_rule.description
     self.severity = oscap_rule.severity
-    self.references = oscap_rule.references
     self
   end
 
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -24,6 +24,7 @@ class Rule < ApplicationRecord
     self.rationale = oscap_rule.rationale
     self.description = oscap_rule.description
     self.severity = oscap_rule.severity
+    self.references = oscap_rule.references
     self
   end
 
--- a/app/services/concerns/xccdf_report/rules.rb
+++ b/app/services/concerns/xccdf_report/rules.rb
@@ -28,8 +28,17 @@ class RuleOscapObject
 
   def references
     @references ||= @rule_xml.css('reference').map do |node|
-      {'href' => node['href'], 'text' => node.text }
+      { href: node['href'], label: node.text }
     end
+  end
+
+  def identifier
+    @identifier ||= @rule_xml.at_css('ident')&.text
+  end
+
+  def identifier_system
+    @identifier_system ||= (ident = @rule_xml.at_css('ident')) &&
+                           ident['system']
   end
 end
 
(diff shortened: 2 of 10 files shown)


app/models/rule.rb
```diff
@@ -24,7 +26,7 @@ def from_oscap_object(oscap_rule)
     self.rationale = oscap_rule.rationale
     self.description = oscap_rule.description
     self.severity = oscap_rule.severity
-    self
+    self.rule_references = RuleReference.from_oscap_objects(oscap_rule.references)
```
akofink: This doesn't work with AR import. I still need to figure out how to save the rule references and then associate them with the rule.

app/services/concerns/xccdf_report/rules.rb
```diff
@@ -73,15 +84,40 @@ def new_rules
...
-          end, recursive: true
-        )
+        new_rule_records = new_rules
+                           .each_with_object([]).map do |oscap_rule, _new_rules|
+          rule_object = Rule.new(profiles: new_profiles).from_oscap_object(oscap_rule)
+          if oscap_rule.identifier
+            rule_object.rule_identifier = RuleIdentifier
+                                          .new(label: oscap_rule.identifier, system: oscap_rule.identifier_system)
+          end
+          rule_object
```
dLobatog: How about creating `RuleReferencesRule` in here instead of through ActiveRecord in `associate_rule_references`?

db/migrate/20190517153452_create_rule_references.rb
```diff
@@ -0,0 +1,8 @@
+class CreateRuleReferences < ActiveRecord::Migration[5.2]
+  def change
+    create_table :rule_references, id: :uuid do |t|
```
dLobatog: Shouldn't these fields have a validation for presence at least?

db/migrate/20190524170042_create_rule_identifier_table.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRuleIdentifierTable < ActiveRecord::Migration[5.2]
+  def change
+    create_table :rule_identifiers, id: :uuid do |t|
```
dLobatog: Ditto

test/services/xccdf_report_parser_test.rb
```diff
@@ -107,7 +107,7 @@ class XCCDFReportParserTest < ActiveSupport::TestCase
                      RuleResult.find(rule_results.ids.sample).host.name
         rule_names = RuleResult.where(id: rule_results.ids).map(&:rule)
                                .pluck(:ref_id)
-        assert rule_names.include?(@report_parser.rule_ids.sample)
+        assert rule_names.include?(@report_parser.rule_results.map(&:id).sample)
```
dLobatog: Any reason for removing the `rule_ids` method?

akofink: It wasn't used in any code, just tests. IMO that's not a valid reason to add a method. If the test should be more DRY, a helper method could be added to the test.

app/services/concerns/xccdf_report/rules.rb
akofink: I tried, but active-record-import was complaining (I forget the error message). I can try again, but there is nothing in active-record-import that claims join tables are supported for `has_many through: :foo`.

db/migrate/20190517153452_create_rule_references.rb
akofink: Wait, you mean like this? https://github.com/RedHatInsights/compliance-backend/pull/146/files#diff-829c4379c551ce7d1b0c9a9e1bbead61R9

db/migrate/20190524170042_create_rule_identifier_table.rb
akofink: I updated the RuleIdentifiers model for presence validations.

db/migrate/20190517153452_create_rule_references.rb
akofink: I was still missing a presence validation on the href, so I've added that.

PR #149, opened 2019-05-20
Set new host ID to metadata ID

akofink:
Metadata from insights-upload should contain ID. We should save our hosts with the same ID, if available.


app/services/xccdf_report_parser.rb
```diff
@@ -98,10 +100,6 @@ def rule_results_rule_ids
     ).pluck(:id)
   end
 
-  def host_id
-    @host_id ||= Host.select(:id).find_by(name: report_host)
```
akofink: This just seems a bit dangerous to me and unnecessary. Why query the DB if we already have the host in memory?

akofink:
@dLobatog this one should be good to go.

PR #154, opened 2019-05-29
 RHICOMPL-181 - Delete hosts when notified by the inventory 

dLobatog:
https://github.com/RedHatInsights/s2i-insights-compliance/pull/12

### DIFF
--- a/app/consumers/inventory_events_consumer.rb
+++ b/app/consumers/inventory_events_consumer.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# Receives messages from the Kafka topic, converts them into jobs
+# for processing
+class InventoryEventsConsumer < ApplicationConsumer
+  subscribes_to Settings.platform_kafka_inventory_topic
+
+  def process(message)
+    msg_value = JSON.parse(message.value)
+    logger.info "Received message, enqueueing: #{message.value}"
+
+    return unless msg_value['type'] == 'delete'
+
+    DeleteHost.perform_async(msg_value)
+  end
+
+  private
+
+  def logger
+    Rails.logger
+  end
+end
--- a/app/jobs/delete_host.rb
+++ b/app/jobs/delete_host.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+# Job meant to delete hosts and associated objects asynchronously
+class DeleteHost
+  include Sidekiq::Worker
+
+  def perform(message)
+    Host.transaction do
+      RuleResult.where(host_id: message['id']).delete_all
+      Host.find(message['id']).destroy
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


app/consumers/inventory_events_consumer.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    return unless msg_value['type'] == 'delete'
+
+    DeleteHost.perform_async(msg_value['id'])
+  end
+
+  private
+
+  def logger
+    Rails.logger
```
akofink: Why rails logger and not Sidekiq?

app/jobs/delete_host.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Job meant to delete hosts and associated objects asynchronously
+class DeleteHost
+  include Sidekiq::Worker
+
+  def perform(host_id)
+    RuleResult.where(host_id: host_id).delete_all
+    Host.find(host_id).destroy
+  end
```
akofink: Let's wrap this in a transaction

app/consumers/inventory_events_consumer.rb
```diff
@@ -0,0 +1,22 @@
...
+class InventoryEventsConsumer < ApplicationConsumer
+  subscribes_to Settings.platform_kafka_inventory_topic
+
+  def process(message)
+    msg_value = JSON.parse(message.value)
+    logger.info "Received message, enqueueing: #{message.value}"
+
+    return unless msg_value['type'] == 'delete'
+
+    DeleteHost.perform_async(msg_value['id'])
```
akofink: I'd much prefer passing the entire message for the logs (on potential failure). It'd also keep us more flexible for future changes (like if we one day support deletion of multiple hosts).

app/consumers/inventory_events_consumer.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# Receives messages from the Kafka topic, converts them into jobs
+# for processing
+class InventoryEventsConsumer < ApplicationConsumer
+  subscribes_to Settings.platform_kafka_inventory_topic
```
akofink: Missing the config file update?

app/consumers/inventory_events_consumer.rb
dLobatog: The `consumer` runs using Racecar, not sidekiq

app/jobs/delete_host.rb
dLobatog: :+1:

app/consumers/inventory_events_consumer.rb
dLobatog: :+1:

app/consumers/inventory_events_consumer.rb
dLobatog: Definitely :sob:

PR #155, opened 2019-05-30
RHICOMPL-186 - Stream the report instead of writing to tempfile

dLobatog:
Before, it seems that we were only deleting the file when the validation failed.
It should not be necessary to do this, because we use `Tempfile` in `SafeDownloader`. That should remove the file automatically once it's no longer used in the ruby process memory context. According to our reports from QE it's not, so this is a way of covering our backs and delete it for sure after `process`

### DIFF
--- a/app/consumers/compliance_reports_consumer.rb
+++ b/app/consumers/compliance_reports_consumer.rb
@@ -30,14 +30,13 @@ class ComplianceReportsConsumer < ApplicationConsumer
   end
 
   def download_file
-    @file = SafeDownloader.download(@msg_value['url'], message_id)
-    @file_contents = @file.read
+    @report_contents = SafeDownloader.download(@msg_value['url'])
   end
 
   def enqueue_job
     if validate == 'success'
       job = ParseReportJob.perform_async(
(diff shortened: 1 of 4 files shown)


app/consumers/compliance_reports_consumer.rb
```diff
@@ -14,6 +14,9 @@ def process(message)
     logger.error "Error parsing report: #{@msg_value['request_id']}"\
       " - #{e.message}"
     send_validation('failure')
+  ensure
+    @file.close
```
akofink: You can delete a file that's not closed without error. This line isn't actually required.

app/consumers/compliance_reports_consumer.rb
```diff
@@ -30,14 +30,13 @@ def message_id
   end
 
   def download_file
-    @file = SafeDownloader.download(@msg_value['url'], message_id)
-    @file_contents = @file.read
+    @report_contents = SafeDownloader.download(@msg_value['url'])
```
akofink: Nice variable rename. It's a better name, indeed.

PR #160, opened 2019-06-03
Add inventory-consumer to docker-compose

akofink:


akofink:
I don't actually think this is useful unless you run the inventory locally :confused: I'd love to figure out how to run this with an inventory from CI.

akofink:
I mean, when you are developing locally, you don't need to delete hosts in this way because you have the rails console and the db available directly; however, to test deletion behavior from the inventory, this is required but only useful with a local inventory.

Glutexo:
@dLobatog As far as I know, the Inventory itself is not dockerized. It uses Docker Compose only to spin up a database and a message queue. I think it will be quite easy to dockerize it, but for me it’s not a “one-liner”. If you (and others) really need it, please create a ticket, so we can grab it in a Sprint.

akofink:
I'd like to merge it for completeness' sake, I guess. Maybe we should just comment it out but leave it in the docker-compose in case someone needs it and is running a local inventory. I don't think we really need to push for a containerized inventory at this point in time.

akofink:
Updated to comment this out. IMO it's ready for merge now.

akofink:
Merging!

PR #163, opened 2019-06-05
RHICOMPL-199 - Cache failed rules results

bastilian:


app/models/schema.rb
```diff
@@ -127,12 +127,14 @@
     type types[RuleType]
     description 'Rules failed by a system'
     resolve lambda { |host, _args, _ctx|
-      Rule.where(
-        id: RuleResult.includes(:rule).where(
-          host: host,
-          result: %w[error fail notchecked]
-        ).pluck(:rule_id).uniq
-      )
+      Rails.cache.fetch("#{host.id}/rule_objects_result", expires_in: 1.week) do
```
dLobatog: Looks good - can you invalidate the cache when a new report comes in? We normally do that in https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/xccdf_report_parser.rb#L72 - this looks like just 1 operation, no need to put it on a loop like that one

bastilian: Thanks, that was what I was looking for!

bastilian:
@dLobatog Added the cache deletion and renamed it to reflect what it contains better.

app/services/xccdf_report_parser.rb
```diff
@@ -71,6 +71,7 @@ def save_all
       rules_already_saved.each do |rule|
         Rails.cache.delete("#{rule.id}/#{@host.id}/compliant")
       end
+      Rails.cache.delete("#{host.id}/failed_rule_objects_result")
```
dLobatog: `@host` instead of host

PR #165, opened 2019-06-06
Fix rubocop, add model delete dependencies

akofink:


app/models/rule.rb
```diff
@@ -10,10 +10,10 @@ class Rule < ApplicationRecord
   has_many :profiles, through: :profile_rules, source: :profile
   has_many :rule_results, dependent: :delete_all
   has_many :hosts, through: :rule_results, source: :host
-  has_many :rule_references_rules
+  has_many :rule_references_rules, dependent: :delete_all
```
dLobatog: Not sure if this actually changes anything vs destroy according to our experience, let's keep it though

akofink: At least it doesn't instantiate Ruby objects, even if it still makes N delete queries.

PR #166, opened 2019-06-07
Handle missing host on delete

akofink:
We should handle errors as well as possible.


akofink:
Merging without review to fix production. It's a fairly straightforward change.

PR #167, opened 2019-06-10
Changed openshift instance; updated Jenkinsfile and slave pod

quarckster:


Jenkinsfile
```diff
@@ -14,36 +14,27 @@ node {
...
-        cloud: "cmqe",
-        name: "compliance-backend-test",
-        label: label,
-        inheritFrom: "compliance-backend-test",
-        serviceAccount: pipelineVars.jenkinsSvcAccount
+    openShift.withNode(
+        cloud: "upshift",
+        namespace: "insights-qe-ci",
+        yaml: "openshift/Jenkins/slave_pod_template.yaml",
+        image: "jenkins-slave-base-centos7-ruby25-openscap:latest",
```
akofink: can we use rhel or UBI?

openshift/Jenkins/slave_pod_template.yaml
```diff
@@ -14,7 +14,7 @@ spec:
       value: db_admin_password
     - name: POSTGRESQL_DATABASE
       value: compliance
-    image: centos/postgresql-10-centos7:latest
+    image: registry.access.redhat.com/rhscl/postgresql-10-rhel7
```
akofink: :+1:

Jenkinsfile
quarckster: This image inherits `docker-registry.engineering.redhat.com/centralci/jnlp-slave-base:1.5` which is required to use inside VPN.

PR #168, opened 2019-06-11
RHICOMPL-178 - Optimize Rule results policy fetching

dLobatog:
This PR fixes the policy method to retrieve rule_results on
api/compliance/rule_results. Before this PR, the policy
was iterating on all of the rule_results, which was incredibly
slow.

Aside from that, output was not paginated unless 'search' was
passed. Fixing the pagination forced me to change other controllers to
match it, as a side effect all other endpoints pagination are
now fixed.

### DIFF
--- a/app/controllers/concerns/metadata.rb
+++ b/app/controllers/concerns/metadata.rb
@@ -12,9 +12,11 @@ module Metadata
       response.headers['Content-Type'] = 'application/vnd.api+json'
     end
 
-    def metadata(opts)
+    def metadata(opts = {})
+      opts[:total] ||= policy_scope(resource).count
       options = {}
-      options[:meta] = { total: opts[:total], search: params[:search] }
+      options[:meta] = { total: opts[:total], search: params[:search],
+                         limit: pagination_limit, offset: pagination_offset }
       options[:links] = links(opts[:total])
       options
     end
--- a/app/controllers/concerns/search.rb
+++ b/app/controllers/concerns/search.rb
@@ -6,11 +6,9 @@ module Search
 
   included do
     def scope_search
-      return policy_scope(resource) unless params[:search]
-
-      policy_scope(resource).search_for(params[:search])
-                            .paginate(page: pagination_offset,
-                                      per_page: pagination_limit)
+      result = policy_scope(resource)
+      result = result.search_for(params[:search]) if params[:search].present?
+      result.paginate(page: pagination_offset, per_page: pagination_limit)
     end
   end
 end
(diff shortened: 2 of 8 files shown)


app/controllers/concerns/metadata.rb
```diff
@@ -12,7 +12,8 @@ def set_headers
       response.headers['Content-Type'] = 'application/vnd.api+json'
     end
 
-    def metadata(opts)
+    def metadata(opts = {})
+      opts[:total] ||= policy_scope(resource).count
```
dLobatog: This is just a convenience, before this, I had to call this method on every controller

app/controllers/concerns/search.rb
```diff
@@ -6,11 +6,9 @@ module Search
 
   included do
     def scope_search
-      return policy_scope(resource) unless params[:search]
```
dLobatog: Notice on the old version, `.paginate` was never called unless you passed the `search` param.

app/controllers/profiles_controller.rb
```diff
@@ -3,11 +3,7 @@
 # API for Profiles
 class ProfilesController < ApplicationController
   def index
-    profiles = scope_search.sort_by(&:score)
-    render json: ProfileSerializer.new(
-      profiles,
-      metadata(total: profiles.count)
-    )
+    render json: ProfileSerializer.new(scope_search.sort_by(&:score), metadata)
```
dLobatog: This and other controllers that are not `rule_results` had associated tests that would not work if I did not adapt them to the new changes.

app/policies/rule_result_policy.rb
```diff
@@ -13,7 +13,8 @@ def show?
   # Only show RuleResults belonging to hosts visible by the current user
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      scope.select { |rule_result| Pundit.policy(user, rule_result.host) }
+      available_hosts = HostPolicy::Scope.new(user, Host).resolve.pluck(:id)
```
dLobatog: Instead of iterating over every `RuleResult` object to find which ones had a `host_id` owned by the current user, now I simply get all `hosts` available for the current user and search on `RuleResult` by that.

If you turn on the SQL logging and check the before/after log, the difference is huge

app/controllers/concerns/metadata.rb
akofink: I think we should include `limit` and `offset` in the `meta` option as well so the user sees what the defaults are if nothing is provided, and they see what the server received if they do provide limit and offset.

PR #169, opened 2019-06-11
Rescue SafeDownloader when downloaded file is empty

dLobatog:
If the file download fails because the file is empty, the message will
be pushed back to kafka and we will consume it endlessly, effectively
turning one of the 2 consumers idle. Instead we should just rescue that
case and invalidate that payload.

### DIFF
--- a/app/services/safe_downloader.rb
+++ b/app/services/safe_downloader.rb
@@ -7,18 +7,27 @@ require 'net/http'
 # protection against the infamous open('| ls')
 class SafeDownloader
   DownloadError = Class.new(StandardError)
+  # Exception class to handle empty reports
+  class EmptyReportError < StandardError
+    def initialize(message = 'report is empty')
+      super(message)
+    end
+  end
 
   DOWNLOAD_ERRORS = [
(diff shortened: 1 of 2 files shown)


app/services/safe_downloader.rb
```diff
@@ -13,7 +13,8 @@ class SafeDownloader
     OpenURI::HTTPError,
     RuntimeError,
     URI::InvalidURIError,
-    DownloadError
+    DownloadError,
+    TypeError
```
dLobatog: This is raised when `IO.read` tries to read something it can't:

```

E, [2019-06-11T13:03:43.412502 #1] ERROR -- : [[compliance-reports-consumer] {platform.upload.compliance: 2, 1, 0}:] Exception raised when processing platform.upload.compliance/1 at offset 5451 -- TypeError: no implicit conversion of StringIO into String
--
  | /opt/app-root/src/app/services/safe_downloader.rb:22:in `read'
  | /opt/app-root/src/app/services/safe_downloader.rb:22:in `download'
```

PR #173, opened 2019-06-19
Extend docker compose setup for development

bastilian:
This extends the current docker setup with:

 * Allows configuring of `REDIS_URL` and `PROMETHEUS_EXPORTER_HOST` as environment variables and set this appropriatly in `docker-compose.yml`
 * Add a build script to bootstrap the containers
 * Remove migrating from `entrypoint.sh`
 * Make `entrypoint.sh` execute given commands and move starting command to compose file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,6 @@ RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
 
 COPY Gemfile* ./
 COPY entrypoint.sh ./
-RUN chmod a+x ./entrypoint.sh
 
 RUN bundle -j4
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,5 +7,7 @@ RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
       postgresql-client netcat
 
 COPY Gemfile* ./
+COPY entrypoint.sh ./
+RUN chmod a+x ./entrypoint.sh
 
 RUN bundle -j4
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-from ruby:2.5
+FROM ruby:2.5
 
-workdir /app
+WORKDIR /app
 
-run apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
+RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
       gstreamer1.0-plugins-base gstreamer1.0-tools gstreamer1.0-x libopenscap-dev \
-      postgresql-client
+      postgresql-client netcat
 
-add Gemfile Gemfile.lock ./
+COPY Gemfile* ./
 
-run bundle -j4
+RUN bundle -j4
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,13 @@ WORKDIR /app
 
 RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
       gstreamer1.0-plugins-base gstreamer1.0-tools gstreamer1.0-x libopenscap-dev \
-      postgresql-client netcat
+      postgresql-client
 
 COPY Gemfile* ./
 COPY entrypoint.sh ./
 RUN chmod a+x ./entrypoint.sh
 
 RUN bundle -j4
+
+ENTRYPOINT ["/app/entrypoint.sh"]
+CMD ["bundle", "exec", "rails", "s", "-b", "0.0.0.0"]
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -52,11 +66,33 @@ services:
...
   redis:
     image: redis:latest
+    ports:
+      - '6379:6379'
+  zookeeper:
+    image: confluentinc/cp-zookeeper
+    environment:
+      - ZOOKEEPER_CLIENT_PORT=32181
+      - ZOOKEEPER_SERVER_ID=1
+  kafka:
```
akofink: Why not just use the docker-compose kafka and zookeeper from insights-upload? It's kind of out of scope of our project IMO.

entrypoint.sh
```diff
@@ -4,6 +4,11 @@ set -e
 # Remove a potentially pre-existing server.pid for Rails.
 rm -f /app/tmp/pids/server.pid
 
-bundle exec rake db:migrate
+while ! nc -z redis 6379;
```
akofink: what's with the semicolons?

bastilian:
@akofink The review comments should be addressed.

The `postgres` volume is back, I removed it because in `/tmp` it will get purged at some point, depending on OS and configuration, maybe this could be configurable via an ENV variable or change it to be in the projects `tmp`?

The entrypoint is now changed, as explained in #183 and all containers based upon `compliance-backend-rails` are updated to provide a command. The practice of using the entrypoint this way is not uncommon for images that are serving multiple purposes. The `postgres/postgres` image[1] has an entrypoint, similar to is proposed here, and provides `postgres` as a command, which can easily be overridden by providing other arguments to `docker(-compose) run`.

One minor thing that got in my way a couple of times is the `restart:` directive, but this might (have) be(en) a local condition. The containers kept running throughout restarts and then were not recognized by `docker-compose` anymore/somehow and it tried to start a second set of some of the containers that then failed. It didn't happen when `restart` was commented out. I left it in though and will see if that keeps happening.

[1] https://github.com/docker-library/postgres/blob/master/Dockerfile-alpine.template#L151

config/initializers/sidekiq.rb
```diff
@@ -1,7 +1,7 @@
 if $0.include?('sidekiq')
   Sidekiq.configure_server do |config|
     config.redis = {
-      url: "redis://#{Settings.redis_url}",
+      url: ENV['REDIS_URL'] || "redis://#{Settings.redis_url}",
```
akofink: I don't think this is necessary. The env variable `SETTINGS__REDIS_URL` will override what is in the settings config file. You can try it like this:

```
docker-compose run -e SETTINGS__REDIS_URL=foo rails rails c
[1] pry(main)> Settings.redis_url
=> "foo"
```

config/initializers/prometheus.rb
```diff
@@ -4,7 +4,7 @@
   require 'prometheus_exporter/middleware'
   require 'prometheus_exporter/instrumentation'
   client = PrometheusExporter::Client.new(
-    host: Settings.prometheus_exporter_host,
+    host: ENV['PROMETHEUS_EXPORTER_HOST'] || Settings.prometheus_exporter_host,
```
akofink: Same here. We can use `SETTINGS__PROMETHEUS_EXPORTER_HOST`

docker-compose.yml
```diff
@@ -3,6 +3,8 @@ services:
   db:
     image: postgres
     restart: on-failure
+    ports:
+      - '5432:5432'
```
akofink: Why though? Are you wanting to access the db from outside the docker-compose images?

docker-compose.yml
```diff
@@ -11,7 +13,13 @@ services:
     tty: true
     stdin_open: true
     restart: on-failure
-    entrypoint: ./entrypoint.sh
+    environment:
+      - RAILS_ENV=development
```
akofink: Some of these values are the default already, no?

Dockerfile
```diff
@@ -1,11 +1,16 @@
...
 
-run apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
+RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
       gstreamer1.0-plugins-base gstreamer1.0-tools gstreamer1.0-x libopenscap-dev \
       postgresql-client
 
-add Gemfile Gemfile.lock ./
+COPY Gemfile* ./
+COPY entrypoint.sh ./
+RUN chmod a+x ./entrypoint.sh
```
akofink: We can remove this line, I think. The entrypoint.sh should be copied with its permissions unchanged, and it's `a+x` in the repo.

bastilian:
@akofink Cleaned up the ENV variables (I was not aware of the `SETTINGS__` vars till now), removed the ports to publish from containers that don't need to and removed the chmodding.

PR #171, opened 2019-06-14
RHICOMPL-187 - GraphQL refactor into classes

dLobatog:

PR #176, opened 2019-06-21
 Add rule identifiers/references to graphql schema

akofink:
Built on top of https://github.com/RedHatInsights/compliance-backend/pull/171

Towards https://redhatmbu.aha.io/epics/COMP-E-99

### DIFF
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -11,7 +11,22 @@ module Types
     field :description, String, null: true
     field :ref_id, String, null: false
     field :compliance_threshold, Float, null: false
-    field :rules, [::Types::Rule], null: true
+    field :rules, [::Types::Rule], null: true do
+      argument :identifier, String,
+               'Rule identifier to filter by', required: false
+      argument :references, [String],
+               'Rule references to filter by', required: false
+    end
+
+    def rules(args = {})
+      rules = object.rules
+
+      rules = rules.with_identifier(args[:identifier]) if args.dig(:identifier)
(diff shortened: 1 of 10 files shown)


akofink:
This should be good for review now.

app/graphql/types/profile.rb
```diff
@@ -11,7 +11,27 @@ class Profile < Types::BaseObject
...
+    field :rules, [::Types::Rule], null: true do
+      argument :identifier, String,
+               'Rule identifier to filter by', required: false
+      argument :references, [String],
+               'Rule references to filter by', required: false
+    end
+    def rules(args = {})
+      rules = object.rules
+
+      if args.dig(:identifier)
```
dLobatog: Seems like these 2 could be moved to a helper method? It's nearly identical code

dLobatog: Also it's separate to the rest of the methods down below

app/graphql/types/profile.rb
```diff
@@ -11,7 +11,22 @@ class Profile < Types::BaseObject
...
+               'Rule identifier to filter by', required: false
+      argument :references, [String],
+               'Rule references to filter by', required: false
+    end
+
+    def rules(args = {})
+      rules = object.rules
+
+      rules = rules.with_identifier(args[:identifier]) if args.dig(:identifier)
+      rules = rules.with_references(args[:references]) if args.dig(:references)
```
akofink: How about this, @dLobatog?

PR #179, opened 2019-07-03
RHICOMPL-207 - Display business objectives for policies

dLobatog:
These are the changes required for https://github.com/RedHatInsights/compliance-frontend/pull/154 to work


app/graphql/mutations/business_objective/create.rb
```diff
@@ -0,0 +1,19 @@
...
+module Mutations
+  module BusinessObjective
+    # Mutation to create business objectives
+    class Create < BaseMutation
+      graphql_name 'createBusinessObjective'
+
+      argument :title, String, required: true
+      field :business_objective, Types::BusinessObjective, null: true
+
+      def resolve(args = {})
```
akofink: This could be `def resolve(title:)` to match others like it. Right?

app/graphql/mutations/profile/edit.rb
```diff
@@ -17,7 +17,10 @@ def resolve(args = {})
           ::Profile.find(args[:id]),
           :edit?
         )
-        profile.update(args.to_h)
+        profile.update(
+          compliance_threshold: args[:compliance_threshold],
```
akofink: We could also use named args in this resolve method to clean up the code a bit. (optional)

app/graphql/mutations/business_objective/create.rb
```diff
@@ -0,0 +1,19 @@
...
+module Mutations
+  module BusinessObjective
+    # Mutation to create business objectives
+    class Create < BaseMutation
+      graphql_name 'createBusinessObjective'
+
+      argument :title, String, required: true
+      field :business_objective, Types::BusinessObjective, null: true
+
+      def resolve(title:)
```
dLobatog: This one can remain a named argument, as the field is 100% required

app/graphql/mutations/profile/edit.rb
dLobatog: I realized we cannot use named args here, because they force the arguments to exist. Therefore, a mutation that contains "ID" and "threshold" but not "businesss objective ID" does not work. All three should exist.

PR #183, opened 2019-07-15
Add a util container to docker-compose

akofink:
This allows running the rails console or rake tasks without waiting for
one of the services to come up. This is very important for migrations,
which disallow some services from coming up in the first place.

For example:

`docker-compose exec util bundle update`
`docker-compose exec util bundle exec rake db:migrate`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,6 +57,11 @@ services:
       - MALLOC_ARENA_MAX=2
   redis:
     image: redis:latest
+  util:
+    image: compliance-backend-rails
+    volumes:
+      - .:/app:z
+    entrypoint: bash -c 'sleep infinity'
 networks:
   default:
     external:


akofink:
@bastilian this is very helpful to get around ruby processes requiring all gems to be installed or migrations ran before starting. I am currently maintaining the container in a local git stash and think it would benefit others.

bastilian:
@akofink A utility container is useful, but we should try to avoid adding another container in `docker-compose` to override the entrypoint and instead change `entrypoint.sh`, to allow the same.

As you pointed out the Docker Compose reference says "An ENTRYPOINT allows you to configure a container that will run as an executable.", at the moment it runs as the `rails server` executable, with `bundle exec rails s -b 0.0.0.0`. 

If we'd want to have the rails container to allow running `rails console` and `rails server`, we would change it to `bundle exec rails` and set `['server', '-b', '0.0.0.0']` as a command either in `Dockerfile` or in `docker-compose.yml`, then the rails container can be run with `docker-compose run rails console` if we'd want to go into a console. 

Now, if we'd wanted to run `rake` in the rails container, we could again remove the `rails` argument from `entrypoint.sh` and move it to command, if we then also want to run commands outside the `bundle exec` scope we can replace the last line of `entrypoint.sh` with `exec "$@"` and allow calling any command in the container. 

What is said in the reference still holds true, `entrypoint.sh` prepares the container to run as an executable, in this case `exec` to allow running as any executable. :)

This also means for the other containers based on `compliance-backend-rails` they only need to have a command and not need to override the entrypoint with `bash -c COMMAND`, since the entrypoint will already do that.

I've updated #173 to have the entrypoint changed and allow the rails container to run as any executable.

akofink:
Ok, that all sounds fine to me. Closing in favor of #173

PR #189, opened 2019-07-25
Disable and delete hosts not in the inventory

dLobatog:


app/models/profile.rb
```diff
@@ -71,8 +71,18 @@ def results(host)
...
 
-    (hosts.count { |host| compliant?(host) }).to_f / hosts.count
+    (hosts_in_inventory.count { |host| compliant?(host) }).to_f /
+      hosts_in_inventory.count
+  end
+
+  # Apparently fixtures relations are set using Array instead of
+  # ActiveRecord associations, so scopes do not work on tests with fixtures
+  # This method is meant to fix such cases.
+  def hosts_in_inventory
```
bastilian: It might be better to mock this method or the scope in the unit test and maybe add a test for `in_inventory` to verify it returns the correct result or simply verify the resulting SQL.

However, there must be a way to avoid having to treat fixtures differently.

app/services/host_inventory_api.rb
```diff
@@ -7,16 +7,17 @@
...
-    @b64_identity = b64_identity
+    @b64_identity = Base64.strict_encode64(account.to_identity_header.to_json)
   end
 
-  def host_already_in_inventory
-    response = connection.get(@url, {}, 'X_RH_IDENTITY' => @b64_identity)
-    find_results(JSON.parse(response.body))
+  def hosts_already_in_inventory(hosts)
+    response = connection.get(
+      "#{@url}/#{hosts.map(&:id).join('%2C')}", {}, 'X_RH_IDENTITY' => @b64_identity
```
dLobatog: To test, you may remove the `X_RH_IDENTITY` here , and override the `connection` with something like:

```ruby
  def connection
    Faraday.new do |f|
    faraday = Faraday.new do |f|
      f.response :raise_error
      f.ssl[:verify] = false
    end
    faraday.basic_auth 'user', 'pass'
    faraday
  end
end
```

app/services/hosts_sync.rb
```diff
@@ -0,0 +1,43 @@
...
+        end
+      end
+    end
+
+    def remove_disabled_hosts
+      disabled_hosts = Host.where(disabled: true,
+                                  last_seen_in_inventory: 1.day.ago)
+      logger.info 'Sweep job: hosts about to be removed - '\
+        "#{disabled_hosts.pluck(:id)}"
+      RuleResult.where(host_id: disabled_hosts.pluck(:id)).destroy_all
```
akofink: Hmm... I don't think this is necessary since we have the `dependent: :delete_all` on the host. Right?

app/services/host_inventory_api.rb
```diff
@@ -7,16 +7,17 @@
...
     @account = account
-    @b64_identity = b64_identity
+    @b64_identity = Base64.strict_encode64(account.to_identity_header.to_json)
   end
 
-  def host_already_in_inventory
-    response = connection.get(@url, {}, 'X_RH_IDENTITY' => @b64_identity)
-    find_results(JSON.parse(response.body))
+  def hosts_already_in_inventory(hosts)
+    response = connection.get(
```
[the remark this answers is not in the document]
dLobatog: The reason why I used 100 is that the inventory is meant to support that number of IDs max - pagination in the UI would not work otherwise with 100 elements

app/services/hosts_sync.rb
dLobatog: @akofink It doesn't seem to work, it still `destroys all` - making a ton of queries

app/services/host_inventory_api.rb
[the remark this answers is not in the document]
akofink: I guess what I mean is, this is a public method - perhaps we should ensure `hosts.length < 100`. It's not that important, but I could see someone in the future (mis)using this method by passing in an unbounded hosts list.

app/graphql/types/profile.rb
```diff
@@ -38,30 +38,32 @@ class Profile < Types::BaseObject
...
     end
 
     def total_host_count
-      object.hosts.count
+      object.hosts.in_inventory.count
     end
 
     def compliant(system_id:)
-      object.compliant?(Host.find(system_id))
+      object.compliant?(Host.in_inventory.find(system_id))
```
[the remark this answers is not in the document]
dLobatog: Not a huge fan of `default_scope` for anything that's not sorting or similar (https://www.ombulabs.com/blog/ruby/rails/best-practices/why-using-default-scope-is-a-bad-idea.html) - I prefer to be explicit about this since it's key for our app

quarckster:
@dLobatog @akofink @bastilian I believe this PR is not needed at all. It seems host-inventory relies on `metadata` field and if it's empty the host won't be created. Perhaps compliance should check if a host exists in the inventory and no to parse reports without metadata.

dLobatog:
@quarckster I've created https://github.com/RedHatInsights/compliance-backend/pull/192 for that too, but in any case I think it's a good practice to ensure we don't have any hosts in our app that are not synced.

quarckster:
I wonder how do other apps solve that issue?

akofink:
That is a very fine question, indeed!

Victoremepunto:

akofink:
Closing as stale. Feel free to reopen to continue work.

PR #194, opened 2019-08-05
RHICOMPL-182 - Parse only reports with entitlements

dLobatog:


app/controllers/concerns/authentication.rb
```diff
@@ -11,10 +9,10 @@ module Authentication
   end
 
   def authenticate_user
-    return unauthenticated unless identity_header
-    return unauthenticated unless entitled?
+    @identity_header = IdentityHeader.new(request.headers['X-RH-IDENTITY'])
```
akofink: Why not keep the `identity_header` helper method and use `||=`?

test/models/identity_header_test.rb
```diff
@@ -0,0 +1,27 @@
...
+        }
+      },
+      'identity': {
+        'account_number': '293912'
+      }
+    }
+  end
+
+  test 'is valid if smart_management entitlement provided' do
+    assert IdentityHeader.new(Base64.encode64(@b64_identity.to_json))
```
akofink: Technically, I think we should be using [`strict_encode64`](https://ruby-doc.org/stdlib-2.5.3/libdoc/base64/rdoc/Base64.html#method-i-strict_encode64) without line feeds. Right?

app/consumers/compliance_reports_consumer.rb
```diff
@@ -1,16 +1,20 @@
...
+
 # Receives messages from the Kafka topic, converts them into jobs
 # for processing
 class ComplianceReportsConsumer < ApplicationConsumer
   subscribes_to Settings.platform_kafka_topic
 
   def process(message)
     @msg_value = JSON.parse(message.value)
-    logger.info "Received message, enqueueing: #{message.value}"
+    raise EntitlementError unless identity.valid?
```
akofink: We could also do something like `identity.validate!` which would raise this error, and that would keep the error defined within the `IdentityHeader` class. (optional)

PR #195, opened 2019-08-05
Fixes 264: Only import new rule references

akofink:
Without the relevant code change, the test fails:

```
ERROR["test_only_new_rule_references_are_saved", #<Minitest::Reporters::Suite:0x000055d2bc8ac048 @name="RulesTest">, 1.3665143339894712]
 test_only_new_rule_references_are_saved#RulesTest (1.37s)
ActiveRecord::NotNullViolation:         ActiveRecord::NotNullViolation: PG::NotNullViolation: ERROR:  null value in column "id" violates not-null constraint
        DETAIL:  Failing row contains (null, , bar).
        : INSERT INTO "rule_references" ("id","href","label") VALUES ('18c74014-06a4-496f-a1f6-63092788d2b5','','foo'),(NULL,'','bar') ON CONFLICT DO NOTHING  RETURNING "id"
            app/services/concerns/xccdf_report/rules.rb:48:in `save_rule_references'
            test/services/concerns/xccdf_report/rules_test.rb:60:in `block (2 levels) in <class:RulesTest>'
            test/services/concerns/xccdf_report/rules_test.rb:59:in `block in <class:RulesTest>'
``` 

### DIFF
--- a/test/services/xccdf_report_parser_test.rb
+++ b/test/services/xccdf_report_parser_test.rb
@@ -112,6 +112,7 @@ class XCCDFReportParserTest < ActiveSupport::TestCase
 
   context 'rule results' do
     should 'save them, associate them with a rule and a host' do
+      @report_parser.stubs(:save_rule_references)
       assert_difference('RuleResult.count', 367) do
         rule_results = @report_parser.save_all
         assert_equal @report_parser.report_host,
--- a/app/services/concerns/xccdf_report/rules.rb
+++ b/app/services/concerns/xccdf_report/rules.rb
@@ -35,14 +35,17 @@ module XCCDFReport
         end
       end
 
-      def save_rule_references
+      def rule_references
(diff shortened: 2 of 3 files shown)


akofink:
Something is weird with the tests - they pass locally but they are getting killed for some reason in CI.

test/services/xccdf_report_parser_test.rb
```diff
@@ -112,6 +112,7 @@ class XCCDFReportParserTest < ActiveSupport::TestCase
 
   context 'rule results' do
     should 'save them, associate them with a rule and a host' do
+      @report_parser.stubs(:save_rule_references)
```
akofink: This significantly speeds up this test (Jenkins was sometimes killing it due to long runtime ~150s), and `save_rule_references` is not relevant for this test.

PR #200, opened 2019-08-06
Refactor xccdf, pull out rule_references

akofink:
This is almost all refactoring. The only functionality change should be that saving a profile raises an exception if saving fails for some reason (i.e. model validation failure).

Built on top of https://github.com/RedHatInsights/compliance-backend/pull/195

### DIFF
--- a/app/services/concerns/xccdf_report/profiles.rb
+++ b/app/services/concerns/xccdf_report/profiles.rb
@@ -8,17 +8,17 @@ module XCCDFReport
 
     included do
       def save_profiles
-        created = []
-        # Only save profiles with an associated TestResult. Otherwise there
-        # could be profiles saved w/o results.
-        @oscap_parser.profiles.each do |ref_id, name|
+        @oscap_parser.profiles.map do |ref_id, name|
           profile = Profile.find_or_initialize_by(name: name, ref_id: ref_id,
                                                   account_id: @account.id)
-          profile.description = @oscap_parser.description
-          profile.save
-          created << profile
+          unless profile.persisted?
+            profile.update!(
+              description: @oscap_parser.description
+            )
+          end
+
+          profile
         end
-        created
       end
 
       def host_new_profiles
--- a/app/services/concerns/xccdf_report/rule_references.rb
+++ b/app/services/concerns/xccdf_report/rule_references.rb
@@ -8,8 +8,9 @@ module XCCDFReport
 
     included do
       def rule_references
-        @rule_references ||= new_rules.map do |rule|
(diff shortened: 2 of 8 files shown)


app/services/concerns/xccdf_report/profiles.rb
```diff
@@ -8,17 +8,17 @@ module Profiles
...
-        # could be profiles saved w/o results.
-        @oscap_parser.profiles.each do |ref_id, name|
+        @oscap_parser.profiles.map do |ref_id, name|
           profile = Profile.find_or_initialize_by(name: name, ref_id: ref_id,
                                                   account_id: @account.id)
-          profile.description = @oscap_parser.description
-          profile.save
-          created << profile
+          unless profile.persisted?
+            profile.update!(
```
akofink: This will raise an exception if the profile isn't valid. I'd argue we want to know/fail when this happens rather than just returning a non-persisted profile that failed to save.

app/services/concerns/xccdf_report/profiles.rb
```diff
@@ -8,17 +8,17 @@ module Profiles
...
+              description: @oscap_parser.description
```
akofink: Using update rather than `profile.description = @oscap_parser.description; profile.save` allows us to easily add other fields we'd like to update in the future.

app/services/concerns/xccdf_report/rule_references.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+
+      def save_rule_references
+        RuleReference.import(new_rule_references,
+                             columns: %i[href label],
+                             ignore: true)
+      end
+
+      def associate_rule_references(new_rules)
+        new_rules.zip(@rule_references || []).each do |rule, references|
```
akofink: Setting `@rule_references` here (in the old code with `||=`) does not seem right to me. Using this `@rule_references || []` accomplishes the same thing without polluting the instance variable.

test/services/concerns/xccdf_report/profiles_test.rb
```diff
@@ -15,20 +15,21 @@ def report_description
...
-    @account = OpenStruct.new(id: 1)
-    @host = OpenStruct.new(id: 2)
+    @account = accounts(:test)
+    @host = hosts(:one)
     @oscap_parser = OpenscapParser::Base.new(
       file_fixture('xccdf_report.xml').read
     )
   end
 
   test 'save_profiles' do
```
akofink: Just a refactoring here :)

PR #201, opened 2019-08-07
Only add rules to profiles within same account

dLobatog:


app/services/concerns/xccdf_report/rules.rb
```diff
@@ -51,7 +51,6 @@ def save_rule_references
       end
 
       def save_rules
-        new_profiles = Profile.where(ref_id: @oscap_parser.profiles.keys)
```
dLobatog: Looks like this variable was irrelevant as it's already extracted to a method

PR #204, opened 2019-08-13
Reduce SQL statements in associate_rule_references

akofink:
This changes from <# of new rules> updates for rule references to 1 update. It shaves about 13 seconds off the initial import, from ~85 seconds to ~72 seconds.

### DIFF
--- a/app/services/concerns/xccdf_report/rules.rb
+++ b/app/services/concerns/xccdf_report/rules.rb
@@ -39,9 +39,8 @@ module XCCDFReport
 
       def save_rules
         add_profiles_to_old_rules(rules_already_saved, new_profiles)
-        rule_import = Rule.import!(new_rule_records, recursive: true)
         associate_rule_references
-        rule_import
+        Rule.import!(new_rule_records, recursive: true)
       end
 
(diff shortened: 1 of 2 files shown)


app/services/concerns/xccdf_report/rules.rb
```diff
@@ -39,9 +39,8 @@ def new_rules
 
       def save_rules
         add_profiles_to_old_rules(rules_already_saved, new_profiles)
-        rule_import = Rule.import!(new_rule_records, recursive: true)
         associate_rule_references
-        rule_import
+        Rule.import!(new_rule_records, recursive: true)
```
akofink: Notice how `associate_rule_references` is moved _before_ we actually save the rules using ActiveRecord import

akofink: Here's what we're dealing with now in terms of runtime cost:

```
                                     user     system      total        real
associate_rule_references       26.206187   0.122374  26.328561 ( 28.748383)
Rule.import!                    35.324798   0.643221  35.968019 ( 37.612515)
```

PR #210, opened 2019-08-26
Fixes 323: Include associations in graphql query

akofink:
The following is the system details graphql query with the standard RHEL7 profile for this system:

Before: 

```
Completed 200 OK in 24344ms (Views: 302.9ms | ActiveRecord: 2026.4ms)


user: root
POST /api/compliance/graphql
USE eager loading detected
  Rule => [:rule_identifier]
  Add to your finder: :includes => [:rule_identifier]
Call stack
  /app/app/controllers/graphql_controller.rb:8:in `query'


user: root
POST /api/compliance/graphql
USE eager loading detected
  Rule => [:rule_references_rules]
  Add to your finder: :includes => [:rule_references_rules]
Call stack
  /app/app/controllers/graphql_controller.rb:8:in `query'


user: root
POST /api/compliance/graphql
USE eager loading detected
  Rule => [:rule_references]
  Add to your finder: :includes => [:rule_references]
Call stack
  /app/app/controllers/graphql_controller.rb:8:in `query'
```

After:

```
Completed 200 OK in 17949ms (Views: 340.4ms | ActiveRecord: 888.1ms)
```

It uses the [lookahead feature of graphql-ruby](https://graphql-ruby.org/queries/lookahead.html).

### DIFF
--- a/app/graphql/types/base_object.rb
+++ b/app/graphql/types/base_object.rb
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 module Types
+  # Definition of the BaseObject type in GraphQL
   class BaseObject < GraphQL::Schema::Object
+    protected
+
+    def lookahead_includes(lookahead, object, selects_includes)
+      selects_includes.each do |selects, includes|
+        object = object.includes(includes) if lookahead.selects?(selects)
+      end
+
+      object
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


bastilian:
@akofink LGTM! It might make sense to add a rails benchmark test, at least for the graphql controller. That way we can keep an eye on that and maybe there is a way to define thresholds to fail a tests if it is exceeded.

PR #214, opened 2019-08-28
Do not instantiate unnecessary objects in graphql

akofink:
From the system details graphql request:

Before:

```
Completed 200 OK in 20178ms (Views: 281.5ms | ActiveRecord: 909.9ms)
```

After:

```
Completed 200 OK in 12679ms (Views: 320.2ms | ActiveRecord: 2061.4ms)
```

At this point, I think the bulk of the time is spent creating/serializing objects rather than SQL execution. This PR trades higher SQL time for creating less objects (essentially via `pluck`). It is about a 7 second improvement in response time, but we're still at around 13 seconds total for this response. :man_shrugging: 

### DIFF
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -18,18 +18,21 @@ module Types
                'Rule references to filter by', required: false
     end
 
+    # rubocop:disable AbcSize
     def rules(args = {})
-      rules = object.rules
+      selected_columns = args[:lookahead].selections.map(&:name) &
+                         ::Rule.column_names.map(&:to_sym)
+      rules = object.rules.select(selected_columns)
 
       rules = rules.with_identifier(args[:identifier]) if args.dig(:identifier)
       rules = rules.with_references(args[:references]) if args.dig(:references)
 
       rules = lookahead_includes(args[:lookahead], rules,
-                                 identifier: :rule_identifier,
-                                 references: :rule_references)
+                                 identifier: :rule_identifier)
 
       rules
     end
+    # rubocop:enable AbcSize
 
     field :hosts, [::Types::System], null: true
     field :business_objective, ::Types::BusinessObjective, null: true
@@ -77,7 +80,7 @@ module Types
     end
 
     def last_scanned(system_id:)
-      rule_ids = object.rules.map(&:id)
+      rule_ids = object.rules.pluck(:id)
       rule_results = RuleResult.where(
         rule_id: rule_ids,
         host_id: Host.find(system_id).id
--- a/app/graphql/types/rule.rb
+++ b/app/graphql/types/rule.rb
@@ -14,13 +14,22 @@ module Types
     field :severity, String, null: false
     field :profiles, [::Types::Profile], null: true
     field :identifier, ::Types::RuleIdentifier, null: true
-    field :references, [::Types::RuleReference], null: true
+    field :references, [::Types::RuleReference], null: true,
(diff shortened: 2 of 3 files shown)


app/graphql/types/system.rb
```diff
@@ -67,8 +67,7 @@ def rule_objects_failed
     def last_scanned(args = {})
       if args[:profile_id].present?
         rule_ids = ::Profile.find(args[:profile_id]).rules.pluck(:id)
-        rule_results = ::RuleResult.where(rule_id: rule_ids,
-                                          host: object.id)
+        rule_results = object.rule_results.where(rule_id: rule_ids)
```
dLobatog: I thought this would convert the query into a left outer join, but AR is doing it's magic well here and translating the query to the original one here. Why changing it if it's the same? Is it not clear enough?

```ruby
irb(main):005:0> Host.last.rule_results.where(rule_id: ["4d00711c-e8c7-430a-9940-67aa79c558e6"]).to_sql
=> "SELECT \"rule_results\".* FROM \"rule_results\" WHERE \"rule_results\".\"host_id\" = 'fe60873d-6585-4226-b7d3-b3d43b020f8e' AND \"rule_results\".\"rule_id\" = '4d00711c-e8c7-430a-9940-67aa79c558e6'" 

irb(main):015:0> rule_results = ::RuleResult.where(rule_id: '4d00711c-e8c7-430a-9940-67aa79c558e6', host_id: Host.last.id).to_sql
=> "SELECT \"rule_results\".* FROM \"rule_results\" WHERE \"rule_results\".\"rule_id\" = '4d00711c-e8c7-430a-9940-67aa79c558e6' AND \"rule_results\".\"host_id\" = 'fe60873d-6585-4226-b7d3-b3d43b020f8e'"


```

app/graphql/types/profile.rb
```diff
@@ -18,19 +18,23 @@ class Profile < Types::BaseObject
                'Rule references to filter by', required: false
     end
 
+    # rubocop:disable AbcSize
     def rules(args = {})
       rules = object.rules
 
+      selected_columns = args[:lookahead].selections.map(&:name) &
+                         ::Rule.column_names.map(&:to_sym)
+      object.rules.select(selected_columns)
```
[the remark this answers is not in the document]
akofink: Nice catch! :laughing: I'll fix this - it should be `rules = ...`

app/graphql/types/system.rb
akofink: Yeah, just a refactor to make it a bit cleaner. I can revert it if you have strong feelings.

dLobatog: No problem, no need to change it if it's easier to read. I probably created this in the first place because I was not aware of the AR magic optimization.

app/graphql/types/rule.rb
```diff
@@ -14,13 +14,22 @@ class Rule < Types::BaseObject
...
 
     def compliant(system_id:)
       object.compliant?(Host.find(system_id))
     end
+
+    def references(lookahead:)
+      selected_columns = lookahead.selections.map(&:name) &
+                         ::RuleReference.column_names.map(&:to_sym)
+      object.references.pluck(selected_columns).map do |reference|
+        selected_columns.zip([reference].flatten).to_h
```
akofink: this `[].flatten` is because `pluck` can return an array of arrays or strings if only one column is selected.

akofink:
Updated

PR #216, opened 2019-08-29
Fixes #236: Remove slug/friendly_id from profiles

akofink:
Profile ref_id is not globally unique. Profiles are unique by ref_id,
account_id, and name, but that doesn't work well for a friendly ID.

https://projects.engineering.redhat.com/browse/RHICOMPL-236


bastilian:
Thanks @akofink!

PR #217, opened 2019-09-03
Enforce a 10 second timeout in graphql

akofink:
10 seconds is terribly slow - anything more than this is in danger of
timing out at the gateway (30 seconds total response time). This graphql
timeout is not the total response time, but just for graphql query
execution.

https://graphql-ruby.org/queries/timeout.html

This makes a nasty UX - I was trying to figure out how to return a 5XX (like 503 Service Unavailable, perhaps), but I haven't yet figured it out. If we return a 5XX then we can handle that easily in the UI. I don't think we should merge this without graceful handling in the UI.


bastilian:
Can (or do) we maybe keep track somewhere of when a query times out?

akofink:
@bastilian This will allow us to see it in the compliance-backend logs. Today, timeouts only occur at the 3scale gateway, for which compliance team does not receive logs.

(I need to fix the rubocop issue)

akofink:
Updated to fix rubocop. Again, let's not merge this until we handle this in the UI.

akofink:
The gateway is 30 seconds for the entire response. This timeout is only for the Graphql part of the timing. I think 25s might still hit the gateway timeout, but 20s may not. 10s is already quite a long response time for graphql IMO and a decent maximum.

PR #219, opened 2019-09-04
Allow an environment variable for the test DB

akofink:
- Update docker-compose with the environment variable

I've been getting these kinds of messages with the commented version of the database.yml in master: 

```
ActiveRecord::EnvironmentMismatchError: You are attempting to modify a database that was last run in `test` environment.                                                                                           
You are running in `development` environment. If you are sure you want to continue, first set the environment using:                                                                                               
                                                                                                                                                                                                                   
        bin/rails db:environment:set RAILS_ENV=development
```

This allows passing another environment variable to specify a separate test database, but it is not required.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -14,19 +14,10 @@ default: &default
 
 test:
   <<: *default
-  # use this config with docker-compose
-  #host: db
-  #username: postgres
-  #password:
-  #database: compliance_test
+  database: <%= ENV['POSTGRESQL_TEST_DATABASE'] || ENV['POSTGRESQL_DATABASE'] %>
 
 production:
   <<: *default
 
 development:
   <<: *default
-  # use this config with docker-compose
-  #host: db
-  #username: postgres
-  #password:
-  #database: compliance_dev
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       - DATABASE_SERVICE_NAME=postgres
       - POSTGRES_SERVICE_HOST=db
       - POSTGRESQL_DATABASE=compliance_dev
+      - POSTGRESQL_TEST_DATABASE=compliance_test
       - POSTGRESQL_USER=postgres
       - SETTINGS__PROMETHEUS_EXPORTER_HOST=prometheus
     ports:
@@ -36,6 +37,7 @@ services:
       - DATABASE_SERVICE_NAME=postgres
       - POSTGRES_SERVICE_HOST=db
       - POSTGRESQL_DATABASE=compliance_dev
+      - POSTGRESQL_TEST_DATABASE=compliance_test
       - POSTGRESQL_USER=postgres
       - KAFKAMQ=kafka:29092
     depends_on:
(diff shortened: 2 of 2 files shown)


bastilian:
Do the report-consumer and sidkiq really need the var set?

On a side-not, we could also add a `.env.example` which has common default vars, to copy it in `bin/build`.

akofink:
I often run tests on whatever random container is up, as they all have the rails environment. i.e.

```
docker-compose exec [sidekiq|rails|reports-consumer] bundle exec rake test:validate
```

akofink:
Since the docker-compose environment is solely for development, my goal is to allow it to 'just work' without copying over any config. So far we've been using environment variables to do this.

bastilian:
Oh, we do already have a `.env` checked in. I thought it was not in the repo, since it is also `.gitignored`...

What I meant was, we should move more common values (including the test db variable) there. docker-compose will pick them up and we don't have to define them for each container in `docker-compose.yml`. 

However, there is already the default database variable that can be overridden with `-e` when using `docker-compose`: 

```
bastilian@triton: compliance-backend - [master●] > docker-compose run -e POSTGRESQL_DATABASE=compliance_test rails bash
Starting compliance-backend_prometheus_1 ... done
Starting compliance-backend_db_1         ... done
root@0c3117eb453d:/app# echo $POSTGRESQL_DATABASE
compliance_test
root@0c3117eb453d:/app#
```

We could add a script to `bin/` to boot a shell or test container with `-e POSTGRESQL_DATABASE` and also adapt `bin/build` to also create and migrate a database for a test environment.

akofink:
@bastilian this PR is just meant to fix the issue mentioned in the description. If you've ever run the tests since the last docker-compose update, https://github.com/RedHatInsights/compliance-backend/commit/7b2e6251ad1cb518de5196a530b035cc89282456, using the env variables introduced, you will see this issue. That change prevents running tests without first updating the config/database.yml. I don't see much benefit in providing extra scripts, as we already have a rake task for tests, and devs should know how to run rake tasks, rails console, etc.

bastilian:
The way I run tests is via the following:

```
$ docker-compose run -e POSTGRESQL_DATABASE=compliance_test rails rake db:create db:migrate
....
$ docker-compose run -e POSTGRESQL_DATABASE=compliance_test rails rake test
```
This still works as expected.

I only want to avoid introducing an environment variable that might not be needed to accomplish that `rake test:validate` works.

Changing 
```
  database: <%= ENV['POSTGRESQL_TEST_DATABASE'] || ENV['POSTGRESQL_DATABASE'] %>
```
to 

```
  database: <%= "ENV['POSTGRESQL_DATABASE']_test" %>
```

would make it work without having to have an explicit `POSTGRESQL_TEST_DATABASE` variable.

It would name the test database `compliance_dev_test`, but that is not untrue. :)

akofink:
Right, I initially implemented like that, but I fear that will change CI/QA testing. Having a separate test database variable allows the flexibility in development without affecting other environments. More environment variables is not a bad thing. By allowing Rails to decide which database to use (test vs. dev) removes the need for the `-e ...` option and instead uses the `Rails.env`, which is `test` automatically when running `rake test`.

bastilian:
> but I fear that will change CI/QA testing. 

It should work as expected. Jenkins at least uses rake tasks as well to prepare the database. 
The QA environment will likely use the `production` environment and shouldn't be affected by that.

PR #225, opened 2019-09-10
RHICOMPL-216 - Hide non pass/fail results

dLobatog:
This is probably not good enough, but it's worth trying it to see how our app would look and feel like with it.

You will likely need to run `Rails.cache.clear` and restart the server & proxy to get the new requests. In general I did not see any inconsistencies on the UI after doing that. I would however go further and extend this by:

 - Adding something to the XCCDF parser to make it stop parsing these rule results entirely
 - Adding a migration that removes the rule results that are `notselected`.


docker-compose.yml
```diff
@@ -32,7 +32,7 @@ services:
     volumes:
       - .:/app:z
     environment:
-      - SETTINGS__REDIS_URL=redis://redis:6379
+      - SETTINGS__REDIS_URL=redis:6379
```
dLobatog: The prefix is automatically added in https://github.com/RedHatInsights/compliance-backend/blob/master/config/initializers/sidekiq.rb#L4 , so it'll fail to load with it

app/models/rule.rb
```diff
@@ -31,6 +31,8 @@ class Rule < ApplicationRecord
     joins(:rule_identifier).where(rule_identifiers: { label: identifier_label })
   }
 
+  default_scope { where(id: RuleResult.pluck(:rule_id)) }
```
dLobatog: Notice this chains with the `RuleResult` default scope, so only rules which also contain at least some valid result will be shown. Also notice this is not scoped by host.. which is problematic, as we could:
0 - Assume 2 hosts, host1 and host2 with reports for a profile called `standard`.
1 - `host1.rules` returns *n* rules, according to it's results in the db. 
2 - `host2` report is uploaded, and it contains *n+1* rules because OpenSCAP parsed more rules.
3 - `host2.rules` returns *n+1* rules.
4 - Now `host1.rules` also returns *n+1* rules as we are just looking for rules with valid results, but we don't care which host do they come from. To get all rules for which a host has reports, you need to do `Rule.where(id: RuleResult.where(host: host1).pluck(:rule_id))`. We don't do this on the GraphQL responses.

Gemfile.lock
```diff
@@ -44,22 +44,22 @@ GEM
       i18n (>= 0.7, < 2)
```
akofink: This isn't relevant to the issue. IMO we should keep Gemfile updates separate or when required.

app/models/profile.rb
```diff
@@ -60,7 +60,7 @@ def results(host)
               ON rules.id = profile_rules.rule_id
               WHERE profile_rules.profile_id = ?)
        ) rule_results WHERE RANK = 1', host.id, id]
-      )
+      ).select { |rr| %w(pass fail notapplicable error unknown).include?(rr.result) }
```
akofink: Does this work?

```suggestion
      ).where(result: %w(pass fail notapplicable error unknown))
```

app/models/rule.rb
akofink: default_scope?? Why not make a normal scope:

```suggestion
    scope :with_results, lambda { |rule_result_ids = RuleResult.pluck(:id)|
      where(id: rule_result_ids)
    }
```

app/models/rule_result.rb
```diff
@@ -10,4 +10,6 @@ class RuleResult < ApplicationRecord
 
   validates :host, presence: true
   validates :rule, presence: true
+
+  default_scope { where(result: %w(pass fail notapplicable error unknown)) }
```
akofink: Same here, default_scope is scary. We're most likely going to need to unscope this when dealing with tailoring work in the future:

```suggestion
  scope :selected { where.not(result: 'notselected') }
```

app/models/profile.rb
bastilian: Using `.where` might not, `find_by_sql` seems to return a plain `Array`, but if the SQL query could/would be transformed into a Arel query it should work.

bastilian: Oh, then it should also be possible to use it with the `:selected` scope, @akofink proposed.

dLobatog: `NoMethodError (undefined method `where' for #<Array:0x000056298f41c670>):`

Then again I can add just the where to the query directly

dLobatog:
gg

PR #226, opened 2019-09-11
RHICOMPL-331 - Import remediations task

dLobatog:


lib/tasks/import_remediations.rake
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+def remove_ref_id_prefix(ref_id)
+  short_ref_id = ref_id.downcase.split('xccdf_org.ssgproject.content_profile_')[1];
+  short_ref_id.present? ? short_ref_id : ref_id
+end
+
+task import_remediations: :environment do
+  account = Account.find_by_account_number(ENV['JOBS_ACCOUNT_NUMBER'])
```
dLobatog: We will need to set one of our accounts here, wanted to keep the number as a secret, so I kept this as an env var.

app/services/remediations_api.rb
```diff
@@ -0,0 +1,67 @@
...
+      Rails.logger.info("Updating rule #{rule_ref_id} "\
+                        "remediation_available: #{value.present?}")
+      remediation_available[value.present?.to_s] << rule_ref_id
+    end
+    remediation_available
+  end
+
+  def update_rules(remediations)
+    remediations.each do |available, rule_ref_ids|
+      # rubocop:disable Rails/SkipsModelValidations
```
dLobatog: If I disable this cop, this will run a loop of 100 updates, which is much slower.

app/services/remediations_api.rb
```diff
@@ -0,0 +1,67 @@
...
+        "|#{rule.ref_id}"
+    end
+  end
+
+  def profile_for_rules(rules)
+    profile_rules = ProfileRule.where(rule_id: rules.pluck(:id))
+                               .pluck(:rule_id, :profile_id).to_h
+    profile_ref_ids = Profile.where(id: profile_rules.values.uniq)
+                             .pluck(:id, :ref_id).to_h
+    profile_rules.transform_values! { |profile_id| profile_ref_ids[profile_id] }
```
dLobatog: This method just ensures we can get the profile `ref_id` for each rule as fast as possible. Without it, in `build_issues_list` I could do:
```
rules.map { |rule| "ssg:rhel7:#{rule.profile.first}|#{rule_ref_id}" } 
``` 
which runs 1 query on every loop. This saves us from that and builds a hash to be used there.

app/services/platform.rb
```diff
@@ -0,0 +1,15 @@
...
+require 'faraday'
+
+# Methods related to connecting to other platform services
+module Platform
+  def self.connection
+    faraday = Faraday.new do |f|
+      f.response :raise_error
+      f.adapter Faraday.default_adapter # this must be the last middleware
+      f.ssl[:verify] = false
+    end
```
dLobatog: You may test this by using `faraday.basic_auth 'foo', 'bar'` and using the public URL for the remediations API.

lib/tasks/import_remediations.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+task import_remediations: :environment do
+  Rails.logger.info('Starting import_remediations job')
+  RemediationsAPI.new(
+    Account.find_by(account_number: ENV['JOBS_ACCOUNT_NUMBER']),
+    ENV['REMEDIATIONS_API']
```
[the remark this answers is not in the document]
akofink: Also, IMO we should add path in the RemediationsAPI class to whatever is specified here, so it should be "https://cloud.redhat.com/api/remediations/v1". This will help ensure RemediationsAPI stays flexible to other remediations API paths.

app/services/remediations_api.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+# This class is meant to contain all calls to the Remediations API.
+class RemediationsAPI
+  def initialize(account, url)
+    @url = URI.parse(url)
+    @b64_identity = Base64.strict_encode64(account.to_identity_header.to_json)
+  end
+
+  def import_remediations
```
akofink: Consider `s/remediations/resolutions/` to match the API endpoint.

app/services/remediations_api.rb
```diff
@@ -0,0 +1,67 @@
...
+    Rule.find_in_batches(batch_size: 100).each do |rules|
+      response = Platform.connection.post(@url) do |req|
+        req.headers['X_RH_IDENTITY'] = @b64_identity
```
akofink: This stuff should come from the connection automatically, right? I think you can override things specified in the connection here that you'd like to change just for this request. Things like 'Content-Type' header could probably be moved to the connection.

app/services/platform.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'faraday'
+
+# Methods related to connecting to other platform services
+module Platform
```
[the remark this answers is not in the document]
bastilian: Or use delegate?

app/services/platform.rb
```diff
@@ -0,0 +1,15 @@
...
+  def self.connection
+    faraday = Faraday.new do |f|
+      f.response :raise_error
+      f.adapter Faraday.default_adapter # this must be the last middleware
+      f.ssl[:verify] = false
```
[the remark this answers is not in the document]
bastilian: Could using `Rails.env.production?` be a problem in CI? It seems to also have "self-signed" certs, but not sure if they are added to be trusted. If it's a problem we could use an ENV variable, set to false in docker-compose, and default to true.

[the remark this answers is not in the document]
akofink: SSL verification is turned on by default in Faraday. If we haven't had problems so far in CI, this shouldn't affect it either.

app/services/platform.rb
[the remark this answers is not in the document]
akofink: Yes, delegate would also work and probably be more flexible, as we'd choose to delegate certain methods rather than all as a subclass.

lib/tasks/import_remediations.rake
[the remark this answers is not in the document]
dLobatog: I'm removing the variable altogether as we can just use `"#{URI.parse(Settings.host_inventory_url)}#{ENV['PATH_PREFIX']}/remediations/v1/resolutions"`

app/services/remediations_api.rb
dLobatog: It would be nice to have some defaults but the list of default params for the connection are very limited - https://lostisland.github.io/faraday/middleware/list - everything else is on a per-request basis.

lib/tasks/import_remediations.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+desc 'Update rule remediation availability by querying the Remdiations API'
```
akofink: IMO it'd also be good to show the usage with the env variable here in the description.

akofink: Small typo as well:

```suggestion
desc 'Update rule remediation availability by querying the Remediations API'
```

lib/tasks/import_remediations.rake
```diff
@@ -0,0 +1,14 @@
...
+task import_remediations: :environment do
+  start_time = Time.now.utc
+  Rails.logger.info("Starting import_remediations job at #{start_time}")
+  RemediationsAPI.new(
+    Account.find_by(account_number: ENV['JOBS_ACCOUNT_NUMBER'])
+  ).import_remediations
+  end_time = Time.now.utc
+  duration = end_time - start_time
+  Rails.logger.info("Finishing import_remediations job at #{end_time} "\
+                    "and last #{duration} seconds ")
```
akofink: Not a huge deal, but this logs to the rails logger. When running a rake task, I expect logs to output to STDOUT from where I'm running the task.

config/racecar.yml
```diff
@@ -1,6 +1,8 @@
 # These config values will be shared by all environments but can be overridden.
 common: &common
   client_id: "compliance_backend"
+  brokers:
+    - localhost:29092
```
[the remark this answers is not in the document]
dLobatog: Actually this was added to the commit by mistake..

app/services/remediations_api.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def remediations_available(response)
+    remediation_available = { 'true' => [], 'false' => [] }
+    JSON.parse(response.body).each do |remediation_id, value|
+      rule_ref_id = remediation_id.split('|')[2]
+      Rails.logger.info("Updating rule #{rule_ref_id} "\
+                        "remediation_available: #{value.present?}")
+      remediation_available[value.present?.to_s] << rule_ref_id
+    end
+    remediation_available
```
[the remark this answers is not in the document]
dLobatog: Good point, let me try that out :smile:

lib/tasks/import_remediations.rake
dLobatog: I thought RAILS_LOG_TO_STDOUT would also force this to print it out, but it doesn't - I'm changing this.

app/services/remediations_api.rb
```diff
@@ -0,0 +1,70 @@
...
+  def profile_for_rules(rules)
+    profile_rules = ProfileRule.where(rule_id: rules.pluck(:id))
+                               .pluck(:rule_id, :profile_id).to_h
+    profile_ref_ids = Profile.where(id: profile_rules.values.uniq)
+                             .pluck(:id, :ref_id).to_h
+    profile_rules.transform_values! { |profile_id| profile_ref_ids[profile_id] }
+  end
+
+  def remediations_available(response)
+    JSON.parse(response.body).each_with_object(
```
dLobatog: `rubocop` suggests not to use inject and use `each_with_object` - it saves 1 step, returning the accumulator at the end of each loop is not needed with it :smile:

lib/tasks/import_remediations.rake
```diff
@@ -0,0 +1,24 @@
...
+  according to whether or not there's a remediation available in the Remediations
+  API.
+
+  Examples:
+    # JOBS_ACCOUNT_NUMBER=000001 rake import_remediations
+END_DESC
+
+task import_remediations: :environment do
+  start_time = Time.now.utc
+  puts "Starting import_remediations job at #{start_time}"
```
dLobatog: Updated to use `puts` and have a better description

app/services/remediations_api.rb
akofink: Oh snap! Even better :)

PR #227, opened 2019-09-16
RHICOMPL-216 - Show only pass/fail results

dLobatog:
Moved from https://github.com/RedHatInsights/compliance-backend/pull/225 

- This version does not rely on default_scope
- It should be followed by a PR that stops parsing irrelevant results (notselected)
- The response time is considerably faster as we cut down results by 90%

### DIFF
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -22,14 +22,13 @@ module Types
     def rules(args = {})
       selected_columns = args[:lookahead].selections.map(&:name) &
                          ::Rule.column_names.map(&:to_sym)
-      rules = object.rules.select(selected_columns)
-
+      rules = object.rules.select(selected_columns << :id).where(
+        id: RuleResult.selected.pluck(:rule_id)
+      )
       rules = rules.with_identifier(args[:identifier]) if args.dig(:identifier)
       rules = rules.with_references(args[:references]) if args.dig(:references)
-
       rules = lookahead_includes(args[:lookahead], rules,
                                  identifier: :rule_identifier)
-
       rules
     end
     # rubocop:enable AbcSize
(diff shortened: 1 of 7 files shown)


app/graphql/types/profile.rb
```diff
@@ -22,14 +22,14 @@ class Profile < Types::BaseObject
     def rules(args = {})
       selected_columns = args[:lookahead].selections.map(&:name) &
                          ::Rule.column_names.map(&:to_sym)
-      rules = object.rules.select(selected_columns)
-
+      selected_columns << :id
```
dLobatog: We have to add the `id` here. Otherwise the queries are done with "rule_id = NULL". So for example, computing compliance is done by (`SELECT * in rule_results ...long rank query... WHERE host_id=whatever AND rule_id=NULL` which will always be empty.

Maybe currently we don't see any issue with this in production because results and compliance are cached, but we will see it at the point that the `compliance` is recomputed.

akofink:
And rubocop isn't happy, of course :)

test/models/profile_test.rb
```diff
@@ -25,10 +25,10 @@ class ProfileTest < ActiveSupport::TestCase
     assert profiles(:one).compliant?(hosts(:one))
   end
 
-  test 'host is not compliant if some rules are "fail" or "notselected"' do
+  test 'host is not compliant if some rules are "fail" or "error"' do
```
dLobatog: notselected, notchecked aren't even showing up anymore

app/models/profile.rb
```diff
@@ -47,19 +47,22 @@ def compliant?(host)
...
+              )
+           FROM rule_results rr2
+           WHERE rr2.host_id = ? AND rr2.result IN (?) AND rr2.rule_id IN
+              (SELECT rules.id FROM rules
+               INNER JOIN profile_rules
+               ON rules.id = profile_rules.rule_id
+               WHERE profile_rules.profile_id = ?)
+          ) rule_results WHERE RANK = 1',
+          host.id, RuleResult::SELECTED, id
+        ]
```
bastilian: This could go into a scope as well, it would allow for a descriptive name and would highlight the parameters passed.

akofink: IMO scopes should return an AR::Relation so they can be chained

PR #234, opened 2019-09-24
311: Allow importing datastreams to compliance DB

akofink:
Now that we can download and extract datastreams, we need to import them to our database! This requires a new `Datastream` model and associations with `Rule` and `Profile`. I'd like to standardize how we import, but maybe that can be a refactoring for later.

```sh
$ time d-c exec reports-consumer bundle exec rake ssg:import DATASTREAM_FILE='./test/fixtures/files/ssg-rhel7-ds.xml' # empty database
2019-10-23T15:22:34.716Z 5556 TID-gqskgfd34 INFO: GitLab reliable fetch activated!
Importing ./test/fixtures/files/ssg-rhel7-ds.xml at 2019-10-23 15:22:34 UTC
Finished importing ./test/fixtures/files/ssg-rhel7-ds.xml in 45.478060306 seconds.
docker-compose exec reports-consumer bundle exec rake ssg:import   1.00s user 0.39s system 2% cpu 49.431 total
 
$ time d-c exec reports-consumer bundle exec rake ssg:import DATASTREAM_FILE='./test/fixtures/files/ssg-rhel7-ds.xml' # second run, no new data
2019-10-23T15:23:27.626Z 5576 TID-gqct9i0i4 INFO: GitLab reliable fetch activated!
Importing ./test/fixtures/files/ssg-rhel7-ds.xml at 2019-10-23 15:23:27 UTC
Finished importing ./test/fixtures/files/ssg-rhel7-ds.xml in 38.478002909 seconds.
docker-compose exec reports-consumer bundle exec rake ssg:import   0.96s user 0.39s system 3% cpu 41.972 total
```

This also affects report parsing:

```sh
2019-10-23T15:26:04.922Z 1 TID-go88o6pel ParseReportJob JID-687351b8d6a0ca2ac640b7a0 INFO: start
2019-10-23T15:26:48.989Z 1 TID-go88o6pel ParseReportJob JID-687351b8d6a0ca2ac640b7a0 INFO: done: 44.067 sec # rhel7 standard
2019-10-23T15:31:44.154Z 1 TID-go88o6pel ParseReportJob JID-683134941b4974fad90876fe INFO: start
2019-10-23T15:32:31.356Z 1 TID-go88o6pel ParseReportJob JID-683134941b4974fad90876fe INFO: done: 47.202 sec # rhel7 hipaa
2019-10-23T15:32:56.365Z 1 TID-go88o6pel ParseReportJob JID-caef0872cf52b0d428fbdbae INFO: start
2019-10-23T15:33:01.786Z 1 TID-go88o6pel ParseReportJob JID-caef0872cf52b0d428fbdbae INFO: done: 5.421 sec # rhel6 standard (much smaller datastream file)
```

Required: ~https://github.com/dLobatog/openscap_parser/pull/19~ merged


app/models/xccdf/benchmark.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Stores information about xccdf benchmarks
```
dLobatog: Just a bit more info about how this is populated from the SSG would be nice

app/models/xccdf/benchmark.rb
```diff
@@ -0,0 +1,11 @@
...
+module Xccdf
+  class Benchmark < ApplicationRecord
+    has_many :profiles, dependent: :destroy
+    has_many :rules, dependent: :destroy
+    validates :ref_id, uniqueness: { scope: %i[version] }, presence: true
```
dLobatog: These are shared by all accounts, right?

db/migrate/20190924171028_create_benchmarks.rb
```diff
@@ -0,0 +1,10 @@
+class CreateBenchmarks < ActiveRecord::Migration[5.2]
+  def change
+    create_table :benchmarks, id: :uuid do |t|
+      t.string :ref_id
```
dLobatog: How about adding `null: false` to the ones that cannot be changed?

app/services/datastream_importer.rb
```diff
@@ -0,0 +1,72 @@
...
+      )
+
+      profile
+    end
+  end
+
+  def save_rules(benchmark: Xccdf::Benchmark.new, rules: [])
+    rules.map do |rule|
+      rule = Rule.find_or_initialize_by(
+        ref_id: profile.id,
```
dLobatog: `rule.id`

app/services/datastream_importer.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    benchmark
+  end
+
+  def save_profiles(benchmark: Xccdf::Benchmark.new, profiles: [])
+    profiles.map do |profile|
+      profile = Profile.find_or_initialize_by(
+        ref_id: profile.id,
+        name: profile.title,
+        account_id: nil,
```
dLobatog: I'm a bit confused by this method. Currently profiles are created at import time, and the account ID is assigned to them at the same time. We could set the benchmark_id to those profiles and that would be it. 

Is this meant to import just the canonical profiles that are in the SSG? If so, we have to introduce the `parent_id` in this PR already.

app/services/datastream_importer.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  private
+
+  def save_benchmark(benchmark: OpenscapParser::Benchmark.new)
+    benchmark = Xccdf::Benchmark.find_or_initialize_by(
+      ref_id: benchmark.id,
+      version: benchmark.version
+    )
+
+    benchmark.update(
```
dLobatog: Nitpick, why not chaining all like:

```ruby
Xccdf::Benchmark.find_or_initialize_by(
      ref_id: benchmark.id,
      version: benchmark.version
).update(
     title: benchmark.title,
     description: benchmark.description
)
```

The current code is not updating the benchmark as far as I can see, as the argument `benchmark` is overshadowed by `benchmark = Xccdf::Benchmark.find_or...`

app/models/xccdf/benchmark.rb
akofink: Yes, that's right.

app/services/datastream_importer.rb
akofink: Good catch about the overshadowing. I wanted this method to take the OpenscapParser benchmark and return an ActiveRecord benchmark because we need the AR benchmark ID to save the rules and profiles later. The `update()` method returns true or false - that's why I've created this intermediate variable.

app/services/xccdf_report_parser.rb
```diff
@@ -46,36 +44,46 @@ def inventory_api
...
+    invalidate_cache
+  end
+
+  def host_profiles
+    host_profile = @profiles.find do |profile|
+      profile.ref_id == @test_result_file.test_result.profile_id
+    end
+
+    unless @host.profiles.pluck(:ref_id).include?(host_profile.ref_id)
+      host_profile = host_profile.dup
```
akofink: @dLobatog I'd like some help figuring this out. At this point in the import, we've imported all the benchmark info, including profiles, and those profiles have `nil` `account_id`. The problem is that all the associated rules (via `ProfileRules` table) are not duplicated, so this dupe profile has no rules. I'm leaning towards adding `parent_id` now like we talked about and only ever have profiles with `nil` accound_id be associated with rules. What do you think?

akofink: I think for now, I'm just going to copy all the rules from the parent profile without adding `parent_id`

akofink:
Tests should be green now (finally) :)

akofink:
Tests will not pass here until a new openscap_parser is released https://github.com/dLobatog/openscap_parser/pull/20

app/models/xccdf.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+Dir[File.join(__dir__, 'xccdf', '*.rb')].each { |f| require f }
+
+# Represents all our models which come directly from the Xccdf XML format for
+# SCAP content
+module Xccdf
```
akofink: My hope is to move Rule, Profile, RuleResult, etc. here as well, but it's a refactoring for another PR.

app/services/concerns/xccdf/rules.rb
```diff
@@ -9,51 +9,56 @@ module Rules
...
+        @new_rules = @op_rules.reject do |rule|
           ref_ids.include? rule.id
         end
       end
 
       def save_rules
-        add_profiles_to_old_rules(rules_already_saved, new_profiles)
+        add_profiles_to_old_rules
         associate_rule_references
         Rule.import!(new_rule_records, recursive: true)
```
dLobatog: With an empty database, I'm seeing `ActiveRecord::RecordInvalid: Validation failed: Rule references is invalid` on this database. After that error, if I rerun the task, the error goes away (I think this happens because rules were already imported properly, so `new_rule_records` is `[]` the 2nd time we run the import, but rule references were lost)

app/models/rule_identifier.rb
```diff
@@ -7,7 +7,10 @@ class RuleIdentifier < ApplicationRecord
   validates :label, presence: true
   validates :system, presence: true
 
-  def self.from_oscap_rule(oscap_rule)
-    new(oscap_rule.identifier) if oscap_rule.identifier.dig(:label)
+  def self.from_openscap_parser(op_rule_identifier)
+    if op_rule_identifier.label # rubocop:disable Style/GuardClause
```
dLobatog: Why disabling this instead of `return if op_rule_identifier_label.blank?`?

test/services/datastream_importer_test.rb
```diff
@@ -0,0 +1,19 @@
...
+    DATASTREAM_FILE = file_fixture('ssg-rhel7-ds.xml')
+  end
+
+  test 'datastream import' do
+    importer = DatastreamImporter.new(DATASTREAM_FILE)
+    importer.expects(:save_benchmark)
+    importer.expects(:save_profiles)
+    importer.expects(:save_rule_references)
+    importer.expects(:save_rules)
+    importer.import!
```
dLobatog: What is this meant to test? Currently it just checks that methods like `@op_benchmark.profiles` don't throw an exception

app/services/xccdf_report_parser.rb
```diff
@@ -54,28 +63,46 @@ def save_host
...
+  def host_profiles
+    test_result_profile = @profiles.find do |profile|
+      profile.ref_id == @test_result_file.test_result.profile_id
+    end
+
+    unless @host.profiles.pluck(:ref_id).include?(test_result_profile.ref_id)
+      host_profile = test_result_profile.dup
+      host_profile.update(account: @account, rules: test_result_profile.rules)
+
+      return @host.profiles | [host_profile]
```
dLobatog: Not sure what exactly do you mean to return early here?

app/services/concerns/xccdf/rules.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+
+      def preexisting_profile_ids
+        @preexisting_profile_ids ||= ::ProfileRule.select(:profile_id).where(
+          rule_id: rules_already_saved.pluck(:id)
+        ).pluck(:profile_id).uniq
+      end
+
+      def new_rules
+        return @new_rules if @new_rules
```
dLobatog: nitpick, `.present?`?

app/services/concerns/xccdf/profiles.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    included do
+      def profiles
+        @profiles ||= @op_profiles.map do |op_profile|
+          profile = ::Profile.find_or_initialize_by(
+            name: op_profile.name, ref_id: op_profile.id,
+            account_id: @account&.id, benchmark_id: @benchmark&.id
+          )
+
+          unless profile.persisted?
```
dLobatog: nitpick, `profile.new_record?`

test/services/datastream_importer_test.rb
akofink: It's just testing the right `save_` methods are called, that's it. Each `save_` method is tested elsewhere (in the Xccdf module tests)

app/services/xccdf_report_parser.rb
akofink: This returns the union of the two arrays. `(The existing host profiles) UNION ([One more profile, maybe new, maybe already existing])`

app/models/rule_identifier.rb
akofink: rubocop gets confused here. The contents of the block, while one line, must wrap due to the line length requirement. When I use a guard clause, I need to wrap the one line across two or three, and rubocop then complains: `Style/MultilineIfModifier: Favor a normal if-statement over a modifier clause in a multiline statement.`. :man_shrugging:

akofink:
Some tests are still failing, but the code should be stable at this point. Importing reports or datastreams in any order should work as expected. Some changes to how this used to work:

- All profiles from a benchmark are importer with nil account_id
- Profile is validated to have both host and account or neither host nor account
- Only test result profiles are associated with a host, not all benchmark profiles
- A host's test result profile is cloned, if it doesn't exist, from an existing profile with nil account_id
- Only selected rules for a profile are associated with a profile

akofink:
Tests should be green now! :crossed_fingers:

akofink:
- We might want to tweak the datastream import task to run as a sidekiq job like you've done with the fix_xccdf task.
- No work has started on any kind of openshift job
- We're still taking in reports from hosts using the scap-security-guide packaged with RHEL, so we won't see any benefit of importing the latest datastream
- We should coordinate with remediations to ensure we're both using the same SSG version and updating the SSG at or around the same time.

PR #235, opened 2019-09-27
RHICOMPL-299 - Provide OpenAPI v3 endpoint

dLobatog:


dLobatog:
@akofink Do you mean add it to https://github.com/domaindrivendev/rswag ? I think the whole thing would need to export to yaml. It doesn't seem so difficult, but I don't know all of the OpenAPI v2 vs v3 differences. It already has an endpoint for it's own JSON at `api/compliance/v1/openapi.json`

akofink:
When you `GET /api/compliance/openapi` it shows all the routes and params that are accepted with their types. What isn't shown is `/api/compliance/openapi` itself:

```
$ http https://ci.foo.redhat.com:1337/api/compliance/openapi.json | grep openapi   
openapi: 3.0.0
```

akofink:
Also, it was just a question. I'm not blocking this PR on this, of course :)

bastilian:
How is this built? Is there maybe a way to automate it to keep v2 & v3 updated?

dLobatog:
@akofink oh got it - yeah, I can add the endpoint to itself, that's not an issue. Notice the controller should pick up `api/compliance/openapi` but also `api/compliance/openapi.yaml`, I used `https://github.com/Mermade/oas-kit/blob/master/packages/swagger2openapi/README.md` I should add this to the README. I would add a rake task but it's a JS standalone script.

PR #236, opened 2019-09-30
Adapting CI to new pod configuration 

quarckster:
Depends on https://github.com/RedHatInsights/insights-pipeline-lib/pull/40


openshift/Jenkins/slave_pod_template.yaml
```diff
@@ -59,16 +60,27 @@ spec:
...
       value: test
-    image: jenkins-slave-rhel7-ruby25-openscap:latest
-    name: jnlp
+    image: docker-registry.default.svc:5000/buildfactory/s2i-insights-compliance
+    name: s2i-insights-compliance
+    command:
+      - cat
+    resources:
+      requests:
+        cpu: 500m
```
akofink: I was told that we should ensure all our requests are set (unset means request == limit) and at the minimum (100m CPU, 100Mi memory I think?) to ensure the pods can spin up under heavy load of the cluster. The limit may still be set to whatever is needed by the service.

PR #237, opened 2019-09-30
RHICOMPL-321 - Block import of xccdf.xml reports

dLobatog:
In draft for now, needs testing.

### DIFF
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -7,6 +7,8 @@ end
 
 # Error to raise if no metadata is available
 class EmptyMetadataError < StandardError; end
+# Error to raise if the format of the report is wrong
+class WrongFormatError < StandardError; end
 
 # Takes in a path to an XCCDF file, returns all kinds of properties about it
 # and saves it in our database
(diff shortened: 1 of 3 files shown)


app/services/xccdf_report_parser.rb
```diff
@@ -25,6 +27,12 @@ def initialize(report_contents, message)
     @metadata = message['metadata']
     @host_inventory_id = message['id']
     @oscap_parser = OpenscapParser::Base.new(report_contents)
+    check_report_format
+  end
+
+  def check_report_format
+    raise WrongFormatError if @oscap_parser.report_xml.
+      at_xpath('.//benchmark/@href').value.match?('.*-xccdf.xml')
```
akofink: To be safe, we should look only for benchmarks nested under a TestResult. xpath supports matching as well:

```rb
.xpath(".//TestResult/benchmark[contains(@href, '-xccdf.xml')]/@href").any?
```

dLobatog: I can do that for sure :smile:  there was only 1 match for this xpath as far as I could see but it's a good point.

akofink: There _should_ only ever be one `<TestResult />` afaik, but I could be wrong.

test/services/xccdf_report_parser_test.rb
```diff
@@ -216,4 +216,21 @@ class XCCDFReportParserTest < ActiveSupport::TestCase
       end
     end
   end
+
+  context 'datastream-based reports only' do
+    should 'raise an error if the report is not coming from a datastream' do
+      fake_report = file_fixture('rhel-xccdf-report-wrong.xml').read
```
dLobatog: This is just a report coming from a xccdf file.

PR #240, opened 2019-10-03
Change remediations API URL & header

dLobatog:


app/services/remediations_api.rb
```diff
@@ -3,15 +3,15 @@
 # This class is meant to contain all calls to the Remediations API.
 class RemediationsAPI
   def initialize(account)
-    @url = URI.parse("#{URI.parse(Settings.host_inventory_url)}"\
+    @url = URI.parse("#{URI.parse(Settings.remediations_url)}"\
```
dLobatog: The remediations URL cannot be created using `Settings.host_inventory_url` in any environment other than production

app/services/remediations_api.rb
```diff
@@ -3,15 +3,15 @@
...
                      "#{ENV['PATH_PREFIX']}/remediations/v1/resolutions")
     @b64_identity = Base64.strict_encode64(account.fake_identity_header.to_json)
   end
 
   def import_remediations
     Rule.find_in_batches(batch_size: 100).each do |rules|
       response = Platform.connection.post(@url) do |req|
-        req.headers['X_RH_IDENTITY'] = @b64_identity
+        req.headers['X-RH-IDENTITY'] = @b64_identity
```
dLobatog: @jharting We noticed that in this app that inventory and remediations use a different version of the token. (underscores vs dashes) - compliance uses underscores too.

bastilian:
Thanks @dLobatog!

akofink:
We also need to update the docker-compose, right?

PR #241, opened 2019-10-09
RHICOMPL-375 - Filter rules by system ID

dLobatog:


app/graphql/types/profile.rb
```diff
@@ -22,9 +24,7 @@ class Profile < Types::BaseObject
     def rules(args = {})
       selected_columns = args[:lookahead].selections.map(&:name) &
                          ::Rule.column_names.map(&:to_sym)
-      rules = object.rules.select(selected_columns << :id).where(
-        id: RuleResult.selected.pluck(:rule_id)
```
dLobatog: This was really the crux of the problem. `RuleResult.selected` will return all rules with a valid result.. no matter the host. So the following could happen:

1. Host A checks in with policy PCI which contains 1025 rules, of which 90 are valid (not `notselected`, `notapplicable` etc..)
2. System Details page looks good because there are no other hosts, and this `RuleResult.selected` just returns the valid rules for the current host.
3. Host B checks in with policy STIG-DISA from the same SSG as Host A. Policy contains 1025 rules, of which 90 are valid. This set of valid rules is DIFFERENT to the one in point 1, it may have some overlap.
4. Now both Host B and Host A system details rules table show "120 rules" instead of "90". Why? Because the rules returned by this query contain all rules with a valid result no matter the policy .

app/graphql/types/profile.rb
```diff
@@ -22,9 +24,7 @@ class Profile < Types::BaseObject
...
-      )
+      rules = rules_for_system(object, selected_columns, args[:system_id])
```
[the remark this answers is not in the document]
dLobatog: Sounds good to me, I'll update it

app/graphql/types/profile.rb
```diff
@@ -71,11 +73,11 @@ def compliant(system_id:)
     end
 
     def rules_passed(system_id:)
-      object.results(Host.find(system_id)).count { |result| result }
+      Host.find(system_id).rules_passed(object)
```
dLobatog: We already have this method implemented in `concerns/system_like.rb`

app/graphql/types/profile.rb
```diff
@@ -71,11 +73,11 @@ def compliant(system_id:)
...
     end
 
     def rules_failed(system_id:)
-      object.results(Host.find(system_id)).count(&:!)
+      Host.find(system_id).rules_failed(object)
```
dLobatog: Ditto

app/models/concerns/system_like.rb
```diff
@@ -40,4 +40,16 @@ def compliance_score
...
   end
+
+  def selected_rules(profile = nil, selected_columns = [])
+    rules_for_system = profile.present? ? profile.rules : rules
+
+    rules_for_system.select(selected_columns).yield_self do |temp_rules|
+      temp_rules.where(
+        id: RuleResult.selected.where(
+          host_id: id
+        ).pluck(:rule_id)
```
akofink: Nit: I think this should be defined as a variable outside the `select` loop so we don't run it N times.

dLobatog: That's a good point - IDK if the interpreter would fix that automatically but regardless I'd rather just make sure. I'll update the PR with that.

PR #244, opened 2019-10-10
RHICOMPL-364 - Migrate broken report ref IDs

dLobatog:
Run with noop with `NOOP=true rake fix_xccdf_reports`

### DIFF
--- a/app/jobs/migrate_xccdf_reports_job.rb
+++ b/app/jobs/migrate_xccdf_reports_job.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Migrates ref IDs from reports to comply with the expected standard.
+# Profile ref IDs should look like "xccdf_org.ssgproject.content_profile_*"
+# Rule ref IDs should look like "xccdf_org.ssgproject.content_rule_*"
+class MigrateXCCDFReportsJob
+  include Sidekiq::Worker
+
+  def perform(account_number, noop)
+    XCCDFReportMigration.new(
+      Account.find_by!(account_number: account_number),
+      noop
+    ).run
+  end
+end
--- a/app/services/xccdf_report_migration.rb
+++ b/app/services/xccdf_report_migration.rb
@@ -1,0 +1,131 @@
+# frozen_string_literal: true
+
+# Migrates ref IDs from reports to comply with the expected standard.
+# Profile ref IDs should look like "xccdf_org.ssgproject.content_profile_*"
+# Rule ref IDs should look like "xccdf_org.ssgproject.content_rule_*"
+class XCCDFReportMigration
+  def initialize(account, noop)
+    @account = account
+    @noop = noop
+  end
+
+  def logger
+    Sidekiq.logger
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def run
+    xccdf_profiles = broken_profiles
+    logger.info "Found #{xccdf_profiles.count} profiles from XCCDF reports."\
+      ' Renaming profiles...'
+    rename_profiles(xccdf_profiles)
+    logger.info "Profiles rename finished \n-------------------------------\n"\
+      'Some profiles might have not been renamed if the name '\
+      'conflicts with another profile in the same account.'\
+      "Let's proceed to migrating the XCCDF profile results into the"\
+      ' DS profile results'
+    xccdf_profiles = broken_profiles
+    logger.info "Found #{xccdf_profiles.count} profiles from XCCDF reports."
+    migrate_profiles(xccdf_profiles)
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  def broken_profiles
+    Profile.where(account_id: @account.id).where.not(
+      'ref_id LIKE ?', 'xccdf_org.ssgproject.content_profile_%'
+    )
+  end
+
+  def rename_profiles(xccdf_profiles)
+    return if @noop
+
+    xccdf_profiles.map do |profile|
+      profile.update(
+        ref_id: "xccdf_org.ssgproject.content_profile_#{profile.ref_id}"
+      )
+      migrate_rules(profile.rules, profile) if profile.errors.empty?
+    end
+  end
+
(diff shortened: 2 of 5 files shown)


lib/tasks/fix_xccdf_reports.rake
```diff
@@ -0,0 +1,18 @@
...
+desc <<-END_DESC
+END_DESC
+
+task fix_xccdf_reports: :environment do
+  noop = ENV['NOOP']
+  start_time = Time.now.utc
+  puts '---------- NOOP --------' if noop
+  puts "Starting fix_xccdf_reports job at #{start_time}"
+
+  MigrateXCCDFReportsJob.perform_later(ENV['ACCOUNT_NUMBER'], noop)
```
akofink: Maybe we should also create another rake task which calls this one with all accounts currently in the database.

lib/tasks/fix_xccdf_reports.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+desc <<-END_DESC
+END_DESC
```
akofink: May be good to fill this in.

akofink:
This doesn't output logs where the rake task is run. It logs inline with the rest of the rails logs. That doesn't seem right to me.

Also, a duplicate profile seems to cause issues: https://paste.fedoraproject.org/paste/-aGnO3nwX3~YHIWg52VfgQ

app/services/xccdf_report_migration.rb
```diff
@@ -0,0 +1,133 @@
...
+    ).first
+  end
+
+  def migrate_profile_hosts(xccdf_profile, ds_profile)
+    return if @noop
+
+    # Justification: Enhance the speed of the mass-migration
+    # rubocop:disable Rails/SkipsModelValidations
+    logger.info ProfileHost.where(profile_id: xccdf_profile.id)
+                           .update_all(profile_id: ds_profile.id)
```
akofink: This can fail if another ProfileHost exists already with this host/profile combo.

app/services/xccdf_report_migration.rb
```diff
@@ -0,0 +1,133 @@
...
+      ds_rule = rename_rule(xccdf_rule, ds_rule)
+      next if @noop
+
+      ds_profile.rules << ds_rule unless ds_profile.rules.include? ds_rule
+      migrate_rule(xccdf_rule, ds_rule)
+    end
+  end
+
+  def destroy_profile(xccdf_profile)
+    xccdf_profile.rules.map(&:destroy)
```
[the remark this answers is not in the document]
dLobatog: I tried that before, but for some reason the test doesn't pass if I do that. I'll dig deeper and find out why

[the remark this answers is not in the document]
akofink: Also, destroy will run callbacks, deleting associated ProfileRule, RuleResult, RuleReferenceRule, and RuleIdentifier objects. Is that what we want (I think so, but double checking)?

app/jobs/migrate_xccdf_reports_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Migrates ref IDs from reports to comply with the expected standard.
+# Profile ref IDs should look like "xccdf_org.ssgproject.content_profile.*"
```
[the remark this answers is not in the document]
dLobatog: The comment is wrong - the code however uses the `_` instead of the `.`

app/services/xccdf_report_migration.rb
```diff
@@ -0,0 +1,133 @@
...
+    xccdf_profiles = broken_profiles
+    logger.info "Found #{xccdf_profiles.count} profiles from XCCDF reports."
+    migrate_profiles(xccdf_profiles)
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  def broken_profiles
+    Profile.where(account_id: @account.id).reject do |profile|
+      profile.ref_id.match?(/xccdf_org.ssgproject.content_profile_.*/)
+    end
```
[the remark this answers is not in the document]
dLobatog: Good point!

lib/tasks/fix_xccdf_reports.rake
dLobatog: :+1:

app/services/xccdf_report_migration.rb
```diff
@@ -0,0 +1,131 @@
...
+      ref_id: "xccdf_org.ssgproject.content_profile_#{xccdf_profile.ref_id}"
+    ).first
+  end
+
+  def migrate_profile_hosts(xccdf_profile, ds_profile)
+    return if @noop
+
+    logger.info(ProfileHost
+      .where(profile_id: xccdf_profile.id)
+      .map { |profile_host| profile_host.update(profile_id: ds_profile.id) })
```
dLobatog: This runs validations - it's probably fine as every Host doesn't have more than a couple profiles usually and this task runs once. It helps us avoid the issue with `update_all` triggering DB validations and therefore an exception that kills the job.

app/services/xccdf_report_migration.rb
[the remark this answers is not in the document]
dLobatog: `ProfileRule`, `RuleResult` we definitely want them destroyed as we have already migrated them by this point. 
As for `RuleReferenceRule` and `RuleIdentifier`, they have to be migrated too - especially in the case where the user uploads the XCCDF-based report first, and then the DS-based one. In that case the XCCDF-based rules contain the identifiers as you can't create them twice - they are unique by label. They seem to be migrated properly with this task already.

app/services/xccdf_report_migration.rb
dLobatog: Fixed

akofink:
Yes, it makes sense to show the logs in Sidekiq, but this rake task isn't asynchronous (i.e. fire and forget), so I think it also makes sense to output from where the rake task is run (why not both?). It's not extremely important IMO, but it's just a bit confusing.

PR #249, opened 2019-10-17
RHICOMPL-363 - Make uniqueness constraint for profile account_id and name

dLobatog:
Currently if many threads start to upload reports with the same profile
for the first time to an account, it could happen that 6 sidekiq pods
(at the moment these are 6 threads but it could change) start processing
the report at the same time.

Since all data is saved on the database at once, the Rails uniqueness
constraints may NOT apply in this case of concurrent parsing. It will
end with multiple profiles with the same ref_id, name and account_id
which is not what is expected.

This PR introduces a DB constraint so that when this happens, only 1 of
the jobs will actually save the profile, the first one to finish
parsing. The rest of them will likely fail - which is EXPECTED. As the
job is going to retry 3 times, it will retry the parsing and the 2nd
time `Profile.find_or_initialize` will find the pre-existing profile, so
it's eventually consistent.

### DIFF
--- a/db/migrate/20191017185340_add_unique_constraint_to_profile_by_account.rb
+++ b/db/migrate/20191017185340_add_unique_constraint_to_profile_by_account.rb
@@ -1,0 +1,5 @@
+class AddUniqueConstraintToProfileByAccount < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:profiles , %i[account_id ref_id benchmark_id], unique: true)
+  end
+end
--- a/test/services/xccdf_report_migration_test.rb
+++ b/test/services/xccdf_report_migration_test.rb
@@ -49,7 +49,7 @@ class XCCDFReportMigrationTest < ActiveSupport::TestCase
     conflict_profile = Profile.create!(
       account: accounts(:test),
       hosts: [hosts(:one)],
-      name: 'footitle',
+      name: 'footitle2',
       benchmark: benchmarks(:one),
       ref_id: "xccdf_org.ssgproject.content_profile_#{@profile.ref_id}"
     )


db/migrate/20191017185340_add_unique_constraint_to_profile_by_account.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueConstraintToProfileByAccount < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:profiles , %i[account_id name], unique: true)
```
akofink: This will very quickly need to be:

```suggestion
    add_index(:profiles , %i[account_id name benchmark_id], unique: true)
```

test/services/xccdf_report_migration_test.rb
```diff
@@ -49,7 +49,7 @@ class XCCDFReportMigrationTest < ActiveSupport::TestCase
     conflict_profile = Profile.create!(
       account: accounts(:test),
       hosts: [hosts(:one)],
-      name: 'footitle',
+      name: 'footitle2',
```
dLobatog: I needed to change the title as this conflicting profile couldn't occur anymore due to the uniqueness constraint. The migration test however is still valid as all we compare here is the ref IDs of `conflict_profile` and `@profile`

db/migrate/20191017185340_add_unique_constraint_to_profile_by_account.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueConstraintToProfileByAccount < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:profiles , %i[account_id name benchmark_id], unique: true)
```
akofink: Profiles aren't even unique by ID in the DB. Maybe we should add that too?

akofink: Also, this doesn't match the model. Name doesn't matter, and ref_id does: https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/profile.rb#L17

```suggestion
    add_index(:profiles , %i[account_id ref_id benchmark_id], unique: true)
```

akofink:
This also doesn't work if one column is nil (i.e. account_id):

```rb
[52] pry(main)> (p=Profile.new(account_id: nil, ref_id: 'foo', name: 'foo', benchmark_id: Xccdf::Benchmark.first.id)).save(validate: false)
  Xccdf::Benchmark Load (1.0ms)  SELECT  "benchmarks".* FROM "benchmarks" ORDER BY "benchmarks"."id" ASC LIMIT $1  [["LIMIT", 1]]
   (0.7ms)  BEGIN
  Profile Create (1.1ms)  INSERT INTO "profiles" ("name", "ref_id", "created_at", "updated_at", "benchmark_id") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["name", "foo"], ["ref_id", "foo"], ["created_at", "2019-10-25 14:04:55.828752"], ["updated_at", "2019-10-25 14:04:55.828752"], ["benchmark_id", "5b836812-4208-48a6-87e6-af2ebfe83bac"]]
   (3.7ms)  COMMIT
=> true
[53] pry(main)> (p=Profile.new(account_id: nil, ref_id: 'foo', name: 'foo', benchmark_id: Xccdf::Benchmark.first.id)).save(validate: false)
  Xccdf::Benchmark Load (1.6ms)  SELECT  "benchmarks".* FROM "benchmarks" ORDER BY "benchmarks"."id" ASC LIMIT $1  [["LIMIT", 1]]
   (0.6ms)  BEGIN
  Profile Create (1.1ms)  INSERT INTO "profiles" ("name", "ref_id", "created_at", "updated_at", "benchmark_id") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["name", "foo"], ["ref_id", "foo"], ["created_at", "2019-10-25 14:04:56.099062"], ["updated_at", "2019-10-25 14:04:56.099062"], ["benchmark_id", "5b836812-4208-48a6-87e6-af2ebfe83bac"]]
   (4.3ms)  COMMIT
=> true
[54] pry(main)> (p=Profile.new(account_id: nil, ref_id: 'foo', name: 'foo', benchmark_id: Xccdf::Benchmark.first.id)).save(validate: false)
```

db/migrate/20191017185340_add_unique_constraint_to_profile_by_account.rb
dLobatog: Not sure if you mean something different here, but Profile ID is the primary key, so they're not_null and unique by it:
```
compliance=# \d profiles
                                 Table "public.profiles"
        Column         |            Type             |             Modifiers
-----------------------+-----------------------------+------------------------------------
 id                    | uuid                        | not null default gen_random_uuid()
 name                  | character varying           |
 ref_id                | character varying           |
 policy_id             | uuid                        |
 created_at            | timestamp without time zone | not null
 updated_at            | timestamp without time zone | not null
 description           | character varying           |
 account_id            | uuid                        |
 compliance_threshold  | double precision            | default 100.0
 business_objective_id | uuid                        |
Indexes:
    "profiles_pkey" PRIMARY KEY, btree (id)
    "index_profiles_on_account_id" btree (account_id)
    "index_profiles_on_business_objective_id" btree (business_objective_id)
    "index_profiles_on_name" btree (name)
    "index_profiles_on_policy_id" btree (policy_id
```

dLobatog: Good catch, the commit was right before https://github.com/RedHatInsights/compliance-backend/commit/8bf4510eca6bcc955ef28ae4ede2fb12607980f6#diff-6ef5e2d67a2bf4ad80d150039dccd3f2L17 - I didn't realize the name constraint was removed

akofink: Ok yeah, idk what I was thinking here :laughing:

akofink: Unique name was essentially a stop-gap solution to the differences between benchmark versions. Now we don't need that anymore! :champagne:

akofink:
The scenario you describe is not completely covered by this PR: remember that these nil account_id profiles are created during report import now as well if the profile didn't already exist. Regardless of how it happens though, I'm not sure postgresql offers a clean solution to constrain uniqueness with nulls. This PR by itself is a big improvement, but I don't think it will completely solve the bug. Maybe instead of using nil account_id, we could use empty string `''` or `'-1'` :man_shrugging: What do you think?

akofink:
By the way, https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/concerns/xccdf/profiles.rb#L15 never imports profiles with non-null account_id. When a test result is found, the import code now clones from one of these canonical profiles imported from the benchmark at https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/concerns/xccdf/hosts.rb#L13.

dLobatog:
@akofink `null` in SQL really means something like 'unknown' rather than 'not assigned' - so Postgres itself does OK by not imposing the uniqueness constraint on something it's 'unknown'. 

I'm a little confused as to what to change here, on the 1st comment you mention:

> remember that these nil account_id profiles are created during report import now as well if the profile didn't already exist.

but on the 2nd one you say:

> By the way, https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/concerns/xccdf/profiles.rb#L15 never imports profiles with non-null account_id

As far as I can see, during import, profiles are set the `account_id` found or created in the XccdfParser, which is the one from the b64_identity, the only case where we create Profiles without any `account_id` is during SSG import. Data seems to confirm this, as in production or QA we don't have any Profile like that, but on CI, where we have imported from the SSG, we do.

akofink:
I think there might be a misunderstanding of what import does wrt profiles. I'll reiterate that the Xccdf::Profiles code never creates profiles with an account. It's only for creating the profile as defined in the benchmark. A TestResult contains a ref_id of the profile tested against, and when the parser sees this, it clones one of the existing (already imported) profiles from the benchmark and assigns a host and account to it.

The two statements are the same - don't get confused with double negatives:

> remember that these nil account_id profiles are created during report import now as well if the profile didn't already exist.

> By the way, https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/concerns/xccdf/profiles.rb#L15 *never* imports profiles with *non-null* account_id

"Never imports profiles with non-null account_id" means "Always creates null account_id profiles." Like I said, profiles with account_ids and hosts are always cloned from an existing account with no hosts or account_id in our DB. I hope this clears things up a bit, but feel free to discuss on Slack.

PR #250, opened 2019-10-19
[WIP] Add limit, offset missing params to paths

Victoremepunto:
- patches rule_results, rules, systems paths

### DIFF
--- a/swagger/v1/openapi.v3.yaml
+++ b/swagger/v1/openapi.v3.yaml
@@ -156,6 +156,29 @@ paths:
           in: header
           schema:
             type: string
+        - name: limit
+          in: query
+          description: The number of items to return
+          required: false
+          schema:
+            type: integer
+            minimum: 1
+            maximum: 100
+            default: 10
+        - name: offset
+          in: query
+          description: The number of items to skip before starting to collect the result set.
+          required: false
+          schema:
+            type: integer
+            minimum: 0
+            default: 0
+        - name: search
+          in: query
+          description: "query string compliant with scoped_search query language: https://github.com/wvanbergen/scoped_search/wiki/Query-language"
+          required: false
+          schema:
+            type: string
       responses:
         "200":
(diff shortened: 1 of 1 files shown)


Victoremepunto:
> Fair enough, thanks @Victoremepunto
> We will have to keep track of these to add them to tests though, as when the new API is generated fron tests it'll have to contain tests for these fields

Yes, actually I don't know it i'll be necessary to generate specific tests to cover all cases, I think fuzzy tests generate tests from the spec, but I don't know if they replace or are complementary to the standard REST api test.

Victoremepunto:
> Looks good - is there no way to auto-generate this?

Actually, as I briefly commented a couple of times with @dLobatog , I think we should change towards a `design first` approach rather than our current `code first` approach regarding the REST API design. Auto-generated specs are okay when the REST API is something that must be delivered really quick and when the consumers are going to be very reduced, or when the API is going to be super limited and simplistic, which I don't think is our case. 

Actually I think it could be great if we can discuss this openly one of these days. I plan to add the REST API client and keep adding changes to this spec, but it requires a lot of time and probably I won't leave it finished. The REST API must be understood as a contract we need to comply to, and hence it shouldn't be changed frequently ( consumers of the API won't be happy about it if we release new versions relatively often ). As I said, this is a topic may be worthy to discuss about it. 

If any of you want to read more about API design differences you can go through [this small article](https://swagger.io/blog/api-design/design-first-or-code-first-api-development/). 

If want to go further, I'm actually enjoying [this book](https://www.manning.com/books/irresistible-apis), even if it's a little dated, I find it pretty useful.

akofink:
I'm fine with design first. Seems to make more sense to ship a stable API definition. Perhaps the auto-generated API doc can serve to help keep us honest in the future, at least.

akofink:
Closing in favor of https://github.com/RedHatInsights/compliance-backend/pull/325

PR #254, opened 2019-10-22
fix stable branch conflicts

Victoremepunto:
@akofink, @dLobatog  please check


Victoremepunto:
> This is a fast-forward, then? Looks good to me. ACK

Yes, I do wanted to review the changes (I'm not sure about if these are the changes that need to be actually merged to stable). Regarding FF , it is, and it _should_ stay like that on further merges from master to stable.

PR #255, opened 2019-10-24
Add a phony default benchmark to profiles + rules

akofink:

### DIFF
--- a/db/migrate/20190924171312_add_benchmark_id_to_profile.rb
+++ b/db/migrate/20190924171312_add_benchmark_id_to_profile.rb
@@ -1,5 +1,13 @@
 class AddBenchmarkIdToProfile < ActiveRecord::Migration[5.2]
+  PHONY_BENCHMARK = Xccdf::Benchmark.find_or_create_by!(
+    ref_id: 'phony_ref_id',
+    version: 'v0.0.0',
+    title: 'phony title',
+    description: 'phony description'
+  )
+
   def change
-    add_column :profiles, :benchmark_id, :uuid, null: false
+    add_column :profiles, :benchmark_id, :uuid, null: false,
+      default: PHONY_BENCHMARK.id
   end
 end
--- a/db/migrate/20190924173755_delete_policies.rb
+++ b/db/migrate/20190924173755_delete_policies.rb
@@ -1,5 +1,14 @@
 class DeletePolicies < ActiveRecord::Migration[5.2]
-  def change
+  def up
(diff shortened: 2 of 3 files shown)


db/migrate/20190924173755_delete_policies.rb
```diff
@@ -1,5 +1,14 @@
 class DeletePolicies < ActiveRecord::Migration[5.2]
-  def change
+  def up
     drop_table :policies
```
akofink: This is not an automatically reversible migration and so requires up/down rather than change if we want to support rollback.

db/migrate/20190924171312_add_benchmark_id_to_profile.rb
```diff
@@ -1,5 +1,11 @@
 class AddBenchmarkIdToProfile < ActiveRecord::Migration[5.2]
+  PHONY_BENCHMARK = Xccdf::Benchmark.find_or_create_by(
+    ref_id: 'phony',
+    version: 'v0.0.0'
```
dLobatog: `null value in column "title" violates not-null constraint`

dLobatog: Also `description`

db/migrate/20190924171312_add_benchmark_id_to_profile.rb
```diff
@@ -1,5 +1,11 @@
...
+  )
+
   def change
-    add_column :profiles, :benchmark_id, :uuid, null: false
+    add_column :profiles, :benchmark_id, :uuid, null: false,
+      default: PHONY_BENCHMARK.id
```
dLobatog: It still fails with `PG::NotNullViolation: ERROR:  column "benchmark_id" contains null values`.
I think you have to `Profile.update_all(benchmark_id: PHONY_BENCHMARK.id)` before you even run the change

dLobatog: Unfortunately I get `PG::NotNullViolation: ERROR:  column "benchmark_id" contains null values
: ALTER TABLE "profiles" ALTER COLUMN "benchmark_id" TYPE uuid, ALTER COLUMN "benchmark_id" SET DEFAULT '5e791411-38aa-4ff0-9618-48cf355ea5a0', ALTER COLUMN "benchmark_id" SET NOT NULL` - as you can see, the default is a valid benchmark ID

PR #256, opened 2019-10-25
Import remediations for rules with profiles only

dLobatog:
Currently there's a problem in `import_remediations`. It looks like the job is failing on CI because some rules do not have profiles, yet we use these profiles to build the request we send to remediations. See the inline comment.


app/services/remediations_api.rb
```diff
@@ -9,20 +9,29 @@ def initialize(account)
...
+    Platform.connection.post(@url) do |req|
+      req.headers['X-RH-IDENTITY'] = @b64_identity
+      req.headers['Content-Type'] = 'application/json'
+      req.body = { 'issues': build_issues_list(rules) }.to_json
+    end
+  end
+
+  def rules_with_profiles
+    Rule.find_in_batches(batch_size: 100).select do |rules|
+      rules.map { |rule| rule.profiles.empty? }
```
dLobatog: Notice in Line 41, we call `"ssg:rhel7|#{remove_ref_id_prefix(profile_rules[rule.id])}"` - this method fails if any of the rules does not have a profile.

dLobatog: The profile is currently required by remediations to make the request.

app/services/remediations_api.rb
```diff
@@ -9,20 +9,27 @@ def initialize(account)
...
+    Platform.connection.post(@url) do |req|
+      req.headers['X-RH-IDENTITY'] = @b64_identity
+      req.headers['Content-Type'] = 'application/json'
+      req.body = { 'issues': build_issues_list(rules) }.to_json
+    end
+  end
+
+  def rules_with_profiles
+    Rule.all.reject { |rule| rule.profiles.empty? }
+  end
```
akofink: This should be a scope on Rule IMO that returns an AR::Relation. Also, `Rule.all` is a Bad Idea. Something like `Rule.with_profiles`.

app/services/remediations_api.rb
```diff
@@ -9,20 +9,27 @@ def initialize(account)
...
 
   def import_remediations
-    Rule.find_in_batches(batch_size: 100).each do |rules|
-      response = Platform.connection.post(@url) do |req|
-        req.headers['X-RH-IDENTITY'] = @b64_identity
-        req.headers['Content-Type'] = 'application/json'
-        req.body = { 'issues': build_issues_list(rules) }.to_json
-      end
-      update_rules(remediations_available(response))
+    rules_with_profiles.each_slice(100) do |rules|
```
akofink: With an AR::Relation, you could use `find_each`

app/services/remediations_api.rb
```diff
@@ -9,20 +9,27 @@ def initialize(account)
...
+      update_rules(remediations_available(remediations_request(rules)))
     end
   rescue Faraday::ClientError => e
     Rails.logger.info("#{e.message} #{e.response}")
   end
 
   private
 
+  def remediations_request(rules)
```
akofink: Weird that the method is named `...request` when it actually returns a response.

app/services/remediations_api.rb
```diff
@@ -9,20 +9,27 @@ def initialize(account)
...
+  def remediations_response(rules)
+    Platform.connection.post(@url) do |req|
+      req.headers['X-RH-IDENTITY'] = @b64_identity
+      req.headers['Content-Type'] = 'application/json'
+      req.body = { 'issues': build_issues_list(rules) }.to_json
+    end
+  end
+
+  def rules_with_profiles
+    Rule.all.reject { |rule| rule.profiles.empty? }
```
akofink: I think you meant to update this one to use the new scope, eh?

akofink: It seems to work fine with the new scope here.

PR #257, opened 2019-10-28
RHICOMPL-354 - GraphQL index benchmarks

dLobatog:
This API is currently used on https://github.com/RedHatInsights/compliance-backend/pull/276 to fetch the profiles for a benchmark.


app/policies/profile_policy.rb
```diff
@@ -17,7 +17,9 @@ def update?
   # Only show hosts in our user account
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      only_matching_account
+      return scope.where('1=0') if user.account_id.blank?
+
+      scope.where(account_id: [nil, user.account_id])
```
dLobatog: I'm not overly convinced about this, but we need to be able to fetch the canonical profiles somehow

akofink: Agreed - it's not very intuitive; if we added a `canonical_profile_id` attribute it would make a lot more sense IMO.

akofink:
This seems good to go to me. Any reason it's still a draft?

app/models/profile.rb
```diff
@@ -17,7 +17,6 @@ class Profile < ApplicationRecord # rubocop:disable Metrics/ClassLength
   validates :name, presence: true
   validates :benchmark_id, presence: true
   validates :compliance_threshold, numericality: true
-  validates :account, absence: true, unless: -> { hosts.any? }
```
dLobatog: Had to remove this validation because technically you should be able to create policies without any hosts according to the UX flow - see https://github.com/RedHatInsights/compliance-frontend/pull/197 for context

app/graphql/types/benchmark.rb
```diff
@@ -0,0 +1,19 @@
...
+  # Definition of Benchmark as a GraphQL type
+  class Benchmark < Types::BaseObject
+    graphql_name 'Benchmark'
+    description 'A representation of a SCAP Security Guide version'
+
+    field :id, ID, null: false
+    field :title, String, null: false
+    field :ref_id, String, null: false
+    field :version, String, null: false
+    field :profiles, [::Types::Profile], null: true
```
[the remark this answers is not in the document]
dLobatog: Done

Victoremepunto:

app/graphql/types/profile.rb
```diff
@@ -11,6 +11,8 @@ class Profile < Types::BaseObject
     field :description, String, null: true
     field :ref_id, String, null: false
     field :compliance_threshold, Float, null: false
+    field :benchmark_id, ID, null: false
+    field :account_id, ID, null: false
```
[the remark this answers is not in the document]
dLobatog: It's a few lines below, wasn't moved

app/policies/benchmark_policy.rb
```diff
@@ -0,0 +1,20 @@
...
+  def index?
+    true
+  end
+
+  def show?
+    true
+  end
+
+  # Only select Rules belonging to profiles visible by the current user
+  # (profile account ID = rule account ID)
```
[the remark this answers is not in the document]
dLobatog: Definitely, it came from the rules policy, thanks for the catch

app/graphql/types/mutation.rb
```diff
@@ -8,6 +8,8 @@ class Mutation < Types::BaseObject
     description 'The mutation root of this schema'
 
     field :updateProfile, mutation: Mutations::Profile::Edit
+    field :createProfile, mutation: Mutations::Profile::Create
+    field :associateSystems, mutation: Mutations::Profile::AssociateSystems
```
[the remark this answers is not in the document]
dLobatog: That sounds reasonable, the only issue I see is that it creates a sort of N+1 issue until we merge graphql-batching (we have to `SELECT * FROM profiles WHERE id = whatever` for every ProfileHost create mutation in that case, N hosts...)

app/graphql/schema.graphql
```diff
@@ -1,3 +1,36 @@
+"""
+A representation of a SCAP Security Guide version
+"""
+type Benchmark implements Node {
```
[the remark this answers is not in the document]
dLobatog: I would say so, can we have this in a separate JIRA ticket & PR?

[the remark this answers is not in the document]
akofink: Yes, of course :)

app/policies/profile_policy.rb
```diff
@@ -3,11 +3,11 @@
 # Policies for accessing Profiles
 class ProfilePolicy < ApplicationPolicy
   def index?
-    match_account?
+    match_account? || record.account_id.blank?
```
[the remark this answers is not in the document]
dLobatog: Fixed

app/policies/benchmark_policy.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# Policies for accessing Benchmarks
+class BenchmarkPolicy < ApplicationPolicy
+  def index?
+    true
```
akofink: I raised this on the wizard frontend PR, but today we allow users to upload any random SSG version and benchmark. This would allow users from different accounts to see and use those benchmarks, which I don't think is what we want. It seems we may need to associate some benchmarks with accounts in the future (we don't need to fix this in this PR necessarily).

app/graphql/mutations/profile/create.rb
```diff
@@ -0,0 +1,50 @@
...
+      argument :description, String, required: false
+      argument :business_objective_id, String, required: false
+      argument :compliance_threshold, Float, required: false
+      field :profile, Types::Profile, null: false
+
+      def resolve(args = {})
+        original_profile = find_original_profile(args[:clone_from_profile_id])
+        profile = ::Profile.new(new_profile_options(args))
+        profile.save
+        profile.add_rules_from(profile: original_profile)
```
dLobatog: I tried for a bit reusing `clone_to` here, but it was more complicated than I thought. 
What I did was changing `clone_to` to take in a hash of all possible arguments to Profile (in this case `new_profile_options`). This worked for the regular use case of parsing XCCDF reports. However `clone_to` expects that either the cloned profile or the clonee is `nil` to update the attributes (https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/profile.rb#L92 , I changed it to use all attributes).    

In any case, the suggestion was good, it allowed me to remove the method I had in the original commit which was almost a copy of `Profile#add_rules_from` :sweat:  thanks for noticing that

akofink: So, correct me if I'm wrong, but this mutation will raise a 500 error if the user tries to create the same profile twice, right? I guess `clone_to` is meant to try and find an existing profile first before creating a new one, and that's not functionality we want in this mutation.

akofink:
@dLobatog Another thing I thought of (since I'm working on creation of tailoring files), we need to be able to point back to the parent profile. A Tailoring element has a `<Profile id='<ref id>' extends='<parent ref_id>' />`. This may (finally) be the time to add a parent_profile_id column to the profiles table.

app/graphql/mutations/profile/create.rb
dLobatog: It raised a 500, correct - I have changed `save` to `save!`- this returns a 422 if you try to create the same profile twice. I have also seen we can catch these with:
```
.catch(
            (err) => {
                <ErrorPage error={err}/>;
            }
        );
```
on the mutation on the wizard, which I'm working on

app/policies/benchmark_policy.rb
dLobatog: That's a good point, we should think about this though, because SSG should be visible to everyone but custom benchmarks should not...

PR #260, opened 2019-11-05
Relay GraphQL (pagination)

dLobatog:
This allows to query connections so that:

```
{
    systems(first: 10, after: "MQ", search: "example") {
        totalCount
        pageInfo {
            startCursor
            endCursor
            hasNextPage
            hasPreviousPage
        },
        edges {
          cursor
          node {
            id
            name
          }
        }
    }
}
```

```
{
  "data": {
    "systems": {
      "totalCount": 200,
      "pageInfo": {
        "startCursor": "Mg",
        "endCursor": "MTE",
        "hasNextPage": true,
        "hasPreviousPage": false
      },
      "edges": [
        {
          "cursor": "Mg",
          "node": {
            "id": "QmFzZU9iamVjdC0wM2YwYTlhNS1jN2U2LTRhZTItOTIyYS00ZWNkMTE5ZmY3NmU=",
            "name": "example-1"
          }
        },
        {
          "cursor": "Mw",
          "node": {
            "id": "QmFzZU9iamVjdC0xOWI0NDk0ZS1jYWQxLTQ1NDEtYmQ4ZS0yMzIyOGU1YWU3NGM=",
            "name": "example-2"
          }
        },
        {
          "cursor": "NA",
          "node": {
            "id": "QmFzZU9iamVjdC0zMjE3OWFmOS03NmEwLTRjZmQtOTJjZC0wM2Y5ODZmYjE0OGM=",
            "name": "example-3"
          }
        },
        {
          "cursor": "NQ",
          "node": {
            "id": "QmFzZU9iamVjdC1mMTczZThjMy02MjRkLTQzZDEtOTQ0ZC1jYjFlN2YwN2MwMWM=",
            "name": "example-4"
          }
        },
        {
          "cursor": "Ng",
          "node": {
            "id": "QmFzZU9iamVjdC1jNjUzOTRiMS1iMjE2LTQwYWEtODNiZS1kZjMzOGNjNzg4Yjk=",
            "name": "example-5"
          }
        },
        {
          "cursor": "Nw",
          "node": {
            "id": "QmFzZU9iamVjdC1mNmRmZjhjYy04Y2MxLTQxNjktYTI4Mi1kNTYxMGJlZGM3MzY=",
            "name": "example-6"
          }
        },
        {
          "cursor": "OA",
          "node": {
            "id": "QmFzZU9iamVjdC0xMjViMmZhMC03Y2RlLTQ0MWMtYWY2OS0yNGRiODRjZTdkMzg=",
            "name": "example-7"
          }
        },
        {
          "cursor": "OQ",
          "node": {
            "id": "QmFzZU9iamVjdC1iZTgxZWQ1OC0wZDVjLTQxNDctODUzYy1hYTVlYTQ0ODgzYTg=",
            "name": "example-8"
          }
        },
        {
          "cursor": "MTA",
          "node": {
            "id": "QmFzZU9iamVjdC1lOWFjOGY1YS04Yzc1LTQ2ZGMtYjJkZi04OTc5MzA1OGU4NGM=",
            "name": "example-9"
          }
        },
        {
          "cursor": "MTE",
          "node": {
            "id": "QmFzZU9iamVjdC0zNzVjNjkwZi1jOWUwLTQxOTctYTYyNy0zZDFkMjk0NDFlMWU=",
            "name": "example-10"
          }
        }
      ]
    }
  }
}
```

This is missing tests, especially for authorization scopes

### DIFF
--- a/app/graphql/resolvers/concerns/collection.rb
+++ b/app/graphql/resolvers/concerns/collection.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+module Resolvers
+  module Concerns
+    # Module for Collection GraphQL fields, contains usual pagination options
+    module Collection
+      extend ActiveSupport::Concern
+
+      included do
+        type ["Types::#{self::MODEL_CLASS}".safe_constantize], null: false
+        argument :search, String, 'Search query', required: false
+        argument :limit, Integer, 'Pagination limit', required: false
+        argument :offset, Integer, 'Pagination offset', required: false
+      end
+
+      def resolve(**kwargs)
+        filters = filter_list(**kwargs)
+        base_scope = authorized_scope
+        filters.reduce(base_scope) { |scope, filter| filter.call(scope) }
+      end
+
+      private
+
+      def filter_list(search: nil, offset: nil, limit: nil)
+        filters = []
+        filters << search_filter(search: search) if search.present?
+
+        if offset.present? || limit.present?
+          filters << pagination_filter(offset: offset, limit: limit)
+        end
+
+        filters
+      end
+
+      def pagination_filter(offset:, limit:)
+        lambda do |scope|
+          begin
+            scope.paginate(page: offset, per_page: limit)
+          rescue ScopedSearch::QueryNotSupported => e
+            raise GraphQL::ExecutionError, e.message
+          end
+        end
+      end
+
+      def search_filter(search:)
+        lambda do |scope|
+          begin
+            scope.search_for(search)
+          rescue ScopedSearch::QueryNotSupported => e
+            raise GraphQL::ExecutionError, e.message
+          end
+        end
+      end
+
+      def model_class
+        self.class::MODEL_CLASS
+      end
+
+      def authorized_scope
+        Pundit.policy_scope(context[:current_user], model_class)
+      end
+    end
+  end
+end
--- a/app/graphql/resolvers/concerns/record.rb
+++ b/app/graphql/resolvers/concerns/record.rb
@@ -1,0 +1,17 @@
+module Resolvers
+  module Concerns
+    module Record
+      extend ActiveSupport::Concern
+
+      included do
+        type "Types::#{self::MODEL_CLASS}".safe_constantize, null: false
+
+        argument :id, String, 'Global ID for Record', required: true
+      end
+
+      def resolve(id:)
+        RecordLoader.for(self.class::MODEL_CLASS).load_by_global_id(id)
+      end
+    end
+  end
+end
--- a/app/graphql/resolvers/concerns/record.rb
+++ b/app/graphql/resolvers/concerns/record.rb
@@ -1,5 +1,8 @@
+# frozen_string_literal: true
+
 module Resolvers
   module Concerns
+    # Contains fields and methods related with GraphQL Relay record field
     module Record
       extend ActiveSupport::Concern
 
@@ -10,7 +13,7 @@ module Resolvers
       end
 
       def resolve(id:)
-        RecordLoader.for(self.class::MODEL_CLASS).load_by_global_id(id)
+        ::RecordLoader.for(self.class::MODEL_CLASS).load_by_global_id(id)
       end
     end
   end
--- a/app/graphql/resolvers/generic.rb
+++ b/app/graphql/resolvers/generic.rb
@@ -1,0 +1,27 @@
+module Resolvers
+  class Generic
+    def initialize(type)
+      @type = type
+    end
+
+    def self.for(type)
+      new(type)
+    end
+
+    def collection
+      return unless model_class
+      base_class.include(Resolvers::Concerns::Collection)
+    end
+
+    private
+
+    attr_reader :type
+    delegate :model_class, to: :type
+
+    def base_class
+      Class.new(Resolvers::BaseResolver).tap do |c|
+        c.const_set('MODEL_CLASS', model_class)
+      end
+    end
+  end
+end
--- a/app/graphql/resolvers/generic.rb
+++ b/app/graphql/resolvers/generic.rb
@@ -1,5 +1,10 @@
+# frozen_string_literal: true
+
 module Resolvers
+  # Base class for resolvers
   class Generic
+    attr_reader :type
+
     def initialize(type)
       @type = type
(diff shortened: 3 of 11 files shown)


dLobatog:
Latest change also adds offset/limit so that queries can be like:

```
{
    systems(limit: 100, offset: 2) {
        totalCount
        pageInfo {
            startCursor
            endCursor
            hasNextPage
            hasPreviousPage
        },
        edges {
          node {
            id
            name
          }
        }
    }
}
```

app/graphql/types/system.rb
```diff
@@ -3,12 +3,13 @@
 module Types
   # Definition of the System GraphQL type
   class System < Types::BaseObject
+    model_class ::Host
```
bastilian: Can we not rename `System` to `Host` or the other way around?

app/graphql/resolvers/generic.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def collection
+      return unless model_class
+      base_class.include(Resolvers::Concerns::Collection)
+    end
+
+    private
+
+    attr_reader :type
```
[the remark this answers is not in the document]
bastilian: The position is fine if it has to be kept `private`, but has it to be private?

dLobatog:
Requires https://github.com/RedHatInsights/compliance-frontend/pull/214

app/graphql/types/system.rb
dLobatog: We can definitely do that in some other PR - the reasoning behind this is that a while ago, there were meant to be two classes of systems, `Host` and `Imagestream` - but that's been delayed for a while

app/graphql/types/base_object.rb
```diff
@@ -2,7 +2,24 @@
 
 module Types
   # Definition of the BaseObject type in GraphQL
-  class BaseObject < GraphQL::Schema::Object
+  class BaseObject < GraphQL::Types::Relay::BaseObject
```
akofink:
```suggestion
  class BaseObject < Relay::BaseObject
```

app/graphql/resolvers/concerns/record.rb
```diff
@@ -0,0 +1,17 @@
...
+      extend ActiveSupport::Concern
+
+      included do
+        type "Types::#{self::MODEL_CLASS}".safe_constantize, null: false
+
+        argument :id, String, 'Global ID for Record', required: true
+      end
+
+      def resolve(id:)
+        RecordLoader.for(self.class::MODEL_CLASS).load_by_global_id(id)
```
akofink: Nit: I wonder if we need to ensure all these constants in a concern are fully qualified (i.e. `::RecordLoader`). It's caused problems in the past with models, but idk if it's a special case with Rails objects.

app/graphql/resolvers/generic.rb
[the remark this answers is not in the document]
akofink: @dLobatog yeah, I really don't understand why we're defining a private `attr_reader`. I don't think that makes any sense, since anything within this class will already have access to the instance variable `@type`. You can `delegate :model_class, to: :@type`

app/graphql/types/query.rb
```diff
@@ -6,6 +6,20 @@ class Query < Types::BaseObject
...
+      def collection_field(name, type)
+        field name, type.connection_type, null: false, resolver: Resolvers::Generic.for(type).collection
+      end
+    end
+
+    collection_field :systems, Types::System
+    collection_field :profiles, Types::Profile
+    record_field :profile, Types::Profile
+
     field :all_systems, [Types::System], null: true do
```
akofink: Shouldn't we remove this? Or is this for backwards compatibility?

akofink:
There are some rubocop issues here

app/graphql/resolvers/concerns/collection.rb
```diff
@@ -0,0 +1,61 @@
...
+      extend ActiveSupport::Concern
+
+      included do
+        type ["Types::#{self::MODEL_CLASS}".safe_constantize], null: false
+        argument :search, String, 'Search query', required: false
+        argument :limit, Integer, 'Pagination limit', required: false
+        argument :offset, Integer, 'Pagination offset', required: false
+      end
+
+      def resolve(**kwargs)
```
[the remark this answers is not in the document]
dLobatog: They don't get called in that case, `resolve` is called on the module directly apparently, and the arguments are included in the base class through `included`.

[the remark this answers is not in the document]
akofink: Ok! Weird :)

app/graphql/types/system.rb
```diff
@@ -30,6 +31,15 @@ class System < Types::BaseObject
                required: false
     end
 
+    def profiles(lookahead:)
+      if %i[rulesPassed rulesFailed compliant lastScanned].any? do |option|
+        lookahead.selects?(option)
+      end
+        context[:parent_system_id] = object.id
+      end
```
akofink: What's with this extra `end` ?

app/graphql/types/profile.rb
```diff
@@ -68,19 +69,23 @@ def total_host_count
       object.hosts.count
     end
 
-    def compliant(system_id:)
+    def compliant(args = {})
+      system_id = args[:system_id] || context[:parent_system_id]
```
akofink: It seems like you should just go ahead and make this a private method with a cached (`||=`) instance variable.

app/graphql/types/query.rb
dLobatog: Yeah, I would like to keep these until we have this in prod so we can phase out the old API when the new one is already used everywhere.

app/graphql/types/system.rb
dLobatog: This is the `end`  of the `if`, the one above is the `do end` block. I'll move it to a separate method for clarity

app/graphql/types/profile.rb
dLobatog: I can make it a private method, but I don't think caching is appropriate here because you may use the argument on certain methods, and rely on context for others.

akofink: Gotcha, that makes sense

PR #266, opened 2019-11-19
RHICOMPL-420 - Introduce uniqueness constraint to RuleReferencesRules

dLobatog:


app/graphql/types/rule.rb
```diff
@@ -30,7 +30,7 @@ def references(lookahead:)
                          ::RuleReference.column_names.map(&:to_sym)
       object.references.pluck(selected_columns).map do |reference|
         selected_columns.zip([reference].flatten).to_h
-      end
+      end.uniq
```
akofink: I'd prefer we used `distinct` in the query itself. `uniq` is on an array.

app/models/rule_references_rule.rb
```diff
@@ -4,4 +4,7 @@
 class RuleReferencesRule < ApplicationRecord
   belongs_to :rule_reference
   belongs_to :rule
+
+  validates :rule, presence: true
+  validates :rule_reference, presence: true, uniqueness: { scope: :rule }
```
akofink: :+1: Should we add a DB index?

app/graphql/types/rule.rb
```diff
@@ -28,7 +28,10 @@ def compliant(system_id:)
     def references(lookahead:)
       selected_columns = lookahead.selections.map(&:name) &
                          ::RuleReference.column_names.map(&:to_sym)
-      object.references.pluck(selected_columns).map do |reference|
+      qualified_columns = selected_columns.map { |c| "rule_references.#{c}" }
+      object.references.pluck(
+        "DISTINCT #{qualified_columns.join(', ')}"
```
akofink: There's got to be a way to do this with ActiveRecord. I think it's `.distinct(selected_columns)`

dLobatog: @akofink I had to add the qualified name here, because `.distinct.pluck(selected_columns)` would not work, unless I qualify the columns, and if I do that, then `selected_columns.zip([reference].flatten).to_h` returns something that is not expected by the frontend.

dLobatog: Just `distinct` or `distinct(selected_columns)` work. But then you cannot `pluck` the result - the arguments you pass to `pluck` from an AssociationRelation have to be fully qualified.

```
[9] pry(#<Types::Rule>)> object.references.class
=> RuleReference::ActiveRecord_Associations_CollectionProxy
[10] pry(#<Types::Rule>)> object.references.distinct.class
=> RuleReference::ActiveRecord_AssociationRelation
```

So for example, if you're thinking of this, `pluck` won't run unless the columns are fully qualified:  
```
[12] pry(#<Types::Rule>)> object.references.distinct.pluck(selected_columns)
DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): [:label, :href]. Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). (called from references at (pry):7)
   (1.2ms)  SELECT DISTINCT label, href FROM "rule_references" INNER JOIN "rule_references_rules" ON "rule_references"."id" = "rule_references_rules"."rule_reference_id" WHERE "rule_references_rules"."rule_id" = $1 ORDER BY "rule_references"."label" ASC  [["rule_id", "7fa6c199-ff7f-4d75-b608-ab5465675d97"]]
  ↳ (pry):7
ActiveRecord::StatementInvalid: PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: ...E "rule_references_rules"."rule_id" = $1 ORDER BY "rule_refe...
                                                             ^
: SELECT DISTINCT label, href FROM "rule_references" INNER JOIN "rule_references_rules" ON "rule_references"."id" = "rule_references_rules"."rule_reference_id" WHERE "rule_references_rules"."rule_id" = $1 ORDER BY "rule_references"."label" ASC
from /usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/postgresql_adapter.rb:611:in `async_exec_params'
Caused by PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: ...E "rule_references_rules"."rule_id" = $1 ORDER BY "rule_refe...
                                                      
```

akofink: Hmm... I think the default order is messing this up. I wonder if `.reorder('')` would fix it?

dLobatog: That works too - I'll change it to that, it's the same query in any case, for some reason ActiveRecord adds that problematic ORDER_BY unless you remove it with that `reorder('')`, I was not aware of that

dLobatog:
Merging since it was approved - the only change is including the suggestion of `object.references.distinct.reorder('').pluck(selected_columns) ` instead of `object.references.pluck("DISTINCT #{qualified_columns.join(', ')}")

app/graphql/types/rule.rb
akofink: It's the pesky default scope: https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/rule_reference.rb#L5

PR #268, opened 2019-11-20
Add parsed host to pre-existing profile in account

dLobatog:
Draft for now as I'm writing tests and checking other parts of the parser that can be improved.

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -104,6 +104,8 @@ class Profile < ApplicationRecord
     new_profile = in_account(account)
     if new_profile.nil?
       (new_profile = dup).update!(account: account, hosts: [host])
+    else
+      new_profile.hosts << host unless new_profile.hosts.include?(host)
     end
     new_profile.add_rules_from(profile: self)
     new_profile
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -79,4 +79,33 @@ class ProfileTest < ActiveSupport::TestCase
     profiles(:one).update(business_objective: nil)
     assert_empty BusinessObjective.where(title: 'abcd')
   end
+
+  context 'cloning profile to account' do
+    should 'create host relation when the profile is created' do
+      assert_difference('ProfileHost.count', 1) do
+        cloned_profile = profiles(:one).clone_to(
+          account: accounts(:one), host: hosts(:two)
+        )
+        assert hosts(:one).profiles.include?(cloned_profile)
+      end
+    end
+
+    should 'create host relation even if profile is already created' do
+      assert_difference('ProfileHost.count', 1) do
+        cloned_profile = profiles(:two).clone_to(
+          account: accounts(:one), host: hosts(:one)
+        )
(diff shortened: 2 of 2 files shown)


app/models/profile.rb
```diff
@@ -106,6 +106,8 @@ def clone_to(account: nil, host: nil)
     new_profile = in_account(account)
     if new_profile.nil?
       (new_profile = dup).update!(account: account, hosts: [host])
```
dLobatog: When the profile is the *first* profile of its kind in an account, we set the host properly up here. But in case the profile already existed in the account, nothing sets it.

app/models/profile.rb
```diff
@@ -106,6 +106,8 @@ def clone_to(account: nil, host: nil)
...
+    else
+      new_profile.hosts << host unless new_profile.hosts.include?(host)
```
dLobatog: ` unless new_profile.hosts.include?(host)` is required in order not to fail with `ActiveRecord::RecordInvalid: Validation failed: Host has already been taken`

test/models/profile_test.rb
```diff
@@ -79,4 +79,33 @@ class ProfileTest < ActiveSupport::TestCase
     profiles(:one).update(business_objective: nil)
     assert_empty BusinessObjective.where(title: 'abcd')
   end
+
+  context 'cloning profile to account' do
+    should 'create host relation when the profile is created' do
+      assert_difference('ProfileHost.count', 1) do
+        cloned_profile = profiles(:one).clone_to(
```
dLobatog: Perhaps it's a bit unclear if you don't see the `setup` method in this file - `profiles(:one)` has `hosts(:one)`, but not `hosts(:two)`

test/models/profile_test.rb
```diff
@@ -79,4 +79,33 @@ class ProfileTest < ActiveSupport::TestCase
...
+          account: accounts(:one), host: hosts(:two)
+        )
+        assert hosts(:one).profiles.include?(cloned_profile)
+      end
+    end
+
+    should 'create host relation even if profile is already created' do
+      assert_difference('ProfileHost.count', 1) do
+        cloned_profile = profiles(:two).clone_to(
```
dLobatog: `profiles(:two)` doesn't have any host

test/models/profile_test.rb
```diff
@@ -79,4 +79,33 @@ class ProfileTest < ActiveSupport::TestCase
...
+          account: accounts(:one), host: hosts(:one)
+        )
+        assert hosts(:one).profiles.include?(cloned_profile)
+      end
+    end
+
+    should 'not create host relation if host is already in profile' do
+      assert_difference('ProfileHost.count', 0) do
+        cloned_profile = profiles(:one).clone_to(
```
dLobatog: In this case, nothing should be done - `profiles(:one)` is already assigned to `hosts(:one)`

test/models/profile_test.rb
bastilian: Asserting that `host(:two)` is not yet included before `clone_to` would make that more obvious.

PR #269, opened 2019-11-20
[RHICOMPL-412] - Improve parsing performance by searching pre-existing rules in bulk

dLobatog:
Before this commit, every time a new report came in, the report would
look for every Rule, RuleIdentifier, etc... by iterating over all of
them and calling SQL on every rule. On top of instantiating a lot of
objects, this is much slower than the alternative we have.

Instead, we can search in bulk which of the Rule, RuleIdentifiers,
etc... have already been saved in DB. If they have already been saved,
then skip saving them (which involved finding them, instantiating the
object and then saving them).

```
sidekiq_1           | 2019-11-20T18:01:30.798Z 1 TID-gshcdmt95 ParseReportJob JID-2c5e207599e41c6b3f43cac0 INFO: start
sidekiq_1           | 2019-11-20T18:01:33.781Z 1 TID-gshcdmt95 ParseReportJob JID-2c5e207599e41c6b3f43cac0 INFO: done: 2.983 sec
sidekiq_1           | 2019-11-20T18:02:20.823Z 1 TID-gshd3wlw9 ParseReportJob JID-49e9604a22cbce9ff3357a10 INFO: start
sidekiq_1           | 2019-11-20T18:02:22.852Z 1 TID-gshd3wlw9 ParseReportJob JID-49e9604a22cbce9ff3357a10 INFO: done: 2.029 sec
```

### DIFF
--- a/app/models/rule_identifier.rb
+++ b/app/models/rule_identifier.rb
@@ -7,11 +7,20 @@ class RuleIdentifier < ApplicationRecord
   validates :label, presence: true
   validates :system, presence: true
 
-  def self.from_openscap_parser(op_rule_identifier, rule_id)
-    if op_rule_identifier.label # rubocop:disable Style/GuardClause
-      find_or_initialize_by(label: op_rule_identifier.label,
-                            system: op_rule_identifier.system,
-                            rule_id: rule_id)
+  class << self
+    def from_openscap_parser(op_rule_identifier, rule_id)
+      if op_rule_identifier.label # rubocop:disable Style/GuardClause
+        find_or_initialize_by(label: op_rule_identifier.label,
+                              system: op_rule_identifier.system,
+                              rule_id: rule_id)
+      end
+    end
+
+    def preexisting_from_oscap_parser(new_rules)
+      ::RuleIdentifier.where(
+        label: new_rules.map { |r| r.op_source.identifier.label },
+        rule_id: new_rules.map(&:id)
+      )
     end
   end
 end
--- a/app/models/rule_reference.rb
+++ b/app/models/rule_reference.rb
@@ -18,16 +18,29 @@ class RuleReference < ApplicationRecord
                      message: 'and label combination already taken'
                    }
 
-  def self.find_from_oscap(oscap_references)
-    oscap_references.inject(where('1=0')) do |rel, reference|
-      rel.or(where(href: reference.href, label: reference.label))
+  class << self
+    def find_from_oscap(oscap_references)
+      oscap_references.inject(where('1=0')) do |rel, reference|
+        rel.or(where(href: reference.href, label: reference.label))
+      end
     end
-  end
 
-  def self.from_openscap_parser(op_rule_reference)
-    find_or_initialize_by(
-      href: op_rule_reference.href,
-      label: op_rule_reference.label
-    )
+    def from_openscap_parser(op_rule_reference)
+      find_or_initialize_by(
+        href: op_rule_reference.href,
+        label: op_rule_reference.label
+      )
+    end
+
+    def new_from_openscap_parser(op_rule_references)
+      preexisting_rule_references = ::RuleReference.where(
+        href: op_rule_references.map(&:href),
+        label: op_rule_references.map(&:label)
+      ).pluck(:href, :label)
+      op_rule_references.reject do |rule_reference|
+        preexisting_rule_references.include?([rule_reference.href,
+                                              rule_reference.label])
+      end
+    end
   end
 end
--- a/app/services/concerns/xccdf/rule_identifiers.rb
+++ b/app/services/concerns/xccdf/rule_identifiers.rb
@@ -7,7 +7,7 @@ module Xccdf
 
     included do
       def save_rule_identifiers
-        @rule_identifiers ||= @new_rules.map do |rule|
+        @rule_identifiers ||= new_rules_with_new_identifiers.map do |rule|
           ::RuleIdentifier.from_openscap_parser(rule.op_source.identifier,
                                                 rule.id)
         end.compact
@@ -15,10 +15,19 @@ module Xccdf
         ::RuleIdentifier.import!(new_rule_identifiers, ignore: true)
       end
 
+      def new_rules_with_new_identifiers
+        preexisting_identifiers = RuleIdentifier
+                                  .preexisting_from_oscap_parser(@new_rules)
+                                  .pluck(:label)
+        @new_rules.reject do |rule|
+          preexisting_identifiers.include?(rule.op_source.identifier.label)
+        end
+      end
+
       private
 
       def new_rule_identifiers
-        @new_rule_identifiers ||= @rule_identifiers.select(&:new_record?)
+        @rule_identifiers.select(&:new_record?)
       end
     end
   end
--- a/app/services/concerns/xccdf/rule_references.rb
+++ b/app/services/concerns/xccdf/rule_references.rb
@@ -7,7 +7,9 @@ module Xccdf
 
     included do
       def save_rule_references
-        @rule_references ||= @op_rule_references.map do |op_reference|
+        @rule_references ||= RuleReference.new_from_openscap_parser(
+          @op_rule_references
+        ).map do |op_reference|
           ::RuleReference.from_openscap_parser(op_reference)
         end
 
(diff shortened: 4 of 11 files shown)


app/services/concerns/xccdf/rule_identifiers.rb
```diff
@@ -7,18 +7,28 @@ module RuleIdentifiers
 
     included do
       def save_rule_identifiers
-        @rule_identifiers ||= @new_rules.map do |rule|
+        @rule_identifiers ||= new_rule_identifiers.map do |rule|
```
dLobatog: This tends to be the issue in all of these - `from_openscap_parser` is a new SQL call.. per record. What I'm doing in here is just making sure we only do these for *new* records, and skip pre-exisiting records.

docker-compose.yml
```diff
@@ -28,13 +28,16 @@ services:
   reports-consumer:
     image: compliance-backend-rails
     restart: on-failure
+    tty: true
```
dLobatog: I found this quite helpful when I was debugging and AFAIK it doesn't cause any problems to include it

test/services/concerns/xccdf/rules_test.rb
```diff
@@ -33,12 +33,19 @@ def initialize(report_contents)
     end
   end
 
-  test 'returns rules saved in the report' do
+  test 'returns only rules saved with the report' do
+    original_rule = ::Rule.from_openscap_parser(@mock.op_rules.sample)
```
dLobatog: The original test here does not match the behavior, previously all rules would be instantiated and saved in `@rules`, now only new ones are.

test/services/xccdf_report_parser_test.rb
```diff
@@ -220,17 +221,17 @@ class TestParser < ::XccdfReportParser
...
       ).save(validate: false)
 
+      assert_equal 367, @report_parser.op_rules.count
       assert_difference('Rule.count', 365) do
         @report_parser.save_rules
       end
 
-      @report_parser.rules = nil
-
+      # All rules already exist so none are saved
```
dLobatog: Setting `@report_parser.rules` will obviously make `save_rules` save 0 rules, so let's make the test a bit more difficult by removing it, so that we check that the parser does not import the same rules twice.

test/services/xccdf_report_parser_test.rb
```diff
@@ -220,17 +221,17 @@ class TestParser < ::XccdfReportParser
...
       assert_difference('Rule.count', 0) do
         @report_parser.save_rules
       end
 
-      assert_equal @report_parser.op_rules.count, @report_parser.rules.count
+      assert_equal @report_parser.op_rules.count - 2, @report_parser.rules.count
```
dLobatog: There are 2 pre-existing rules, so the total number of rules we import, and the total number of rules in general is `@report_parser.rules` - 365

app/services/concerns/xccdf/rule_results.rb
```diff
@@ -19,7 +19,10 @@ def save_rule_results
     private
 
     def rule_ids
-      @rule_ids ||= @rules.map { |r| [r.ref_id, r.id] }.to_h
+      @rule_ids = ::Rule.where(
+        ref_id: selected_rule_results.map(&:id),
+        benchmark_id: @benchmark&.id
+      ).pluck(Arel.sql('ref_id, id')).to_h
```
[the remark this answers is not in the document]
dLobatog: It's the same, but [:ref_id, :id] throws an ugly warning:
```
DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): [:label, :href]. Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). 
```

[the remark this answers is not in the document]
akofink: `pluck([:ref_id, :id])`, sure - but what about `pluck(:ref_id, :id)` ?

akofink:
A related approach to speed up subsequent parsing: https://github.com/RedHatInsights/compliance-backend/pull/270

test/services/xccdf_report_parser_test.rb
dLobatog: That's a good point - I actually changed it at a point when I removed the `||=` cache, I reverted that change but not this test chnage

test/services/concerns/xccdf/rules_test.rb
```diff
@@ -33,12 +33,19 @@ def initialize(report_contents)
...
+    assert ::Rule.where(ref_id: original_rule.ref_id).present?
+  end
+
+  test 'does not return rules in the report that were saved previously' do
     rule = Rule.from_openscap_parser(@mock.op_rules.sample,
                                      benchmark_id: @mock.benchmark.id)
     assert rule.save
     @mock.save_rules
-    assert_includes @mock.rules, rule
+    assert_not_includes @mock.rules, rule
```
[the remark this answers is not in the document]
akofink: Yes, I agree this should be configurable, probably as an option or attribute on the `Xccdf` mixins. Ideally `import_datastream` would pretty much always try and parse the entire file, as time isn't really as important as accuracy here; during report import, though, I think we never want to import the entire benchmark (untrusted user input and all that) if we can help it.

app/models/rule_identifier.rb
```diff
@@ -7,11 +7,20 @@ class RuleIdentifier < ApplicationRecord
...
+                              system: op_rule_identifier.system,
+                              rule_id: rule_id)
+      end
+    end
+
+    def preexisting_from_oscap_parser(new_rules)
+      ::RuleIdentifier.where(
+        label: new_rules.map { |r| r.op_source.identifier.label },
+        rule_id: new_rules.map(&:id)
+      )
```
akofink: This query is not correct and should be updated to use `RuleIdentifier.where(label: 'one label', rule_id: 'one rule id').or(RuleIdentifier.where(...)).or(...)` (constructed with `inject` like I showed before)

app/models/rule_reference.rb
```diff
@@ -18,16 +18,29 @@ class RuleReference < ApplicationRecord
...
                    }
 
-  def self.find_from_oscap(oscap_references)
-    oscap_references.inject(where('1=0')) do |rel, reference|
-      rel.or(where(href: reference.href, label: reference.label))
+  class << self
+    def find_from_oscap(oscap_references)
+      oscap_references.inject(where('1=0')) do |rel, reference|
+        rel.or(where(href: reference.href, label: reference.label))
+      end
```
akofink: @dLobatog this is the `inject` style query to find all of a collection of objects from the database.

app/models/rule_reference.rb
```diff
@@ -18,16 +18,29 @@ class RuleReference < ApplicationRecord
...
-    )
+    def new_from_openscap_parser(op_rule_references)
+      preexisting_rule_references = ::RuleReference.where(
+        href: op_rule_references.map(&:href),
+        label: op_rule_references.map(&:label)
+      ).pluck(:href, :label)
+      op_rule_references.reject do |rule_reference|
+        preexisting_rule_references.include?([rule_reference.href,
+                                              rule_reference.label])
+      end
```
akofink: `find_from_oscap` already does this, and with the correct query, no?

app/services/concerns/xccdf/rule_references.rb
```diff
@@ -7,7 +7,9 @@ module RuleReferences
 
     included do
       def save_rule_references
-        @rule_references ||= @op_rule_references.map do |op_reference|
+        @rule_references ||= RuleReference.new_from_openscap_parser(
+          @op_rule_references
+        ).map do |op_reference|
           ::RuleReference.from_openscap_parser(op_reference)
```
akofink: IMO `from_openscap_parser` could be for 1 record - `all_from_openscap_parser` could be the same thing but accept a collection of op objects (and call `from_openscap_parser`) and return AR objects.

akofink:
So, should we close this? Parsing performance for exsting data is now much quicker after https://github.com/RedHatInsights/compliance-backend/pull/270

dLobatog:
Closing as it's not relevant except for parsing new benchmarks.

PR #270, opened 2019-11-21
[RHICOMPL-412] - Don't parse already parsed benchmarks

akofink:
https://projects.engineering.redhat.com/browse/RHICOMPL-412

Initial and subsequent parse of RHEL 7 standard profile:

```
2019-11-21T17:03:05.957Z 1 TID-gni5eoqyl ParseReportJob JID-580531214732438c34fe3e78 INFO: start
2019-11-21T17:03:52.511Z 1 TID-gni5eoqyl ParseReportJob JID-580531214732438c34fe3e78 INFO: done: 46.553 sec
2019-11-21T17:04:02.524Z 1 TID-gni5eoqyl ParseReportJob JID-0910570d946ff49f5b2d9e31 INFO: start
2019-11-21T17:04:03.860Z 1 TID-gni5eoqyl ParseReportJob JID-0910570d946ff49f5b2d9e31 INFO: done: 1.336 sec
```

TODO:

- [x] tests

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -22,6 +22,8 @@ class Profile < ApplicationRecord
 
   after_update :destroy_orphaned_business_objective
 
+  scope :canonical, lambda { where(account_id: nil) }
+
   def self.from_openscap_parser(op_profile, benchmark_id: nil, account_id: nil)
     profile = find_or_initialize_by(
       ref_id: op_profile.id,
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # OpenSCAP profile
-class Profile < ApplicationRecord
+class Profile < ApplicationRecord # rubocop:disable Metrics/ClassLength
   scoped_search on: %i[id name ref_id account_id compliance_threshold]
 
   has_many :profile_rules, dependent: :delete_all
@@ -22,7 +22,7 @@ class Profile < ApplicationRecord
 
   after_update :destroy_orphaned_business_objective
 
-  scope :canonical, lambda { where(account_id: nil) }
+  scope :canonical, -> { where(account_id: nil) }
 
   def self.from_openscap_parser(op_profile, benchmark_id: nil, account_id: nil)
     profile = find_or_initialize_by(
--- a/app/services/concerns/xccdf/benchmarks.rb
+++ b/app/services/concerns/xccdf/benchmarks.rb
@@ -7,9 +7,16 @@ module Xccdf
 
     included do
       def save_benchmark
-        @benchmark = ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
-        ::Xccdf::Benchmark.import!([@benchmark].select(&:new_record?),
+        ::Xccdf::Benchmark.import!([benchmark].select(&:new_record?),
                                    ignore: true)
+      end
+
+      def benchmark_saved?
+        benchmark.persisted?
+      end
+
+      def benchmark
+        @benchmark ||= ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
       end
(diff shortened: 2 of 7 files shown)


app/services/concerns/xccdf/benchmarks.rb
```diff
@@ -7,10 +7,15 @@ module Benchmarks
...
-        @benchmark = ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
+        @benchmark ||= ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
         ::Xccdf::Benchmark.import!([@benchmark].select(&:new_record?),
                                    ignore: true)
       end
+
+      def benchmark_saved?
+        @benchmark ||= ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
+        @benchmark.persisted?
+      end
```
dLobatog: Looks like you could have a helper method to initialize `@benchmark`:
```ruby
      def save_benchmark
        ::Xccdf::Benchmark.import!([benchmark].select(&:new_record?),
                                   ignore: true)
      end

      def benchmark_saved?
        benchmark.persisted?
      end

      def benchmark
        @benchmark ||= ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
      end
```

app/services/concerns/xccdf/hosts.rb
```diff
@@ -20,9 +20,8 @@ def report_host
     private
 
     def test_result_profile
-      @profiles.find do |profile|
-        profile.ref_id == @test_result_file.test_result.profile_id
-      end
+      ::Profile.find_by(ref_id: @test_result_file.test_result.profile_id,
```
dLobatog: This is unrelated, right? It makes sure the profile found is the "template" profile. Maybe we should have a helper for that in the model

akofink: This isn't unrelated. If we don't save the benchmark, `@profiles` never gets set. It's no problem, though, since this method just selects the profile that is referenced in the `TestResult`. This change just changes from selecting from the existing array of profiles from the benchmark parsing code to querying directly from the database. Now it works for both parsing benchmarks and skipping benchmark parsing.

app/models/profile.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # OpenSCAP profile
-class Profile < ApplicationRecord
+class Profile < ApplicationRecord # rubocop:disable Metrics/ClassLength
```
akofink: idk, this model doesn't feel too big to me. I'd be willing to bump up the max class length from 100 lines to 150.

app/services/concerns/xccdf/util.rb
```diff
@@ -17,13 +17,15 @@ module Util
...
 
       def save_all_benchmark_info
-        save_benchmark
-        save_profiles
-        save_rules
-        save_rule_identifiers
-        save_profile_rules
-        save_rule_references
-        save_rule_references_rules
+        unless benchmark_saved? # rubocop:disable Style/GuardClause
```
akofink: Rubocop is confused here. idk why it's trying to make me move this to be a guard clause when the block is multiple lines long :man_shrugging:

dLobatog: I don't know what does it have to do with the length of the block? It just wants to return early and have the rest of the method unindented I think  
```
return unless benchmark_saved?
```

akofink: Ohh, yeah that makes sense :)

akofink:
@dLobatog updated!

app/models/profile.rb
dLobatog: I'd say it'd be best not to start with that - maybe in this class it's OK though since the `result` method is quite a long SQL string

PR #272, opened 2019-11-25
Batching POC

dLobatog:
Don't take this one too seriously yet - the goal is to defer search of references and query them all at once in the systemDetails with:

1. select * from profiles where id = whatever
2. select * from rules where id in (a,b,c,d...)
3. select * from rule_references_rules where rule_id in (a,b,c,d)
4. select * from rule_references where id in (result from 3)

This works, but right now it's not taking into account `selected` rules, which makes it unusable as it's querying ~1000 rules and a huge number of rule references. Also we need to deal with duplicate RuleReferences I see in my DB, we need a constraint for that.

### DIFF
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -23,7 +23,7 @@ module Types
     # rubocop:disable AbcSize
     def rules(args = {})
       selected_columns = (args[:lookahead].selections.map(&:name) &
-                         ::Rule.column_names.map(&:to_sym)) << :id
+                          ::Rule.column_names.map(&:to_sym)) << :id
       host = Host.find(args[:system_id]) if args[:system_id].present?
       rules = object.rules_for_system(host, selected_columns) if host.present?
       rules = object.rules.select(selected_columns) if host.blank?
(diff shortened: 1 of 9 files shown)


app/graphql/types/profile.rb
```diff
@@ -63,32 +63,43 @@ def rules(args = {})
...
     def total_host_count
-      object.hosts.count
+      CollectionLoader.for(object.class, :hosts).load(object).then do |hosts|
+        hosts.count
+      end
     end
 
     def compliant(args = {})
-      object.compliant?(Host.find(system_id(args)))
+      RecordLoader.for(Host).load(system_id(args)).then do |host|
```
akofink:
```suggestion
      RecordLoader.for(::Host).load(system_id(args)).then do |host|
```

app/graphql/types/profile.rb
```diff
@@ -63,32 +63,43 @@ def rules(args = {})
...
+        object.compliant?(host)
+      end
     end
 
     def rules_passed(args = {})
-      Host.find(system_id(args)).rules_passed(object)
+      RecordLoader.for(Host).load(system_id(args)).then do |host|
```
akofink:
```suggestion
      RecordLoader.for(::Host).load(system_id(args)).then do |host|
```

app/graphql/types/profile.rb
```diff
@@ -63,32 +63,43 @@ def rules(args = {})
...
 
     def last_scanned(args = {})
-      rule_ids = object.rules.pluck(:id)
-      rule_results = RuleResult.where(
-        rule_id: rule_ids,
-        host_id: Host.find(system_id(args)).id
-      )
-      rule_results.maximum(:end_time)&.iso8601 || 'Never'
+      CollectionLoader.for(object.class, :rules).load(object).then do |rules|
+        rule_results = RuleResult.where(
```
akofink:
```suggestion
        rule_results = ::RuleResult.where(
```

PR #274, opened 2019-11-26
RHICOMPL-398 - Add TestResult model and migration

dLobatog:
Just a draft for now - we can create some initial test results from the latest RuleResults. The goal with this is just to massively improve the easiness of finding the appropriate rules, scores, etc..

### DIFF
--- a/app/models/test_result.rb
+++ b/app/models/test_result.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class TestResult < ApplicationRecord
+  belongs_to :profile
+  belongs_to :host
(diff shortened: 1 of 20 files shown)


dLobatog:
I'm going to split this one in 2 PRs for easier reviewing

app/models/test_result.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class TestResult < ApplicationRecord
+  belongs_to :profile
```
akofink: Technically, this should be `has_one :profile` (profile_id should be stored on the TestResult) and should be optional

image

PR #275, opened 2019-12-03
RHICOMPL-398 - Add TestResult model and migration

dLobatog:
Still in draft - basically https://github.com/RedHatInsights/compliance-backend/pull/274 but without the graphql batching

### DIFF
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -24,13 +24,14 @@ module Types
     def rules(args = {})
       selected_columns = (args[:lookahead].selections.map(&:name) &
                          ::Rule.column_names.map(&:to_sym)) << :id
-      host = ::Host.find(args[:system_id]) if args[:system_id].present?
-      rules = object.rules_for_system(host, selected_columns) if host.present?
-      rules = object.rules.select(selected_columns) if host.blank?
+
+      rules = rules_for_system(args, selected_columns) if system_id(args)
+      rules = object.rules.select(selected_columns) if rules.blank?
       rules = rules.with_identifier(args[:identifier]) if args.dig(:identifier)
       rules = rules.with_references(args[:references]) if args.dig(:references)
       rules = lookahead_includes(args[:lookahead], rules,
                                  identifier: :rule_identifier)
+      parent_profile(rules)
       rules
     end
     # rubocop:enable AbcSize
@@ -71,30 +72,49 @@ module Types
     end
 
     def compliant(args = {})
-      object.compliant?(::Host.find(system_id(args)))
+      host_results = latest_test_result(args)&.rule_results
+      host_results.present? &&
+        (
+          host_results.where(result: 'pass').count /
+          host_results.count.to_f
+        ) >=
+          (object.compliance_threshold / 100.0)
     end
 
     def rules_passed(args = {})
-      ::Host.find(system_id(args)).rules_passed(object)
+      return 0 if (@latest_test_result = latest_test_result(args)).blank?
+
+      @latest_test_result.rule_results.passed.count
     end
 
     def rules_failed(args = {})
-      ::Host.find(system_id(args)).rules_failed(object)
+      return 0 if (@latest_test_result = latest_test_result(args)).blank?
+
+      @latest_test_result.rule_results.failed.count
     end
 
     def last_scanned(args = {})
-      rule_ids = object.rules.pluck(:id)
-      rule_results = ::RuleResult.where(
-        rule_id: rule_ids,
-        host_id: ::Host.find(system_id(args)).id
-      )
-      rule_results.maximum(:end_time)&.iso8601 || 'Never'
+      latest_test_result(args)&.end_time&.iso8601 || 'Never'
     end
 
     private
 
     def system_id(args)
       args[:system_id] || context[:parent_system_id]
+    end
+
+    def parent_profile(rules)
+      context[:parent_profile_id] ||= {}
+      rules.each { |rule| context[:parent_profile_id][rule.id] = object.id }
+    end
+
+    def latest_test_result(args)
+      TestResult.latest(object.id, system_id(args))
+    end
+
+    def rules_for_system(args, selected_columns)
+      host = Host.find(system_id(args)) if system_id(args).present?
+      object.rules_for_system(host, selected_columns) if host.present?
     end
   end
 end
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -52,26 +52,21 @@ class Rule < ApplicationRecord
     rule
   end
 
-  # Disabling MethodLength because it measures things wrong
-  # for a multi-line string SQL query.
-  # rubocop:disable Metrics/MethodLength
-  def compliant?(host)
+  def compliant?(host, profile)
     Rails.cache.fetch("#{id}/#{host.id}/compliant", expires_in: 1.week) do
-      latest_result = RuleResult.find_by_sql(
-        ['SELECT rule_results.* FROM (
-          SELECT rr2.*,
-             rank() OVER (
-                    PARTITION BY rule_id, host_id
-                    ORDER BY end_time DESC, created_at DESC
-             )
-          FROM rule_results rr2
-          WHERE rr2.host_id = ? AND rr2.result IN (?) AND rr2.rule_id = ?
-       ) rule_results WHERE RANK = 1', host.id, RuleResult::SELECTED, id]
-      ).last
-      return false if latest_result.blank?
+      return false unless profile.present? && profile.rules.include?(self)
 
-      %w[pass notapplicable notselected].include? latest_result.result
+      latest_rule_result = latest_result(host, profile)
+      return false if latest_rule_result.blank?
+
+      %w[pass notapplicable notselected].include? latest_rule_result.result
     end
   end
-  # rubocop:enable Metrics/MethodLength
+
+  def latest_result(host, profile)
+    test_result = TestResult.latest(profile.id, host.id)
+    return nil if test_result.blank?
+
+    test_result.rule_results.find_by(rule_id: id)
+  end
 end
--- a/app/models/rule_result.rb
+++ b/app/models/rule_result.rb
@@ -7,21 +7,28 @@ class RuleResult < ApplicationRecord
   scoped_search on: %i[id rule_id host_id result]
   belongs_to :host
   belongs_to :rule
+  belongs_to :test_result
 
+  validates :test_result, presence: true
   validates :host, presence: true
   validates :rule, presence: true
 
   SELECTED = %w[pass fail notapplicable error unknown].freeze
   FAIL = %w[fail error unknown notchecked].freeze
+  PASSED = %w[pass notapplicable].freeze
 
+  scope :passed, -> { where(result: PASSED) }
   scope :selected, -> { where(result: SELECTED) }
   scope :failed, -> { where(result: FAIL) }
   scope :for_system, ->(host_id) { where(host_id: host_id) }
 
-  def self.from_openscap_parser(op_rule_result, rule_ids: {}, host_id: nil,
-                                start_time: nil, end_time: nil)
-    find_or_initialize_by(rule_id: rule_ids[op_rule_result.id],
-                          host_id: host_id, result: op_rule_result.result,
-                          start_time: start_time, end_time: end_time)
+  def self.from_openscap_parser(op_rule_result, test_result_id: nil,
+                                rule_id: nil, host_id: nil)
+    find_or_initialize_by(
+      test_result_id: test_result_id,
+      rule_id: rule_id,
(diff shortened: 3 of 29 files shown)


app/graphql/types/profile.rb
```diff
@@ -71,30 +71,42 @@ def total_host_count
...
     private
 
     def system_id(args)
       args[:system_id] || context[:parent_system_id]
     end
+
+    def latest_test_result(args)
+      TestResult.where(
+        profile_id: object.id, host_id: system_id(args)
+      ).order('created_at DESC').includes(:rule_results)&.first
```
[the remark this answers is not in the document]
dLobatog: Besides - I am calling `latest.rule_results` or `latest.end_time` in other places here (which don't work with `limit(1)` or `first(1)` as both return a collection, not a single instance)

dLobatog:
I am seeing some odd compliant :heavy_check_mark:  results, please do not merge yet :sweat:

app/graphql/types/profile.rb
```diff
@@ -71,30 +71,40 @@ def total_host_count
     end
 
     def compliant(args = {})
-      object.compliant?(Host.find(system_id(args)))
+      host_results = latest_test_result(args)&.rule_results
+      host_results.present? &&
+        (host_results.count(true) / host_results.count.to_f) >=
```
dLobatog: This `count` is wrong

dLobatog:
Should be fixed now

db/migrate/20191126180559_create_test_results.rb
```diff
@@ -0,0 +1,15 @@
+class CreateTestResults < ActiveRecord::Migration[5.2]
+  def change
+    create_table :test_results, id: :uuid do |t|
+      t.datetime :start_time
+      t.datetime :end_time
+      t.decimal :score
+      t.references :profile, type: :uuid, index: true
+      t.references :host, type: :uuid, index: true
```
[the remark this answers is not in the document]
dLobatog: I can add benchmark here, but what is the point? `test_result.profile.benchmark` will always get you the benchmark of the profile associated, which has to be the same one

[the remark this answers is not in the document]
akofink: I guess the reason the XCCDF model has both is because it references both by ref_id, not by uuid. Since we're using uuids, we can keep the association relationship via profile like you have, but I'd also opt for adding that a TestResult `belongs_to :benchmark, through: profile`.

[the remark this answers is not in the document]
dLobatog: That makes sense - will put it there

app/models/rule.rb
```diff
@@ -52,26 +52,22 @@ def self.from_openscap_parser(op_rule, benchmark_id: nil)
...
+      latest_rule_result = latest_result(host, profile)
+      return false if latest_rule_result.blank?
+
+      %w[pass notapplicable notselected].include? latest_rule_result.result
     end
   end
-  # rubocop:enable Metrics/MethodLength
+
+  def latest_result(host, profile)
+    test_result = TestResult.where(host: host, profile: profile)
```
akofink: `TestResult.latest`

app/models/rule_result.rb
```diff
@@ -18,10 +20,13 @@ class RuleResult < ApplicationRecord
...
 
-  def self.from_openscap_parser(op_rule_result, rule_ids: {}, host_id: nil,
-                                start_time: nil, end_time: nil)
-    find_or_initialize_by(rule_id: rule_ids[op_rule_result.id],
-                          host_id: host_id, result: op_rule_result.result,
-                          start_time: start_time, end_time: end_time)
+  def self.from_openscap_parser(op_rule_result, test_result, rule_id: nil,
+                                host_id: nil)
+    find_or_initialize_by(
+      test_result: test_result,
```
akofink: Let's be consistent:

```suggestion
      test_result_id: test_result_id,
```

app/models/rule_result.rb
```diff
@@ -18,10 +20,13 @@ class RuleResult < ApplicationRecord
   scope :failed, -> { where(result: FAIL) }
   scope :for_system, ->(host_id) { where(host_id: host_id) }
 
-  def self.from_openscap_parser(op_rule_result, rule_ids: {}, host_id: nil,
-                                start_time: nil, end_time: nil)
-    find_or_initialize_by(rule_id: rule_ids[op_rule_result.id],
-                          host_id: host_id, result: op_rule_result.result,
-                          start_time: start_time, end_time: end_time)
+  def self.from_openscap_parser(op_rule_result, test_result, rule_id: nil,
```
akofink:
```suggestion
  def self.from_openscap_parser(op_rule_result, test_result_id: nil, rule_id: nil,
```

app/models/test_result.rb
```diff
@@ -0,0 +1,19 @@
...
+# Representation of the TestResult XML property in an OpenSCAP report. Holds the
+# basic report properties, such as dates, host, and results.
+class TestResult < ApplicationRecord
+  belongs_to :profile
+  belongs_to :host
+  has_one :benchmark, through: :profile
+  has_many :rule_results, dependent: :delete_all
+
+  validates :host_id, presence: true
+  validates :profile_id, presence: true
```
akofink: What's the uniqueness validation here? We should define it and also add a DB constraint or index.

app/models/test_result.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# Representation of the TestResult XML property in an OpenSCAP report. Holds the
+# basic report properties, such as dates, host, and results.
+class TestResult < ApplicationRecord
+  belongs_to :profile
+  belongs_to :host
```
akofink: This would also be nice to have: `has_many :rules, through: :rule_results`

app/services/xccdf_report_migration.rb
```diff
@@ -3,29 +3,31 @@
...
+# rubocop:disable Metrics/ClassLength
 class XCCDFReportMigration
   def initialize(account, noop)
     @account = account
     @noop = noop
-  end
-
-  def logger
-    Sidekiq.logger
+    @logger = Sidekiq.logger
```
akofink: Why the new instance variable? The method wasn't sufficient?

db/migrate/20191126180559_create_test_results.rb
```diff
@@ -0,0 +1,15 @@
...
+
+      t.timestamps
```
akofink: Let's also add a uniqueness index/constraint.

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -0,0 +1,60 @@
+class SeedTestResults < ActiveRecord::Migration[5.2]
+  def change
+    migrated = 0
+    ProfileHost.find_each do |profile_host|
+      host = profile_host.host
+      profile = profile_host.profile
+      binding.pry if profile.ref_id == "xccdf_org.ssgproject.content_profile_cjis"
```
akofink: :see_no_evil:

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -0,0 +1,60 @@
...
+      test_result = TestResult.create(
+        host: host,
+        profile: profile
+      )
+
+      puts(" - Migrating results for host #{host.name} - profile #{profile.name}")
+
+      rule_result_ids = []
+      latest_scan_results(host, profile).each do |rule_result|
```
akofink:
```suggestion
      latest_scan_results(host, profile).map(&:id)
```

db/migrate/20191126181009_seed_test_results.rb
akofink: I'm actually shocked rubocop doesn't catch this.

db/migrate/20191126180559_create_test_results.rb
dLobatog: Added on line 14

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -0,0 +1,63 @@
...
+      profile = profile_host.profile
+      test_result = TestResult.create(
+        host: host,
+        profile: profile
+      )
+
+      puts(" - Migrating results for host #{host.name}"\
+           " - profile #{profile.name}")
+      rule_results_by_start_time = scan_results(host, profile)
+        .group_by(&:start_time)
```
dLobatog: This line here is key - I am grouping rule results by start time, host, profile so that we can infer "TestResults" from these. This allows us to migrate all rule results, not just the latest ones

Victoremepunto:

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -0,0 +1,63 @@
...
+      rule_results_by_start_time.each do |start_time, rule_results|
+        rule_result_ids = rule_results.pluck(:id)
+        puts("   - Number of RuleResults found: #{rule_result_ids.count}"\
+             " at time #{start_time}")
+        latest_rule_results = RuleResult.where(id: rule_result_ids)
+        latest_rule_results.update_all(test_result_id: test_result.id)
+        test_result.update(
```
akofink: Let's be sure to raise errors if anything bad happens.

```suggestion
        test_result.update!(
```

db/migrate/20191203091219_remove_time_fields_from_rule_results.rb
```diff
@@ -0,0 +1,17 @@
...
+    remove_column :rule_results, :end_time
+  end
+
+  def down
+    add_column :rule_results, :start_time, :datetime
+    add_column :rule_results, :end_time, :datetime
+    TestResult.find_each do |test_result|
+      rule_result_ids = test_result.rule_results.pluck(:id)
+      RuleResult.where(id: rule_result_ids).update_all(
+        start_time: test_result.start_time, end_time: test_result.end_time
```
akofink: The migration and rollback seems to work as expected (nice work)!

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -0,0 +1,63 @@
+class SeedTestResults < ActiveRecord::Migration[5.2]
+  def up
+    migrated = 0
+    ProfileHost.find_each do |profile_host|
+      host = profile_host.host
+      profile = profile_host.profile
+      test_result = TestResult.create(
+        host: host,
+        profile: profile
+      )
```
akofink: This looks like it may be the issue. We're creating only one `TestResult` per `ProfileHost`, but there will likely be more than one `TestResult` for each combination of profile and host as new scans are performed for the same profile/host combination.

db/migrate/20191126180559_create_test_results.rb
```diff
@@ -0,0 +1,16 @@
...
+    end
+
+    add_reference :rule_results, :test_result, type: :uuid, index: true
+    add_index :test_results, %i[profile_id host_id], unique: true
```
akofink: This is unexpected. I'd have thought a TestResult would be unique by profile, host, and end_time or similar. Try the following:

1. Scan a system and upload the report
1. Re-run the same scan and upload that report

```
sidekiq_1           | 2020-01-03T18:42:25.758Z 1 TID-grvelhdgl WARN: ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_test_results_on_profile_id_and_host_id"                                                                                                                                                                
sidekiq_1           | DETAIL:  Key (profile_id, host_id)=(4a6eb9d2-69af-4f9f-801b-ac0a4b5b2710, b34c2911-9c07-4278-a58a-8c2ba70b24f4) already exists.                                                                                                                                                                                                                                         
sidekiq_1           | : INSERT INTO "test_results" ("start_time", "end_time", "score", "profile_id", "host_id", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"
```

app/models/xccdf.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-Dir[File.join(__dir__, 'xccdf', '*.rb')].each { |f| require f }
+Dir[File.join(__dir__, 'xccdf', '*.rb')].sort.each { |f| require f }
```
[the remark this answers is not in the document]
dLobatog: New rubocop requirement so that we make sure the files are always in the same order

db/migrate/20191126181009_seed_test_results.rb
dLobatog: Good catch

db/migrate/20191126181009_seed_test_results.rb
dLobatog: Let's move this `.create` to the `rule_results_by_start_time` block. That would allow us to create 1 test result per group of rule_results in a host/profile group.

PR #276, opened 2019-12-03
 RHICOMPL-419 - GraphQL batching to query rules by references 

dLobatog:
:warning:  This PR should be merged *after* https://github.com/RedHatInsights/compliance-backend/pull/275

Test with a string like:
```
query System($systemId: String!) {
  system(id: $systemId) {
    id
    name
    profiles {
      id
      name
      refId
      rulesFailed
      rulesPassed
      lastScanned
      compliant
      rules {
        title
        severity
        rationale
        refId
        description
        remediationAvailable
        compliant
        references
        identifier
      }
    }
  }
}
```
I would also disable `Bullet` on development for this one as the performance penalty is enormous


app/graphql/types/rule.rb
```diff
@@ -14,24 +14,57 @@ class Rule < Types::BaseObject
...
+    def references
+      if context[:"rule_references_#{object.id}"].nil?
+        ::CollectionLoader.for(::Rule, :rule_references).load(object).then do |references|
+          references
+        end
+      else
+        ::RecordLoader.for(::RuleReference)
+          .load_many(context[:"rule_references_#{object.id}"])
+          .then do |references|
+          references.map { |ref| [ref.href, ref.label] }
```
dLobatog: It's insane how much faster this is - in a request that took 10s, it takes 5s now, all those 5 seconds were spent serializing `::Type::RuleReference` instances by GraphQL-ruby

app/graphql/types/rule.rb
```diff
@@ -14,24 +14,57 @@ class Rule < Types::BaseObject
...
 
-    def references(lookahead:)
-      selected_columns = lookahead.selections.map(&:name) &
-                         ::RuleReference.column_names.map(&:to_sym)
-      object.references.distinct.reorder('')
-            .pluck(selected_columns).map do |reference|
-        selected_columns.zip([reference].flatten).to_h
+    def identifier
+      ::CollectionLoader.for(::Rule, :rule_identifier).load(object).then do |identifier|
+        identifier&.label
```
dLobatog: This shaves off ~500ms too

db/migrate/20191203091219_remove_time_fields_from_rule_results.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveTimeFieldsFromRuleResults < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :rule_results, :start_time
```
dLobatog: These fields are irrelevant as they are the same as the one in the TestResult

app/graphql/types/profile.rb
```diff
@@ -62,39 +48,99 @@ def rules(args = {})
...
+          &.where(result: %w[pass notapplicable notselected])&.count
+      end
     end
 
     def rules_failed(args = {})
-      Host.find(system_id(args)).rules_failed(object)
+      latest_test_result_batch(args).then do |latest_test_result|
+        latest_test_result
+          &.rule_results
+          &.where&.not(result: %w[pass notapplicable notselected])&.count
```
akofink: All these safe operators, but all these methods are guaranteed to return an AR::Relation and never `nil`.

app/graphql/types/interfaces/rules_preload.rb
```diff
@@ -0,0 +1,68 @@
...
+# Class to deal with preloading various attributes for rules, such as
+# rule results, compliant,and references
+module RulesPreload
+  include GraphQL::Schema::Interface
+
+  def rules(args = {})
+    context[:parent_profile_id] ||= {}
+    return all_rules if system_id(args).blank?
+
+    latest_test_result_batch(args).then do |latest_test_result|
```
dLobatog: The whole idea here is not only to batch the results, but also to pre-set all of the rule related fields whenever the profile is found

app/graphql/types/rule.rb
```diff
@@ -14,37 +14,66 @@ class Rule < Types::BaseObject
...
 
-    def references(lookahead:)
-      selected_columns = lookahead.selections.map(&:name) &
-                         ::RuleReference.column_names.map(&:to_sym)
-      object.references.distinct.reorder('')
-            .pluck(selected_columns).map do |reference|
-        selected_columns.zip([reference].flatten).to_h
+    def references_from_context
+      ::RecordLoader.for(::RuleReference)
+                    .load_many(context[:"rule_references_#{object.id}"])
```
dLobatog: Here is the key - if we already have all references loaded here, we can simply load them in one SELECT. 
Before this, if you just have the field defiend as `field :rule_references, [RuleReference]`, GraphQL would do this:
1. Find the profile "foo"
2. Find rule A in profile "foo". 
3. Find all rule references related to rule A, one by one to render them as RuleReference (SELECT * FROM rule_references WHERE id = single rule reference ID)
4. Repeat step 3 until all references are found and rendered.
5. Repeat step 2 until all rules are found

Whereas right now, we already preload nearly everything:

1. Find profile "foo"
2. Batch find all rules in profile "foo" (one SELECT for all rules, even if we query for multiple profiles are queried)
3. Batch find all references in rules found in 2. (one SELECT for all rule references)

app/graphql/types/rule.rb
```diff
@@ -14,37 +14,66 @@ class Rule < Types::BaseObject
...
+                    .then do |references|
+        references.map { |ref| [ref.href, ref.label] }.to_json
```
dLobatog: The reason why the references are printed as simple strings is that I could not find a way of rendering them as RuleReference (the GraphQL type) without forcing GQL to find them in the db again. Finding them on the db triggers a SELECT per rule_reference, which is a huge amount when you want to get all references for all rules/profiles in a host (system details page)

app/graphql/types/system.rb
```diff
@@ -51,11 +51,15 @@ def profile_names
     end
 
     def rules_passed(args = {})
-      object.rules_passed(::Profile.find_by(id: args[:profile_id]))
+      ::RecordLoader.for(::Profile).load(args[:profile_id]).then do |profile|
+        object.rules_passed(profile)
```
dLobatog: No changes here - this just batches the `Profile` find call in case you are finding multiple profiles, e.g: if you're getting all profiles for a host.

app/graphql/types/profile.rb
```diff
@@ -66,37 +54,55 @@ def rules(args = {})
...
-        ) >=
-          (object.compliance_threshold / 100.0)
+      latest_test_result_batch(args).then do |latest_test_result|
+        host_results = latest_test_result&.rule_results
+        host_results.present? &&
+          (
+            host_results.where(result: 'pass').count /
+            host_results.count.to_f
+          ) >=
+            (object.compliance_threshold / 100.0)
```
akofink: I thought we no longer need to do this kind of calculation since we have the TestResult, right?

app/graphql/schema.rb
```diff
@@ -13,4 +13,8 @@ class Schema < GraphQL::Schema
   use ComplianceTimeout, max_seconds: 20
   query Types::Query
   mutation Types::Mutation
+  lazy_resolve(Promise, :sync)
```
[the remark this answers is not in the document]
dLobatog: It allows lazy loaders like CollectionLoader or RecordLoader to resolve the promises lazily (after all IDs are collected) so that the SQL queries are batched. https://github.com/rmosolgo/graphql-ruby/blob/master/guides/schema/lazy_execution.md#example-batched-find

app/graphql/types/profile.rb
dLobatog: That's right - I had this previous to this change, I'll change it to `latest_test_result.score >= compliance_threshold`

PR #278, opened 2019-12-09
Mount compliance app at both prefixes

akofink:
Mounted at both /api and /r/insights/platform


akofink:
:heavy_check_mark:

PR #279, opened 2019-12-11
Import remediations after report parse

akofink:
After importing a benchmark for the first time, we should import remediations rather than waiting until 8am the next morning.

### DIFF
--- a/app/services/datastream_importer.rb
+++ b/app/services/datastream_importer.rb
@@ -16,6 +16,15 @@ class DatastreamImporter
   def import!
     Xccdf::Benchmark.transaction do
       save_all_benchmark_info
+      import_remediations
     end
+  end
+
+  private
+
+  def import_remediations
+    RemediationsAPI.new(
+      Account.find_by!(account_number: ENV['JOBS_ACCOUNT_NUMBER'])
+    ).import_remediations
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/datastream_importer.rb
```diff
@@ -16,6 +16,15 @@ def initialize(datastream_filename)
...
       save_all_benchmark_info
+      import_remediations
     end
   end
+
+  private
+
+  def import_remediations
+    RemediationsAPI.new(
+      Account.find_by!(account_number: ENV['JOBS_ACCOUNT_NUMBER'])
```
dLobatog: I was going to comment on this - but you got it covered on https://github.com/RedHatInsights/e2e-deploy/commit/464d17cd4edd8be2711d8578a41179d8710977b0

app/services/xccdf_report_parser.rb
```diff
@@ -44,6 +44,13 @@ def save_all
     Host.transaction do
       save_all_benchmark_info
       save_all_test_result_info
+      import_remediations
```
dLobatog: I understand importing remediations after every datastream import, but why after every report import?

akofink: Well, tbh I think the QA pipeline uploads reports without ever importing the datastream first, and this is certainly a possibility in our app as well if someone uploads a report from a datastream we don't know about already. When this happens, no remediations are found in the frontend/from the perspective of the user. It seems this could be fixed by setting `remediation_available` to true/false if cached or nil if it's not cached.

akofink:
Updated to only import after a datastream import and not after a report parse.

PR #286, opened 2019-12-18
[RHICOMPL-444] - Handle tailored test result parsing

akofink:
https://projects.engineering.redhat.com/browse/RHICOMPL-444

### DIFF
--- a/app/services/concerns/xccdf/hosts.rb
+++ b/app/services/concerns/xccdf/hosts.rb
@@ -9,20 +9,40 @@ module Xccdf
       @host.update!(name: inventory_host['fqdn'])
     end
 
-    def save_profile_host
-      test_result_profile.clone_to(account: @account, host: @host)
+    def host_profile
+      @host_profile ||= test_result_profile.clone_to(
+        account: @account, host: @host
+      )
     end
+    alias save_profile_host host_profile
 
     def report_host
-      @metadata&.dig('fqdn') || @test_result_file.host
+      @metadata&.dig('fqdn') || @test_result_file.test_result.host
+    end
+
+    def associate_rules_from_rule_results
+      ::ProfileRule.import!(
+        ::Rule.where(
+          ref_id: selected_op_rule_results.map(&:id),
+          benchmark_id: @benchmark.id
+        ).pluck(:id).map do |rule_id|
+          ::ProfileRule.new(profile_id: @host_profile.id,
+                            rule_id: rule_id)
+        end, ignore: true
+      )
     end
 
     private
 
     def test_result_profile
       ::Profile.canonical
-               .find_by(ref_id: @test_result_file.test_result.profile_id,
-                        benchmark: @benchmark)
+               .where(ref_id: @test_result_file.test_result.profile_id,
+                      benchmark: @benchmark).first ||
+        ::Profile.find_or_initialize_by(
+          ref_id: @test_result_file.test_result.profile_id,
+          name: @test_result_file.test_result.profile_id,
+          benchmark_id: @benchmark.id
+        )
     end
 
     def inventory_host
--- a/app/services/concerns/xccdf/rule_results.rb
+++ b/app/services/concerns/xccdf/rule_results.rb
@@ -4,7 +4,7 @@ module Xccdf
   # Methods related to saving RuleResults from openscap_parser
   module RuleResults
     def save_rule_results
-      @rule_results = selected_rule_results.map do |op_rule_result|
+      @rule_results = selected_op_rule_results.map do |op_rule_result|
         ::RuleResult.from_openscap_parser(
           op_rule_result,
           rule_ids: rule_ids, host_id: @host.id,
@@ -16,18 +16,18 @@ module Xccdf
       ::RuleResult.import!(@rule_results.select(&:new_record?), ignore: true)
     end
 
+    def selected_op_rule_results
+      @op_rule_results.reject do |rule_result|
+        rule_result.result == 'notselected'
+      end
+    end
+
     private
 
     def rule_ids
       @rule_ids ||= Rule.where(
-        benchmark: @benchmark, ref_id: selected_rule_results.map(&:id)
+        benchmark: @benchmark, ref_id: selected_op_rule_results.map(&:id)
       ).pluck(:ref_id, :id).to_h
-    end
-
-    def selected_rule_results
-      @op_rule_results.reject do |rule_result|
-        rule_result.result == 'notselected'
-      end
     end
   end
 end
--- a/app/services/concerns/xccdf/util.rb
+++ b/app/services/concerns/xccdf/util.rb
@@ -30,9 +30,20 @@ module Xccdf
 
       def save_all_test_result_info
         save_host
-        @host_profile = save_profile_host
+        save_profile_host
         save_rule_results
+        associate_rules_from_rule_results
         invalidate_cache
+      end
+
+      def set_openscap_parser_data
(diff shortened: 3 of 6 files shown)


app/services/concerns/xccdf/hosts.rb
```diff
@@ -9,20 +9,40 @@ def save_host
...
     end
 
-    def save_profile_host
-      test_result_profile.clone_to(account: @account, host: @host)
+    def host_profile
+      @host_profile ||= test_result_profile.clone_to(
+        account: @account, host: @host
+      )
     end
+    alias save_profile_host host_profile
```
akofink: rubocop needs the memoized variable name to match the method name

app/services/concerns/xccdf/util.rb
```diff
@@ -30,11 +30,22 @@ def save_all_benchmark_info
...
 
+      def set_openscap_parser_data
+        @op_benchmark = @test_result_file.benchmark
+        @op_test_result = @test_result_file.test_result
+        @op_profiles = @op_benchmark.profiles
+        @op_rules = @op_benchmark.rules
+        @op_rule_references =
+          @op_benchmark.rule_references.reject { |rr| rr.label.empty? }
+        @op_rule_results = @op_test_result.rule_results
+      end
```
akofink: I decided to move this to util because it's required to set up all the rest of the xccdf parsing methods (and we can conveniently use it in tests).

app/services/concerns/xccdf/hosts.rb
```diff
@@ -9,20 +9,40 @@ def save_host
...
 
     def test_result_profile
       ::Profile.canonical
-               .find_by(ref_id: @test_result_file.test_result.profile_id,
-                        benchmark: @benchmark)
+               .where(ref_id: @test_result_file.test_result.profile_id,
+                      benchmark: @benchmark).first ||
+        ::Profile.find_or_initialize_by(
+          ref_id: @test_result_file.test_result.profile_id,
+          name: @test_result_file.test_result.profile_id,
```
dLobatog: I think we can get the profile name and description correctly from 
```
<Profile id='foo'>
  <title>$profilename</title>
  <description>$profiledescription</description>
```

app/services/concerns/xccdf/hosts.rb
```diff
@@ -9,20 +9,40 @@ def save_host
...
+    alias save_profile_host host_profile
 
     def report_host
-      @metadata&.dig('fqdn') || @test_result_file.host
+      @metadata&.dig('fqdn') || @test_result_file.test_result.host
+    end
+
+    def associate_rules_from_rule_results
+      ::ProfileRule.import!(
+        ::Rule.where(
```
dLobatog: As far as I can tell, this is going to attempt to relate rules with profiles when the rules are already in our DB.

So the following 2 scenarios are not right if I understood correctly:

1 - Report uploaded, we find a canonical profile for it. This would still attempt to add more rules to the profile, but we don't need that as we have the canonical profile (it would be just a waste of time)
2 - Report uploaded, we don't find a canonical profile for it. The profile is not just a tailored profile but something completely custom - no rules are imported and we import an empty profile. 

I think both could be fixed by checking if the result from `test_result_profile` is a initialized object or not, perhaps even a helper for that would be helpful.

app/services/concerns/xccdf/hosts.rb
```diff
@@ -9,20 +9,40 @@ def save_host
...
+    def host_profile
+      @host_profile ||= test_result_profile.clone_to(
+        account: @account, host: @host
+      )
     end
+    alias save_profile_host host_profile
 
     def report_host
-      @metadata&.dig('fqdn') || @test_result_file.host
+      @metadata&.dig('fqdn') || @test_result_file.test_result.host
```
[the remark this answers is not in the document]
akofink: This is a drive-by bug fix. `test_result_file.host` is undefined. We've just never been bitten by this because metadata always exists.

app/services/concerns/xccdf/rule_results.rb
```diff
@@ -4,7 +4,7 @@ module Xccdf
   # Methods related to saving RuleResults from openscap_parser
   module RuleResults
     def save_rule_results
-      @rule_results = selected_rule_results.map do |op_rule_result|
+      @rule_results = selected_op_rule_results.map do |op_rule_result|
```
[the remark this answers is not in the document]
akofink: Yep, just making very clear what's an AR model and what's from openscap_parser

app/services/concerns/xccdf/hosts.rb
```diff
@@ -9,20 +9,40 @@ def save_host
...
     def test_result_profile
       ::Profile.canonical
-               .find_by(ref_id: @test_result_file.test_result.profile_id,
-                        benchmark: @benchmark)
+               .where(ref_id: @test_result_file.test_result.profile_id,
+                      benchmark: @benchmark).first ||
+        ::Profile.find_or_initialize_by(
+          ref_id: @test_result_file.test_result.profile_id,
+          name: @test_result_file.test_result.profile_id,
+          benchmark_id: @benchmark.id
```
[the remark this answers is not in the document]
akofink: Hosts and account are associated [here](https://github.com/RedHatInsights/compliance-backend/pull/286/files#diff-3450e41fcc987c8952a1ffe42f72771cR14), just like for non-tailored results.

app/services/concerns/xccdf/hosts.rb
akofink: If there were a profile with a refId that matched this customized profile refID, we could do that. The reason we have to make this PR in the first place is because there is no 'canonical' profile defined in the benchmark with the custom refId. We don't know the parent profile ID of this customized profile, and there's no way to figure it out with just the result output (we'd need the tailoring file as well).

app/services/concerns/xccdf/hosts.rb
akofink: Note that no rules are created here at all - we're only creating `ProfileRule` records in the DB. Rule creation only ever happens during the benchmark parse - this doesn't change for tailored profiles.

It turns out `ProfileRule` records are unique by profile_id and rule_id, so it's impossible to associate a profile to a rule twice. This method is run for non-customized test result profiles, but it's essentially a no-op (yes, it wastes some time, but premature optimization and all that...); however, if a rule result exists for this test result that isn't actually associated with the named canonical profile (an inconsistency in the generated report for some reason, likely never to happen), then those rules would be associated with this profile.

app/services/concerns/xccdf/hosts.rb
```diff
@@ -9,20 +9,40 @@ def save_host
...
     end
 
     private
 
     def test_result_profile
       ::Profile.canonical
-               .find_by(ref_id: @test_result_file.test_result.profile_id,
-                        benchmark: @benchmark)
+               .where(ref_id: @test_result_file.test_result.profile_id,
+                      benchmark: @benchmark).first ||
```
akofink: @dLobatog This `||` condition checks that there's no existing canonical profile.

app/services/concerns/xccdf/hosts.rb
dLobatog: We cannot tell who the parent is, that's right - but it looks like we can tell what the name and description are isn't it? Look for `<Profile><title>` in your sample fixture for example

akofink: There is no `<Profile id="xccdf_org.ssgproject.content_profile_C2S_customized">` in this benchmark with nested title/description. I'm quite sure we can't get that information without the tailoring file.

PR #290, opened 2020-01-03
RHICOMPL-446 - RBAC light implementation

dLobatog:
~~:warning: Due to the fact the service does not seem to contain the principal for compliance yet, it has not been tested properly yet. Please hold off merging until the service becomes GA in CI.~~

:traffic_light: :green_apple:  The service is now available on CI so it can be merged already


app/services/rbac_api.rb
```diff
@@ -0,0 +1,29 @@
...
+  def initialize(account)
+    @url = URI.parse("#{URI.parse(Settings.remediations_url)}"\
+                     "#{ENV['PATH_PREFIX']}/rbac/v1/access")
+    @b64_identity = Base64.strict_encode64(account.fake_identity_header.to_json)
+  end
+
+  def check_user
+    body = JSON.parse(access_check_response)
+    body['data'].find do |access_check|
+      return true if access_check['permission'] == 'compliance:*:*'
```
dLobatog: For now, that's the only check we have to make - https://github.com/RedHatInsights/rbac-config/blob/master/configs/compliance.json

app/services/rbac_api.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# This class is meant to handle calls to the RBAC light API
+class RBACAPI
+  def initialize(account)
+    @url = URI.parse("#{URI.parse(Settings.remediations_url)}"\
```
akofink: I think we need a new setting. In prod, this could look like: `remediations_url: http://remediations.remediations-ci.svc.cluster.local:9002`

app/services/rbac_api.rb
```diff
@@ -0,0 +1,29 @@
...
+                     "#{ENV['PATH_PREFIX']}/rbac/v1/access")
+    @b64_identity = Base64.strict_encode64(account.fake_identity_header.to_json)
+  end
+
+  def check_user
+    body = JSON.parse(access_check_response)
+    body['data'].find do |access_check|
+      return true if access_check['permission'] == 'compliance:*:*'
+    end
+    false
```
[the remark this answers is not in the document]
dLobatog: Fair enough, nil is 'falsey' so it's good enough to be used as a boolean in the caller (`Authentication#rbac_allowed?`)

app/services/rbac_api.rb
dLobatog: Good one - yup  indeed we needed it

app/controllers/concerns/authentication.rb
```diff
@@ -41,6 +43,11 @@ def find_or_create_user(username)
     user
   end
 
+  def rbac_allowed?
+    rbac_api ||= ::RBACAPI.new(account_from_header)
```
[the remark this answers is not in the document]
dLobatog: Fixed

app/services/rbac_api.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# This class is meant to handle calls to the RBAC light API
+class RBACAPI
```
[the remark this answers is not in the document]
dLobatog: Fixed

akofink:
FYI: the ssl option change locally in platform.rb is no longer necessary: https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/platform.rb#L11

config/appsignal.yml
```diff
@@ -0,0 +1,40 @@
...
+
+# Configuration per environment, leave out an environment or set active
+# to false to not push metrics for that environment.
+development:
+  <<: *defaults
+  active: true
+
+production:
+  <<: *defaults
+  active: true
```
akofink: This doesn't seem related

dLobatog: Oops - I'll remove this one.

PR #291, opened 2020-01-06
Add a unique db constraint for benchmarks

akofink:
The migration migrates a duplicate benchmark's rules and profiles and associated records (namely host_profiles and rule results) and then adds the unique benchmark DB constraint.

Test scenario ideas:

- Import 1 benchmark, duplicate it and save, run the migration
- Import 2 benchmarks of different versions, update the versions to match, run the migration
- Import 1 RHEL 7 benchmark, import 1 RHEL 6 benchmark, update ref_id and version to match, run the migration

For all cases: confirm 1 benchmark exists and no errors occurred
These counts should change: `[Xccdf::Benchmark.count, Profile.count, Rule.count, RuleIdentifier.count]`
These may or may not change: `[RuleReference.count, ProfileHost.count, ProfileRule.count]`
These should not change: `[RuleResult.count, Host.count, TestResult.count]`

### DIFF
--- a/db/migrate/20200106134953_add_unique_index_to_benchmarks.rb
+++ b/db/migrate/20200106134953_add_unique_index_to_benchmarks.rb
@@ -1,0 +1,11 @@
+class AddUniqueIndexToBenchmarks < ActiveRecord::Migration[5.2]
+  def up
+    DuplicateBenchmarkResolver.run!
+
+    add_index(:benchmarks, %i[ref_id version], unique: true)
+  end
+
+  def down
+    remove_index(:benchmarks, %i[ref_id version])
+  end
+end
--- a/test/services/duplicate_benchmark_resolver_test.rb
+++ b/test/services/duplicate_benchmark_resolver_test.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require './db/migrate/20200106134953_add_unique_index_to_benchmarks'
+
+class DuplicateBenchmarkResolverTest < ActiveSupport::TestCase
+  setup do
+    @migration = AddUniqueIndexToBenchmarks.new
+    @migration.down # remove db constraint
+
+    (@dupe_bm = benchmarks(:one).dup).save(validate: false)
+
+    @dupe_rules = rules.map do |r|
+      dupe_r = r.dup
+      dupe_r.benchmark = @dupe_bm
+      dupe_r.save(validate: false)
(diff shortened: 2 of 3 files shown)


db/migrate/20200106134953_add_unique_index_to_benchmarks.rb
```diff
@@ -0,0 +1,48 @@
...
+      Xccdf::Benchmark.includes(:rules, :profiles).find_each do |bm|
+        if benchmarks[[bm.ref_id, bm.version]]
+          migrate_rules(benchmarks[[bm.ref_id, bm.version]], bm)
+          migrate_profiles(benchmarks[[bm.ref_id, bm.version]], bm)
+          bm.destroy! # profiles, rules
+        else
+          benchmarks[[bm.ref_id, bm.version]] = bm
+        end
+      end
+    end
```
dLobatog: Can we get this block out into a service? It would allow to get some tests to ensure this works on all scenarios and ensure we don't break this with future commits before it gets deployed to prod.

test/services/duplicate_benchmark_resolver_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    (@dupe_bm = benchmarks(:one).dup).save(validate: false)
+
+    @dupe_rules = rules.map do |r|
+      dupe_r = r.dup
+      dupe_r.benchmark = @dupe_bm
+      dupe_r.save(validate: false)
+      dupe_r
+    end
+
+    @dupe_profiles = profiles.map do |p|
```
akofink: We want to test having the same profiles and rules in two different benchmarks

test/services/duplicate_benchmark_resolver_test.rb
```diff
@@ -0,0 +1,56 @@
...
+      dupe_p = p.dup
+      dupe_p.benchmark = @dupe_bm
+      dupe_p.rules = [@dupe_rules.sample]
+      dupe_p.save(validate: false)
+      dupe_p
+    end
+
+    @dupe_test_results = test_results.map do |tr|
+      dupe_tr = tr.dup
```
akofink: We also want to test having the same test result twice in a single benchmark.

akofink:
Actually, let's wait a sec here. I think the added fixtures are messing with other tests.

akofink:
Ok - should be good to go now. I changed it to just use existing fixtures.

PR #292, opened 2020-01-08
Use real X-RH-IDENTITY coming from original request

dLobatog:
There is no need to send the account fake identity header when we can send directly the request header.

### DIFF
--- a/spec/swagger_helper.rb
+++ b/spec/swagger_helper.rb
@@ -159,7 +159,7 @@ RSpec.configure do |config|
 end
 
 def encoded_header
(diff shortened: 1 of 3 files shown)


spec/swagger_helper.rb
```diff
@@ -159,7 +159,7 @@
 end
 
 def encoded_header
-  Base64.encode64(x_rh_identity.to_json)
+  Base64.strict_encode64(x_rh_identity.to_json)
```
dLobatog: @akofink This was missing - `strict` removes linebreaks which was apparently causing problems with Swagger specs

PR #295, opened 2020-01-15
Add kafka producer for payload tracker

akofink:
https://projects.engineering.redhat.com/browse/RHICOMPL-432


app/producers/payload_tracker.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# A Kafka producer client for payload-tracker
+# https://www.rubydoc.info/gems/ruby-kafka/Kafka/Producer
+class PayloadTracker < Kafka::Client
+  BROKERS = [ENV['KAFKAMQ']].compact.freeze
+  CLIENT_ID = 'compliance-payload-tracker-producer'
+  TOPIC = 'platform.payload-status'
```
dLobatog: I don't know if this is the payload that's actually in the cluster - double check with QE for this. The repo says that https://github.com/RedHatInsights/payload-tracker - but I had to change this to `payload_tracker` to make it work

app/consumers/compliance_reports_consumer.rb
```diff
@@ -29,6 +31,14 @@ def send_validation(validation)
 
   private
 
+  def notify_payload_tracker(status, status_msg = '')
+    PayloadTracker.deliver(
+      account: @msg_value['account'], system_id: @msg_value['id'],
+      payload_id: @msg_value['payload_id'], status: status,
```
dLobatog: This should be `@msg_value['request_id']` - there is no `payload_id` field and the delivery always fails if the field is missing

app/jobs/parse_report_job.rb
```diff
@@ -25,4 +21,29 @@ def cancelled?
...
+
+  def parser
+    @parser ||= XccdfReportParser.new(ActiveSupport::Gzip.decompress(@file),
+                                      @msg_value)
+  end
+
+  def notify_payload_tracker(status, status_msg = '')
+    PayloadTracker.deliver(
+      account: @msg_value['account'], system_id: @msg_value['id'],
+      payload_id: @msg_value['payload_id'], status: status,
```
dLobatog: `payload_id: @msg_value['request_id']`

app/producers/payload_tracker.rb
dLobatog: Checked on openshift and this TOPIC is correct - nothing to change here :smile:

PR #296, opened 2020-01-20
Allow to consume multiple reports from a tar archive

dLobatog:
This is meant to support upload of multiple reports at once in a tar -  https://github.com/RedHatInsights/insights-core/pull/2415


app/consumers/compliance_reports_consumer.rb
```diff
@@ -42,14 +42,14 @@ def download_file
...
+
+    logger.info "Received message, enqueueing: #{@msg_value}"
+    @report_contents.each do |report|
       job = ParseReportJob.perform_async(
-        ActiveSupport::Gzip.compress(@report_contents), @msg_value
+        ActiveSupport::Gzip.compress(report), @msg_value
       )
       logger.info "Message enqueued: #{message_id} as #{job}"
-    else
-      logger.error "Error parsing report: #{message_id}"
```
dLobatog: Notice this is the *exact* same error as the one that's logged in `validation_message`  if there's a failure. It makes no sense to log the same thing twice.

app/services/reports_tar_reader.rb
```diff
@@ -0,0 +1,25 @@
...
+require 'rubygems/package'
+require 'zlib'
+
+# Finds the files in a insights tar and returns them as an array
+class ReportsTarReader
+  def initialize(file)
+    @file = file
+  end
+
+  def reports
```
dLobatog: I'll add a sample tarfile here created with `insights-core` to test this one

app/services/reports_tar_reader.rb
```diff
@@ -0,0 +1,25 @@
...
+    tar_extract = Gem::Package::TarReader.new(Zlib::GzipReader.open(@file))
+    tar_extract.rewind # The extract has to be rewinded after every iteration
+    files = []
+    tar_extract.each do |entry|
+      files << entry.read
+    end
+    tar_extract.close
```
akofink: This should probably be in an `ensure` block at the end of this method.

akofink: This seems like a great opportunity to use `yield` and wrap all this TarReader stuff in its own method which can contain the `ensure` to close the archive.

app/consumers/compliance_reports_consumer.rb
```diff
@@ -60,7 +60,9 @@ def validate
   end
 
   def validation_message
-    XccdfReportParser.new(@report_contents, @msg_value)
+    @report_contents.each do |report|
+      XccdfReportParser.new(report, @msg_value)
+    end
```
akofink: Should we allow partial processing of an archive if one of the reports fails? The StandardError rescue below will halt parsing for all remaining reports in the archive if that error is raised for any one report.

PR #297, opened 2020-01-20
RHICOMPL-454 - Find latest benchmarks

dLobatog:
This should be used by the frontend to only display the latest benchmarks available on the CreatePolicy wizard

### DIFF
--- a/app/models/xccdf/benchmark.rb
+++ b/app/models/xccdf/benchmark.rb
@@ -31,11 +31,15 @@ module Xccdf
       end
 
       def latest
-        all.map(&:ref_id).uniq.map { |ref_id| find_latest(ref_id) }
+        select(
+          'DISTINCT ref_id, version, id, title'
+        ).group_by(&:ref_id).map do |_, benchmarks|
+          find_latest(benchmarks)
+        end
       end
 
-      def find_latest(ref_id)
-        ::Xccdf::Benchmark.where(ref_id: ref_id).max_by do |benchmark|
+      def find_latest(benchmarks)
+        benchmarks.max_by do |benchmark|
           Gem::Version.new(benchmark.version)
         end
       end
--- a/app/models/xccdf/benchmark.rb
+++ b/app/models/xccdf/benchmark.rb
@@ -29,6 +29,16 @@ module Xccdf
       def policy_class
         BenchmarkPolicy
       end
+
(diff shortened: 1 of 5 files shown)


test/fixtures/benchmarks.yml
```diff
@@ -4,4 +4,4 @@ one:
   ref_id: xccdf_org.ssgproject.content_benchmark_RHEL-7
   title: Benchmark One
   description: The first benchmark
-  version: v0.1.43
+  version: 0.1.43
```
akofink: Nice catch. We could add a format validation on `version`. I think it'd be a good idea (in a different PR).

app/models/xccdf/benchmark.rb
```diff
@@ -29,6 +29,16 @@ def from_openscap_parser(op_benchmark)
       def policy_class
         BenchmarkPolicy
       end
+
+      def latest
+        all.map(&:ref_id).uniq.map { |ref_id| find_latest(ref_id) }
```
akofink: Let's try to leverage the DB as much as possible.

```suggestion
        select(:ref_id).distinct.find_each.map { |ref_id| find_latest(ref_id) }
```

app/models/xccdf/benchmark.rb
```diff
@@ -29,6 +29,16 @@ def from_openscap_parser(op_benchmark)
...
+      end
+
+      def find_latest(ref_id)
+        ::Xccdf::Benchmark.where(ref_id: ref_id).max_by do |benchmark|
```
akofink:
```suggestion
        where(ref_id: ref_id).max_by do |benchmark|
```

app/models/xccdf/benchmark.rb
```diff
@@ -29,6 +29,20 @@ def from_openscap_parser(op_benchmark)
       def policy_class
         BenchmarkPolicy
       end
+
+      def latest
+        select(
+          'DISTINCT ref_id, version, id, title'
```
dLobatog: I had to do it like this because your suggestion would limit the fields retrieved to just ref_id. That's fine, but I need `version` later to compare versions en `Gem::Version` and also in GraphQL, a user should be able to query these fields at least (I didn't include description as we shouldn't need to query that for a SSG)

PR #300, opened 2020-01-24
Change payload_id to request_id

dLobatog:


app/producers/payload_tracker.rb
```diff
@@ -14,7 +14,7 @@ def deliver(payload_id:, status:, account:, system_id:, status_msg: nil)
         date: DateTime.now.iso8601, service: SERVICE,
         account: account, system_id: system_id,
         payload_id: payload_id, status: status,
-        status_msg: status_msg
+        request_id: payload_id, status_msg: status_msg
```
dLobatog: This is on purpose - otherwise a new request_id will be generated on payload_tracker and we will not be able to track it further

PR #301, opened 2020-01-30
Notify Slack on Rake job exception

dLobatog:
ExceptionNotification only notifies on 500 errors on the controllers. For Rake jobs, we have to rescue any error and send it over to Slack. We have missed some errors on jobs without this.


Victoremepunto:
Looks good, but does this identify the environment where the notification comes from ? . we will receiving notifications from `CI`, `QA` and `prod`, at the same time

PR #305, opened 2020-02-05
RHICOMPL-476 - Include search in metadata and fix total count

dLobatog:


app/controllers/concerns/metadata.rb
```diff
@@ -13,7 +13,7 @@ def set_headers
     end
 
     def metadata(opts = {})
-      opts[:total] ||= policy_scope(resource).count
+      opts[:total] ||= scope_search.count
```
dLobatog: Notice the definition of this method doesn't include the search if it's not in the request

app/controllers/concerns/metadata.rb
```diff
@@ -61,5 +59,15 @@ def next_offset(last_offset)
...
       (total / pagination_limit.to_f).ceil
     end
+
+    private
+
+    def base_link
+      base_url = "#{path_prefix}/#{ENV['APP_NAME']}"\
+                 "/#{controller_name}"
+      base = "#{base_url}?limit=#{pagination_limit}"
+      base += "&search=#{params[:search]}" if params[:search].present?
```
dLobatog: This line is the critical one here - it adds the search param everywhere

test/controllers/concerns/metadata_test.rb
```diff
@@ -16,10 +16,19 @@ def authenticate
...
+    3.times do
+      Profile.create(ref_id: "foo#{SecureRandom.uuid}", name: SecureRandom.uuid,
+                     benchmark: benchmarks(:one), hosts: [hosts(:one)],
+                     account: accounts(:test))
+    end
     profiles(:one).update(account: accounts(:test), hosts: [hosts(:one)])
-    search_query = 'name=profile1'
-    get profiles_url, params: { search: search_query }
+    search_query = 'ref_id~foo'
+    get profiles_url, params: { search: search_query, limit: 1, offset: 2 }
```
dLobatog: Line 32 of this test was supposed to be testing the total count using the search param, the test passed in the past because there was only 1 profile in the db. So I'm making the db and results a bit larger so that if you remove the changes I made on the controller to compute the count based on the scope + search, the test would fail.

PR #306, opened 2020-02-05
RHICOMPL-420 - Add references, identifier to Rules scoped_search

dLobatog:
To search, you may use `api/compliance/rules?search=reference=4.3.3.5.4`

### DIFF
--- a/app/policies/rule_policy.rb
+++ b/app/policies/rule_policy.rb
@@ -14,10 +14,12 @@ class RulePolicy < ApplicationPolicy
   # (profile account ID = rule account ID)
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      ids = scope.all.select do |rule|
-        rule.profiles.pluck(:account_id).include? user.account_id
-      end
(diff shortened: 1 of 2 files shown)


app/policies/rule_policy.rb
```diff
@@ -14,10 +14,12 @@ def show?
   # (profile account ID = rule account ID)
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      ids = scope.all.select do |rule|
-        rule.profiles.pluck(:account_id).include? user.account_id
-      end
-      scope.where(id: ids)
+      return Rule.where('false') if user.account_id.blank?
```
dLobatog: Not strictly a necessary change but the performance is drastically improved, especially in prod where each rule has 100s of profiles. This returns the result in 0.5s vs the N+1 query we had here before, which took 8s to respond because of this.

dLobatog: This API is hit continuously by the remediations API to get rules description and titles, so improving this should improve the overall SLO too.

PR #307, opened 2020-02-05
RHICOMPL-446 - RBAC light implementation (#290)

dLobatog:
RBAC-light phase 2 has begun. This means RBAC light should be in QA and tested some time within the next 2 weeks. This looks like a fairly minor change to stable so hopefully it won't cause too much of a disruption.


dLobatog:
https://github.com/RedHatInsights/compliance-backend/pull/292 should also be included here

dLobatog:
Failing because e2e does not include RBAC yet.

PR #317, opened 2020-02-17
Empty commit to trigger stable image creation

dLobatog:


Victoremepunto:
Looks like this should work as `smoke` already deactivates `rbac`:
https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/compliance/env/smoke.yml#L10

PR #318, opened 2020-02-18
Create host from inventory on policy association

akofink:
Create host from the inventory on policy assocate_systems only if the host doesn't already exist in compliance.

Related to https://projects.engineering.redhat.com/browse/RHICOMPL-306
Reported at https://projects.engineering.redhat.com/browse/RHICOMPL-485

Requires https://github.com/RedHatInsights/compliance-backend/pull/319

### DIFF
--- a/app/graphql/mutations/profile/associate_systems.rb
+++ b/app/graphql/mutations/profile/associate_systems.rb
@@ -24,15 +24,51 @@ module Mutations
 
       def find_profile(profile_id)
         ::Pundit.authorize(
-          context[:current_user],
+          current_user,
           ::Profile.find(profile_id),
           :edit?
         )
       end
 
       def find_hosts(system_ids)
-        ::Pundit.policy_scope(context[:current_user], ::Host)
-                .where(id: system_ids)
+        existing_systems = ::Pundit.policy_scope(current_user, ::Host)
+                                   .where(id: system_ids)
+        save_hosts(system_ids - existing_systems.pluck(:id))
+        existing_systems
+      end
+
+      def save_hosts(ids)
+        ids.map do |id|
+          save_host(id)
+        end
+      end
+
+      def save_host(id)
+        i_host = inventory_host(id)
+        host = ::Host.find_or_initialize_by(
+          id: i_host['id'],
+          account_id: current_user.account.id
+        )
+
+        host.update!(
+          name: i_host['fqdn']
+        )
+
+        host
+      end
+
+      def current_user
+        @current_user ||= context[:current_user]
+      end
+
+      def inventory_host(id)
+        ::HostInventoryAPI.new(
+          id,
+          nil, # unknown hostname
+          current_user.account,
+          ::Settings.host_inventory_url,
+          nil # infer identity from account
+        ).inventory_host
       end
     end
   end
--- a/app/services/host_inventory_api.rb
+++ b/app/services/host_inventory_api.rb
@@ -11,7 +11,7 @@ class HostInventoryAPI
     @hostname = hostname
     @url = "#{URI.parse(url)}#{ENV['PATH_PREFIX']}/inventory/v1/hosts"
     @account = account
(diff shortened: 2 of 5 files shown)


app/services/host_inventory_api.rb
```diff
@@ -45,7 +45,7 @@ def inventory_host
   def find_results(body)
     body['results'].find do |host|
       host['account'] == @account.account_number && (
-        host['id'] == @host_inventory_id ||
+        host['id'] == @id ||
```
akofink: This was never correct (`@host_inventory_id` was always `nil`). We've been matching on hostname this whole time for report parsing.

app/services/host_inventory_api.rb
```diff
@@ -36,7 +36,7 @@ def create_host_in_inventory
 
   def inventory_host
     @inventory_host ||= host_already_in_inventory(@id) ||
-                        host_already_in_inventory(@hostname) ||
+                        (@hostname && host_already_in_inventory(@hostname)) ||
```
akofink: This allows skipping the hostname check if we don't know the hostname. Just pass nil to the HostInventoryApi.

app/graphql/mutations/profile/associate_systems.rb
```diff
@@ -24,15 +24,51 @@ def resolve(args = {})
...
           :edit?
         )
       end
 
       def find_hosts(system_ids)
-        ::Pundit.policy_scope(context[:current_user], ::Host)
-                .where(id: system_ids)
+        existing_systems = ::Pundit.policy_scope(current_user, ::Host)
+                                   .where(id: system_ids)
+        save_hosts(system_ids - existing_systems.pluck(:id))
```
dLobatog: I guess you do this just to avoid a few unnecessary requests to the inventory?

app/graphql/mutations/profile/associate_systems.rb
```diff
@@ -24,15 +24,51 @@ def resolve(args = {})
...
+
+      def save_hosts(ids)
+        ids.map do |id|
+          save_host(id)
+        end
+      end
+
+      def save_host(id)
+        i_host = inventory_host(id)
+        host = ::Host.find_or_initialize_by(
```
dLobatog: This `find` is only meant to happen to avoid a host being initialized twice (two concurrent mutations) or when would that happen?

app/services/host_inventory_api.rb
```diff
@@ -11,7 +11,7 @@ def initialize(id, hostname, account, url, b64_identity)
     @hostname = hostname
     @url = "#{URI.parse(url)}#{ENV['PATH_PREFIX']}/inventory/v1/hosts"
     @account = account
-    @b64_identity = b64_identity
+    @b64_identity = b64_identity || @account.b64_identity
```
dLobatog: You could probably make then `b64_identity` an optional argument from now on

app/services/host_inventory_api.rb
dLobatog: Makes sense, thanks

app/graphql/mutations/profile/associate_systems.rb
akofink: Right - no need to find from inventory and save a host we already have in our DB.

app/graphql/mutations/profile/associate_systems.rb
akofink: It's a safeguard that _shouldn't_ ever be hit, but there could be a race if the host is created between lines 47 and 48 in some other process.

app/services/host_inventory_api.rb
akofink: It's implicitly optional (you may pass nil) like this. I was trying to avoid changing the constructor signature and every usage in the code base, but I can do that if you'd like.

akofink:
Updated to resolve conflicts

PR #319, opened 2020-02-18
Find inventory systems by passed ID before hostname

akofink:
Finding hosts by ID has never worked correctly until this change. It has been verifying a match with the inventory only based on the fqdn returned by the insights-client metadata.

Without the change, the ID test fails:

```
 FAIL["test_find_results_matches_on_ID", #<Minitest::Reporters::Suite:0x000055a0d359d5c0 @name="HostInventoryApiTest">, 0.18526139999812585]
 test_find_results_matches_on_ID#HostInventoryApiTest (0.19s)
        find_results did not return the expected match by ID.
        Expected nil to not be nil.
        test/services/host_inventory_api_test.rb:56:in `block in <class:HostInventoryApiTest>'
```

### DIFF
--- a/test/services/host_inventory_api_test.rb
+++ b/test/services/host_inventory_api_test.rb
@@ -51,4 +51,28 @@ class HostInventoryApiTest < ActiveSupport::TestCase
     @api.expects(:create_host_in_inventory).returns(@host)
     assert_equal @host, @api.inventory_host
   end
+
+  test 'find_results matches on ID' do
+    assert_equal(
+      @api.send(
+        :find_results, 'results' => [
+          { 'account' => @account.account_number, 'id' => @host.id }
+        ]
+      )['id'],
+      @host.id,
+      'find_results did not return the expected match by ID'
+    )
+  end
+
+  test 'find_results matches on hostname' do
+    assert_equal(
+      @api.send(
+        :find_results, 'results' => [
+          { 'account' => @account.account_number, 'fqdn' => @host.name }
+        ]
+      )['fqdn'],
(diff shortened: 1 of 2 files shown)


test/services/host_inventory_api_test.rb
```diff
@@ -51,4 +51,20 @@ class HostInventoryApiTest < ActiveSupport::TestCase
     @api.expects(:create_host_in_inventory).returns(@host)
     assert_equal @host, @api.inventory_host
   end
+
+  test 'find_results matches on ID' do
+    assert_not_nil @api.send(
```
bastilian: 🤔 (there is no friendly suspicious looking emoji)

Asserting that it returns a result with the matching `id` instead of `not_nil` would be much more appropiate.

akofink: This is checking that, given this response with the matching ID as set up on line 12, the find_results method properly matches based on ID. Checking again that the ID matches doesn't actually test any code, as that is exactly what find_results is doing.

dLobatog: It's true whatever the `find_results` method returns is irrelevant, as in `host_already_in_inventory` we're not even checking the response, we just return the result if there was some matching, or nil otherwise. 

It'd make sense to me too to at least add the check here that `find_results` is matching the right host, and perhaps also making sure `host_already_in_inventory`  does too, it is not supposed to break any tests.

akofink: Ok, I've updated these tests to validate the ID or name matches

PR #320, opened 2020-02-18
RHICOMPL-292 - Endpoint to associate multiple policies to a host

dLobatog:
Can be tested alongside https://github.com/RedHatInsights/compliance-frontend/pull/371


test/graphql/mutations/associate_profiles_test.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class AssociateSystemsMutationTest < ActiveSupport::TestCase
```
akofink: This should fix the tests:

```suggestion
class AssociateProfilesMutationTest < ActiveSupport::TestCase
```

PR #323, opened 2020-02-19
Determine canonical profiles based on parent_profile presence

akofink:
With the added parent_profile_id field, we no longer need to check existence of an account to determine if a profile is canonical or not.

### DIFF
--- a/test/graphql/queries/profile_query_test.rb
+++ b/test/graphql/queries/profile_query_test.rb
@@ -40,7 +40,8 @@ class ProfileQueryTest < ActiveSupport::TestCase
       }
     GRAPHQL
 
-    profiles(:one).update account: accounts(:test)
+    profiles(:one).update account: accounts(:test),
+                          parent_profile: profiles(:two)
     users(:test).update account: nil
 
     assert_raises(Pundit::NotAuthorizedError) do
--- a/test/services/xccdf_report_parser_test.rb
+++ b/test/services/xccdf_report_parser_test.rb
@@ -262,6 +262,7 @@ class XccdfReportParserTest < ActiveSupport::TestCase
         benchmark: @report_parser.benchmark,
         name: @report_parser.op_profiles.first.name,
         account: accounts(:one),
+        parent_profile: profiles(:one),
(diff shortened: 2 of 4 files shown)


test/graphql/queries/profile_query_test.rb
```diff
@@ -40,7 +40,8 @@ class ProfileQueryTest < ActiveSupport::TestCase
       }
     GRAPHQL
 
-    profiles(:one).update account: accounts(:test)
+    profiles(:one).update account: accounts(:test),
+                          parent_profile: profiles(:two)
```
akofink: Canonical (`parent_profile == nil`) profiles are always visible by everyone based on our current Pundit policy. We need to write a data migration to set parent_profile, matched via benchmark and ref_id, on existing profiles that are associated with an account, or this Pundit policy will give unauthorized access.

test/services/xccdf_report_parser_test.rb
```diff
@@ -262,6 +262,7 @@ class TestParser < ::XccdfReportParser
         benchmark: @report_parser.benchmark,
         name: @report_parser.op_profiles.first.name,
         account: accounts(:one),
+        parent_profile: profiles(:one),
```
akofink: Same thing here.

akofink:
This one's good for re-review

test/graphql/queries/profile_query_test.rb
dLobatog: I was going to merge this one to CI, but if I do, it looks like the policy will block the "create policy" workflow, as there will not be any policies with `parent_profile_id` set for any user. It looks like we need the migration either on this PR or some PR that is merged before this one in order to be able to merge without blocking that workflow.

akofink:
Makes sense. I'll start work on a migration so we can merge this.

akofink:
Here's the migration https://github.com/RedHatInsights/compliance-backend/pull/334

akofink:
This one should be good for merge.

akofink:
Ok sorry - tests are fixed!

PR #324, opened 2020-02-19
Tailoring file service accepts rule selections

akofink:
Test like this:

```rb
XccdfTailoringFile.new(
  profile: Profile.where.not(parent_profile_id: nil).first,
  rule_ref_ids: {'rule_ref_1' => true, 'rule_ref_2' => false}
).to_xml
```

which should return something like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<xccdf:Tailoring xmlns:xccdf="http://checklists.nist.gov/xccdf/1.2" id="xccdf_csfr-compliance_tailoring_default">
  <xccdf:benchmark id="xccdf_org.ssgproject.content_benchmark_RHEL-7"/>
  <xccdf:version time="2020-02-19T18:16:24+00:00">1</xccdf:version>
  <xccdf:Profile id="xccdf_org.ssgproject.content_profile_standard" extends="xccdf_org.ssgproject.content_profile_standard">
    <xccdf:title xmlns:xhtml="http://www.w3.org/1999/xhtml" xml:lang="en-US" override="true">Standard System Security Profile for Red Hat Enterprise Linux 7</xccdf:title>
    <xccdf:description xmlns:xhtml="http://www.w3.org/1999/xhtml" xml:lang="en-US" override="true">This profile contains rules to ensure standard security baseline
of a Red Hat Enterprise Linux 7 system. Regardless of your system's workload
all of these checks should pass.</xccdf:description>
    <xccdf:select idref="rule_ref_1" selected="true"/>
    <xccdf:select idref="rule_ref_2" selected="false"/>
  </xccdf:Profile>
</xccdf:Tailoring>
```

### DIFF
--- a/app/services/xccdf_tailoring_file.rb
+++ b/app/services/xccdf_tailoring_file.rb
@@ -6,7 +6,7 @@ class XccdfTailoringFile
 
   XCCDF = 'xccdf'
 
-  def initialize(profile:, rule_ref_ids: [], set_values: {})
+  def initialize(profile:, rule_ref_ids: {}, set_values: {})
     @profile = profile
     @rule_ref_ids = rule_ref_ids
     @set_values = set_values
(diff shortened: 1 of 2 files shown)


app/services/xccdf_tailoring_file.rb
```diff
@@ -20,13 +20,31 @@ def builder
     @builder ||= create_builder do |xml|
       xml[XCCDF].benchmark(id: @profile.benchmark.ref_id)
```
dLobatog: It looks like the benchmark elements needs a `href` to be validated by the oscap parser, otherwise you get the following error when testing:

```
OpenSCAP Error: File 'tailoring_c2s_with_backend.xml' line 3: Element '{http://checklists.nist.gov/xccdf/1.2}benchmark': The attribute 'href' is required but missing.
 [xccdf_session.c:614]
Invalid XCCDF Tailoring (1.2) content in tailoring_c2s_with_backend.xml. [oscap_source.c:343]
Invalid XCCDF Tailoring (1.2) content in tailoring_c2s_with_backend.xml [xccdf_session.c:1172]
```

akofink: I've added href now - it's not actually used for anything, but I'm using the inferred_os_major_version on a benchmark to create it in the format of `ssg-rhel7-ds.xml`.

app/services/xccdf_tailoring_file.rb
```diff
@@ -18,15 +18,41 @@ def initialize(profile:, rule_ref_ids: [], set_values: {})
...
-      xml[XCCDF].Profile(id: @profile.ref_id,
-                         extends: @profile.parent_profile.ref_id)
+      profile_builder(xml)
+    end
+  end
+
+  def benchmark_builder(xml)
+    xml[XCCDF].benchmark(
+      id: @profile.benchmark.ref_id,
+      version: @profile.benchmark.version,
```
akofink: I also added version, since we save this field. idk, it might be useful.

app/services/xccdf_tailoring_file.rb
```diff
@@ -18,15 +18,41 @@ def initialize(profile:, rule_ref_ids: [], set_values: {})
 
   def builder
     @builder ||= create_builder do |xml|
-      xml[XCCDF].benchmark(id: @profile.benchmark.ref_id)
+      benchmark_builder(xml)
```
akofink: btw, funnily enough, leaving off the benchmark completely is valid for the oscap eval run; I feel like we'll want to know which benchmark a tailoring file was generated from, though, so I'd rather have a kind of weird href value than no benchmark info at all.

akofink:
@dLobatog this one should be good for re-review

akofink:
@dLobatog it looks like maybe you're not using the latest version of this PR or maybe your scan wasn't against the most recently generated file. Your generated tailoring file should have both benchmark href and version now, and I don't see the version attribute. Additionally, your href attribute is not generated from my code - it should be of the format `ssg-rhelX-ds.xml` without the rest of the path. I'm not sure which tailoring file you're showing me.

akofink:
The `Profile` section of your test results will not change based on tailoring, but the TestResult will, of course.

PR #325, opened 2020-02-19
Update rswag specs to include limit/offset/search

akofink:
@Victoremepunto this one's for you :)

Just so we're all on the same page, the files in `spec` contain both the actual API definition as well as some tests for the API. This is just the way rswag works, but it's still an API definition first model (i.e. we're not generating the API from our controller code in `app/`, and there's no guarantee that our API behaves the way we defined it). `bundle exec rake rswag` generates the `swagger/v1/openapi.json` from what's defined in `spec`. It's much nicer to be able to define resources separately (and in Ruby) than in one big yaml or json file manually.

@Victoremepunto compare against your changes in https://github.com/RedHatInsights/compliance-backend/pull/250

### DIFF
--- a/spec/integration/profiles_spec.rb
+++ b/spec/integration/profiles_spec.rb
@@ -8,10 +8,12 @@ describe 'Profiles API' do
       fixtures :profiles
       tags 'profile'
       description 'Lists all profiles requested'
-      consumes 'application/vnd.api+json'
-      produces 'application/vnd.api+json'
       operationId 'ListProfiles'
-      parameter name: :'X-RH-IDENTITY', in: :header, schema: { type: :string }
+
+      content_types
+      auth_header
+      pagination_params
+      search_params
 
       response '200', 'lists all profiles requested' do
         let(:'X-RH-IDENTITY') { encoded_header }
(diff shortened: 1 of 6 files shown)


spec/integration/profiles_spec.rb
```diff
@@ -8,10 +8,12 @@
...
       description 'Lists all profiles requested'
-      consumes 'application/vnd.api+json'
-      produces 'application/vnd.api+json'
       operationId 'ListProfiles'
-      parameter name: :'X-RH-IDENTITY', in: :header, schema: { type: :string }
+
+      content_types
+      auth_header
+      pagination_params
+      search_params
```
akofink: Common things like this are now defined in `swagger_helper.rb` to DRY out the specs.

swagger/v1/openapi.json
```diff
@@ -291,7 +423,7 @@
                   "$ref": "#/definitions/links"
                 },
                 "data": {
-                  "type": "array",
+                  "type": "object",
```
Victoremepunto: I think this "array" object is correct, and what was required was to specify the "type: object" for each of the elements in the array : https://github.com/RedHatInsights/compliance-backend/pull/250/files#diff-bbc70b65028c3e55922508735e913f6eR275

In addition, can we update the openapi.json to be OpenAPI ver. 3 ? ( and keep the .json instead of .yaml file ? )

dLobatog: The current rswag gem does not support that yet, so we have to make the conversion to v3 and JSON manually. However, there is a fork we might be able to use for the time being https://github.com/jdanielian/open-api-rswag

Victoremepunto: okay, anything that helps in the meantime to get an updated `openapi V3` file will work I guess.

dLobatog:
I'll be updating the openapi v3 we have in https://github.com/RedHatInsights/compliance-backend/tree/master/swagger/v1 soon

akofink:
Ok - I was going to also look into how we can output that file in json with rswag

PR #326, opened 2020-02-24
Remove destructive dependency and remove results only optionally

bastilian:
See https://github.com/RedHatInsights/compliance-frontend/pull/392 for the UI part that uses this.

### DIFF
--- a/app/graphql/mutations/profile/delete.rb
+++ b/app/graphql/mutations/profile/delete.rb
@@ -17,7 +17,6 @@ module Mutations
           :destroy?
         )
         profile.delete_all_test_results = args[:delete_all_test_results]
-        profile.inspect
         profile.destroy!
         { profile: profile }
       end
--- a/app/graphql/mutations/profile/delete.rb
+++ b/app/graphql/mutations/profile/delete.rb
@@ -7,6 +7,7 @@ module Mutations
       graphql_name 'deleteProfile'
 
       argument :id, ID, required: true
+      argument :delete_all_test_results, Boolean, required: false
       field :profile, Types::Profile, null: false
 
       def resolve(args = {})
@@ -15,6 +16,7 @@ module Mutations
           ::Profile.find(args[:id]),
           :destroy?
         )
+        profile.delete_all_test_results = args[:delete_all_test_results]
         profile.destroy!
         { profile: profile }
       end
--- a/app/graphql/mutations/profile/delete.rb
+++ b/app/graphql/mutations/profile/delete.rb
@@ -17,6 +17,7 @@ module Mutations
           :destroy?
         )
         profile.delete_all_test_results = args[:delete_all_test_results]
+        profile.inspect
         profile.destroy!
         { profile: profile }
       end
--- a/app/jobs/delete_test_results_job.rb
+++ b/app/jobs/delete_test_results_job.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Job deletes test results of a specified profile
+class DeleteTestResultsJob
+  include Sidekiq::Worker
+
+  def perform(profile_id)
+    TestResult.where(profile_id: profile_id).delete_all
+  end
+end
--- a/app/jobs/delete_test_results_job.rb
+++ b/app/jobs/delete_test_results_job.rb
@@ -5,6 +5,6 @@ class DeleteTestResultsJob
   include Sidekiq::Worker
 
   def perform(profile_id)
-    TestResult.where(profile_id: profile_id).delete_all
+    TestResult.where(profile_id: profile_id).destroy_all
   end
 end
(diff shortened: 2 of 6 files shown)


test/graphql/mutations/delete_profile_mutation_test.rb
```diff
@@ -3,23 +3,29 @@
...
+      }
+  GRAPHQL
 
+  setup do
     users(:test).update account: accounts(:test)
     profiles(:one).update(account: accounts(:test))
+    result = test_results(:one).update(
+      profile: profiles(:one), host: hosts(:one)
+    )
+    result.save
```
bastilian: Locally, some of these tests won't succeed, while on Jenkins this raises errors, with results appearently just being a trueclass....
```
ERROR["test_delete_a_profile_and_test_results_when_deleteAllTestResults_is_set", #<Minitest::Reporters::Suite:0x000055bec18afec8 @name="DeleteProfileMutationTest">, 4.455383786000311]

 test_delete_a_profile_and_test_results_when_deleteAllTestResults_is_set#DeleteProfileMutationTest (4.46s)

NoMethodError:         NoMethodError: undefined method `save' for true:TrueClass
...
```

app/graphql/mutations/profile/delete.rb
```diff
@@ -15,6 +16,8 @@ def resolve(args = {})
           ::Profile.find(args[:id]),
           :destroy?
         )
+        profile.delete_all_test_results = args[:delete_all_test_results]
+        profile.inspect
```
dLobatog: Debugging leftovers?

app/models/profile.rb
```diff
@@ -54,6 +59,10 @@ def destroy_orphaned_business_objective
     business_objective.destroy if business_objective.profiles.empty?
   end
 
+  def destroy_all_test_results
+    test_results.destroy_all
```
dLobatog: Could we make a job out of this and submit it on deletion? The problem I see with this is that unfortunately some pods are OOM when massive deletions like this happen (this would trigger the deletion *with callbacks and associations* of 1500 RuleResults + 10 TestResults assuming reports with 150 rules)

app/models/profile.rb
```diff
@@ -22,6 +26,7 @@ class Profile < ApplicationRecord
   validates :account, presence: true, if: -> { hosts.any? }
 
   after_update :destroy_orphaned_business_objective
+  before_destroy :destroy_all_test_results, if: -> { delete_all_test_results }
```
dLobatog: Why setting an attribute instead of manually calling `profile.test_results.destroy_all` and submitting the job I mention below as part of the mutation?

test/graphql/mutations/delete_profile_mutation_test.rb
```diff
@@ -28,4 +34,30 @@ class DeleteProfileMutationTest < ActiveSupport::TestCase
       assert_equal profiles(:one).id, result['id']
     end
   end
+
+  test 'delete a profile provided an ID but not its test_results' do
```
dLobatog: :+1:

app/models/profile.rb
dLobatog: We've seen this happening when in QE anyone calls https://github.com/RedHatInsights/compliance-backend/blob/master/app/controllers/systems_controller.rb#L22 - removing all RuleResults is a tough task that can make a pod OOM

app/graphql/mutations/profile/delete.rb
bastilian: yep.

app/models/profile.rb
bastilian: The reason is that it doesn't feel like the responsibility of the mutation to remove a profiles test results.
 It shouldn't even know about the `delete_all_test_results` attribute, actually. At the moment it just passes it on, so its ...fine. :)

The way I see it, the mutation and GraphQL endpoint is the mediatior between User intput and Business Logic. And so, to me it's ok for it to know how to delete a `Profile`, but it shouldn't know how to delete their `TestResult`s, that is only for the Profile to know.

app/models/profile.rb
bastilian: 👍 Will turn this into running in a job.

bastilian:
@dLobatog Thanks for testing! It's now updated to run the deletion in a job. And tests are passing.

app/jobs/delete_test_results_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Job deletes test results of a specified profile
+class DeleteTestResultsJob
+  include Sidekiq::Worker
+
+  def perform(profile_id)
+    TestResult.where(profile_id: profile_id).delete_all
```
dLobatog: @bastilian I think this one needs to be `destroy_all` to trigger the `has_many :rule_results, dependent :delete_all`

PR #332, opened 2020-02-26
Search by references with external method

dLobatog:


app/controllers/concerns/search.rb
```diff
@@ -7,8 +7,16 @@ module Search
   included do
     def scope_search
       result = policy_scope(resource)
+      escape_references_search
       result = result.search_for(params[:search]) if params[:search].present?
       result.paginate(page: pagination_offset, per_page: pagination_limit)
     end
+
+    def escape_references_search
+      return unless params[:search]&.match(/reference=/)
```
akofink: This regex isn't quite right. It needs to be able to match any [valid query](https://github.com/wvanbergen/scoped_search/wiki/Query-language) for a string field.

```suggestion
      return unless params[:search]&.match(/reference\s?(=|!=|<>|~|!~|\^|!\^)\s?/)
```

app/controllers/concerns/search.rb
```diff
@@ -7,8 +7,16 @@ module Search
...
+
+      search_term = CGI.escape(params[:search].split('=')[1])
```
akofink: CGI escape is overkill. We can just surround the search term in double quotes.
```suggestion
      search_term = "\"#{params[:search].split('=')[1]}\""
```

app/controllers/concerns/search.rb
akofink: This is also problematic if users don't specify `reference=`. Scoped search is meant to search every searchable field when just the keyword is given.

dLobatog: We can make it `only_explicit` though, and allow only `=` - otherwise we'll have to take into account every possibility in the API for what I believe is likely a corner case

akofink: There's no reason an API user shouldn't be able to search by references as a generic keyword; disabling that functionality with `only_explicit` just for this corner case doesn't seem like the best way to solve this. Having special behavior for this field will make the API more difficult to document and maintain. API users, including our own frontend, should just properly double-quote such search terms.

akofink: By the way, this search term is a collection, so users must use the 'includes' operator `^` and cannot use the string comparison operators mentioned above.

app/models/rule.rb
```diff
@@ -41,18 +42,32 @@ class Rule < ApplicationRecord
...
+                             rationale: op_rule.rationale,
+                             severity: op_rule.severity)
 
-    rule
+      rule
+    end
+
+    def search_by_reference(_, operator, value)
+      conditions = sanitize_sql_for_conditions(
+        ["rule_references.label #{operator} (?)", CGI.unescape(value)]
```
dLobatog: I've added now here `(?)` and pass the operator (scoped search automatically translates the operator into a SQL operator), so all possible operators work

test/models/rule_test.rb
```diff
@@ -41,4 +41,31 @@ class RuleTest < ActiveSupport::TestCase
                .include?(rules(:one)),
            'Expected rule not found by identifier'
   end
+
+  context 'references' do
```
akofink: The types of URLs we'll get depend on the browser used. For the record, though, `CGI.encode` is _not_ what happens in the browser (Firefox examples shown below, but Chrome seems the same), even though for decoding purposes, `CGI.decode` covers the most ground.

In: `https://ci.foo.redhat.com:1337/api/compliance/rules?&search=reference = 164.308(a)(1)(ii)(B)`
Browser fills: `https://ci.foo.redhat.com:1337/api/compliance/rules?&search=reference%20=%20164.308(a)(1)(ii)(B)`

PR #334, opened 2020-02-26
Migrate profiles using account_id to parent_profile_id

akofink:
Since the parent_profile migration hasn't made it past CI and given it's the latest migration in the repo, I propose we rollback CI 1 migration and re-migrate to get this data migration in. I think it's more consistent with how we've done this in the past than to create another migration.

### DIFF
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# A service class to associate profiles to their parent profiles
+class ParentProfileAssociator
+  class << self
+    def run!
+      Profile.transaction do
+        Profile.where.not(account: nil).find_each do |profile|
+          profile.update!(parent_profile: find_parent(profile))
+        end
+      end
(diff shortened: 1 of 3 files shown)


app/services/parent_profile_associator.rb
```diff
@@ -0,0 +1,31 @@
...
+        Profile.where.not(account: nil).find_each do |profile|
+          profile.update!(parent_profile: find_parent(profile))
+        end
+      end
+    end
+
+    private
+
+    def find_parent(profile)
+      parents = Profile.where(ref_id: profile.ref_id,
```
dLobatog: Just a nitpick, why not `Profile.find` so you just get 1 result?

akofink: I want to raise an error if the response is not 1 record. `find` only accepts an ID, not ref_id, account_id, and benchmark_id. `find` also only raises an error if no records are found; `find_by` is the same. Neither raises an error if multiple records are found.

PR #338, opened 2020-02-28
Add test result graphql endpoint

akofink:
This is a very early version that definitely needs work.


akofink:
This doesn't currently work at all. @dLobatog does anything here stick out in particular?

akofink:
Tests are :cactus:

PR #340, opened 2020-03-02
Find all rules by slug using LIKE

dLobatog:
Apparently our DB contains multiple slugged rules, one per benchmark. In that case, the regular `friendly.find` will not find it.  

Instead of that, when a user looks for a ref_id we can just find all of the ref_ids that look like it within the user scope. In that case, if you look for `xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies` but only `xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-98d17f35-6094-4071-b099-cb4bca02615d` is in the scope, it will still be found. 

```ruby
Rule.where(ref_id: "xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_tcp_syncookies").map(&:slug)
  Rule Load (3.6ms)  SELECT "rules".* FROM "rules" WHERE "rules"."ref_id" = $1  [["ref_id", "xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_tcp_syncookies"]]
=> ["xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-98d17f35-6094-4071-b099-cb4bca02615d",
 "xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-85c2e772-8b39-48ad-8146-2813ce479ee5",
 "xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies",
 "xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-7c478946-f34a-45eb-a086-05e4b214bb8d",
 "xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-9eb475b7-9c5e-41e6-b5b5-ee0ea9485031",
 "xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-5c0699a9-1ba9-4453-bcbf-4294eb8ddb68",
 "xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-e00e81f0-971b-4ce3-b628-654200888e44",
 "xccdf_org-ssgproject-content_rule_sysctl_net_ipv4_tcp_syncookies-d4e05553-08b2-4c37-8bf0-439bd8516843"]
```

@Victoremepunto @akofink I'm adding also `slug` to the serializer so it's easier for QE to find these broken slugs. To avoid this in the future, a proper fix IMO would be to use:
```ruby
  friendly_id :ref_id, :use => :scoped, :scope => :benchmark
``` 
, which would also include removing the uniqueness constraint on the slug column, and a migration to change that. I don't want to get that change in right now as that would be problematic to get to production. 

http://norman.github.io/friendly_id/file.Guide.html#Uniqueness

### DIFF
--- a/app/controllers/rules_controller.rb
+++ b/app/controllers/rules_controller.rb
@@ -29,15 +29,12 @@ class RulesController < ApplicationController
   end
 
   def search_by_ref_id
-    rule = canonical.where(
+    rule = Rule.canonical.where(
       'rules.slug LIKE ?',
       "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
     )
     raise ActiveRecord::RecordNotFound if rule.blank?
-    rule.first
-  end
 
-  def canonical
-    Rule.where(benchmark_id: ::Xccdf::Benchmark.latest.pluck(:id))
+    rule.first
   end
 end
--- a/app/controllers/rules_controller.rb
+++ b/app/controllers/rules_controller.rb
@@ -7,9 +7,12 @@ class RulesController < ApplicationController
   end
 
   def show
-    rule = ::Pundit.policy_scope(User.current, ::Rule).friendly.find(
-      params[:id]
-    )
+    rule = ::Pundit.policy_scope(User.current, ::Rule).where(
+      'rules.slug LIKE ?',
+      "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
+    ).first
+    raise ActiveRecord::RecordNotFound if rule.blank?
+
     render json: RuleSerializer.new(rule)
   end
 
--- a/app/controllers/rules_controller.rb
+++ b/app/controllers/rules_controller.rb
@@ -29,7 +29,7 @@ class RulesController < ApplicationController
   end
 
   def search_by_ref_id
-    rule = Rule.canonical.where(
+    rule = Rule.latest.where(
       'rules.slug LIKE ?',
       "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
     )
--- a/app/controllers/rules_controller.rb
+++ b/app/controllers/rules_controller.rb
@@ -7,10 +7,12 @@ class RulesController < ApplicationController
   end
 
   def show
-    rule = ::Pundit.policy_scope(User.current, ::Rule).where(
-      'rules.slug LIKE ?',
-      "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
-    ).first
+    rule = if ::UUID.validate(params[:id])
+             search_by_id
+           else
+             search_by_ref_id
+           end
+
     raise ActiveRecord::RecordNotFound if rule.blank?
 
     render json: RuleSerializer.new(rule)
@@ -20,5 +22,22 @@ class RulesController < ApplicationController
 
   def resource
     Rule
+  end
+
+  def search_by_id
+    ::Pundit.policy_scope(User.current, ::Rule).friendly.find(params[:id])
+  end
+
+  def search_by_ref_id
+    rule = canonical.where(
+      'rules.slug LIKE ?',
+      "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
+    )
+    raise ActiveRecord::RecordNotFound if rule.blank?
+    rule.first
+  end
+
+  def canonical
+    Rule.where(benchmark_id: ::Xccdf::Benchmark.latest.pluck(:id))
   end
 end
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -41,6 +41,10 @@ class Rule < ApplicationRecord
     joins(:profile_rules).where.not(profile_rules: { profile_id: nil }).distinct
   }
 
+  scope :canonical, lambda {
+    where(benchmark_id: ::Xccdf::Benchmark.latest.pluck(:id))
+  }
+
   def self.from_openscap_parser(op_rule, benchmark_id: nil)
     rule = find_or_initialize_by(ref_id: op_rule.id,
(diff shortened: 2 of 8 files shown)


app/controllers/rules_controller.rb
```diff
@@ -7,9 +7,12 @@ def index
...
 
   def show
-    rule = ::Pundit.policy_scope(User.current, ::Rule).friendly.find(
-      params[:id]
-    )
+    rule = ::Pundit.policy_scope(User.current, ::Rule).where(
+      'rules.slug LIKE ?',
+      "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
+    ).first
+    raise ActiveRecord::RecordNotFound if rule.blank?
```
akofink: IMO `show` must guarantee uniqueness.

```suggestion
    rules = ::Pundit.policy_scope(User.current, ::Rule).where(
      'rules.slug LIKE ?',
      "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
    )
    raise ActiveRecord::RecordNotFound unless rules.one?
```

app/controllers/rules_controller.rb
```diff
@@ -7,9 +7,12 @@ def index
   end
 
   def show
-    rule = ::Pundit.policy_scope(User.current, ::Rule).friendly.find(
-      params[:id]
-    )
+    rule = ::Pundit.policy_scope(User.current, ::Rule).where(
+      'rules.slug LIKE ?',
+      "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
```
akofink: Let's also be sure not to break our API here when requesting a Rule by ID rather than slug.

app/controllers/rules_controller.rb
```diff
@@ -7,9 +7,14 @@ def index
   end
 
   def show
-    rule = ::Pundit.policy_scope(User.current, ::Rule).friendly.find(
-      params[:id]
-    )
+    rule = if ::UUID.validate(params[:id])
```
dLobatog: This was easier and cleaner than 
`.where('rules.slug LIKE ? OR rules.id = ?', params[:id], params[:id])`, as we would have to validate anyway. Postgres seems to error out if you compare a String with a UUID type, which would happen if anyone searched for a ref_id

app/models/rule.rb
```diff
@@ -41,6 +41,10 @@ class Rule < ApplicationRecord
     joins(:profile_rules).where.not(profile_rules: { profile_id: nil }).distinct
   }
 
+  scope :canonical, lambda {
+    where(benchmark_id: ::Xccdf::Benchmark.latest.pluck(:id))
```
dLobatog: I didn't limit this to RHEL7 only because I found during testing that some number rules with remediations are part of the RHEL 8 benchmark only, e.g `xccdf_org-ssgproject-content_rule_package_audispd `

akofink: This makes sense, but I don't like the 'canonical' name. So far, that has meant anything that we've created rather than users. This also returns duplicate rules, which isn't super useful for our purposes. What if replaced this with a `latest` scope, or similarly named, that gives a unique list of rules from the latest benchmarks?

db/migrate/20200303102215_change_phony_benchmark.rb
```diff
@@ -0,0 +1,9 @@
+class ChangePhonyBenchmark < ActiveRecord::Migration[5.2]
+  def up
+    ::Xccdf::Benchmark.find_by(ref_id: 'phony_ref_id').update(version: '0.0.0')
+  end
+
+  def down
+    ::Xccdf::Benchmark.find_by(ref_id: 'phony_ref_id').update(version: 'v0.0.0')
+  end
+end
```
akofink: I'd appreciate a slightly more descriptive migration name. Also, down and up shouldn't really be symmetrical for this case (down should be no-op probably); if they should be though, we should just use `change` rather than up/down.

test/controllers/rules_controller_test.rb
```diff
@@ -19,16 +22,30 @@ class RulesControllerTest < ActionDispatch::IntegrationTest
...
 
-  test 'does not find a rule outside of the user scope' do
-    profiles(:one).update rules: [rules(:one)]
+  test 'finds a rule by ID' do
+    get rule_url(rules(:one).id)
+
+    assert_response :success
+  end
+
+  test 'finds rules not within the user scope but within canonical' do
```
akofink: :100:

app/models/rule.rb
akofink: This still returns duplicate rules, but I think it's fine as is.

db/migrate/20200303102215_change_phony_benchmark_version.rb
```diff
@@ -0,0 +1,7 @@
+class ChangePhonyBenchmarkVersion < ActiveRecord::Migration[5.2]
+  # The `.latest` method of ::Xccdf::Benchmark is using Gem::Version, which
+  # expects the version to have the same versioning syntax as a gemspec.
+  def change
+    ::Xccdf::Benchmark.find_by(ref_id: 'phony_ref_id').update(version: '0.0.0')
```
akofink: This isn't quite what I suggested. `down` should be no-op, don't you think? Also, we should use `update!` here to raise any errors.

dLobatog: @akofink Sure, I've updated it to `update!` - I thought this would be a no-op as Rails wouldn't try run this on rollback if it doesn't know how to do it (it skips changes that it doesn't know how to revert for its migration DSL)

PR #342, opened 2020-03-02
Allow downloading tailoring of a profile

akofink:
~It works without https://github.com/RedHatInsights/compliance-backend/pull/324 (merged), but you won't be able to generate any useful tailoring file with selected/unselected rules without it.~

### DIFF
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -7,12 +7,28 @@ class ProfilesController < ApplicationController
   end
 
   def show
-    profile = Profile.find(params[:id])
     authorize profile
     render json: ProfileSerializer.new(profile)
   end
 
+  def tailoring_file
+    return unless profile.tailored?
+
+    send_data XccdfTailoringFile.new(
+      profile: profile,
+      rule_ref_ids: profile.tailored_rule_ref_ids
+    ).to_xml, filename: tailoring_filename, type: Mime[:xml]
+  end
+
   private
+
+  def tailoring_filename
+    "#{profile.benchmark.ref_id}__#{profile.ref_id}__tailoring.xml"
+  end
+
+  def profile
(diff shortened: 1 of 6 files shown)


app/controllers/profiles_controller.rb
```diff
@@ -7,13 +7,29 @@ def index
...
+  end
+
   private
 
+  def tailoring_filename
+    "#{profile.benchmark.ref_id}__#{profile.ref_id}__tailoring.xml"
+  end
+
+  def profile
+    @profile ||= Profile.find(params[:id])
```
dLobatog: Can we change this one to `Pundit.policy_scope(current_user, ::Profile).find(params[:id])` and remove the `authorize` line? No need to do it in two steps and search in a larger table when we can restrict it to user owned profiles

app/controllers/profiles_controller.rb
```diff
@@ -7,13 +7,29 @@ def index
...
   def show
-    profile = Profile.find(params[:id])
     authorize profile
     render json: ProfileSerializer.new(profile)
   end
 
+  def tailoring_file
+    return unless profile.tailored?
+
+    send_data XccdfTailoringFile.new(
```
dLobatog: Just a thought, maybe we could include a link to this endpoint in the regular `show` endpoint - there's info on how to do this on https://github.com/Netflix/fast_jsonapi (doesn't need to be in this PR)

akofink: Seems useful. I'll create a card to track it: https://projects.engineering.redhat.com/browse/RHICOMPL-509

app/controllers/profiles_controller.rb
akofink: This PR addresses your comment: https://github.com/RedHatInsights/compliance-backend/pull/347

PR #349, opened 2020-03-09
Correct sidekiq configuration

bastilian:
Currently sidekiq is not configured correctly for running in a `rails` process.
This causes issues when trying to create a job from the backend, since it will try to connect to the default `127.0.0.1` address. 

Since the two blocks were identical it's now one in a lambda.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,46 +1,30 @@
-if $0.include?('sidekiq')
-  Sidekiq.configure_server do |config|
-    config.redis = {
-      url: "redis://#{Settings.redis_url}",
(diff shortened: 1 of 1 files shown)


config/initializers/sidekiq.rb
```diff
@@ -1,46 +1,30 @@
...
+    PrometheusExporter::Client.default.stop(wait_timeout_seconds: 10)
   end
+end
+
+if $0.include?('sidekiq')
+  Sidekiq.configure_server(&sidekiq_config)
+end
 
+if Rails.env != 'test'
+  Sidekiq.configure_client(&sidekiq_config)
```
dLobatog: For onlookers, here's the key. This LOC was not called before on compliance-backend

PR #350, opened 2020-03-10
Update stable with master (except for a breaking change)

dLobatog:
:wave: 

This PR contains all of the updates between `stable` and `master`. As far as I can tell, most of these will be compatible with what we have in prod. There is one change in master though, which I could not put in here:

https://github.com/RedHatInsights/compliance-backend/commit/28b9b9efbe5a7f071ae68d799858aa89e93a1076

This commit, if added to stable, would pose a problem in production, because users have no way to manually delete Policies there, so they have to be deleted when they are orphaned for now. 

There's a change: https://github.com/RedHatInsights/compliance-backend/commit/ef37da2fd342ac651a58fce38d88eebf275fec9b which fundamentally changes the way rule references and rule identifiers are pulled from the frontend. To put this change into prod, we would have to update `frontend-components-inventory-compliance` in the frontend to some version `> 0.0.33`.


Victoremepunto:
@akofink  @dLobatog  it seems there's some legit failure in the smoke tests. the test that is failing currently checks that a report that is sent without canonical facts is not created, yet the test fails because apparently it does gets created. this doesn't seem to happen on `master` branch though. any ideas ?

PR #352, opened 2020-03-11
RHICOMPL-502 - Tailoring mutation

dLobatog:
Given a list of refIds, tailor the set of rules in the policy to it.

Enables https://github.com/RedHatInsights/compliance-frontend/pull/423

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -110,7 +110,7 @@ class Profile < ApplicationRecord
     else
       new_profile.hosts << host unless new_profile.hosts.include?(host)
     end
-    new_profile.add_rules_from(profile: self)
+    new_profile.add_rules(rules.map(&:ref_id))
     new_profile
   end
 
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -119,12 +119,10 @@ class Profile < ApplicationRecord
                     benchmark_id: benchmark_id)
   end
 
-  def add_rules_from(profile: nil)
-    new_profile_rules = profile.profile_rules - profile_rules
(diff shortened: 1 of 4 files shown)


app/models/profile.rb
```diff
@@ -110,7 +110,7 @@ def clone_to(account: nil, host: nil)
     else
       new_profile.hosts << host unless new_profile.hosts.include?(host)
     end
-    new_profile.add_rules_from(profile: self)
+    new_profile.add_rules(rules.map(&:ref_id))
```
akofink:
```suggestion
    new_profile.add_rules(rules.pluck(:ref_id))
```

app/models/profile.rb
```diff
@@ -119,12 +119,10 @@ def in_account(account)
                     benchmark_id: benchmark_id)
   end
 
-  def add_rules_from(profile: nil)
-    new_profile_rules = profile.profile_rules - profile_rules
-    ProfileRule.import!(new_profile_rules.map do |profile_rule|
-      new_profile_rule = profile_rule.dup
-      new_profile_rule.profile = self
-      new_profile_rule
+  def add_rules(ref_ids)
```
akofink: Explicit naming goes a long way towards readability

```suggestion
  def add_rule_ref_ids(ref_ids)
```

or (my preference)

```suggestion
  def add_rules(ref_ids:)
```

Otherwise, the assumption is that this method is meant to receive Rule objects.

akofink:
You can test this and the frontend PR end to end with this client PR https://github.com/RedHatInsights/insights-core/pull/2498

PR #353, opened 2020-03-12
RHICOMPL-515 - Calculate score from test results

akofink:
I noticed this issue while trying to create REST API examples. The score for a profile is a simple average across all hosts' latest test_result scores. This PR contains a scope for latest TestResults, `TestResult.latest`, which we need for almost every TestResult query.

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -97,9 +97,9 @@ class Profile < ApplicationRecord
   end
 
   def score
-    return 1 if hosts.blank?
+    return 1 if test_results.latest.blank?
 
-    (hosts.count { |host| compliant?(host) }).to_f / hosts.count
+    (scores = test_results.latest.pluck(:score)).sum / scores.size
   end
 
   def clone_to(account: nil, host: nil)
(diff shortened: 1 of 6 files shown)


app/models/profile.rb
```diff
@@ -97,9 +97,9 @@ def results(host)
   end
 
   def score
-    return 1 if hosts.blank?
+    return 1 if test_results.latest.blank?
 
-    (hosts.count { |host| compliant?(host) }).to_f / hosts.count
+    (scores = test_results.latest.pluck(:score)).sum / scores.size
```
akofink: This change represents the actual fix of the bug I encountered in the API.

app/models/test_result.rb
```diff
@@ -12,9 +12,15 @@ class TestResult < ApplicationRecord
...
 
-  def self.latest(profile_id, host_id)
-    where(host_id: host_id, profile_id: profile_id)
-      .order('created_at DESC')
-      .includes(:rule_results).first
+  scope :latest, lambda {
+    joins("JOIN (#{latest_without_ids.to_sql}) as tr on "\
+          'test_results.profile_id = tr.profile_id AND '\
+          'test_results.host_id = tr.host_id AND '\
+          'test_results.end_time = tr.end_time')
```
akofink: This join is the SQL magic that makes `TestResult.latest` work. We should definitely add a DB constraint and index across these three fields.

PR #356, opened 2020-03-17
Add scoped search definition for system search

dLobatog:


app/models/profile.rb
```diff
@@ -7,6 +7,8 @@ class Profile < ApplicationRecord
   attribute :delete_all_test_results, :boolean, default: false
 
   scoped_search on: %i[id name ref_id account_id compliance_threshold]
+  scoped_search relation: :hosts, on: :id, aliases: [:system_ids, :systems]
+  scoped_search relation: :hosts, on: :name, aliases: [:system_names, :systems]
```
akofink: When you do this, it adds search terms for the `on: ` term (i.e. `id` or `name`) unless you use `rename` (`aliases` adds more aliases but does not rename the original).

dLobatog:
Cool, thanks @akofink  for the fast review - I'm adapting the response in the client a bit, as the format has changed a bit (e.g ref_id is fetched from `policy['attributes']` and id is fetched from `policy['id']` now)

PR #357, opened 2020-03-18
Associate parent profiles outside of own benchmark

dLobatog:
Some profiles cannot be associated to parents within its own benchmark,
such as profiles that come from the phony benchmark (which has no
canonical profiles).

### DIFF
--- a/app/services/duplicate_benchmark_resolver.rb
+++ b/app/services/duplicate_benchmark_resolver.rb
@@ -16,6 +16,9 @@ class DuplicateBenchmarkResolver
     private
 
     def migrate_benchmark(existing_bm, duplicate_bm)
+      logger.info(
+        "Duplicate benchmark found for #{existing_bm.id} - #{duplicate_bm.id}"
+      )
       migrate_rules(existing_bm, duplicate_bm)
       migrate_profiles(existing_bm, duplicate_bm)
       duplicate_bm.delete
@@ -37,13 +40,23 @@ class DuplicateBenchmarkResolver
     # we expect validations to fail due to nonunique rules/profiles
     # accept duplicate rules for now
     def migrate_rules(existing_bm, duplicate_bm)
+      logger.info(
+        "Migrating rules for #{existing_bm.id} - #{duplicate_bm.id}..."
+      )
       duplicate_bm.rules.update_all(benchmark_id: existing_bm.id)
     end
 
     # accept duplicate profiles for now
     def migrate_profiles(existing_bm, duplicate_bm)
+      logger.info(
+        "Migrating profiles for #{existing_bm.id} - #{duplicate_bm.ref_id}..."
+      )
       duplicate_bm.profiles.update_all(benchmark_id: existing_bm.id)
     end
     # rubocop:enable Rails/SkipsModelValidations
+
+    def logger
+      Rails.logger
+    end
   end
 end
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -21,7 +21,7 @@ class ParentProfileAssociator
         ref_id: profile.ref_id,
         benchmark_id: parent.benchmark.id
       ).count
-      return false unless duplicate_profiles > 1
+      return false unless duplicate_profiles >= 1
 
       profile.delete_all_test_results = true
       profile.destroy
@@ -51,7 +51,13 @@ class ParentProfileAssociator
       Profile.where(name: profile.name,
                     ref_id: profile.ref_id,
                     description: profile.description,
-                    account_id: nil)
+                    account_id: nil).or(
+                      Profile.where(
+                        name: profile.name,
+                        ref_id: profile.ref_id,
+                        account_id: nil
+                      )
+                    )
     end
 
     def find_most_recent_parent(parents)
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -6,7 +6,7 @@ class ParentProfileAssociator
     def run!
       Profile.transaction do
         Profile.where.not(account: nil).find_each do |profile|
-          next if profile_already_in_account(profile)
+          next if profile_already_in_account!(profile)
 
           parent = find_parent(profile)
           profile.update!(parent_profile: parent, benchmark: parent.benchmark)
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -17,11 +17,18 @@ class ParentProfileAssociator
     private
 
     def profile_already_in_account!(profile, parent)
-      duplicate_profiles = profile.account.profiles.where(
+      possible_duplicate_profiles = profile.account.profiles.where(
         ref_id: profile.ref_id,
         benchmark_id: parent.benchmark.id
-      ).count
-      return false unless duplicate_profiles >= 1
+      )
+      # If multiple profiles are returned with the same ref_id and prospective
+      # benchmark_id, `.update!` will not be able to run. If only 1 profile is
+      # found with these ref_id/benchmark_id - it's the same one we're passing
+      # as an argument, so it's not a problem to keep it. Otherwise, remove it
+      # as there's a profile already in the account with the right attributes.
+      return false if (possible_duplicate_profiles.count == 1 &&
+                       possible_duplicate_profiles.include?(profile)) ||
+                      possible_duplicate_profiles.count == 0
 
       profile.delete_all_test_results = true
       profile.destroy
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -6,23 +6,25 @@ class ParentProfileAssociator
     def run!
       Profile.transaction do
         Profile.where.not(account: nil).find_each do |profile|
-          if profile.account.profiles.where(ref_id: profile.ref_id).count > 1
-            profile.delete_all_test_results = true
-            profile.destroy
-            next
-          end
+          next if profile_already_in_account(profile)
+
           parent = find_parent(profile)
           profile.update!(parent_profile: parent, benchmark: parent.benchmark)
         end
-        empty_benchmarks = ::Xccdf::Benchmark.where.not(id: Profile.select(:benchmark_id).distinct)
-        empty_benchmarks.each do |empty_benchmark|
-          empty_benchmark.rules.delete_all
-        end
-        empty_benchmarks.destroy_all
       end
     end
 
     private
+
+    def profile_already_in_account!(profile)
+      duplicate_profiles = profile.account.profiles.where(
+        ref_id: profile.ref_id
+      ).count
+      return false unless duplicate_profiles > 1
+
+      profile.delete_all_test_results = true
+      profile.destroy
+    end
 
     def find_parent(profile)
       logger.info("Finding parents for profile #{profile.id}")
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -6,8 +6,19 @@ class ParentProfileAssociator
     def run!
       Profile.transaction do
         Profile.where.not(account: nil).find_each do |profile|
-          profile.update!(parent_profile: find_parent(profile))
+          if profile.account.profiles.where(ref_id: profile.ref_id).count > 1
+            profile.delete_all_test_results = true
+            profile.destroy
+            next
+          end
+          parent = find_parent(profile)
+          profile.update!(parent_profile: parent, benchmark: parent.benchmark)
         end
+        empty_benchmarks = ::Xccdf::Benchmark.where.not(id: Profile.select(:benchmark_id).distinct)
+        empty_benchmarks.each do |empty_benchmark|
+          empty_benchmark.rules.delete_all
+        end
+        empty_benchmarks.destroy_all
       end
     end
 
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -14,18 +14,46 @@ class ParentProfileAssociator
     private
 
     def find_parent(profile)
-      parents = Profile.where(ref_id: profile.ref_id,
-                              benchmark_id: profile.benchmark_id,
-                              account_id: nil)
-      raise not_found(profile) unless parents.one?
+      logger.info("Finding parents for profile #{profile.id}")
+      parents = find_in_same_benchmark(profile)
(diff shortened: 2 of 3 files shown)


app/services/duplicate_benchmark_resolver.rb
```diff
@@ -34,13 +37,17 @@ def benchmarks
     end
 
     def migrate_rules(existing_bm, duplicate_bm)
+      logger.info(
```
dLobatog: All of these break rubocop - since this is temporary code only meant to run on a specific migration, and it's just adding logger code, would you be OK with disabling rubocop here?

app/services/parent_profile_associator.rb
```diff
@@ -14,18 +14,46 @@ def run!
...
+    def find_in_same_benchmark(profile)
+      Profile.where(ref_id: profile.ref_id,
+                    benchmark_id: profile.benchmark_id,
+                    account_id: nil)
+    end
+
+    # Profiles in the phony-benchmark will have to be assigned to
+    # a real benchmark if it can be found.
+    def find_in_other_benchmarks(profile)
+      Profile.where(name: profile.name,
```
dLobatog: I'm matching both the name and the ref_id here as that seems a more accurate way of finding the parent profile outside of its own benchmark

app/services/parent_profile_associator.rb
```diff
@@ -14,18 +14,46 @@ def run!
...
+                    ref_id: profile.ref_id,
+                    account_id: nil)
+    end
+
+    def find_most_recent_parent(parents)
+      latest_benchmark = ::Xccdf::Benchmark.find_latest(
```
dLobatog: Given many prospective parents are found, return the one with the most recent version

akofink:
What do you think about just removing the phony benchmark at this point? If users still have profiles associated to it, they would be related to very old test results. New uploads from the same host would have the SSG version parsed and associated to an actual Benchmark.

app/services/parent_profile_associator.rb
```diff
@@ -14,18 +14,46 @@ def run!
...
+        parents.map(&:benchmark)
```
akofink: Red flags :upside_down_face: But, we have < 500 profiles in prod, and `find_*_parent` methods return 1-6 profiles, so the fan out isn't huge.

app/services/parent_profile_associator.rb
akofink: We could (maybe should) add an index on name, since we seem to need to search by name like this periodically. After this uniqueness stuff is solved, I plan on going through our schema and adding indexes on any searched fields

app/services/parent_profile_associator.rb
dLobatog: `find_latest` doesn't take in an AR relation, I can change it to that though

akofink:
Based on the error, we would definitely benefit from a profile uniqueness constraint before this migration is run.

akofink:
Sadly, removing the benchmark uniqueness constraint migration doesn't help:

```
-- add_reference(:profiles, :parent_profile, {:type=>:uuid, :foreign_key=>{:to_table=>:profiles}})                                                                                 [104/45525]
   -> 0.0189s                                                                                                                                                                                 
rake aborted!                                                                                                                                                                                 
StandardError: An error has occurred, this and all later migrations canceled:                                                                                                                 
                                                                                                                                                                                              
Validation failed: Ref has already been taken                                                                                                                                                 
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/validations.rb:80:in `raise_validation_error'                                                                                   
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/validations.rb:52:in `save!'                                                                                                    
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:315:in `block in save!'                                                                                         
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:387:in `block in with_transaction_returning_status'                                                             
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/connection_adapters/abstract/database_statements.rb:265:in `transaction'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:212:in `transaction'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:385:in `with_transaction_returning_status'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:315:in `save!'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/suppressor.rb:48:in `save!'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/persistence.rb:441:in `block in update!'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:387:in `block in with_transaction_returning_status'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/connection_adapters/abstract/database_statements.rb:265:in `transaction'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:212:in `transaction'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/transactions.rb:385:in `with_transaction_returning_status'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/persistence.rb:439:in `update!'
/app/app/services/parent_profile_associator.rb:9:in `block (2 levels) in run!'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/relation/batches.rb:70:in `block (2 levels) in find_each'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/relation/batches.rb:70:in `each'
/usr/local/bundle/gems/activerecord-5.2.4.1/lib/active_record/relation/batches.rb:70:in `block in find_each'
```

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -13,24 +15,23 @@ def up
...
-          profile: profile
+        rules_passed = rule_results.count { |rr| RuleResult::PASSED.include?(rr.result) }
+        rules_failed = rule_results.count { |rr| RuleResult::FAIL.include?(rr.result) }
+        rule_results_set << rule_results
+        test_results << TestResult.new(
+          host_id: host.id,
+          profile_id: profile.id,
+          start_time: rule_results[0].start_time || Time.now.utc,
+          end_time: rule_results[0].end_time || Time.now.utc,
+          score: compliance_score(rules_passed, rules_failed)
```
akofink: This method doesn't protect against `ZeroDivisionError`, and I don't see any checks. If a user scanned a system against a profile from a benchmark of a different OS, it will have 0 rule results.

db/migrate/20200327170640_remove_empty_benchmarks.rb
```diff
@@ -0,0 +1,23 @@
...
+        empty_rule_references_rules = RuleReferencesRule.where(
+          rule_id: empty_benchmark.rules.select(:id)
+        )
+        RuleReferences.where(
+          id: empty_rule_references_rules.select(:rule_reference_id)
+        ).delete_all
+        empty_rule_references_rules.delete_all
+        empty_benchmark.rules.delete_all
+      end
+      empty_benchmarks.destroy_all
```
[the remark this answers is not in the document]
dLobatog: Line 18 deletes the rules, `         empty_benchmark.rules.delete_all `

I did it like this because otherwise it's making a ton of statements and it is way slower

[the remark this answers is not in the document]
akofink: Ok, idk how I missed line 18 :)

app/services/parent_profile_associator.rb
```diff
@@ -6,26 +6,69 @@ class << self
...
     def find_parent(profile)
-      parents = Profile.where(ref_id: profile.ref_id,
-                              benchmark_id: profile.benchmark_id,
-                              account_id: nil)
-      raise not_found(profile) unless parents.one?
+      logger.info("Finding parents for profile #{profile.id}")
+      parents = find_in_same_benchmark(profile)
+      parents = find_in_other_benchmarks(profile) if parents.blank?
+
+      raise not_found(profile) unless parents.any?
```
akofink: So I have discovered some bad data in the phony benchmark; We have 51 profiles in the prod dump that have an incorrect description - it's the benchmark description rather than the profile description. There are no other profiles outside the phony benchmark with this problem, so it seems we've fixed the bug in production already; however, we need to handle these profiles.

Here's an example profile from the prod dump with this problem:

```rb
=> #<Profile:0x000055872542f200
 id: "0096f7cd-aebb-49a6-a91e-8caba583800b",
 name: "OSPP - Protection Profile for General Purpose Operating Systems v. 4.2",
 ref_id: "xccdf_org.ssgproject.content_profile_ospp42",
 created_at: Thu, 09 May 2019 21:21:45 UTC +00:00,
 updated_at: Fri, 04 Oct 2019 08:32:09 UTC +00:00,
 description:
  "This guide presents a catalog of security-relevant\nconfiguration settings for Red Hat Enterprise Linux 7. It is a rendering of\ncontent structured in the eXtensible Configuration Checklist Description Format (XCCDF)\nin order to support security automation.  The SCAP content is\nis available in the scap-security-guide package which is developed at\n\n    https://www.open-scap.org/security-policies/scap-security-guide.\n\nProviding system administrators with such guidance informs them how to securely\nconfigure systems under their control in a variety of network roles. Policy\nmakers and baseline creators can use this catalog of settings, with its\nassociated references to higher-level security control catalogs, in order to\nassist them in security baseline creation. This guide is a catalog, not a\nchecklist, and satisfaction of every item is not likely to be possible or\nsensible in many operational scenarios. However, the XCCDF format enables\ngranular selection and adjustment of settings, and their association with OVAL\nand OCIL content provides an automated checking capability. Transformations of\nthis document, and its associated automated checking content, are capable of\nproviding baselines that meet a diverse set of policy objectives. Some example\nXCCDF Profiles, which are selections of items that form checklists and\ncan be used as baselines, are available with this guide. They can be\nprocessed, in an automated fashion, with tools that support the Security\nContent Automation Protocol (SCAP). The DISA STIG for Red Hat Enterprise Linux 7,\nwhich provides required settings for US Department of Defense systems, is\none example of a baseline created from this guidance.\n",
 account_id: "9b22a83d-57af-4b67-9430-148fc4bdc251",
 compliance_threshold: 100.0,
 business_objective_id: nil,
 benchmark_id: "fe07c3bc-0f58-47ac-bdf6-e2b137ac5677">
```

As you can see, this description is for the benchmark, not the profile. Of the 51 profiles with this issue, 3 are hipaa and thus their benchmark cannot be determined. We need to decide to either associate them to a random benchmark, ignore them, or delete them. Those are the 3 options I see as a way forward. If we keep this `raise`, we won't be able to finish running this migration.

akofink: This could be one solution (tested and seems to work fine):

```diff
diff --git a/app/services/parent_profile_associator.rb b/app/services/parent_profile_associator.rb
index 76e9ca0..0e00844 100644
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -7,6 +7,7 @@ class ParentProfileAssociator
       Profile.transaction do
         Profile.where.not(account: nil).find_each do |profile|
           parent = find_parent(profile)
+          next unless parent
           next if profile_already_in_account!(profile, parent)
 
           profile.update!(parent_profile: parent, benchmark: parent.benchmark)
@@ -32,7 +33,7 @@ class ParentProfileAssociator
       parents = find_in_same_benchmark(profile)
       parents = find_in_other_benchmarks(profile) if parents.blank?
 
-      raise not_found(profile) unless parents.any?
+      return unless parents.any?
 
       parent = find_most_recent_parent(parents)
       logger.info("Found parent #{parent.id} for profile #{profile.id}")
```

app/services/parent_profile_associator.rb
```diff
@@ -6,26 +6,74 @@ class << self
...
+    # Profiles in the phony-benchmark will have to be assigned to
+    # a real benchmark if it can be found.
+    def find_in_other_benchmarks(profile)
+      parent_matching_description = Profile.where(
+        name: profile.name,
+        ref_id: profile.ref_id,
+        description: profile.description,
+        account_id: nil
+      )
+      return parent_matching_description if parent_matching_description.present?
```
dLobatog: Regarding @akofink comment on https://github.com/RedHatInsights/compliance-backend/pull/357/files#r399620519 - this fixes it IMO - if we can find the exact parent based on name, ref_id and description, good, we do it. Otherwise, we just match for the next better option, which is just name/ref_id. 

Initially I did this with a `or` query, but the AR result doesn't really fit properly, so I'm resolving each AR relation in order (it's not terribly slow on the DB dump we have). Notice `parent_matching_description` is `[]` if there are no results, so I can't do `a || b` either.  

```
[17] pry(main)> Profile.where(name: profile['name'], ref_id: profile['ref_id'], description: profile['description'], account_id: nil).or(Profile.where(name: profile['name'], ref_id: profile['ref_id'], account_id: nil)).to_sql
=> "SELECT \"profiles\".* FROM \"profiles\" WHERE \"profiles\".\"name\" = 'OSPP - Protection Profile for General Purpose Operating Systems v. 4.2' AND \"profiles\".\"ref_id\" = 'xccdf_org.ssgproject.content_profile_ospp42' AND \"profiles\".\"account_id\" IS NULL"

```

akofink: My only concern is that we really only want this for profiles with a description that matches the benchmark description (i.e. `This guide presents a catalog of security-relevant\nconfiguration settings...`), but from my testing, those are the only profiles that won't be found in another benchmark anyway.

app/services/parent_profile_associator.rb
```diff
@@ -6,26 +6,81 @@ class << self
...
+      # benchmark_id, `.update!` will not be able to run. If only 1 profile is
+      # found with these ref_id/benchmark_id - it's the same one we're passing
+      # as an argument, so it's not a problem to keep it. Otherwise, remove it
+      # as there's a profile already in the account with the right attributes.
+      return false if (possible_duplicate_profiles.count == 1 &&
+                       possible_duplicate_profiles.include?(profile)) ||
+                      possible_duplicate_profiles.count >= 1
+
+      profile.delete_all_test_results = true
+      profile.destroy
```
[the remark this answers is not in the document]
dLobatog: Both are removed - if any profile gets to this stage, it means that there's a *newer* profile which was reported in the same account (same ref_id, correct benchmark).

PR #361, opened 2020-03-24
Destroy phony benchmark before parent profile association

akofink:
We introduced the phony benchmark Sept 24, 2019. If any data is still associated to the phony benchmark in prod now, it is laughably old. Let's make our lives easier and just delete the phony benchmark for good.

### DIFF
--- a/app/services/parent_profile_associator.rb
+++ b/app/services/parent_profile_associator.rb
@@ -17,7 +17,7 @@ class ParentProfileAssociator
       parents = Profile.where(ref_id: profile.ref_id,
                               benchmark_id: profile.benchmark_id,
                               account_id: nil)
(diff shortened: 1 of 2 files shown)


app/services/parent_profile_associator.rb
```diff
@@ -17,7 +17,7 @@ def find_parent(profile)
       parents = Profile.where(ref_id: profile.ref_id,
                               benchmark_id: profile.benchmark_id,
                               account_id: nil)
-      raise not_found(profile) unless parents.one?
+      raise not_found(profile) unless parents.any?
```
akofink: Expecting to find only one profile without a DB constraint on uniqueness is a Bad Idea. We should update this code back to `one?` once we add the constraint for profiles.

PR #362, opened 2020-03-25
Update more score calculations to use TestResult

akofink:
This PR updates the compliant? calculation to use the TestResult score. Hopefully there are no other locations where we're counting rule results to calculate score.

### DIFF
--- a/test/fixtures/test_results.yml
+++ b/test/fixtures/test_results.yml
@@ -2,8 +2,10 @@ one:
   host_id: one
   profile_id: one
   score: 98
+  end_time: 2020-03-25T14:51:17+00:00
 
 two:
   host_id: two
   profile_id: two
   score: 98
+  end_time: 2020-03-25T14:52:39+00:00
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -18,15 +18,6 @@ class ProfileTest < ActiveSupport::TestCase
     assert_not profiles(:one).compliant?(hosts(:one))
   end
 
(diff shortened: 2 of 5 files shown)


test/fixtures/test_results.yml
```diff
@@ -2,8 +2,10 @@ one:
   host_id: one
   profile_id: one
   score: 98
+  end_time: 2020-03-25T14:51:17+00:00
 
 two:
   host_id: two
   profile_id: two
   score: 98
+  end_time: 2020-03-25T14:52:39+00:00
```
akofink: `TestResult.latest` uses the end_time to figure out which results to return

test/models/profile_test.rb
```diff
@@ -18,15 +18,6 @@ class ProfileTest < ActiveSupport::TestCase
...
   end
 
-  test 'host is compliant if all rules are "pass" or "notapplicable"' do
-    test_results(:one).update(host: hosts(:one), profile: profiles(:one))
-    RuleResult.create(rule: rules(:one), host: hosts(:one), result: 'pass',
-                      test_result: test_results(:one))
-    RuleResult.create(rule: rules(:two), host: hosts(:one),
-                      test_result: test_results(:one), result: 'notapplicable')
-    assert profiles(:one).compliant?(hosts(:one))
-  end
```
akofink: This test no longer makes sense, as compliance is based solely on the TestResult score

akofink:
Thanks for the review! Updated to return 0 for no results - I didn't even have to update any test!

PR #363, opened 2020-03-26
Add Pundit policy for TestResult

dLobatog:
This is meant to allow the `collection_field` TestResult GQL API
endpoint to return a set of filtered TestResults - without it, Pundit
will just return nil.

### DIFF
--- a/app/policies/test_result_policy.rb
+++ b/app/policies/test_result_policy.rb
@@ -3,20 +3,21 @@
 # Policies for accessing TestResults
 class TestResultPolicy < ApplicationPolicy
   def index?
-    Pundit.policy(user, record.profile)
+    Pundit.policy(user, record.profile) || Pundit.policy(user, record.host)
   end
 
   def show?
-    Pundit.policy(user, record.profile)
+    Pundit.policy(user, record.profile) || Pundit.policy(user, record.host)
   end
 
   # Only show TestResults belonging to profiles owned by the current user
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      available_profiles = HostPolicy::Scope.new(
-        user, ::Profile
-      ).resolve.pluck(:id)
-      scope.where(profile_id: available_profiles)
+      available_profiles = ProfilePolicy::Scope.new(user, ::Profile)
+      available_hosts = HostPolicy::Scope.new(user, ::Host)
+      scope.where(profile_id: available_profiles.resolve).or(
+        scope.where(host_id: available_hosts.resolve)
+      )
     end
   end
 end
--- a/app/policies/test_result_policy.rb
+++ b/app/policies/test_result_policy.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# Policies for accessing TestResults
+class TestResultPolicy < ApplicationPolicy
+  def index?
+    Pundit.policy(user, record.profile)
+  end
+
+  def show?
+    Pundit.policy(user, record.profile)
+  end
+
+  # Only show TestResults belonging to profiles owned by the current user
+  class Scope < ::ApplicationPolicy::Scope
+    def resolve
+      available_profiles = HostPolicy::Scope.new(user, ::Profile).resolve.pluck(:id)
+      scope.where(profile_id: available_profiles)
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


app/policies/test_result_policy.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# Policies for accessing TestResults
+class TestResultPolicy < ApplicationPolicy
+  def index?
+    Pundit.policy(user, record.profile)
```
akofink: :thinking: if the profile is deleted but not the report, this breaks

```suggestion
    Pundit.policy(user, record.profile) || Pundit.policy(user, record.host)
```

app/policies/test_result_policy.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  # Only show TestResults belonging to profiles owned by the current user
+  class Scope < ::ApplicationPolicy::Scope
+    def resolve
+      available_profiles = HostPolicy::Scope.new(
+        user, ::Profile
+      ).resolve.pluck(:id)
+      scope.where(profile_id: available_profiles)
+    end
```
akofink: It's much better to include `available_profiles` as a sub-query than a list of IDs. The resulting query is of this form:

```sql
SELECT "test_results".* FROM "test_results" WHERE "test_results"."profile_id" IN (SELECT "profiles"."id" FROM "profiles" WHERE "profiles"."account_id" = '0cda8d0d-0ef5-4b00-92b5-17be60ed68b9')
```

```suggestion
    def resolve
      available_profiles = HostPolicy::Scope.new(
        user, ::Profile
      ).resolve
      scope.where(profile: available_profiles)
    end
```

app/policies/test_result_policy.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def show?
+    Pundit.policy(user, record.profile)
+  end
+
+  # Only show TestResults belonging to profiles owned by the current user
+  class Scope < ::ApplicationPolicy::Scope
+    def resolve
+      available_profiles = HostPolicy::Scope.new(
```
akofink: Why is this `HostPolicy`? Shouldn't it be `ProfilePolicy`?

app/policies/test_result_policy.rb
```diff
@@ -0,0 +1,22 @@
...
+        user, ::Profile
+      ).resolve.pluck(:id)
+      scope.where(profile_id: available_profiles)
```
akofink: We should be checking host _or_ profile accessibility:

```suggestion
      scope.where(profile: available_profiles).or(TestResult.where(host: hosts))
```

app/policies/test_result_policy.rb
```diff
@@ -0,0 +1,23 @@
...
+  def show?
+    Pundit.policy(user, record.profile) || Pundit.policy(user, record.host)
+  end
+
+  # Only show TestResults belonging to profiles owned by the current user
+  class Scope < ::ApplicationPolicy::Scope
+    def resolve
+      available_profiles = ProfilePolicy::Scope.new(user, ::Profile)
+      available_hosts = HostPolicy::Scope.new(user, ::Host)
+      scope.where(profile_id: available_profiles.resolve).or(
```
dLobatog: This works exclusively through scopes, instead of using `pluck`, `resolve` returns the AR relation without resolving it (yes, the naming is a bit strange) - the final SQL called is:
```
SELECT "test_results".* FROM "test_results" WHERE ("test_results"."profile_id" IN (SELECT "profiles"."id" FROM "profiles" WHERE "profiles"."account_id" = 'b428b5d9-df19-5bb9-a1dc-115e071b836c') OR "test_results"."host_id" IN (SELECT "hosts"."id" FROM "hosts" WHERE "hosts"."account_id" = 'b428b5d9-df19-5bb9-a1dc-115e071b836c'))
```

test/policies/test_result_policy_test.rb
```diff
@@ -0,0 +1,49 @@
...
+    profiles(:one).account = accounts(:test)
+    profiles(:one).save
+    assert_includes Pundit.policy_scope(users(:test), Profile), profiles(:one)
+    test_results(:one).profile = profiles(:one)
+    rule_results(:one).host = hosts(:one)
+    rule_results(:one).rule = rules(:one)
+    test_results(:one).rule_results = [rule_results(:one)]
+    # Save without validations to force a TestResult with a profile
+    # but without a host
+    test_results(:one).save(validate: false)
```
dLobatog: I had to do this here and in the other test to ensure we could test the 2nd code path where a TestResult has a Profile but no Host - this shouldn't be possible according to validations though.

PR #364, opened 2020-03-26
Remove legacy GQL queries not used anymore

dLobatog:


test/graphql/queries/benchmark_query_test.rb
```diff
@@ -5,8 +5,8 @@
 class BenchmarkQueryTest < ActiveSupport::TestCase
   test 'query benchmark owned by the user' do
     query = <<-GRAPHQL
-      query allBenchmarks {
-          allBenchmarks {
+      query latestBenchmarks {
+          latestBenchmarks {
```
akofink: Fine for this PR, but this has me thinking that we should match the TestResult API and allow `bool latest` as an argument.

app/graphql/types/query.rb
```diff
@@ -23,32 +23,15 @@ def collection_field(name, type)
...
       description 'Details for a system'
       argument :id, String, required: true
     end
 
-    field :all_image_streams, [Types::System], null: true do
-      description 'All image streams visible by the user'
-      argument :search, String, 'Search query', required: false
-    end
-
     field :all_profiles, [Types::Profile], null: true do
```
akofink: This one will be removed soon as well, right?

test/graphql/queries/benchmark_query_test.rb
dLobatog: Agreed - ideally all fields can be defined as `collection_field :systems, Types::System` - and just parameterized via the "search" field

app/graphql/types/query.rb
dLobatog: Yup, I've seen it used on 1 place still though

PR #365, opened 2020-03-26
RHICOMPL-292 - Filter by hosts with test results

dLobatog:
This allows us to only display hosts with test results on certain pages, such as "Reports by system" or "Report details" - where we do not want systems that have no reports.

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -8,8 +8,8 @@ class Host < ApplicationRecord
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
                 only_explicit: true
-  scoped_search on: :has_test_results, ext_method: 'has_test_results',
-                only_explicit: true
+  scoped_search on: :has_test_results, ext_method: 'test_results?',
+                only_explicit: true, operators: ['=']
   scoped_search relation: :profile_hosts, on: :profile_id
   has_many :rule_results, dependent: :delete_all
   has_many :rules, through: :rule_results, source: :rule
@@ -27,7 +27,11 @@ class Host < ApplicationRecord
         host.compliant.values.all?(ActiveModel::Type::Boolean.new.cast(value))
       end
       ids = ids.pluck(:id).map { |id| "'#{id}'" }
-      scoped_search_sql_conditions(ids, operator)
+
+      return { conditions: '1=0' } if ids.empty?
+
+      operator = operator == '<>' ? 'NOT' : ''
+      { conditions: "hosts.id #{operator} IN(#{ids.join(',')})" }
     end
 
     def filter_by_compliance_score(_filter, operator, score)
@@ -40,20 +44,13 @@ class Host < ApplicationRecord
       { conditions: "hosts.id IN(#{ids.join(',')})" }
     end
 
-    def has_test_results(_filter, operator, value)
-      ids = Host.select do |h|
-        h.test_results.present? == ActiveModel::Type::Boolean.new.cast(value)
-      end
-      ids = ids.pluck(:id).map { |id| "'#{id}'" }
-      scoped_search_sql_conditions(ids, operator)
-    end
-
-    def scoped_search_sql_conditions(ids, operator)
-      return { conditions: '1=0' } if ids.empty?
-      operator = operator == '<>' ? 'NOT' : ''
-      { conditions: "hosts.id #{operator} IN(#{ids.join(',')})" }
+    def test_results?(_filter, _operator, value)
+      operator = ActiveModel::Type::Boolean.new.cast(value) ? '' : 'NOT'
+      host_ids = TestResult.select(:host_id).distinct.where.not(host_id: nil)
+      {
+        conditions: "hosts.id #{operator} "\
+                    "IN(#{host_ids.to_sql})"
+      }
     end
   end
-
-  private_class_method :scoped_search_sql_conditions
 end
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -8,6 +8,8 @@ class Host < ApplicationRecord
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
                 only_explicit: true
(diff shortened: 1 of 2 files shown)


app/models/host.rb
```diff
@@ -8,6 +8,8 @@ class Host < ApplicationRecord
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
                 only_explicit: true
+  scoped_search on: :has_test_results, ext_method: 'has_test_results',
+                only_explicit: true
```
akofink: We should limit available operators:

```suggestion
                only_explicit: true, operators: ['=', '!=']
```

app/models/host.rb
```diff
@@ -41,5 +39,21 @@ def filter_by_compliance_score(_filter, operator, score)
...
+      ids = Host.select do |h|
+        h.test_results.present? == ActiveModel::Type::Boolean.new.cast(value)
+      end
+      ids = ids.pluck(:id).map { |id| "'#{id}'" }
+      scoped_search_sql_conditions(ids, operator)
+    end
+
+    def scoped_search_sql_conditions(ids, operator)
+      return { conditions: '1=0' } if ids.empty?
+      operator = operator == '<>' ? 'NOT' : ''
```
akofink: `<>` is not a valid operator. scoped_search really ought to have a query validator :/

```suggestion
      operator = operator == '!=' ? 'NOT' : ''
```

app/models/host.rb
```diff
@@ -41,5 +39,21 @@ def filter_by_compliance_score(_filter, operator, score)
 
       { conditions: "hosts.id IN(#{ids.join(',')})" }
     end
+
+    def has_test_results(_filter, operator, value)
+      ids = Host.select do |h|
+        h.test_results.present? == ActiveModel::Type::Boolean.new.cast(value)
+      end
```
akofink: Prefer SQL - against prod-sized data, `Host.search_for('has_test_results = true')` takes about 8 seconds :turtle: 

```suggestion
      ids = TestResult.pluck(:host_id)
```

or

```suggestion
      ids = Host.joins(:test_results).distinct.pluck(:id)
```

One is faster than the other:

```rb
[41] pry(main)> Set.new(Host.joins(:test_results).distinct.pluck(:id)) == Set.new(TestResult.pluck(:host_id))
   (15.8ms)  SELECT DISTINCT "hosts"."id" FROM "hosts" INNER JOIN "test_results" ON "test_results"."host_id" = "hosts"."id"
   (8.4ms)  SELECT "test_results"."host_id" FROM "test_results"
=> true
```

app/models/host.rb
```diff
@@ -41,5 +39,21 @@ def filter_by_compliance_score(_filter, operator, score)
...
+      ids = ids.pluck(:id).map { |id| "'#{id}'" }
+      scoped_search_sql_conditions(ids, operator)
+    end
+
+    def scoped_search_sql_conditions(ids, operator)
+      return { conditions: '1=0' } if ids.empty?
+      operator = operator == '<>' ? 'NOT' : ''
+      { conditions: "hosts.id #{operator} IN(#{ids.join(',')})" }
```
akofink: We cannot build queries like this - there is a maximum SQL query limit. I'm not sure if our graphql framework really supports this, but we should be working in ActiveRecord::Relations up until the point of execution against the DB. Arrays , string `join`s, and `map`s are all red flags.

test/models/host_test.rb
```diff
@@ -61,5 +61,22 @@ class HostTest < ActiveSupport::TestCase
...
+      assert_includes Host.search_for('has_test_results = true'), hosts(:one)
+      assert_not_includes(Host.search_for('has_test_results = true'),
+                          hosts(:two))
+      assert_includes Host.search_for('has_test_results = false'), hosts(:two)
+      assert_not_includes(Host.search_for('has_test_results = false'),
+                          hosts(:one))
+      assert_includes Host.search_for('has_test_results != true'), hosts(:two)
+      assert_not_includes(Host.search_for('has_test_results != true'),
+                          hosts(:one))
+      assert_includes Host.search_for('has_test_results != false'), hosts(:one)
```
akofink: Oh! I'm a bit surprised these `!=` tests pass, since our code checks for `<>`.

dLobatog: scoped_search automatically translates `!=` to `<>` as the operator

app/models/host.rb
dLobatog: https://github.com/RedHatInsights/compliance-backend/pull/365#discussion_r398627244

test/models/host_test.rb
akofink: Ah! nice, TIL

app/models/host.rb
```diff
@@ -8,6 +8,8 @@ class Host < ApplicationRecord
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
                 only_explicit: true
+  scoped_search on: :has_test_results, ext_method: 'has_test_results',
+                only_explicit: true, operators: ['=']
```
dLobatog: I'm limiting it to just `=` true or false as that's really all we need from the frontend POV (even from the user POV I'd argue)

app/models/host.rb
```diff
@@ -41,5 +39,22 @@ def filter_by_compliance_score(_filter, operator, score)
 
       { conditions: "hosts.id IN(#{ids.join(',')})" }
     end
+
+    def has_test_results(_filter, _operator, value)
+      operator = ActiveModel::Type::Boolean.new.cast(value) ? '' : 'NOT'
+      host_ids = TestResult.select(:host_id).distinct.where.not(:host_id => nil)
+      {
+        conditions: "hosts.id #{operator} "\
+                    "IN(#{host_ids.to_sql})"
```
dLobatog: This generates a SQL subquery, so the whole thing is executed at once. I guess it's a valid workaround since scoped_search will just take this and `Host.where(conditions)`, so we can't really pass AR relations around here

akofink: Yep, `to_sql` is a valid way to do this :+1:

PR #366, opened 2020-03-27
Compute test results score directly on the migration

dLobatog:


db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -13,24 +15,23 @@ def up
       rule_results_by_end_time = scan_results(host, profile)
         .group_by(&:end_time)
       rule_results_by_end_time.each do |end_time, rule_results|
-        test_result = TestResult.create!(
-          host: host,
-          profile: profile
+        rules_passed = rule_results.count { |rr| RuleResult::PASSED.include?(rr.result) }
```
dLobatog: This is not a relation - just a plain old array

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -13,24 +15,23 @@ def up
...
+        rules_failed = rule_results.count { |rr| RuleResult::FAIL.include?(rr.result) }
+        rule_results_set << rule_results
+        test_results << TestResult.new(
+          host_id: host.id,
+          profile_id: profile.id,
+          start_time: rule_results[0].start_time || Time.now.utc,
+          end_time: rule_results[0].end_time || Time.now.utc,
+          score: compliance_score(rules_passed, rules_failed)
```
dLobatog: I tried before including `rule_results` here but the resulting query when running `.import` was incredibly slow

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -13,24 +15,23 @@ def up
...
-          end_time: latest_rule_results[0].end_time || Time.now.utc,
-          score: compliance_score(host.rules_passed(profile),
-                                  host.rules_failed(profile))
-        ) unless latest_rule_results.empty?
       end
       migrated += 1
     end
+    imported_test_results = TestResult.import test_results
+    imported_test_results.ids.zip(rule_results_set).each do |imported_test_result, rule_results|
+      RuleResult.where(id: rule_results.pluck(:id)).update_all(test_result_id: imported_test_result)
```
akofink: This line seems to be a bottleneck during my testing. We could probably use more SQL here rather than pluck if we want to further optimize. There's also some danger here of hitting the max SQL query length.

db/migrate/20191126181009_seed_test_results.rb
akofink: This compliance_score is not computed correctly, as it counts notselected results as passed. This PR fixes it: https://github.com/RedHatInsights/compliance-backend/pull/368

PR #370, opened 2020-03-28
Add unique profile DB constraint

akofink:
Profiles should be constrained to be unique by the database. We must resolve duplicates before adding such an index. We must resolve associated `test_results`, `profile_hosts`, and child `profiles`. `profile_rules` are destroyed without checking equivalence, as they should match on the existing profile.

image

### DIFF
--- a/app/services/duplicate_profile_resolver.rb
+++ b/app/services/duplicate_profile_resolver.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+# A service class to merge duplicate Profile objects
+class DuplicateProfileResolver
+  class << self
+    def run!
+      @profiles = nil
+      duplicate_profiles.find_each do |profile|
+        if existing_profile(profile)
+          migrate_profile(existing_profile(profile), profile)
+        else
+          self.existing_profile = profile
+        end
+      end
+    end
+
+    private
+
+    def existing_profile(profile)
+      profiles[[profile.ref_id, profile.account_id, profile.benchmark_id]]
+    end
+
+    def existing_profile=(profile)
+      profiles[[profile.ref_id,
+                profile.account_id,
+                profile.benchmark_id]] = profile
+    end
+
+    def profiles
+      @profiles ||= {}
+    end
+
+    def migrate_profile(existing_p, duplicate_p)
+      migrate_profile_hosts(existing_p, duplicate_p)
+      migrate_test_results(existing_p, duplicate_p)
+      migrate_child_profiles(existing_p, duplicate_p)
+      duplicate_p.destroy # BusinessObjectives, ProfileRules
+    end
+
+    def duplicate_profiles
+      Profile.joins(
+        "JOIN (#{grouped_nonunique_profile_tuples.to_sql}) as p on "\
+        'profiles.ref_id = p.ref_id AND '\
+        'profiles.account_id is not distinct from p.account_id AND '\
+        'profiles.benchmark_id = p.benchmark_id'
+      )
+    end
+
+    def grouped_nonunique_profile_tuples
+      Profile.select(:ref_id, :account_id, :benchmark_id)
+             .group(:ref_id, :account_id, :benchmark_id)
+             .having('COUNT(id) > 1')
+    end
+
+    # rubocop:disable Rails/SkipsModelValidations
+    def migrate_profile_hosts(existing_p, duplicate_p)
+      duplicate_p.profile_hosts.where.not(host: existing_p.hosts)
+                 .update_all(profile_id: existing_p.id)
+    end
+
+    def migrate_test_results(existing_p, duplicate_p)
+      duplicate_p.test_results.update_all(profile_id: existing_p.id)
+    end
+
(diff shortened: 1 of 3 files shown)


akofink:
This should be ready for review and testing. Rules uniqueness to follow.

app/services/duplicate_profile_resolver.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+# A service class to merge duplicate Profile objects
+class DuplicateProfileResolver
+  class << self
+    def run!
+      @profiles = nil
```
akofink, line 7: Note that this is a class variable. Subsequent runs of this resolver on the same thread will cache whatever is stored here, so we need to clear it. This is especially true of the tests.

app/services/duplicate_profile_resolver.rb
```diff
@@ -0,0 +1,71 @@
...
+        'profiles.account_id is not distinct from p.account_id AND '\
+        'profiles.benchmark_id = p.benchmark_id'
+      )
+    end
+
+    def grouped_nonunique_profile_tuples
+      Profile.select(:ref_id, :account_id, :benchmark_id)
+             .group(:ref_id, :account_id, :benchmark_id)
+             .having('COUNT(id) > 1')
+    end
```
dLobatog, line 53: Currently from prod:
```
irb(main):017:0> duplicate_profiles.to_sql
=> "SELECT \"profiles\".* FROM \"profiles\" JOIN (SELECT \"profiles\".\"ref_id\", \"profiles\".\"account_id\", \"profiles\".\"benchmark_id\" FROM \"profiles\" GROUP BY \"profiles\".\"ref_id\", \"profiles\".\"account_id\", \"profiles\".\"benchmark_id\" HAVING (COUNT(id) > 1)) as p on profiles.ref_id = p.ref_id AND profiles.account_id is not distinct from p.account_id AND profiles.benchmark_id = p.benchmark_id"
irb(main):018:0> duplicate_profiles
=> #<ActiveRecord::Relation []>
```

app/services/duplicate_profile_resolver.rb
```diff
@@ -0,0 +1,71 @@
...
+                profile.benchmark_id]] = profile
+    end
+
+    def profiles
+      @profiles ||= {}
+    end
+
+    def migrate_profile(existing_p, duplicate_p)
+      migrate_profile_hosts(existing_p, duplicate_p)
+      migrate_test_results(existing_p, duplicate_p)
```
dLobatog, line 35: You can skip these 2 if `profile.canonical`

app/services/duplicate_profile_resolver.rb
```diff
@@ -0,0 +1,71 @@
...
+      migrate_child_profiles(existing_p, duplicate_p)
```
dLobatog, line 36: You can skip this one if `!profile.canonical`

db/migrate/20200325185540_add_unique_index_to_profiles.rb
```diff
@@ -0,0 +1,11 @@
+class AddUniqueIndexToProfiles < ActiveRecord::Migration[5.2]
+  def up
+    DuplicateProfileResolver.run!
+
+    add_index(:profiles, %i[ref_id account_id benchmark_id], unique: true)
```
dLobatog, line 5: :+1:

dLobatog (approved):
LGTM - I have minor suggestions inline but I'm OK with getting this in, at the end of the day there are no duplicates apparently on the db already, and tests run fine :+1:  Thanks @akofink

app/services/duplicate_profile_resolver.rb
akofink: I suspect this won't be the case after running the duplicate benchmark resolver, but my latest PR with tweaks to that one should be a bit smarter about creating duplicates: https://github.com/RedHatInsights/compliance-backend/pull/397

app/services/duplicate_profile_resolver.rb
akofink: Well, at first I agreed with you for the sake of optimization, but these methods are already highly efficient, only generating a single SQL query; adding an additional check seems like premature optimization. IMO the simplicity here is very important to ensure correctness - one of the mistakes I made when first implementing the duplicate benchmark resolver.

I guess, one last comment: canonical profiles _shouldn't_ have profile_hosts or test_results, but why would we check `canonical` and rely on that assumption instead of checking for `profile_hosts` or `test_results` directly?

app/services/duplicate_profile_resolver.rb
akofink: With the updated benchmark uniqueness PR, https://github.com/RedHatInsights/compliance-backend/pull/397, this should continue to be true, hopefully resulting in very fast migrations. :100:

app/services/duplicate_profile_resolver.rb
dLobatog: I'd say if that happens it should straightaway break the migration to log it - no hosts or test results should ever be assigned to these

PR #374, opened 2020-04-01
Seed test results in batches of 200 each (#373)

dLobatog:


db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -1,36 +1,38 @@
...
-    ProfileHost.find_each do |profile_host|
-      host = profile_host.host
-      profile = profile_host.profile
-      if host.nil? || profile.nil?
-        profile_host.destroy
-        next
+    ProfileHost.find_in_batches(batch_size: 200) do |profile_host_group|
+      test_results = []
+      rule_results_set = []
+      profile_host_group.each do |profile_host|
```
dLobatog, line 7: Resetting the arrays here is really the only meaningful change

db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -1,36 +1,38 @@
...
+            start_time: rule_results[0].start_time || Time.now.utc,
+            end_time: rule_results[0].end_time || Time.now.utc,
+            score: compliance_score(rules_passed, rules_failed)
+          }
+        end
+        migrated += 1
+        imported_test_results = TestResult.import test_results
+        imported_test_results.ids.zip(rule_results_set).each do |imported_test_result, rule_results|
+          RuleResult.where(id: rule_results.pluck(:id)).update_all(test_result_id: imported_test_result)
+        end
```
dLobatog, line 34: And now this runs on every batch of 200, instead of at the end of the whole thing

dLobatog:
It looks like this PR is broken after yesterday's issues with Github. let's close and reopen..

akofink:
Oh hey, no smoke came out this time!

PR #375, opened 2020-04-01
Profile graphql returns score

akofink:
We should _never_ be calculating score ourselves. We don't know how to do it (NIST 7275 r.4, section 7.3.2 Scoring Models), and we're not NIST certified.

https://projects.engineering.redhat.com/browse/RHICOMPL-515


akofink:
Correct. Front end PR coming

PR #376, opened 2020-04-02
Performance improvements to SeedTestResults migration

dLobatog:
* Batch size is down to 30 as it seems to be the fastest I found by manually trying different values on Openshift, it runs on underpowered pods, but the value isn't really important. 

* Host and Profile tables are not called anymore, instead we use host_id
  and profile_id directly.

* TestResult inserts are wrapped in a transaction to avoid:

```
BEGIN
  insert into ...
COMMIT
... repeat for as many TestResult objects as needed
```

instead we do

```
BEGIN
  insert into ...
  insert into ...
  insert into ...
  insert into ...
COMMIT
```

  which is more performant and just as safe as everything is already
  wrapped by an upper level nested transaction as it's a migration.

* Also I added a counter to keep track of how many HostProfile objects
  are left and measure the progress.

### DIFF
--- a/db/migrate/20191126181009_seed_test_results.rb
+++ b/db/migrate/20191126181009_seed_test_results.rb
@@ -1,18 +1,18 @@
 class SeedTestResults < ActiveRecord::Migration[5.2]
   def up
     migrated = 0
-    ProfileHost.find_in_batches(batch_size: 200) do |profile_host_group|
+    ProfileHost.find_in_batches(batch_size: 30) do |profile_host_group|
       test_results = []
       rule_results_set = []
       profile_host_group.each do |profile_host|
-        host = profile_host.host
-        profile = profile_host.profile
+        host = profile_host.host_id
+        profile = profile_host.profile_id
         if host.nil? || profile.nil?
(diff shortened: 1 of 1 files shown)


db/migrate/20191126181009_seed_test_results.rb
```diff
@@ -1,33 +1,35 @@
...
     migrated = 0
-    ProfileHost.find_in_batches(batch_size: 200) do |profile_host_group|
+    ProfileHost.find_in_batches(batch_size: 30) do |profile_host_group|
       test_results = []
       rule_results_set = []
       profile_host_group.each do |profile_host|
-        host = profile_host.host
-        profile = profile_host.profile
+        host = profile_host.host_id
+        profile = profile_host.profile_id
```
akofink, line 9: Let's rename these variables to indicate they're just IDs

```suggestion
        host_id = profile_host.host_id
        profile_id = profile_host.profile_id
```

dLobatog:
Updated to use `host_id/profile_id` - thanks for the prompt review @akofink!

PR #377, opened 2020-04-02
Add test result delete mutation

akofink:
This is almost working - I haven't yet figured out how to return the testResults collection.


app/graphql/mutations/test_result/delete.rb
```diff
@@ -0,0 +1,45 @@
...
+      argument :profile_id, ID, required: true
+      field :profile, Types::Profile, null: false
+      collection_field :test_results, Types::TestResult
+
+      def resolve(args = {})
+        profile = scoped_profiles.find(args[:profile_id])
+        test_results = authorized_test_results(args)
+        test_results.map(&:destroy)
+
+        { profile: profile, test_results: ::RecordLoader.for(::TestResult).load_many(test_results)
```
akofink, line 20: @dLobatog do you happen to know how I can return this array of test results? This doesn't do it :confused:

app/graphql/mutations/test_result/delete.rb
```diff
@@ -0,0 +1,45 @@
...
+      graphql_name 'deleteTestResults'
+
+      argument :profile_id, ID, required: true
+      field :profile, Types::Profile, null: false
+      collection_field :test_results, Types::TestResult
+
+      def resolve(args = {})
+        profile = scoped_profiles.find(args[:profile_id])
+        test_results = authorized_test_results(args)
+        test_results.map(&:destroy)
```
akofink, line 18: Part of the reason `test_results` is an array is because we have to authorize `:delete?` on it - I didn't see a good way to authorize `:delete?` on an AR::Relation.

app/graphql/mutations/test_result/delete.rb
dLobatog: It works for me, maybe you're querying it wrong? If you provide a Relay style resolver for that collection field, you need to query it in Relay style too:

```
mutation deleteTestResults($input: deleteTestResultsInput!) {
  deleteTestResults(input: $input) {
    profile {
      id
    }
    testResults {
      edges {
        node {
          id
        }
      }
    }
  }
}
```

test/graphql/mutations/delete_test_result_mutation_test.rb
```diff
@@ -0,0 +1,46 @@
...
+              testResults {
+                  edges {
+                      node {
+                          id
+                      }
+                  }
+              }
+          }
+      }
+  GRAPHQL
```
akofink, line 24: @dLobatog  This test will fail. Is it not the right query?

akofink:
Thanks for the help, @dLobatog! It should be ready to go now.

PR #380, opened 2020-04-06
Serialise score as float

bastilian:
The score either gives a integer (0) or a string, but the score should be a float in the API.

Disabling `encode_big_decimal_as_string` does not work with the serializer used.


bastilian:
@akofink That makes much more sense! `.score` now returns a float.

Should we also round it?

dLobatog:
Running tests manually at https://jenkins-jenkins.5a9f.insights-dev.openshiftapps.com/job/compliance-backend/view/change-requests/job/PR-380/2/

PR #383, opened 2020-04-07
Use SQL only for BO pundit policy

akofink:
BusinessObjective queries should be _much_ faster now.

https://projects.engineering.redhat.com/browse/RHICOMPL-434

### DIFF
--- a/app/policies/business_objective_policy.rb
+++ b/app/policies/business_objective_policy.rb
@@ -3,20 +3,17 @@
 # Policies for accessing BusinessObjectives
 class BusinessObjectivePolicy < ApplicationPolicy
   def index?
-    record.profiles.pluck(:account_id).include? user.account_id
+    record.account_ids.include? user.account_id
   end
 
   def show?
-    record.profiles.pluck(:account_id).include? user.account_id
+    record.account_ids.include? user.account_id
   end
 
   # Only show hosts in our user account
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      ids = scope.all.select do |business_objective|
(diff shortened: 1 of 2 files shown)


app/policies/business_objective_policy.rb
```diff
@@ -3,20 +3,17 @@
 # Policies for accessing BusinessObjectives
 class BusinessObjectivePolicy < ApplicationPolicy
   def index?
-    record.profiles.pluck(:account_id).include? user.account_id
+    record.in_account(user.account_id)
   end
 
   def show?
-    record.profiles.pluck(:account_id).include? user.account_id
```
akofink, line 10: This is the kind of thing we won't notice in a dev/CI environment where we only have a handful of accounts.

app/policies/business_objective_policy.rb
```diff
@@ -3,20 +3,17 @@
...
+    record.in_account(user.account_id)
   end
 
   # Only show hosts in our user account
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      ids = scope.all.select do |business_objective|
-        business_objective.profiles.pluck(:account_id).include? user.account_id
-      end
-      scope.where(id: ids)
```
akofink, line 19: This part was flawed in the same way as above, but it also has the added problems of using ruby-based select and a `where` with an unbounded number of ids, which might also be problematic for maximum SQL query length.

dLobatog: This is the real issue I'd say - the query is nearly identical for the `index?` and `show?` before and after, it gets all account IDs through `profiles`, then checks if the current account_id is there. Thanks @akofink

PR #384, opened 2020-04-08
Sync compliance DB with inventory task

dLobatog:
This rake task simply sends requests to the inventory for every host,
and checks whether the host exists in the inventory. If it doesn't
exist, it deletes it from our DB.

I'm working on an improved version that fetches hosts in batches of 50, but this allows to get us off the hook now that there are some hosts like that, and in -beta envs, they show as hosts with no name in the Systems table.

### DIFF
--- a/lib/tasks/sync_with_inventory.rake
+++ b/lib/tasks/sync_with_inventory.rake
@@ -2,9 +2,9 @@
 
 desc 'Remove systems in Compliance DB which are not in the inventory'
 task sync_with_inventory: [:environment] do
-  ::Account.all.find_each do |account|
+  ::Account.includes(:hosts).find_each do |account|
     puts "Starting to review account #{account.account_number}"
-    account.hosts.find_each do |host|
+    account.hosts.each do |host|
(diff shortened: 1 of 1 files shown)


lib/tasks/sync_with_inventory.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+desc 'Remove systems in Compliance DB which are not in the inventory'
+task sync_with_inventory: [:environment] do
+  ::Account.all.find_each do |account|
```
akofink, line 5: Is this an N+1 case? I can't convince myself by testing that this is faster; what do you think?

```suggestion
  ::Account.includes(:hosts).find_each do |account|
```

akofink:
@dLobatog This one is waiting on you - the changes requested are minor.

akofink:
Ah! nice, that seems like a good solution :+1:

akofink:
I guess the next step is to write a cron job in e2e

PR #385, opened 2020-04-09
Remove user requrement in identity header

akofink:
User information is not guaranteed to be present in the identity header. We cannot rely on user existing, only account. This PR removes any database calls to user and only uses the User model as a proxy to access the account. It's a minimal change that should fix our problems, but we could remove all user-related code in the future; it'd be hundreds of lines of code changed, not something we want to deploy right before GA. :)

Fixes https://projects.engineering.redhat.com/browse/RHICOMPL-560

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -13,8 +13,7 @@ module Authentication
   def authenticate_user
     return unauthenticated unless identity_header.valid? && rbac_allowed?
 
-    user = find_or_create_user(identity_header.identity['user']['username'])
-    return if performed? || !user.persisted?
+    return if performed?
 
     User.current = user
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/authentication.rb
```diff
@@ -33,14 +32,8 @@ def unauthenticated(error = 'X-RH-IDENTITY header should be provided')
...
-  def find_or_create_user(username)
-    user = User.find_by(username: username, account: account_from_header)
-    if user.present?
-      logger.info "User authentication SUCCESS: #{identity_header.identity}"
-    else
-      user = create_user
-    end
-    user
+  def user
+    User.new(account: account_from_header)
```
akofink, line 36: This is only in memory, not persisted to the DB.

PR #395, opened 2020-04-15
Relax benchmark constraint to match on latest

dLobatog:


Victoremepunto:
> Wow, is this really what we talked about? I don't agree with this at all - I thought we were just making frontend changes, not changes to how reports are parsed. I guess I misunderstood?

I thought it was more than just UI changes, as the Policies won't be shown if they were external. Making changes to the parsing is dangerous so close to deadline , isn't it ?

PR #397, opened 2020-04-16
Resolve only duplicate benchmarks.

akofink:
The test updates give a much higher level of confidence that we're doing
what we want to do with this resolver.

I've also updated the migration for adding a uniqueness constraint so
that it runs just before the profiles and rules uniqueness migrations

### DIFF
--- a/app/services/duplicate_benchmark_resolver.rb
+++ b/app/services/duplicate_benchmark_resolver.rb
@@ -4,54 +4,89 @@
 class DuplicateBenchmarkResolver
   class << self
     def run!
-      each_benchmark do |bm|
-        if benchmarks[[bm.ref_id, bm.version]]
-          migrate_benchmark(benchmarks[[bm.ref_id, bm.version]], bm)
+      @benchmarks = nil
+      duplicate_benchmarks.find_each do |bm|
+        if existing_benchmark(bm)
+          migrate_benchmark(existing_benchmark(bm), bm)
         else
-          benchmarks[[bm.ref_id, bm.version]] = bm
+          self.existing_benchmark = bm
         end
       end
     end
 
     private
 
-    def migrate_benchmark(existing_bm, duplicate_bm)
-      logger.info(
-        "Duplicate benchmark found for #{existing_bm.id} - #{duplicate_bm.id}"
-      )
-      migrate_rules(existing_bm, duplicate_bm)
-      migrate_profiles(existing_bm, duplicate_bm)
-      duplicate_bm.delete
+    def existing_benchmark(benchmark)
+      benchmarks[[benchmark.ref_id, benchmark.version]]
     end
 
-    def each_benchmark
-      Xccdf::Benchmark.includes(:rules, :profiles).find_each do |bm|
-        Xccdf::Benchmark.transaction do
-          yield bm
-        end
-      end
+    def existing_benchmark=(benchmark)
+      benchmarks[[benchmark.ref_id, benchmark.version]] = benchmark
     end
 
     def benchmarks
       @benchmarks ||= {}
     end
 
+    def duplicate_benchmarks
+      Xccdf::Benchmark.joins(
+        "JOIN (#{grouped_nonunique_benchmark_tuples.to_sql}) as bm on "\
+        'benchmarks.ref_id = bm.ref_id AND '\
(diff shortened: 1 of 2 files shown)


app/services/duplicate_benchmark_resolver.rb
```diff
@@ -4,54 +4,69 @@
...
+                  .update_all(benchmark_id: existing_bm.id)
     end
 
-    # accept duplicate profiles for now
     def migrate_profiles(existing_bm, duplicate_bm)
       logger.info(
         "Migrating profiles for #{existing_bm.id} - #{duplicate_bm.ref_id}..."
       )
-      duplicate_bm.profiles.update_all(benchmark_id: existing_bm.id)
+      duplicate_bm.profiles
```
dLobatog, line 67: Doesn't this also have to migrate the `parent_profile_id` to the canonical profile in the new benchmark?

dLobatog (changes requested):
Looks good in general, however I think `parent_profile_id` also has to be migrated from the profiles (if there are any non-canonical profiles in the benchmark). Thanks @akofink

app/services/duplicate_benchmark_resolver.rb
akofink: Good point! Yes, this is an oversight.

app/services/duplicate_benchmark_resolver.rb
```diff
@@ -4,54 +4,89 @@
...
-      duplicate_bm.rules.update_all(benchmark_id: existing_bm.id)
+      duplicate_bm.rules.where.not(ref_id: existing_bm.rules.select(:ref_id))
+                  .update_all(benchmark_id: existing_bm.id)
+    end
+
+    def migrate_parent_profiles(existing_bm, duplicate_bm)
+      duplicate_bm.profiles.where.not(
+        ref_id: existing_bm.profiles.select(:ref_id),
+        parent_profile_id: nil
+      ).find_each do |profile|
```
akofink, line 69: This isn't efficient, but I can't think of a better way to do it; also, if we don't have any duplicate profiles in prod, efficiency doesn't make much of a difference.

app/services/duplicate_benchmark_resolver.rb
```diff
@@ -4,54 +4,89 @@
...
+    end
+
+    def migrate_benchmark(existing_bm, duplicate_bm)
+      logger.info(
+        "Duplicate benchmark found for #{existing_bm.id} - #{duplicate_bm.id}"
+      )
+      migrate_rules(existing_bm, duplicate_bm)
+      migrate_parent_profiles(existing_bm, duplicate_bm)
+      migrate_profiles(existing_bm, duplicate_bm)
+      remove_remaining_parent_profiles(duplicate_bm)
```
akofink, line 52: Any profiles that are referenced as parent profiles by another profile cannot be deleted because of the foreign key constraint on the DB. This method clears out any other parent_profile_ids for those profiles we're about to delete.

Here's an example of the error from AR:

```
ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "profiles" violates foreign key constraint "fk_rails_ba3d1c4f86
" on table "profiles"                                                                                                                                                                         
        DETAIL:  Key (id)=(6021fe64-4c07-4f60-8edb-318ae06cf139) is still referenced from table "profiles".                                                                                   
        : DELETE FROM "profiles" WHERE "profiles"."id" = $1
```

app/services/duplicate_benchmark_resolver.rb
```diff
@@ -4,54 +4,69 @@
...
+        'benchmarks.ref_id = bm.ref_id AND '\
+        'benchmarks.version = bm.version'
+      )
+    end
+
+    def grouped_nonunique_benchmark_tuples
+      Xccdf::Benchmark.select(:ref_id, :version)
+                      .group(:ref_id, :version)
+                      .having('COUNT(id) > 1')
+    end
```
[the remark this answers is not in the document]
akofink: Are you sure? I'm seeing 18 in the latest dump - I haven't ran it against prod directly, but there's no way for users to delete benchmarks from our DB.

dLobatog (approved):

PR #398, opened 2020-04-16
RHICOMPL-586 - Batch profile query on ruleObjectsFailed

dLobatog:


app/graphql/types/system.rb
```diff
@@ -63,8 +63,7 @@ def rules_failed(args = {})
     end
 
     def rule_objects_failed
-      ::Rails.cache.fetch("#{object.id}/failed_rule_objects_result",
-                          expires_in: 1.week) do
+      ::Rails.cache.fetch("#{object.id}/failed_rule_objects_result") do
```
dLobatog, line 66: This is rightfully expired at https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/concerns/xccdf/util.rb#L53

PR #399, opened 2020-04-16
Determine policy from a profile

akofink:
External profiles can be associated to policies (an internal profile associated to a user account). We should get the business objective and compliance threshold from the internal policy that the user defined.

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -61,6 +61,26 @@ class Profile < ApplicationRecord
     parent_profile_id.blank?
   end
 
+  def policy
+    return self unless external
+
+    Profile.includes(:benchmark)
+           .find_by(account: account_id, external: false, ref_id: ref_id,
+                    benchmarks: { ref_id: benchmark.ref_id })
+  end
+
+  def business_objective
+    BusinessObjective.find(business_objective_id) if business_objective_id
+  end
+
+  def business_objective_id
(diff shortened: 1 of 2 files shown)


app/models/profile.rb
```diff
@@ -61,6 +61,12 @@ def canonical?
     parent_profile_id.blank?
   end
 
+  def policy
+    Profile.includes(:benchmark)
```
dLobatog, line 65: This can probably return safely just `self` if external is false, isn't it? There can't be more policies with the same ref_id/account_id/benchmark_ref_id in the same account

akofink: Yep! Good point.

app/models/profile.rb
```diff
@@ -61,6 +61,26 @@ def canonical?
...
+    return self unless external
+
+    Profile.includes(:benchmark)
+           .find_by(account: account_id, external: false, ref_id: ref_id,
+                    benchmarks: { ref_id: benchmark.ref_id })
+  end
+
+  def business_objective
+    BusinessObjective.find(business_objective_id) if business_objective_id
+  end
```
akofink, line 74: I'm not sure if there's a better way to do this or how AR does it behind the scenes, but the column name here is actually `business_objective_id`, not `business_objective`.

akofink:
Ok, this one should be good for review.

PR #401, opened 2020-04-16
Migrate existing profiles during db migration

akofink:
Any profiles created before we released policy creation changes to
prod-beta should be external. Others created after our release should
stay external=false in order to show in the SCAP policies UI.

### DIFF
--- a/app/services/external_profile_updater.rb
+++ b/app/services/external_profile_updater.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# A service class to migrate profiles to be external before a certain date
+class ExternalProfileUpdater
+  class << self
+    # rubocop:disable Rails/SkipsModelValidations
+    def run!(date = DateTime.now)
+      Profile.where('created_at < ?', date)
+             .update_all(external: true,
+                         compliance_threshold: 100,
+                         business_objective_id: nil)
+    end
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+end
--- a/db/migrate/20200414143250_add_external_to_profiles.rb
+++ b/db/migrate/20200414143250_add_external_to_profiles.rb
@@ -1,6 +1,12 @@
 class AddExternalToProfiles < ActiveRecord::Migration[5.2]
-  def change
+  def up
     add_column :profiles, :external, :boolean, default: false, null: false
     add_index :profiles, :external
+    ExternalProfileUpdater.run!(DateTime.parse('2020-04-13T10:44:00+00:00'))
+  end
+
+  def down
+    remove_index :profiles, :external
+    remove_column :profiles, :external
   end
 end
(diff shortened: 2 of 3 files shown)


db/migrate/20200414143250_add_external_to_profiles.rb
```diff
@@ -1,6 +1,12 @@
 class AddExternalToProfiles < ActiveRecord::Migration[5.2]
-  def change
+  def up
     add_column :profiles, :external, :boolean, default: false, null: false
     add_index :profiles, :external
+    ExternalProfileUpdater.update!(DateTime.parse('2020-04-13T10:44:00+00:00'))
```
akofink, line 5: Since we haven't deployed this migration yet, I think it makes sense to perform the data migration here.

test/services/external_profile_updator_test.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class ExternalProfileUpdatorTest < ActiveSupport::TestCase
+  test 'updates profiles before a certain date to be external with default '\
+       'threshold and business objective' do
+    expected_change = Profile.count
+    date = DateTime.now
+
+    assert_difference('Profile.where(external: false).count' => 1) do
+      profiles(:one).dup.update(ref_id: 'foo')
+    end
```
akofink, line 13: This profile represents someone using prod-beta today to create a policy in the UI.

akofink:
Rebased against master

db/migrate/20200414143250_add_external_to_profiles.rb
akofink: @dLobatog did you double check this date and time are correct?

dLobatog: It looks right - the jenkins job finished at 12:45 minus 2h ~> 10:44

dLobatog: You have to add a `require` to this migration - it's failing to load this class

akofink:
There were actually a couple embarrassing issues :P (mentioned inline)

app/services/external_profile_updater.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# A service class to migrate profiles to be external before a certain date
+class ExternalProfileUpdater
```
akofink, line 4: s/Updator/Updater/ and renamed the file as well in the same way

db/migrate/20200414143250_add_external_to_profiles.rb
```diff
@@ -1,6 +1,12 @@
 class AddExternalToProfiles < ActiveRecord::Migration[5.2]
-  def change
+  def up
     add_column :profiles, :external, :boolean, default: false, null: false
     add_index :profiles, :external
+    ExternalProfileUpdater.run!(DateTime.parse('2020-04-13T10:44:00+00:00'))
```
akofink, line 5: s/update!/run!/

akofink:
@dLobatog tests are :cactus:

PR #403, opened 2020-04-16
Destroy all profiles in a policy on policy destroy

akofink:

### DIFF
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+# Methods that are related to a profile's policy
+module ProfilePolicyAssociation
+  extend ActiveSupport::Concern
+
+  included do
+    after_destroy :destroy_policy_test_results
+
+    def policy
+      return self unless external
+
+      Profile.includes(:benchmark)
+             .find_by(account: account_id, external: false, ref_id: ref_id,
+                      benchmarks: { ref_id: benchmark.ref_id })
+    end
+
+    def policy_profiles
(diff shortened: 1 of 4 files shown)


akofink:
I'm working on the rubocop failure. :see_no_evil:

app/models/concerns/profile_policy_association.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+# Methods that are related to a profile's policy
+module ProfilePolicyAssociation
```
akofink, line 4: Rubocop was complaining about the length of `profile.rb`

akofink:
This one should be good for review.

app/models/concerns/profile_policy_association.rb
```diff
@@ -0,0 +1,40 @@
...
+    def business_objective_id
+      (policy || self).read_attribute(:business_objective_id)
+    end
+
+    def compliance_threshold
+      (policy || self).read_attribute(:compliance_threshold)
+    end
+
+    def destroy_policy_test_results
+      policy_profiles.destroy_all
```
dLobatog, line 37: This is potentially OOM, if it has to destroy a couple of 1000s test results, can we do this asynchronously? Something like `app/jobs/delete_host.rb` would work

akofink:
Updated to destroy all other profiles in this policy asynchronously. The first profile, though, is still destroy synchronously so the UI updates. Let me know if it's good enough.

app/models/concerns/profile_policy_association.rb
dLobatog: :+1: it's definitely tidier

PR #404, opened 2020-04-17
Find policy for profile

dLobatog:
This adds the following changes to stable:

* There is a new task (no job runs it yet, so it's only available via manual runs) to sync the DB with the inventory, and drop any records not found in the inventory
* Cache expiration date for failed rules, compliant, and results is 'never' - we only clear this cache when a new report comes in.
* Methods to determine a Policy from a Profile, and added Policy to GQL. 
* Destroy all Profiles if the "parent" Policy is destroyed.  

In addition to that, there are 2 commits which we couldn't put into prod until we got rid of current prod-stable. These changes have been already in CI for a long time.

* Don't destroy orphaned profiles (we do this in prod-stable, but it shouldn't be necessary in the "new app")
* Remove GQL endpoints that are unused

### DIFF
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'test_helper'
+require 'sidekiq/testing'
 
 class ProfileTest < ActiveSupport::TestCase
   should validate_uniqueness_of(:ref_id)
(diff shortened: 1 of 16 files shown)


test/models/profile_test.rb
```diff
@@ -83,6 +84,41 @@ class ProfileTest < ActiveSupport::TestCase
...
+    (bm = benchmarks(:one).dup).update!(version: '0.1.47')
+    (external_profile = profiles(:one).dup).update!(benchmark: bm,
+                                                    compliance_threshold: 100,
+                                                    external: true)
+    profiles(:one).update!(compliance_threshold: 30)
+    assert_equal external_profile.policy, profiles(:one)
+    assert_equal 30, external_profile.compliance_threshold
+  end
+
+  test "destroying a profile also destroys its policy's profiles" do
```
dLobatog, line 109: I'm not sure this unit test is stable - if I run it manually through `rake test TEST=test/models/profile_test.rb` , it always passes. If it runs as part of `rake test`, sometimes it doesn't.

PR #405, opened 2020-04-17
Heavy caching GQL

dLobatog:
Just a draft for now. This requires us to manually warm up the cache first, or during parse. I think that would make sense though. The problem with `.fetch` is that it can't capture batched requests, as it only works for arrays or other basic data types (it'd be crazy if it could store AR relations).

Warm the cache with:

```

        def warm_cache
          Profile.find_each do |profile|
            Rails.cache.write("#{profile&.id}/rules", profile.rules)
            profile.hosts.find_each do |host|
              Rails.cache.write("#{profile.id}/#{host.id}/results", profile.results(host))
              Rails.cache.write("#{host.id}/#{profile.id}/rules_passed", host.rules_passed(profile))
              Rails.cache.write("#{host.id}/#{profile.id}/rules_failed", host.rules_passed(profile))
              Rails.cache.write("#{host.id}/#{profile.id}/score", profile.score(host: host))
            end
            profile.rules.each do |rule|
              Rails.cache.write("#{rule.id}/references", rule.references.map { |ref| [ref.href, ref.label] }.to_json)
              Rails.cache.write("#{rule.id}/identifier", { label: rule.identifier&.label, system: rule.identifier&.system }.to_json)
              Rails.cache.write("#{rule.id}/profiles", rule.profiles)
            end
          end
        end

```

### DIFF
--- a/app/graphql/types/concerns/test_results.rb
+++ b/app/graphql/types/concerns/test_results.rb
@@ -23,6 +23,7 @@ module Types
 
     def score(args = {})
       return cached_score if cached_score(object.id, system_id(args))
+
       latest_test_result_batch(args).then do |latest_test_result|
         if latest_test_result.blank?
           0
@@ -41,7 +42,9 @@ module Types
     end
 
     def rules_passed(args = {})
-      return cached_rules_passed(object.id, system_id(args)) if cached_rules_passed(object.id, system_id(args))
+      return cached_rules_passed(object.id, system_id(args)) if
+             cached_rules_passed(object.id, system_id(args))
+
       latest_test_result_batch(args).then do |latest_test_result|
         if latest_test_result.blank?
           0
@@ -52,7 +55,9 @@ module Types
     end
 
     def rules_failed(args = {})
-      return cached_rules_failed(object.id, system_id(args)) if cached_rules_failed(object.id, system_id(args))
+      return cached_rules_failed(object.id, system_id(args)) if
+             cached_rules_failed(object.id, system_id(args))
+
       latest_test_result_batch(args).then do |latest_test_result|
         if latest_test_result.blank?
           0
--- a/app/graphql/types/concerns/test_results.rb
+++ b/app/graphql/types/concerns/test_results.rb
@@ -6,11 +6,13 @@ module Types
     extend ActiveSupport::Concern
 
     def score(args = {})
-      latest_test_result_batch(args).then do |latest_test_result|
-        if latest_test_result.blank?
-          0
-        else
-          latest_test_result.score
+      ::Rails.cache.fetch("#{system_id(args)}/#{object.id}/score") do
+        latest_test_result_batch(args).then do |latest_test_result|
+          if latest_test_result.blank?
+            0
+          else
+            latest_test_result.score
+          end
         end
       end
     end
@@ -24,21 +26,25 @@ module Types
     end
 
     def rules_passed(args = {})
-      latest_test_result_batch(args).then do |latest_test_result|
-        if latest_test_result.blank?
-          0
-        else
-          latest_test_result.rule_results.passed.count
+      ::Rails.cache.fetch("#{system_id(args)}/#{object.id}/rules_passed") do
+        latest_test_result_batch(args).then do |latest_test_result|
+          if latest_test_result.blank?
+            0
+          else
+            latest_test_result.rule_results.passed.count
+          end
         end
       end
     end
 
     def rules_failed(args = {})
-      latest_test_result_batch(args).then do |latest_test_result|
-        if latest_test_result.blank?
-          0
-        else
-          latest_test_result.rule_results.failed.count
+      ::Rails.cache.fetch("#{system_id(args)}/#{object.id}/rules_failed") do
+        latest_test_result_batch(args).then do |latest_test_result|
+          if latest_test_result.blank?
+            0
+          else
+            latest_test_result.rule_results.failed.count
+          end
         end
       end
     end
--- a/app/graphql/types/concerns/test_results.rb
+++ b/app/graphql/types/concerns/test_results.rb
@@ -22,7 +22,8 @@ module Types
     end
 
     def score(args = {})
-      return cached_score if cached_score(object.id, system_id(args))
+      return cached_score(object.id, system_id(args)) if
+             cached_score(object.id, system_id(args))
 
       latest_test_result_batch(args).then do |latest_test_result|
         if latest_test_result.blank?
--- a/app/graphql/types/concerns/test_results.rb
+++ b/app/graphql/types/concerns/test_results.rb
@@ -5,14 +5,29 @@ module Types
   module TestResults
     extend ActiveSupport::Concern
 
+    def cached_score(profile, host)
+      ::Rails.cache.read(profile: profile, host: host, attribute: 'score')
+    end
+
+    def cached_rules_passed(profile, host)
+      ::Rails.cache.read(
+        profile: profile, host: host, attribute: 'rules_passed'
+      )
+    end
+
+    def cached_rules_failed(profile, host)
+      ::Rails.cache.read(
+        profile: profile, host: host, attribute: 'rules_failed'
+      )
+    end
+
     def score(args = {})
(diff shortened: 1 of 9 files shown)


app/graphql/types/concerns/test_results.rb
```diff
@@ -6,11 +6,13 @@ module TestResults
     extend ActiveSupport::Concern
 
     def score(args = {})
-      latest_test_result_batch(args).then do |latest_test_result|
-        if latest_test_result.blank?
-          0
-        else
-          latest_test_result.score
+      ::Rails.cache.fetch("#{system_id(args)}/#{object.id}/score") do
```
akofink, line 9: If we're using ids, why do we need this hierarchical cache path? In this case, we could just cache `"#{object.id}/score"`, right?

app/graphql/types/rule.rb
```diff
@@ -24,21 +24,27 @@ class Rule < Types::BaseObject
     end
 
     def compliant(args = {})
-      system_id(args) &&
-        profile_id(args) &&
-        %w[pass notapplicable notselected].include?(
-          context[:rule_results][object.id][profile_id(args)]
-        )
+      ::Rails.cache.fetch(
+        "#{system_id(args)}/#{profile_id(args)}/#{object.id}/compliant"
```
akofink, line 28: Ok, this one does make sense, since multiple systems can be tested against a single rule+profile combination.

app/graphql/types/rule.rb
```diff
@@ -24,21 +24,27 @@ class Rule < Types::BaseObject
...
+          )
+      end
     end
 
     def references
-      if context[:"rule_references_#{object.id}"].nil?
-        ::CollectionLoader.for(::Rule, :rule_references)
-                          .load(object).then do |references|
-          references.map { |ref| [ref.href, ref.label] }.to_json
+      ::Rails.cache.fetch("#{object.id}/references") do
```
akofink, line 39: :+1:

app/services/concerns/xccdf/util.rb
```diff
@@ -52,10 +52,25 @@ def set_openscap_parser_data
 
       def invalidate_cache
         Rails.cache.delete("#{@host&.id}/failed_rule_objects_result")
-        Rails.cache.delete("#{@new_host_profile&.id}/#{@host&.id}/results")
+        Rails.cache.delete("#{@host_profile&.id}/#{@host&.id}/results")
+        Rails.cache.delete("#{@host_profile&.id}/rules")
+        Rails.cache.delete("#{@host.id}/#{@host_profile&.id}/rules_passed")
+        Rails.cache.delete("#{@host.id}/#{@host_profile&.id}/rules_failed")
+        Rails.cache.delete("#{@host.id}/#{@host_profile&.id}/score")
         @host_profile.rules.each do |rule|
```
akofink, line 60: Wow, maybe we could break this into a method, like `invalidate_cache_for(host:, profile:)`

app/services/concerns/xccdf/util.rb
```diff
@@ -52,10 +52,25 @@ def set_openscap_parser_data
...
+          Rails.cache.delete("#{rule.id}/identifier")
+          Rails.cache.delete("#{rule.id}/profiles")
         end
+
+        @host_profile.rules.each do |rule|
+          Rails.cache.write("#{rule.id}/references", rule.references.map { |ref| [ref.href, ref.label] }.to_json)
+          Rails.cache.write("#{rule.id}/identifier", { label: rule.identifier&.label, system: rule.identifier&.system }.to_json
+                           )
+          Rails.cache.write("#{rule.id}/profiles", rule.profiles)
+        end
```
akofink, line 72: I'd prefer we split this into a `warm_cache` method. What's the most expensive operation here? References seems good to warm; I thought rule results were bad too, right?

dLobatog: Yeah - don't even bother to look too closely :) this is just cruft I had

app/graphql/types/rule.rb
```diff
@@ -23,41 +23,65 @@ class Rule < Types::BaseObject
...
 
+    # rubocop:disable Metrics/AbcSize
     def references
+      return cached_references if cached_references
+
       if context[:"rule_references_#{object.id}"].nil?
         ::CollectionLoader.for(::Rule, :rule_references)
                           .load(object).then do |references|
-          references.map { |ref| [ref.href, ref.label] }.to_json
+          references.map { |ref| { href: ref.href, label: ref.label } }.to_json
```
dLobatog, line 49: Notice I'm bringing this back in because current frontend-components isn't working: if we get it in, we can save this https://github.com/RedHatInsights/frontend-components/pull/471

app/services/cache_helper.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
+# rubocop:disable Metrics/ModuleLength
```
akofink, line 5: This might be a sign of something :rotating_light: You should probably break this up into multiple modules, and that might give you more room to place cache-related methods that you've put directly into graphql types.

app/graphql/types/rule.rb
```diff
@@ -23,41 +23,65 @@ class Rule < Types::BaseObject
...
+          )
+      end
+    end
+    # rubocop:enable Metrics/AbcSize
+
+    def cached_references
+      ::Rails.cache.read(rule: object.id, attribute: 'references')
     end
 
+    # rubocop:disable Metrics/AbcSize
```
akofink, line 42: :see_no_evil:

bastilian:
I agree with @akofink. `graphql-cache` would be really nice or something like it where we can easily enable caching for certain queries or parts or results, but maybe we can add something similar that is not as comprehensive, but gives us just enough for now.

A solution i'm imagining is to have this 
```ruby
  cached_fields :rules, :profiles
```
in the GraphQL types and warming the cache would just be a matter of executing the GraphQL for the objects. The `CacheHelper` module in that case would have one method that wraps the methods declared in `cached_rules` to fetch or write cache and be the only place where the caching logic sits. The cache keys could be a combination of method/field name, object id and passed arguments. This would also make it very easy to add more caching to the application.

bastilian:
On a side note, do we have these, or will we have similar performance issues on the REST API?

dLobatog:
I have a proof of concept at https://github.com/dlobatog/compliance-backend/tree/gql-cache  - there are many things I haven't got around yet. 

1. It has problems with caching any field that returns a batched field. This is probably a hard blocker but I don't know yet.
2. The keys it generates are absurd for our use cases and cannot be easily warmed or invalidated:
```
127.0.0.1:6379> keys compliance*
1) "compliance:System:id:Host:hosts/3d1a82c5-33ce-4d4f-8fbe-2dc119532cdb-20200227230913603410"
2) "compliance:Profile:name:Profile:profiles/b8b6d412-c1ce-46e2-9fe4-9274691eb832-20200407133203625444"
3) "compliance:Profile:id:Profile:profiles/2712fd29-d56b-4b10-bf67-ba8d3aae7970-20200227230913631469"
4) "compliance:Profile:id:Profile:profiles/b8b6d412-c1ce-46e2-9fe4-9274691eb832-20200407133203625444"
5) "compliance:System:profiles:Host:hosts/3d1a82c5-33ce-4d4f-8fbe-2dc119532cdb-20200227230913603410"
6) "compliance:Profile:refId:Profile:profiles/b8b6d412-c1ce-46e2-9fe4-9274691eb832-20200407133203625444"
7) "compliance:Profile:name:Profile:profiles/2712fd29-d56b-4b10-bf67-ba8d3aae7970-20200227230913631469"
8) "compliance:System:name:Host:hosts/3d1a82c5-33ce-4d4f-8fbe-2dc119532cdb-20200227230913603410"
9) "compliance:Profile:description:Profile:profiles/b8b6d412-c1ce-46e2-9fe4-9274691eb832-20200407133203625444"
```
I'm looking into how to solve both, because it's extremely convenient to just write `cache: true` next to the field...

dLobatog:
Closing for https://github.com/RedHatInsights/compliance-backend/pull/411

PR #406, opened 2020-04-17
Add missing indexes and uniqueness validators

akofink:
Unique indexes and validations, where missing, for:
- BusinessObjective
- Host
- RuleIdentifier
- RuleReference
- RuleResult
- TestResult

### DIFF
--- a/app/services/duplicate_rule_reference_resolver.rb
+++ b/app/services/duplicate_rule_reference_resolver.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+# A service class to merge duplicate RuleReference objects
+class DuplicateRuleReferenceResolver
+  class << self
+    def run!
+      @rule_references = nil
+      duplicate_rule_references.find_each do |rule_reference|
+        if existing_rule_reference(rule_reference)
+          migrate_rule_reference(existing_rule_reference(rule_reference),
+                                 rule_reference)
+        else
+          self.existing_rule_reference = rule_reference
+        end
+      end
+    end
+
+    private
+
+    def existing_rule_reference(rule_reference)
+      rule_references[[rule_reference.href, rule_reference.label]]
+    end
+
+    def existing_rule_reference=(rule_reference)
+      rule_references[[rule_reference.href,
+                       rule_reference.label]] = rule_reference
+    end
+
+    def rule_references
+      @rule_references ||= {}
+    end
+
+    def duplicate_rule_references
+      RuleReference.joins(
+        "JOIN (#{grouped_nonunique_rule_reference_tuples.to_sql}) as rr on "\
+        'rule_references.href = rr.href AND '\
+        'rule_references.label = rr.label'
+      )
+    end
+
+    def grouped_nonunique_rule_reference_tuples
+      RuleReference.select(:href, :label).group(:href, :label)
+                   .having('COUNT(id) > 1')
+    end
+
+    def migrate_rule_reference(existing_rr, duplicate_rr)
+      migrate_rule_reference_rules(existing_rr, duplicate_rr)
+      duplicate_rr.destroy
+    end
+
+    # rubocop:disable Rails/SkipsModelValidations
+    def migrate_rule_reference_rules(existing_rr, duplicate_rr)
+      RuleReferencesRule.where(rule_reference: duplicate_rr.id)
+                        .where.not(rule: existing_rr.rules)
+                        .update_all(rule_reference_id: existing_rr.id)
+    end
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+end
--- a/app/services/duplicate_test_result_resolver.rb
+++ b/app/services/duplicate_test_result_resolver.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+# A service class to merge duplicate TestResult objects
+class DuplicateTestResultResolver
+  class << self
+    def run!
+      @test_results = nil
+      duplicate_test_results.find_each do |test_result|
+        if existing_test_result(test_result)
+          migrate_test_result(existing_test_result(test_result),
+                              test_result)
+        else
+          self.existing_test_result = test_result
+        end
+      end
+    end
+
+    private
+
+    def existing_test_result(test_result)
+      test_results[[test_result.host_id, test_result.profile_id,
+                    test_result.end_time]]
+    end
+
+    def existing_test_result=(test_result)
+      test_results[[test_result.host_id, test_result.profile_id,
+                    test_result.end_time]] = test_result
+    end
+
+    def test_results
+      @test_results ||= {}
+    end
+
+    def duplicate_test_results
+      TestResult.joins(
+        "JOIN (#{grouped_nonunique_test_result_tuples.to_sql}) as tr on "\
+        'test_results.host_id = tr.host_id AND '\
+        'test_results.profile_id = tr.profile_id AND '\
+        'test_results.end_time = tr.end_time'
+      )
+    end
+
+    def grouped_nonunique_test_result_tuples
+      TestResult.select(:host_id, :profile_id, :end_time)
+                .group(:host_id, :profile_id, :end_time)
+                .having('COUNT(id) > 1')
+    end
+
+    def migrate_test_result(existing_tr, duplicate_tr)
+      migrate_rule_results(existing_tr, duplicate_tr)
+      duplicate_tr.destroy
+    end
+
+    # rubocop:disable Rails/SkipsModelValidations
+    def migrate_rule_results(existing_tr, duplicate_tr)
+      duplicate_tr.rule_results.where.not(host_id: existing_tr.host_id).or(
+        duplicate_tr.rule_results.where.not(
+          rule_id: existing_tr.rule_results.select(:rule_id)
+        )
+      ).update_all(test_result_id: existing_tr.id)
+    end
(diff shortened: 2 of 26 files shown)


akofink:
The test failure here should be fixed with https://github.com/RedHatInsights/compliance-backend/pull/407:

```
========== FAIL["test_destroying_a_profile_also_destroys_its_policy's_profiles", #<Minitest::Reporters::Suite:0x00005587114acf20 @name="ProfileTest">, 7.134416306391358]

 test_destroying_a_profile_also_destroys_its_policy's_profiles#ProfileTest (7.13s)

        Expected: 1

          Actual: 2

        test/models/profile_test.rb:117:in `block in <class:ProfileTest>'
```

db/migrate/20200417174651_add_unique_index_to_rule_identifiers.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToRuleIdentifiers < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:rule_identifiers, %i[label system rule_id], unique: true)
```
dLobatog, line 3: This one made me think for a second, what happens for Rules with different SSG, there would be multiple identifiers for those - but there are multiple rule_ids for those anyway so this still holds.

db/migrate/20200417174333_add_unique_index_to_hosts.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToHosts < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:hosts, %i[name account_id], unique: true)
```
dLobatog, line 3: I don't think that's right - you can have multiple hosts with the same name in the same account, ID is the only field that should be unique for hosts. We actually have a few of these already in QA at least

db/migrate/20200417174053_add_unique_index_to_business_objectives.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToBusinessObjectives < ActiveRecord::Migration[5.2]
+  def change
+    add_index :business_objectives, :title, unique: true
```
dLobatog, line 3: I don't think this is right, multiple accounts should be able to have multiple business objectives with the same title. I tried this and the moment you try to create a BO from an account that may not even have any yet, you get `:title=>["has already been taken"]`.

dLobatog (changes requested):
@akofink The BusinessObjective and Host validations are wrong I think, :+1:  to the rest!

db/migrate/20200417174053_add_unique_index_to_business_objectives.rb
akofink: Makes sense :+1: I can't think of a good way to validate title is unique within a profile via the DB, so I guess we'll skip this one.

akofink:
Updated with your suggestions. I'm testing all the migration PRs together, so please do not merge this yet.

db/migrate/20200417174948_add_unique_index_to_rule_references.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToRuleReferences < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:rule_references, %i[href label], unique: true)
```
akofink, line 3: It seems we have some duplicates here

db/migrate/20200417175851_add_unique_index_to_rule_results.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToRuleResults < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:rule_results, %i[host_id rule_id test_result_id], unique: true)
```
akofink, line 3: This one, too. It shows about 63K duplicate records

dLobatog (changes requested):
@akofink There are still a few leftovers from the previous version, also I'm unclear as to why the `joins` is necessary, after fixing these I think it'll be on the right way to merge it. Thanks!

app/services/duplicate_rule_reference_resolver.rb
```diff
@@ -0,0 +1,58 @@
...
+      rule_references[[rule_reference.href,
+                       rule_reference.label]] = rule_reference
+    end
+
+    def rule_references
+      @rule_references ||= {}
+    end
+
+    def duplicate_rule_references
+      RuleReference.joins(
```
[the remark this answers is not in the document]
akofink: `grouped_nonunique_rule_reference_tuples` does not contain ID, as it's grouped on the fields that the table should be unique by (href and label in this case). This is identical to how we've been finding duplicates in all of these PRs.

db/migrate/20200417175851_add_unique_index_to_rule_results.rb
akofink: :shrug: a report could have 2 rule results referencing the same rule, perhaps.

akofink:
Updated - there are some missing tests on this PR still for duplicate resolvers.

db/migrate/20200417174333_add_unique_index_to_hosts.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToHosts < ActiveRecord::Migration[5.2]
+  def change
+    add_index(:hosts, %i[id account_id], unique: true)
```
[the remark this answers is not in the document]
akofink: Deleted

app/services/duplicate_test_result_resolver.rb
```diff
@@ -0,0 +1,64 @@
...
+      migrate_rule_results(existing_tr, duplicate_tr)
+      duplicate_tr.destroy
+    end
+
+    # rubocop:disable Rails/SkipsModelValidations
+    def migrate_rule_results(existing_tr, duplicate_tr)
+      duplicate_tr.rule_results.where.not(host_id: existing_tr.host_id).or(
+        duplicate_tr.rule_results.where.not(
+          rule_id: existing_tr.rule_results.select(:rule_id)
+        )
```
akofink, line 59: This query changed to use or. Before it was using AND, which was incorrect (and found via the added tests).

test/fixtures/rule_results.yml
```diff
@@ -1,11 +1,13 @@
 # Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
 
 one:
-  host_id: one
-  rule_id: one
+  host: one
+  rule: one
+  test_result: one
```
akofink, line 6: This is how to define relations in fixtures. Before, host and rule were just set to `nil`. This required updates to only 2 other tests which were relying on host and rule to be `nil`.

test/services/duplicate_rule_reference_resolver_test.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require './db/migrate/20200417174948_add_unique_index_to_rule_references'
+require 'sidekiq/testing'
+
+class DuplicateRuleReferenceResolverTest < ActiveSupport::TestCase
```
akofink, line 7: New tests for the new resolvers were added. All the tests follow the same format.

akofink:
Okay! So, this PR should be good for final review. Some tests needed some updates due to the new DB constraints - `(validate: false)` doesn't work when the DB is validating fields :stuck_out_tongue_closed_eyes:

spec/integration/rules_spec.rb
```diff
@@ -98,8 +98,9 @@
           )
           user = User.from_x_rh_identity(x_rh_identity[:identity])
           user.save
-          profiles(:one).update(account: user.account, hosts: [hosts(:one)])
-          rules(:one).update(profiles: [profiles(:one)])
+          profiles(:one).update!(account: user.account, hosts: [hosts(:one)])
+          rules(:one).rule_results.destroy_all
+          rules(:one).update!(profiles: [profiles(:one)])
```
akofink, line 103: This `update` was silently failing all the sudden due to the newly associated rule_result fixtures. I've updated them to use `update!` so this kind of thing is easier to spot in the future. The update was failing because the associated rule results were invalid - deleting them from the rule fixed the test.

test/services/duplicate_rule_resolver_test.rb
```diff
@@ -49,10 +49,9 @@ class DuplicateRuleResolverTest < ActiveSupport::TestCase
 
   test 'resolves rule_results from a duplicate rule' do
     assert_difference('RuleResult.count' => 2) do
-      (rr = rule_results(:one).dup).save(validate: false)
-      (rr2 = rule_results(:one).dup).save(validate: false)
-      rules(:one).rule_results << rr
-      @dup_rule.rule_results << rr2
+      rule_results(:one).dup.update!(rule: rules(:one), host: hosts(:two),
+                                     test_result: test_results(:two))
+      rule_results(:one).dup.update!(rule: @dup_rule, host: hosts(:two))
```
akofink, line 54: This test wasn't working anymore due to the new DB constraints. Updating the associated test_result and host to be different from `rule_results(:two)` fixture was enough to fix this test.

dLobatog (approved):
Ready to merge IMO , appreciate the effort with tests & covering all possible models, thanks @akofink !

akofink:
:champagne: Thanks for the rounds of reviews, @dLobatog! One more PR in this milestone remaining!

PR #410, opened 2020-04-21
Add full profiles endpoint

akofink:
WIP for create and delete and also more tests.

`PUT profiles/<profile_id>/ < data.json`

data.json:

```json
{
	"data": {
		"attributes": {
			"description": "This profile contains rules to ensure standard security baseline\nof a Red Hat Enterprise Linux 7 system. Regardless of your system's workload\nall of these checks should pass.",
			"name": "Standard System Security Profile for Red Hat Enterprise Linux 7",
			"compliance_threshold": 90,
			"business_objective": "Japan"
		}
	}
}
```

DO NOT MERGE - this was just a PoC

### DIFF
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# Methods that are related to profile searching
+module ProfileSearching
+  extend ActiveSupport::Concern
+
+  included do
+    scoped_search on: %i[id name ref_id account_id compliance_threshold
+                         external parent_profile_id]
+    scoped_search relation: :hosts, on: :id, rename: :system_ids
+    scoped_search relation: :hosts, on: :name, rename: :system_names
+    scoped_search on: :has_test_results, ext_method: 'test_results?',
+                  only_explicit: true, operators: ['=']
+    scoped_search on: :canonical, ext_method: 'canonical?', only_explicit: true,
+                  operators: ['=']
+  end
+
+  class_methods do
+    def canonical?(_filter, _operator, value)
+      operator = ActiveModel::Type::Boolean.new.cast(value) ? '' : 'NOT'
+      { conditions: "parent_profile_id IS #{operator} NULL" }
+    end
+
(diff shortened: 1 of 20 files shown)


app/serializers/profile_serializer.rb
```diff
@@ -13,4 +14,7 @@ class ProfileSerializer
   attribute :compliant_host_count do |profile|
     profile.hosts.count { |host| profile.compliant?(host) }
   end
+  attribute :business_objective do |profile|
+    profile.business_objective.title
```
akofink, line 18: To match the UI in which we have business objective that looks like a string field on a profile, the API as written allows updating business_objective as a string. We find or create the BO for the user. Traditional REST APIs would force users to first create a BO with a BO API, get the ID, then pass `business_objective_id` to the profiles API. I'm open to both methods, but the way it's written is much more user friendly; I wouldn't probably do this if a profile had many BOs, but since it only has one, this seemed OK.

app/models/concerns/profile_searching.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  included do
+    scoped_search on: %i[id name ref_id account_id compliance_threshold
+                         external parent_profile_id]
+    scoped_search relation: :hosts, on: :id, rename: :system_ids
+    scoped_search relation: :hosts, on: :name, rename: :system_names
+    scoped_search on: :has_test_results, ext_method: 'test_results?',
+                  only_explicit: true, operators: ['=']
+    scoped_search on: :canonical, ext_method: 'canonical?', only_explicit: true,
+                  operators: ['=']
```
akofink, line 15: We should also be able to search by OS version here, similar to how the wizard works via graphql.

Victoremepunto: Thanks Andrew. Yes it will definitely help us to identify uniquely a policy (not more than 1 entities returned) when searching by ref_id, external=False, and the RHEL major version.

akofink:
Closing - This PR was broken apart and has been submitted/merged in many other PRs.

PR #411, opened 2020-04-21
Introduce graphql-cache

dLobatog:
This PR introduces graphql-cache. I'll comment in each file what are the relevant changes.


Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'puma', '~> 3.12'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 # gem 'jbuilder', '~> 2.5'
 # Use Redis adapter to run Action Cable in production
-gem 'redis', '~> 4.0'
+gem 'hiredis'
```
dLobatog, line 14: This is just a more performant version of the redis gem

Gemfile
```diff
@@ -68,13 +68,14 @@ gem 'prometheus_exporter'
 
 gem 'activerecord-import'
 gem 'oj'
-gem 'newrelic_rpm'
```
dLobatog, line 71: We are simply not using it and it adds overhead to every API call

app/graphql/schema.rb
```diff
@@ -15,6 +15,5 @@ class Schema < GraphQL::Schema
   mutation Types::Mutation
   lazy_resolve(Promise, :sync)
   use GraphQL::Batch
-  use GraphQL::Execution::Interpreter
-  use GraphQL::Analysis::AST
+  use GraphQL::Cache
```
dLobatog, line 18: Unfortunately the new interpreter is incompatible with gql-cache (https://graphql-ruby.org/queries/interpreter.html)  (https://github.com/stackshareio/graphql-cache/issues/71)

config/initializers/graphql_cache.rb
```diff
@@ -0,0 +1,36 @@
+GraphQL::Cache.configure do |config|
+  config.namespace = 'compliance' # Cache key prefix for keys generated by graphql-cache
+  config.cache     = Rails.cache      # The cache object to use for caching
+  config.logger    = Rails.logger     # Logger to receive cache-related log messages
+  config.expiry    = 15552000         # 6 months (in seconds)
```
dLobatog, line 5: We should probably discuss this and even think more seriously about introducing an LRU cache redis instance directly. Fields are cached alongside its updated_at, e.g: `compliance:Rule:rules/30b9d4c7-4924-4996-bd89-7e29117fccf7-20200227230834209106:title` to force us to realize whether the key can be used or not. It's a sort of 'automatic invalidation' as any time a rule is updated, the old entries will be useless.

config/initializers/graphql_cache.rb
```diff
@@ -0,0 +1,36 @@
...
+end
+
+module GraphQL
+  module Cache
+    module DeconstructorExtensions
+      def perform
+        return raw.value if method == 'lazy'
```
dLobatog, line 12: `GraphQL::Cache::Deconstructor` figures out which sort of value each field is returning. Based on that, the current methods cannot handle lazy methods, or Promises. This is just a way to handle those.

config/initializers/graphql_cache.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+
+    class Deconstructor
+      prepend DeconstructorExtensions
+    end
+
+    class Key
+      def to_s
+        @to_s ||= [
```
dLobatog, line 27: This overrides the way the keys are set - I'll comment more on this in a few

app/graphql/types/interfaces/rules_preload.rb
```diff
@@ -4,6 +4,7 @@
 # rule results, compliant,and references
 module RulesPreload
   include GraphQL::Schema::Interface
+  field_class GraphQL::Cache::Field
```
akofink, line 7: Is this required for some reason? This module is implemented from classes which inherit from `BaseObject`, so I thought the field_class would be set already.

config/initializers/graphql_cache.rb
akofink: So the old entries with older `updated_at` in the key are kept in redis?

dLobatog: Correct, that's why I'd want to go with an instance like https://redis.io/topics/lru-cache

dLobatog: Correct, that's why I'd want to go with an instance like https://redis.io/topics/lru-cache

PR #412, opened 2020-04-21
Task to import supported SSG benchmarks

dLobatog:

### DIFF
--- a/lib/tasks/import_datastream.rake
+++ b/lib/tasks/import_datastream.rake
@@ -29,7 +29,7 @@ namespace :ssg do
     begin
       ENV['DATASTREAMS'] = ::Xccdf::Benchmark::LATEST_SUPPORTED_VERSIONS
         .map do |ref_id, version|
-        "v#{version}:#{ref_id.to_s.downcase.split('_').last.sub('-', '')}"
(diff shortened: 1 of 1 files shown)


lib/tasks/import_datastream.rake
```diff
@@ -23,6 +23,22 @@ namespace :ssg do
     end
   end
 
+  desc 'Update compliance DB with the supported SCAP Security Guide versions'
+  task import_rhel_supported: [:environment] do
+    # DATASTREAM_FILENAMES from openscap_parser's ssg:sync
+    begin
+      ENV['DATASTREAMS'] = ::Xccdf::Benchmark::LATEST_SUPPORTED_VERSIONS
+        .map do |ref_id, version|
+        "v#{version}:#{ref_id.to_s.downcase.split('_').last.sub('-', '')}"
```
akofink, line 32: This works fine, but I can't help but suggest a simpler regex version:

```suggestion
        "v#{version}:rhel#{ref_id[/\d+$/]}"
```

akofink, line 32: This works fine, but I can't help but suggest a simpler regex version:

```suggestion
        "v#{version}:rhel#{ref_id[/\d+$/]}"
```

PR #416, opened 2020-04-22
Add benchmark REST API docs

akofink:


test/fixtures/benchmarks.yml
```diff
@@ -1,5 +1,6 @@
 # Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
-
+_fixture:
+  model_class: Xccdf::Benchmark
```
akofink, line 3: This is set here now globally, so we don't need any more `fixture_class` declarations.

PR #420, opened 2020-04-22
Use request_id instead of payload_id

brantleyr:
The Payload Tracker is transitioning to using the new "request_id" terminology instead of the legacy "payload_id." This renames that field.

### DIFF
--- a/test/consumers/compliance_reports_consumer_test.rb
+++ b/test/consumers/compliance_reports_consumer_test.rb
@@ -35,31 +35,6 @@ class ComplianceReportsConsumerTest < ActiveSupport::TestCase
     end
   end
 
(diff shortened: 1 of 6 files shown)


test/consumers/compliance_reports_consumer_test.rb
```diff
@@ -36,21 +35,20 @@ class ComplianceReportsConsumerTest < ActiveSupport::TestCase
     end
   end
 
-  test 'report message is parsed and job is enqueued with payload_id' do
+  test 'report message is parsed and job is enqueued with request_id' do
```
akofink, line 38: Just delete this test. There's an identical one now at line 13

PR #421, opened 2020-04-23
Add Faraday::Retry middleware

dLobatog:
This should retry 3 times if it finds any of the expected exceptions. Unfortunately we are hitting sometimes remediations or RBAC when the pods are restarting or similar, you can see the #alerts-cloudservices-apps for an example of this. It should go:

1. Try request between 0.05+0.025 seconds
2. Try request between 0.1+0.05 seconds
3. Try request between 0.2+0.1 seconds

The goal is to have less "false alarms" in the channel and actually succeed with some requests that shouldn't fail.

### DIFF
--- a/app/services/platform.rb
+++ b/app/services/platform.rb
@@ -9,9 +9,9 @@ module Platform
     interval: 0.05,
     interval_randomness: 0.5,
     backoff_factor: 2,
+    methods: %i[get],
     exceptions: [
-      Faraday::ConnectionFailed, Faraday::TimeoutError, Errno::ETIMEDOUT,
-      Timeout::Error, Error::TimeoutError
+      *Faraday::Request::Retry::DEFAULT_EXCEPTIONS, Faraday::ConnectionFailed
     ]
   }.freeze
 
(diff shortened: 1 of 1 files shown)


bastilian:
Could we limit this to only the RBAC API requests?

I don't see any place where we do use an API at the moment to create records, but maybe we will and imagining a scenario where a request would timeout creating something and then retry the same request, that could fail if the first just timed out, but still created some records.

app/services/platform.rb
```diff
@@ -4,9 +4,21 @@
...
 # Methods related to connecting to other platform services
 module Platform
+  RETRY_OPTIONS = {
+    max: 3,
+    interval: 0.05,
+    interval_randomness: 0.5,
+    backoff_factor: 2,
+    methods: %i[get],
+    exceptions: [
+      *Faraday::Request::Retry::DEFAULT_EXCEPTIONS, Faraday::ConnectionFailed
```
dLobatog, line 14: I've changed this now to use the DEFAULT_EXCEPTIONS, which include calls that return `RetryAfter`, Timeout, etc.. It seems cleaner than writing out all of the possibilities ourselves

PR #428, opened 2020-04-27
Remove unused gql fields

dLobatog:
Please double-check first, but these fields are not used anywhere in our frontend I believe, and some are just wrong - e.g: there's no concept of "Compliant" system, profile-names should be extracted from `profiles { name }` and same goes for rule objects failed.


akofink:
We get to use the new label! :man_dancing:

dLobatog:
Discussed the plan with QE - notice it won't make it past smoke_tests, merging

PR #429, opened 2020-04-27
RHICOMPL-631 - Update hostnames based on 'update' message

dLobatog:
On Apr 28th, there will be a new event to tell apps to update hostnames if the inventory hostname changed. On the UI, this shouldn't affect us as we already always display the inventory hostname (unless the host is unsynced for whichever reason and is only on compliance, this should not happen in prod though).

On the API, this would be visible as we don't contact the inventory for anything there. So we should update the names according to the message to serve the right name.

https://platform-docs.cloud.paas.psi.redhat.com/backend/inventory.html#host-patching for more info on the message format

### DIFF
--- a/app/jobs/delete_host.rb
+++ b/app/jobs/delete_host.rb
@@ -9,7 +9,7 @@ class DeleteHost
     Host.transaction do
       RuleResult.where(host_id: message['id']).delete_all
       rescue_not_found do
-        Host.find_by!(id: message['id'])&.destroy
+        Host.find_by!(id: message['id']).destroy
       end
     end
   end
--- a/app/jobs/delete_host.rb
+++ b/app/jobs/delete_host.rb
@@ -3,16 +3,14 @@
 # Job meant to delete hosts and associated objects asynchronously
 class DeleteHost
   include Sidekiq::Worker
+  include RecordNotFound
 
   def perform(message)
     Host.transaction do
       RuleResult.where(host_id: message['id']).delete_all
-      Host.find_by(id: message['id'])&.destroy
+      rescue_not_found do
+        Host.find_by!(id: message['id'])&.destroy
+      end
     end
-  rescue ActiveRecord::RecordNotFound => e
-    Sidekiq.logger.info(
-      "#{e.message} (#{e.class}) "\
-      '- this host ID was not registered in Compliance'
-    )
   end
 end
--- a/app/jobs/inventory_host_updated_job.rb
+++ b/app/jobs/inventory_host_updated_job.rb
@@ -11,7 +11,7 @@ class InventoryHostUpdatedJob
     rescue_not_found do
       Host.find_by!(
         id: message['host']['id']
-      ).update(name: message['host']['display_name'])
+      ).update!(name: message['host']['display_name'])
     end
   end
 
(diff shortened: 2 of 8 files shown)


app/jobs/inventory_host_updated_job.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# Updates the host according to the inventory definition
+class InventoryHostUpdatedJob
+  include Sidekiq::Worker
+  include RecordNotFound
+
+  def perform(message)
+    if message['host'] && message['host']['display_name']
```
akofink, line 9: There's a nicer way, if you want:

```suggestion
    if message.dig('host', 'display_name')
```

app/jobs/inventory_host_updated_job.rb
```diff
@@ -0,0 +1,26 @@
...
+      rescue_not_found do
+        Host.find_by!(
+          id: message['host']['id']
+        )&.update(name: message['host']['display_name'])
```
akofink, line 13: This safe operator `&.` won't ever be used because `find_by!` raises an exception, never returning `nil`.

app/jobs/inventory_host_updated_job.rb
```diff
@@ -0,0 +1,26 @@
...
+# Updates the host according to the inventory definition
+class InventoryHostUpdatedJob
+  include Sidekiq::Worker
+  include RecordNotFound
+
+  def perform(message)
+    if message['host'] && message['host']['display_name']
+      rescue_not_found do
+        Host.find_by!(
+          id: message['host']['id']
```
akofink, line 12: We should validate all parts of the message IMO. Maybe in another method.

app/jobs/delete_host.rb
```diff
@@ -3,16 +3,14 @@
...
 class DeleteHost
   include Sidekiq::Worker
+  include RecordNotFound
 
   def perform(message)
     Host.transaction do
       RuleResult.where(host_id: message['id']).delete_all
-      Host.find_by(id: message['id'])&.destroy
+      rescue_not_found do
+        Host.find_by!(id: message['id'])&.destroy
```
akofink, line 12: Again, `find_by!` will never return nil. The safe operator `&.` does nothing.

app/jobs/inventory_host_updated_job.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    rescue_not_found do
+      Host.find_by!(
+        id: message['host']['id']
+      ).update(name: message['host']['display_name'])
+    end
+  end
+
+  def valid_message_format(message)
+    message.dig('host', 'display_name') && message.dig('host', 'id')
```
akofink, line 19: :+1:

app/jobs/inventory_host_updated_job.rb
```diff
@@ -0,0 +1,28 @@
...
+  include Sidekiq::Worker
+  include RecordNotFound
+
+  def perform(message)
+    return wrong_format_warning(message) unless valid_message_format(message)
+
+    rescue_not_found do
+      Host.find_by!(
+        id: message['host']['id']
+      ).update(name: message['host']['display_name'])
```
akofink, line 14: Do you want this to silently fail?

```suggestion
      ).update!(name: message['host']['display_name'])
```

PR #432, opened 2020-04-27
Change entitlement check to 'insights'

dLobatog:


akofink:
You're going to need to update the swagger_helper https://github.com/RedHatInsights/compliance-backend/blob/master/spec/swagger_helper.rb#L207

PR #436, opened 2020-04-28
Add index to test_results by 'latest' fields

dLobatog:
We are constantly making queries by these 3 fields. Adding a field to speed up `SELECT "test_results".* FROM "test_results" JOIN (SELECT "test_results"."profile_id", "test_results"."host_id", MAX(end_time) as end_time FROM "test_results" GROUP BY "test_results"."profile_id", "test_results"."host_id") as tr on test_results.profile_id = tr.profile_id AND test_results.host_id = tr.host_id AND test_results.end_time = tr.end_time` queries is a dead giveaway in this case we don't have. The cons of adding the index (INSERT/DELETE become less performant) are not visible to the user as those happen through Sidekiq.

### DIFF
--- a/db/migrate/20200428110354_add_index_to_test_results.rb
+++ b/db/migrate/20200428110354_add_index_to_test_results.rb
@@ -1,0 +1,5 @@
+class AddIndexToTestResults < ActiveRecord::Migration[5.2]
+  def change
+    add_index :test_results, [:profile_id, :host_id, :end_time]
+  end
+end
--- a/db/migrate/20200428110354_add_index_to_test_results.rb
+++ b/db/migrate/20200428110354_add_index_to_test_results.rb
@@ -1,5 +1,5 @@
 class AddIndexToTestResults < ActiveRecord::Migration[5.2]
   def change
-    add_index :test_results, [:profile_id, :host_id, :end_time]
+    add_index :test_results, [:profile_id, :host_id, :end_time], unique: true
   end
 end
--- a/app/graphql/types/interfaces/rules_preload.rb
+++ b/app/graphql/types/interfaces/rules_preload.rb
@@ -41,9 +41,9 @@ module RulesPreload
     rules.each do |rule|
       context[:parent_profile_id][rule.id] = object.id
     end
-    #if args[:lookahead].selects?(:references)
-      #initialize_rule_references_context(rule_results)
-    #end
+    if args[:lookahead].selects?(:references)
+      initialize_rule_references_context(rule_results)
+    end
     return unless args[:lookahead].selects?(:compliant)
 
     context[:rule_results] ||= {}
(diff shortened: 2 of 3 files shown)


db/migrate/20200428110354_add_index_to_test_results.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToTestResults < ActiveRecord::Migration[5.2]
+  def change
+    add_index :test_results, [:profile_id, :host_id, :end_time]
```
akofink, line 3: The only missing index should be `end_time`. You can check in db/schema.rb:

```
  create_table "test_results", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
    t.datetime "start_time"
    t.datetime "end_time"
    t.decimal "score"
    t.uuid "profile_id"
    t.uuid "host_id"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.index ["host_id"], name: "index_test_results_on_host_id"
    t.index ["profile_id"], name: "index_test_results_on_profile_id"
  end
```

dLobatog: This is not adding 3 indices - it's adding a combined index on the 3 fields which is used everytime we call `latest` https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/test_result.rb#L16

akofink: Ok, I see - this should be `unique: true` as I've done here: https://github.com/RedHatInsights/compliance-backend/pull/406/files#diff-277bf976622928a3a96e1377f1359bb6R5

Otherwise, we will need to remove this index and re-add it as unique when #406 is merged.

dLobatog (commented):
@akofink Thanks - updated to add the uniqueness constraint to the index

akofink (commented):
Ok, looks good, but this might fail in prod if there are any duplicates. Why don't you copy over the DuplicateTestResultResolver here? https://github.com/RedHatInsights/compliance-backend/pull/406/files#diff-647cd69e51b33c9bee0403118886d982R4

From prod:

```
TestResult.joins("JOIN (#{TestResult.select(:host_id, :profile_id, :end_time).group(:host_id, :profile_id, :end_time).having('COUNT(id) > 1').to_sql}) as tr on test_results.host_id = tr.host_id AND test_results.profile_id = tr.profile_id AND test_results.end_time = tr.end_time").count
=> 68
```

dLobatog:
@akofink We can mark it as part of the DB review milestone and merge all of them together as you mentioned

akofink:
Right, but this is already included in one of the PRs: https://github.com/RedHatInsights/compliance-backend/pull/406/files#diff-277bf976622928a3a96e1377f1359bb6R5

dLobatog:
Let's mark this one as dependent on https://github.com/RedHatInsights/compliance-backend/pull/406 , there's no rush to get this index to prod I think

dLobatog (commented):
Anything missing here @akofink ?

akofink:
This is already on master via #406: https://github.com/RedHatInsights/compliance-backend/blob/master/db/migrate/20200417180344_add_unique_index_to_test_results.rb

We can just close this one.

PR #437, opened 2020-04-28
Revert "Add unique profile DB constraint (#370)"

akofink:
This reverts commit d045ad761be8d304f9adc08ea066ec548e3e2c02.

The change causes this issue when trying to delete a profile that is still referenced by a child profile:

```
PG::ForeignKeyViolation: ERROR:  update or delete on table "profiles" violates foreign key constraint "fk_rails_ba3d1c4f86" on table "profiles"                                               
DETAIL:  Key (id)=(61d96c79-0a07-4414-bfe1-3e1ddb0cf428) is still referenced from table "profiles".
```

We didn't catch it because the error logs in sidekiq and doesn't fail the migration.


akofink:
Closing in favor of https://github.com/RedHatInsights/compliance-backend/pull/438

PR #438, opened 2020-04-28
Only delete policy_profiles for non-canonical profiles

akofink:
- Inline the sidekiq jobs for this migration
- Only delete profile_policies for non-canonical (account_id != nil) profiles

```
$ d-c exec rails bundle exec rake db:migrate                                                                         
2020-04-28T15:40:22.087Z 101 TID-go3cbv1r5 INFO: GitLab reliable fetch activated!                                                                                                             
**************************************************                                                                                                                                            
⛔️ WARNING: Sidekiq testing API enabled, but this is not the test environment.  Your jobs will not go to Redis.                                                                               
**************************************************                                                                                                                                            
== 20200325185540 AddUniqueIndexToProfiles: migrating =========================                                                                                                               
-- add_index(:profiles, [:ref_id, :account_id, :benchmark_id], {:unique=>true})                                                                                                               
   -> 0.0095s                                                                                                                                                                                 
== 20200325185540 AddUniqueIndexToProfiles: migrated (0.1736s) ================
```

### DIFF
--- a/db/migrate/20200325185540_add_unique_index_to_profiles.rb
+++ b/db/migrate/20200325185540_add_unique_index_to_profiles.rb
@@ -1,4 +1,12 @@
+require 'profile'
+
 class AddUniqueIndexToProfiles < ActiveRecord::Migration[5.2]
+  class ::Profile < ApplicationRecord
+    def destroy_policy_test_results
+      DestroyProfilesJob.new.perform(policy_profiles.pluck(:id))
+    end
+  end
+
   def up
     DuplicateProfileResolver.run!
 
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -16,6 +16,8 @@ module ProfilePolicyAssociation
     end
 
     def policy_profiles
+      return Profile.none if account_id.nil?
+
       Profile.includes(:benchmark)
              .where(account: account_id, ref_id: ref_id,
                     benchmarks: { ref_id: benchmark.ref_id })
--- a/db/migrate/20200106134953_add_unique_index_to_benchmarks.rb
+++ b/db/migrate/20200106134953_add_unique_index_to_benchmarks.rb
@@ -1,4 +1,12 @@
+require 'profile'
+
 class AddUniqueIndexToBenchmarks < ActiveRecord::Migration[5.2]
+  class ::Profile < ApplicationRecord
+    def destroy_policy_test_results
+      DestroyProfilesJob.new.perform(policy_profiles.pluck(:id))
+    end
+  end
+
   def up
     DuplicateBenchmarkResolver.run!
 


db/migrate/20200325185540_add_unique_index_to_profiles.rb
```diff
@@ -1,5 +1,9 @@
+require 'sidekiq/testing'
+
 class AddUniqueIndexToProfiles < ActiveRecord::Migration[5.2]
   def up
+    Sidekiq::Testing.inline!
```
akofink, line 5: This causes sidekiq jobs to run synchronously, which is probably desired for all migrations. If they're run in redis and something fails, the migration will still pass, and the DB transaction will not get rolled back.

dLobatog: How about calling `.perform` directly? It's what `:inline` does, without needing to require the testing files

dLobatog (changes requested):
Looks good to me, one minor comment inline, thanks @akofink

db/migrate/20200325185540_add_unique_index_to_profiles.rb
akofink: How do you propose doing that? We want this to be async most of the time, just not during this migration.

dLobatog: The `destroy_policy_test_result` method can be overridden in the migration for the Profile class to call `perform`, unless there's some other job called I'm not aware of

dLobatog (changes requested):

akofink:
@dLobatog I've been trying to do what you suggest, overriding the method in profile, but it doesn't work. The profile model is still loaded, and the async job is run. Am I missing something here?

```rb
require 'profile'

class AddUniqueIndexToProfiles < ActiveRecord::Migration[5.2]
  class Profile < ApplicationRecord
    def destroy_policy_test_results
      DestroyProfilesJob.perform(policy_profiles.pluck(:id))
    end
  end

  def up
    DuplicateProfileResolver.run!

    add_index(:profiles, %i[ref_id account_id benchmark_id], unique: true)
  end

  def down
    remove_index(:profiles, %i[ref_id account_id benchmark_id])
  end
end
```

dLobatog:
@akofink You can prepend the overridden module, I've tested this and works:

```
require 'profile'

class AddUniqueIndexToProfiles < ActiveRecord::Migration[5.2]
  module PolicyTestResults
    def destroy_policy_test_results
      DestroyProfilesJob.new.perform(policy_profiles.pluck(:id))
    end
  end

  class ::Profile < ApplicationRecord
    prepend PolicyTestResults
  end
```

dLobatog:
@akofink actually your code is also valid, but you need to define `class ::Profile` - otherwise it's going to think it's `AddUniqueIndexToProfiles::Profile`

akofink:
@dLobatog How did you test? I'm not able to get it to work either with the prepended module or directly defining the Profile class.

I've been testing by putting a break point in the `profile_policy_association.rb` `destroy_policy_test_results` method - if the break point is hit, it's not working correctly. I've pushed what I have, maybe I'm still missing something :shrug:

akofink:
It also appears like this change breaks a couple tests related to profile deletion.

akofink:
Ok! I realized I was triggering the sidekiq job via the unique benchmarks index migration, not the unique profiles migration (both destroy profiles). I've added this `perform` change to both. Thanks for the help, @dLobatog!

akofink:
I'm still not convinced we can override the Profile class just for these migrations. The modification leaks out to the rest of the app, thus the failing tests. Those failing tests are expecting 1 profile deleted immediately and 1 job queued by Sidekiq, but instead they see 2 deleted profiles immediately and fail.

By using `Sidekiq.inline!` in the `up` method of the migration, it doesn't leak to the rest of the app and is only run during migrations. Another option would be a global `async` setting or similar where we can run migrations with `async: false` and normal operation would be async.

akofink:
Here's this PR with a PoC for the async setting: https://github.com/RedHatInsights/compliance-backend/pull/445

akofink:
https://github.com/RedHatInsights/compliance-backend/pull/445 was merged instead of this one

PR #439, opened 2020-04-29
Remove millions of duplicates from RuleReferencesRule

dLobatog:
This commit removes around 63M rows from RuleReferencesRule. Even though
there is a soft-validation in ActiveRecord on rule_id/reference_id, it's
completely skipped at the only place where we insert these records (in
the parser).

This greatly speeds up the SystemDetails page query, as we need to load
all RuleReferences for all rules there. The query, a simple SELECT *
from rule_references_rules WHERE rule_id IN (list of 1600 rules) goes
from 8 seconds to miliseconds. It also allows us to remove the
"DISTINCT" from the query, which was making us lose nearly 3s
sorting the records first.

As for the code itself:

1. I *must* add a primary key in order to identify the duplicate keys.
   Otherwise, you can identify them, but you cannot remove them (what
   are you even removing in that case)? It takes a while to add a
   primary key to 63M records but it's bearable (~5m)

(255122.1ms)  SELECT MIN(id) FROM "rule_references_rules" GROUP BY
rule_id, rule_reference_id
=> 456040
[4] pry(main)> ids = RuleReferencesRule.count
(1503.3ms)  SELECT COUNT(*) FROM "rule_references_rules"
=> 63694827

2. Removing the records directly, whether in batches or directly, is
   simply too slow. It would take around 160 hours to do it. Even after
   removing all indices from the table (which slow down the DELETE),
   it's too much.
   Each batch of 1000 records takes ~8s to delete, so
   (63238787/1000)*8/60/60 -> 140h!!!

   Instead, a usual method to solve this issue is to:
     1. Find all rows you want to keep.
     2. Build a new_rule_references_rules table
     3. Insert all of the records found on 1 in 2.
     4. Drop the original rule_references_rules table
     5. Rename the new_rule_references_rules
     6. Add an uniqueness index so this doesn't happen again.

   This takes around 21 minutes to run.
   == 20200428205535 UniquenessRuleReferencesRuleIndex: migrated
   (1300.7111s) ====

Before:
Screenshot from 2020-04-28 16-44-52
Screenshot from 2020-04-28 23-14-40
Screenshot from 2020-04-28 23-15-10
After:
Screenshot from 2020-04-29 15-28-41
Screenshot from 2020-04-29 15-39-40

:warning: If these rules were added through the mass INSERT ignoring soft-validations, it's likely now that will raise an error, so we might have to adapt the parser to that.

:heavy_check_mark: This *greatly* improves finding references, so the query to find RuleReferences in the system details view should be way faster.

### DIFF
--- a/db/migrate/20200428143250_add_primary_key_to_rules_references.rb
+++ b/db/migrate/20200428143250_add_primary_key_to_rules_references.rb
@@ -1,0 +1,5 @@
+class AddPrimaryKeyToRulesReferences < ActiveRecord::Migration[5.2]
+  def change
+    add_column :rule_references_rules, :id, :primary_key
+  end
+end
--- a/db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
+++ b/db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
@@ -1,7 +1,7 @@
 class UniquenessRuleReferencesRuleIndex < ActiveRecord::Migration[5.2]
   class NewRuleReferencesRule < ApplicationRecord; end
 
-  def change
+  def up
     ids_to_keep = RuleReferencesRule.group(
       'rule_id, rule_reference_id'
     ).having('count(id) > 1').select(Arel.sql('MIN(id)'))
--- a/db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
+++ b/db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
@@ -1,0 +1,28 @@
+class UniquenessRuleReferencesRuleIndex < ActiveRecord::Migration[5.2]
+  class NewRuleReferencesRule < ApplicationRecord; end
+
+  def change
+    ids_to_keep = RuleReferencesRule.group(
+      'rule_id, rule_reference_id'
+    ).having("count(*) > 1").pluck(Arel.sql('MIN(id)'))
+    create_table :new_rule_references_rules, id: false do |t|
+      t.references :rule, type: :uuid, index: true, null: false
+      t.references :rule_reference, type: :uuid, index: true, null: false
+    end
+    columns = [:rule_id, :rule_reference_id]
+    batch_size = 1000
+    to_keep = RuleReferencesRule.where(id: ids_to_keep)
+    batches_to_run = to_keep.count/batch_size
+    to_keep.in_batches(of: batch_size).each_with_index do |batch, index|
+      puts "Inserting RRR batch, #{index} out of #{batches_to_run}"
+      NewRuleReferencesRule.import(
+        columns,
+        batch.pluck(:rule_id, :rule_reference_id)
+      )
+    end
+    drop_table :rule_references_rules
+    rename_table :new_rule_references_rules, :rule_references_rules
+    add_index :rule_references_rules, %i[rule_id rule_reference_id],
+      unique: true
+  end
+end
--- a/db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
+++ b/db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
@@ -4,18 +4,18 @@ class UniquenessRuleReferencesRuleIndex < ActiveRecord::Migration[5.2]
   def change
     ids_to_keep = RuleReferencesRule.group(
       'rule_id, rule_reference_id'
-    ).having("count(*) > 1").pluck(Arel.sql('MIN(id)'))
+    ).having('count(id) > 1').select(Arel.sql('MIN(id)'))
     create_table :new_rule_references_rules, id: false do |t|
       t.references :rule, type: :uuid, index: true, null: false
       t.references :rule_reference, type: :uuid, index: true, null: false
     end
     columns = [:rule_id, :rule_reference_id]
-    batch_size = 1000
+    batch_size = 10000
     to_keep = RuleReferencesRule.where(id: ids_to_keep)
     batches_to_run = to_keep.count/batch_size
     to_keep.in_batches(of: batch_size).each_with_index do |batch, index|
       puts "Inserting RRR batch, #{index} out of #{batches_to_run}"
-      NewRuleReferencesRule.import(
+      NewRuleReferencesRule.import!(
         columns,
         batch.pluck(:rule_id, :rule_reference_id)
       )


db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
```diff
@@ -0,0 +1,28 @@
+class UniquenessRuleReferencesRuleIndex < ActiveRecord::Migration[5.2]
+  class NewRuleReferencesRule < ApplicationRecord; end
+
+  def change
```
akofink, line 4: We should add other methods here besides `change` for readability. Also, is this _really_ reversible, or should we use `up`/`down`?

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
```diff
@@ -0,0 +1,28 @@
...
+    ids_to_keep = RuleReferencesRule.group(
```
akofink, line 5:
```suggestion
    ids_to_keep = RuleReferencesRule.select(:rule_id, :rule_reference_id).group(
```

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
```diff
@@ -0,0 +1,28 @@
...
+      'rule_id, rule_reference_id'
+    ).having("count(*) > 1").pluck(Arel.sql('MIN(id)'))
```
akofink, line 7:
```suggestion
    ).having("count(id) > 1").pluck(Arel.sql('MIN(id)'))
```

akofink, line 7: This is a nice way to find unique records :+1: If you use `select` instead of `pluck`, we will never load the ID records into Ruby memory.

```suggestion
    ).having("count(*) > 1").select(Arel.sql('MIN(id)'))
```

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
```diff
@@ -0,0 +1,28 @@
...
+    create_table :new_rule_references_rules, id: false do |t|
+      t.references :rule, type: :uuid, index: true, null: false
+      t.references :rule_reference, type: :uuid, index: true, null: false
+    end
+    columns = [:rule_id, :rule_reference_id]
+    batch_size = 1000
+    to_keep = RuleReferencesRule.where(id: ids_to_keep)
+    batches_to_run = to_keep.count/batch_size
+    to_keep.in_batches(of: batch_size).each_with_index do |batch, index|
```
akofink, line 16: `find_each` by default has a batch size of 1000. You can use `find_each.with_index`

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
```diff
@@ -0,0 +1,28 @@
...
+      puts "Inserting RRR batch, #{index} out of #{batches_to_run}"
```
akofink, line 17: I'd be curious how this `puts` affects the runtime

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
```diff
@@ -0,0 +1,28 @@
...
+      NewRuleReferencesRule.import(
```
akofink, line 18: This will raise an error, right? AR::Import is confusing as it's backwards from AR where `!` means raise errors iirc.

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
```diff
@@ -0,0 +1,28 @@
...
+        columns,
+        batch.pluck(:rule_id, :rule_reference_id)
```
akofink, line 20: I'd be curious about select vs. pluck here as well.

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
dLobatog: This one doesn't make sense I think - I only want the IDs here

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
dLobatog: AR relations can't be passed to the `.insert` method, it throws an InvalidArgumentsError.

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
akofink: No, the duplicate resolver should be idempotent, so running it a second time should be a no-op. By putting this in a `change` method instead of just `up`, you're telling Rails that this is reversible, but it's not - thus my comment.

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
akofink: Ok, then `select(:id)` - the idea is that limiting your select instead of the default `select('*')` makes the response quicker.

db/migrate/20200428205535_uniqueness_rule_references_rule_index.rb
akofink: Geez, maybe it'd make more sense to write an `INSERT` query yourself with an AR subquery using `to_sql`

akofink: I really doubt trying to do this like I suggested would make any difference. We'd still need to `pluck` to get all the records to insert.

db/migrate/20200428143250_add_primary_key_to_rules_references.rb
```diff
@@ -0,0 +1,5 @@
+class AddPrimaryKeyToRulesReferences < ActiveRecord::Migration[5.2]
+  def change
+    add_column :rule_references_rules, :id, :primary_key
```
akofink, line 3: @dLobatog what does this do?

dLobatog: This allows us to delete the duplicates. Without it, you can find the duplicates, but you can't delete them - think about:
```
DELETE FROM "rule_references_rules" WHERE rule_id = foo AND rule_reference_id = bar 
```
or any statement you could construct would delete *all* of the records for a rule_id/rule_reference_id pair. Without a primary key, you can delete either by rule_id (all of the records for a same rule_id), rule_reference_id (same) or both (same).

akofink: Ok, that makes sense. Thanks for the explanation! :+1:

PR #440, opened 2020-04-29
Add account_number to Openshift info

dLobatog:
Currently, the notifications we are getting contain the User object,
which is useless. Instead, we should get the account number to
help us understand better under which circumstances the exception
happened.


Victoremepunto:
@dLobatog  can we please add the `NAMESPACE` env variable to the service's env and hence to the notification ? just to get the environment where the alarm was triggered. would that make sense ? at least is the quick solution I thought to get the source ENV

PR #441, opened 2020-04-29
RBAC failure results in 403, not 401

akofink:
Towards https://projects.engineering.redhat.com/browse/RHICOMPL-640

UI changes are also needed.


test/controllers/concerns/authentication_test.rb
```diff
@@ -57,7 +57,7 @@ class AuthenticationTest < ActionDispatch::IntegrationTest
     end
 
     should 'missing RBAC access for compliance' do
-      RbacApi.any_instance.expects(:check_user).returns(false)
+      RbacApi.any_instance.expects(:check_user)
```
akofink, line 60: `check_user` returns `nil` in this case, not `false`

PR #442, opened 2020-04-30
RHICOMPL-641 - Keep external profiles on associateProfiles

dLobatog:
Currently if you associate profiles to a host, it removes the associations to external profiles. Those associations shouldn't be removed.

### DIFF
--- a/app/graphql/mutations/system/associate_profiles.rb
+++ b/app/graphql/mutations/system/associate_profiles.rb
@@ -12,8 +12,9 @@ module Mutations
 
       def resolve(args = {})
         host = find_hosts([args[:id]]).first
-        profiles = find_profiles(args[:profile_ids])
-        host.update(profiles: profiles)
+        external_profiles = host.profiles.where(external: true)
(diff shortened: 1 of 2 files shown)


app/graphql/mutations/system/associate_profiles.rb
```diff
@@ -12,8 +12,9 @@ class AssociateProfiles < BaseMutation
 
       def resolve(args = {})
         host = find_hosts([args[:id]]).first
-        profiles = find_profiles(args[:profile_ids])
-        host.update(profiles: profiles)
+        external_profiles = host.profiles.where(external: true)
+        internal_profiles = find_profiles(args[:profile_ids])
+        host.update(profiles: internal_profiles + external_profiles)
```
dLobatog, line 17: I'd agree using `internal_profiles.or(external_profiles)` here is a better solution, but:
`ArgumentError: Relation passed to #or must be structurally compatible. Incompatible values: [:joins]`

PR #445, opened 2020-04-30
Only delete policy_profiles for non-canonical profiles

akofink:
An alternative to #438

### DIFF
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -16,6 +16,8 @@ module ProfilePolicyAssociation
     end
 
     def policy_profiles
+      return Profile.none if account_id.nil?
+
       Profile.includes(:benchmark)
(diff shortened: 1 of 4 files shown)


app/models/concerns/profile_policy_association.rb
```diff
@@ -34,7 +36,11 @@ def compliance_threshold
     end
 
     def destroy_policy_test_results
-      DestroyProfilesJob.perform_async(policy_profiles.pluck(:id))
+      if Settings.async
+        DestroyProfilesJob.perform_async(policy_profiles.pluck(:id))
+      else
+        DestroyProfilesJob.new.perform(policy_profiles.pluck(:id))
+      end
```
akofink, line 43: We can refactor this later, ideally to move it to ApplicationJob or something.

akofink:
Thanks, @dLobatog!

PR #447, opened 2020-05-04
RHICOMPL-645 - Prometheus collector for business metrics

dLobatog:
:warning: Has a co-dependency on https://github.com/RedHatInsights/s2i-insights-compliance/pull/17 

Adding a collector for prometheus_exporter. This runs every time
/metrics is called, which is every 60s, from the prometheus_exporter
deployment. Most other metrics are calculated from the data send to
/send-metrics by the other pods, but this sort of data isn't really
related to any pods runtime.

Timeout has been greatly increased just to prevent any failures,
although benchmarks show the `collect` method is really fast.

```ruby
[3] pry(#<BusinessCollector>)> Benchmark.measure { collect }
=> #<Benchmark::Tms:0x00007f22216bb288
        @cstime=0.0,
	@cutime=0.0,
	@label="",
	@real=0.030418214002565946,
	@stime=0.002605999999999997,
	@total=0.021810000000000218,
	@utime=0.01920400000000022>
```

In addition to that, I added binding to 0.0.0.0 as the latest prometheus
binds to 127.0.0.1, which makes the data only visible through the pod.
The docker-compose also now contains the necessary variables to access
the Rails database. --verbose helps us see which collectors are loaded
when prometheus_exporter boots.


lib/prometheus/business_collector.rb
```diff
@@ -0,0 +1,82 @@
...
+  require File.expand_path('../../config/environment', __dir__)
+end
+
+# rubocop:disable Metrics/MethodLength
+# rubocop:disable Metrics/AbcSize
+# Collects stats relevent for the business from the database
+# when /metrics is called. This runs directly on prometheus_exporter.
+class BusinessCollector < PrometheusExporter::Server::TypeCollector
+  def initialize
+    @total_accounts = PrometheusExporter::Metric::Gauge.new(
```
dLobatog, line 13: These are all `Gauge` vs `Counter` as theoretically they can all go up and down. We don't remove unused accounts, but we could.

lib/prometheus/business_collector.rb
```diff
@@ -0,0 +1,82 @@
...
+    )
+    @client_accounts.observe client_accounts.count
+    @client_accounts_with_hosts.observe(
+      Host.where(
+        account_id: client_accounts.pluck(:id)
+      ).select(:account_id).distinct.count
+    )
+    @total_policies.observe Profile.count
+    @client_policies.observe(
+      Profile.where(account_id: client_accounts.pluck(:id)).count
```
akofink, line 60:
```suggestion
      Profile.where(account: client_accounts).count
```

lib/prometheus/business_collector.rb
```diff
@@ -0,0 +1,82 @@
...
+    )
+    @total_systems.observe Host.count
+    @client_systems.observe(
+      Host.where(account_id: client_accounts.pluck(:id)).count
```
akofink, line 64:
```suggestion
      Host.where(account: client_accounts).count
```

lib/prometheus/business_collector.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def type
+    'business'
+  end
+
+  def collect
+    @total_accounts.observe Account.count
+    client_accounts = Account.distinct.where(
+      id: User.select(:account_id).where.not('email LIKE ?', '%redhat%')
```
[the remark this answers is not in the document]
dLobatog: That was removed to allow for cert-auth from the API, perhaps we can go back to saving them. But overall I think the problem is that we create accounts without reading `internal`  from the identity header, just using the account_number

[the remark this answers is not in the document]
akofink: We can certainly discuss user creation, though it's not a blocker for this PR, of course. There is no reason for us to create users that I know of. We definitely have to address RBAC for cert-based auth. Check out these if you haven't seen them already:

https://projects.engineering.redhat.com/browse/RHICOMPL-643
https://projects.engineering.redhat.com/browse/RHICOMPL-644

Using the `internal` part of the header seems like something we :100: should implement for this purpose of filtering out RH accounts.

PR #450, opened 2020-05-04
Bump codecov to fix coverage issues

akofink:
We have some random files (i.e. app/services/safe_downloader.rb) not seen as covered in our tests by codecov even though they are. I'm hoping updating the gem will fix the issue for us.


akofink:
It didn't help :( Still at 87.02% and not counting app/services/safe_downloader.rb

PR #456, opened 2020-05-05
Find RuleReferencesRule if they're present in the db

dLobatog:
If we don't find them first, we will be trying to introduce duplicates,
which with the current uniqueness index it will cause a failure.

### DIFF
--- a/app/services/concerns/xccdf/rule_references_rules.rb
+++ b/app/services/concerns/xccdf/rule_references_rules.rb
@@ -9,10 +9,9 @@ module Xccdf
       def save_rule_references_rules
         @rule_references_rules ||= @op_rules.flat_map do |op_rule|
           rule_references_for(op_rule: op_rule).map do |reference|
-            ::RuleReferencesRule.find_or_initialize_by(
-              rule_reference_id: reference.id,
-              rule_id: rule_for(ref_id: op_rule.id).id
-            )
+            ::RuleReferencesRule
+              .find_or_initialize_by(rule_reference_id: reference.id,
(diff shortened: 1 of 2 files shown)


app/services/concerns/xccdf/rule_references_rules.rb
```diff
@@ -9,7 +9,8 @@ module RuleReferencesRules
       def save_rule_references_rules
         @rule_references_rules ||= @op_rules.flat_map do |op_rule|
           rule_references_for(op_rule: op_rule).map do |reference|
-            ::RuleReferencesRule.new(rule_reference_id: reference.id,
+            ::RuleReferencesRule
+              .find_or_initialize_by(rule_reference_id: reference.id,
                                      rule_id: rule_for(ref_id: op_rule.id).id)
```
akofink, line 14: This has the potential to really slow this method down, since it now executes N queries. If it becomes problematic, maybe we can introduce an `existing_rule_references_rule` method or something that uses `where` to find them all in one query.

test/services/concerns/xccdf/rule_references_rules_test.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'xccdf/rule_references'
+
+# A class to test saving RuleReferencesRules from OpenscapParser
+class RuleReferencesRulesTest < ActiveSupport::TestCase
```
akofink, line 7: :1st_place_medal:

akofink:
Now, where's the codecov report?? :thinking:

PR #457, opened 2020-05-05
[stable] Adjust rule result pass/fail determination (#451)

akofink:


akofink:
I have a feeling this may need smoke test, or at least some part of IQE, adjustments.

akofink:
I'll need to investigate what's going on with the smoke tests.

PR #462, opened 2020-05-06
Update to Puma 4.3.3

dLobatog:
According to https://github.com/puma/puma/issues/1828 , Puma 4 has no breaking changes. It updated it's major version to 4.0 because it changed it's backend to use nio4r.

https://www.schneems.com/2019/06/26/puma-4-new-io-4-your-server/ for a more readable summary on major changes.


Gemfile
```diff
@@ -7,7 +7,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 gem 'rails', '~> 5.2.1'
 gem 'pg'
 # Use Puma as the app server
-gem 'puma', '~> 3.12'
+gem 'puma', '~> 4'
```
akofink, line 10: I'm pretty sure you want this:

```suggestion
gem 'puma', '~> 4.0'
```

dLobatog: [They're equivalent](https://guides.rubygems.org/patterns/#pessimistic_version_constraint):
  > It will only use the latest version from the 2.x series – so 2.3.0 – and not 3.0.0. 
unless I'm missing something else

akofink: Ah, you're right - sorry :) I just tested it with `gem install puma -v '~> 3'`, and it installed 3.12.4

PR #463, opened 2020-05-06
RHICOMPL-622 - Delete hosts when they become orphaned

dLobatog:
I still need to test this actually on the UI, but in theory this is all that's needed not to leave orphaned hosts around in our database when they no longer have any profile assigned.

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -14,7 +14,7 @@ class Profile < ApplicationRecord
 
   has_many :profile_rules, dependent: :delete_all
   has_many :rules, through: :profile_rules, source: :rule
-  has_many :profile_hosts, dependent: :delete_all
+  has_many :profile_hosts, dependent: :destroy
   has_many :hosts, through: :profile_hosts, source: :host
   has_many :test_results, dependent: :destroy
   belongs_to :account, optional: true
--- a/app/models/profile_host.rb
+++ b/app/models/profile_host.rb
@@ -12,6 +12,12 @@ class ProfileHost < ApplicationRecord
   after_destroy :destroy_orphaned_host
 
   def destroy_orphaned_host
-    DeleteHost.perform_async(id: host.id) if host.profiles.empty?
+    return unless host.profiles.empty?
+
+    if Settings.async
+      DeleteHost.perform_async(id: host.id)
+    else
+      DeleteHost.new.perform(id: host.id)
+    end
   end
 end
--- a/app/models/profile_host.rb
+++ b/app/models/profile_host.rb
@@ -8,4 +8,10 @@ class ProfileHost < ApplicationRecord
 
   validates :profile, presence: true
   validates :host, presence: true, uniqueness: { scope: :profile }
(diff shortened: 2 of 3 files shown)


app/models/profile.rb
```diff
@@ -14,7 +14,7 @@ class Profile < ApplicationRecord
 
   has_many :profile_rules, dependent: :delete_all
   has_many :rules, through: :profile_rules, source: :rule
-  has_many :profile_hosts, dependent: :delete_all
+  has_many :profile_hosts, dependent: :destroy
```
akofink, line 17: I'm pretty sure this has performance implications. We originally had destroy here and changed it to delete_all. I can't quite remember why though.

test/models/profile_host_test.rb
```diff
@@ -1,11 +1,27 @@
...
     @host = hosts(:one)
     @profile_host = ProfileHost.new(profile: @profile, host: @host)
   end
+
+  test 'destroys associated hosts if host has no more policies assigned' do
+    profile = Profile.create(name: 'foo', benchmark: benchmarks(:one),
+                          ref_id: 'foo')
+    host = Host.create(profiles: [profile], name: 'bar',
+                       account: accounts(:one))
+    assert_equal 0, DeleteHost.jobs.size
```
akofink, line 18: You should put `DeleteHost.clear` in the setup method to ensure no other tests have queued jobs on this test object.

test/models/profile_host_test.rb
```diff
@@ -1,11 +1,27 @@
...
+    profile.destroy
+    assert_equal 1, DeleteHost.jobs.size
+    assert_difference('Host.count', -1) do
+      DeleteHost.drain
+    end
+    assert_empty Host.where(id: host.id)
+    assert_equal 0, DeleteHost.jobs.size
```
akofink, line 25: `Sidekiq.inline!` is an option if you haven't already considered it. TIL about `drain` - pretty neat. :)

app/models/profile_host.rb
```diff
@@ -8,4 +8,10 @@ class ProfileHost < ApplicationRecord
 
   validates :profile, presence: true
   validates :host, presence: true, uniqueness: { scope: :profile }
+
+  after_destroy :destroy_orphaned_host
+
+  def destroy_orphaned_host
+    DeleteHost.perform_async(id: host.id) if host.profiles.empty?
+  end
```
akofink, line 16: This is going to affect our migrations that delete profiles which haven't yet landed in QA or prod. We will want to make the migrations synchronous so that they can be rolled back if anything bad happens.

app/models/profile.rb
dLobatog: https://github.com/RedHatInsights/compliance-backend/pull/99 no apparent reason, this only runs when the profile or report are deleted

test/models/profile_host_test.rb
dLobatog: Updated

PR #465, opened 2020-05-08
Commit db/schema.rb

dLobatog:
Currently our pods don't have this, and we suspect it might be causing
issues as there's no authoritative source for the schema. It's being
looked at on the fly. Committing this provides a way for Rails to look
at how the schema should be from a file. Notice the top of the file even
recommends checking in the schema into the VCS


akofink:
@dLobatog this one's good to go once we remove it from the `.gitignore`

PR #471, opened 2020-05-14
[stable] 622, 650, Puma 4, dev changes, remove unused GQL fields

dLobatog:
This PR includes most of the latest changes to stable except for the uniqueness migrations. It's getting a bit difficult to keep track of these if every PR merged to stable creates a merge commit instead of rebasing the existing ones from master to stable. I've changed the config of the repo to allow linear history only at least in the stable branch. Still, please don't squash. 

Summary of changes:

**APIs**
* Profiles API returns parent_profile_ref_id 
* Remove unused GQL fields

**Bugfixes**
* RHICOMPL-622 - Delete hosts when they become orphaned
* RHICOMPL-650 - Delete cached latest_supported_rhel benchmarks
* RHICOMPL-644 - Allow cert-based auth to skip RBAC
* Migrate all profiles before summit release to "external" - https://github.com/RedHatInsights/compliance-backend/pull/401

**Dev related**
* Make profile mutation test deterministic
* Do not initialize Sidekiq or prometheus_exporter from prometheus itself
* Bump devel Dockerfile to Ruby 2.6
* docker-compose requires redis for rails now. 
* Ignore lib in codecov
* Allow pending migrations in RAILS_ENV=development

**Other**
* Update Puma to 4.3.3


app/models/profile_host.rb
```diff
@@ -8,4 +8,16 @@ class ProfileHost < ApplicationRecord
...
   validates :profile, presence: true
   validates :host, presence: true, uniqueness: { scope: :profile }
+
+  after_destroy :destroy_orphaned_host
+
+  def destroy_orphaned_host
+    return unless host.profiles.empty?
+
+    if Settings.async
+      DeleteHost.perform_async(id: host.id)
```
akofink, line 18: Where is the change for `config/settings.yml` to add this setting?

dLobatog: It was missing indeed and it was making the test for this fail, I just cp-ed a6150a2b8cc23f11c41357148c9b83e91ca196f9

akofink:
Alright! No smoke! :1st_place_medal:

akofink:
NB: This one was 'Rebased and Merged', as requested, so we can keep better track of what's on stable.

PR #472, opened 2020-05-18
RHICOMPL-695 - Fix 'undefined method id' error on SystemDetails

dLobatog:
We keep on getting an alert on "undefined method 'id' for nil:NilClass"
on the #alerts-cloudservices-apps.

After investigating, this alert happens because the system details page
is visited. We try to retrieve the latest Test Results for the profiles
for the System, and some RuleResults don't have a correct rule_id FK -
they're pointing to unknown rule_ids.

> irb(main):013:0> rule_ids.count
> => 4405
> irb(main):014:0> rule_ids = RuleResult.select(:rule_id).distinct.pluck(:rule_id); rule_ids.count
> => 4405
> irb(main):015:0> Rule.where(id: rule_ids).count
> => 3283

In our GQL API code, when we try to get the results for these
rules, they're showing up as "nil" and therefore they fail to
load.

### DIFF
--- a/app/graphql/types/interfaces/rules_preload.rb
+++ b/app/graphql/types/interfaces/rules_preload.rb
@@ -12,8 +12,8 @@ module RulesPreload
     latest_test_result_batch(args).then do |latest_test_result|
       latest_rule_results_batch(latest_test_result).then do |rule_results|
         rules_for_rule_results_batch(rule_results).then do |rules|
-          initialize_rules_context(rules, rule_results, args)
-          rules
+          initialize_rules_context(rules.compact, rule_results, args)
+          rules.compact
         end
       end
     end
--- a/test/graphql/queries/system_query_test.rb
+++ b/test/graphql/queries/system_query_test.rb
@@ -223,6 +223,53 @@ class SystemQueryTest < ActiveSupport::TestCase
     assert graphql_host.profiles.pluck(:id).include?(profiles(:one).id)
   end
 
+  test 'query system rules when results contain wrong rule_ids' do
+    query = <<-GRAPHQL
+    query System($systemId: String!){
+        system(id: $systemId) {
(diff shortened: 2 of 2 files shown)


app/graphql/types/interfaces/rules_preload.rb
```diff
@@ -12,8 +12,8 @@ def rules(args = {})
     latest_test_result_batch(args).then do |latest_test_result|
       latest_rule_results_batch(latest_test_result).then do |rule_results|
         rules_for_rule_results_batch(rule_results).then do |rules|
-          initialize_rules_context(rules, rule_results, args)
-          rules
+          initialize_rules_context(rules.compact, rule_results, args)
+          rules.compact
```
dLobatog, line 16: We never got this error as we never got this far in the code when `rules` contains a nil value, but if you do, you still want to get rid of it or you get `{"message"=>"Cannot return null for non-nullable field Profile.rules"}]` in the response

app/graphql/types/interfaces/rules_preload.rb
```diff
@@ -12,8 +12,8 @@ def rules(args = {})
     latest_test_result_batch(args).then do |latest_test_result|
       latest_rule_results_batch(latest_test_result).then do |rule_results|
         rules_for_rule_results_batch(rule_results).then do |rules|
-          initialize_rules_context(rules, rule_results, args)
-          rules
+          initialize_rules_context(rules.compact, rule_results, args)
```
dLobatog, line 15: This is the culprit of the alerts we get - the method called does: `      context[:parent_profile_id][rule.id] = object.id` for a nil `rule`

dLobatog:
@akofink Indeed - I'm trying to understand now how did this happen based on the already existing mismatches first

PR #476, opened 2020-05-19
Skip RBAC for certain endpoints with cert based auth

akofink:
The two endpoints so far we need to allow via cert based auth are
profiles#index and profiles#tailoring_file. Cert based auth may contain
system information but does not contain user informatoin, so this should
be a good check to use to detect cert vs. user auth.

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -6,6 +6,11 @@ require 'rbac_api'
 module Authentication
   extend ActiveSupport::Concern
 
+  ALLOWED_CERT_BASED_RBAC_ACTIONS = [
+    { controller: 'profiles', action: 'index' },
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/authentication.rb
```diff
@@ -47,13 +52,24 @@ def user
...
   private
 
+  def skip_rbac_for_cert_based_auth?
+    cert_based_auth? &&
+      ALLOWED_CERT_BASED_RBAC_ACTIONS.include?(controller: controller_name,
+                                               action: action_name)
+  end
+
+  def cert_based_auth?
+    identity_header.identity.dig('user').nil?
```
dLobatog, line 70: Seems rather weak, but since the identity_header is something we can only get from 3scale, I guess it's fine

PR #480, opened 2020-05-21
Get test coverage up to 90%

dLobatog:
A few tests to get back to 90% and get rid of the :x:


test/jobs/parse_report_job_test.rb
```diff
@@ -13,33 +13,37 @@ class ParseReportJobTest < ActiveSupport::TestCase
   test 'payload tracker is notified about successful processing' do
     XccdfReportParser.stubs(:new).returns(@parser)
     @parser.stubs(:save_all)
-    @parse_report_job.stubs(:cancelled?)
+    Sidekiq.stubs(:redis).returns(false)
```
akofink, line 16: Why not `require 'sidekiq/testing'`?

app/services/xccdf_report_migration.rb
```diff
@@ -1,141 +0,0 @@
...
-
-# Migrates ref IDs from reports to comply with the expected standard.
-# Profile ref IDs should look like "xccdf_org.ssgproject.content_profile_*"
-# Rule ref IDs should look like "xccdf_org.ssgproject.content_rule_*"
-#
-# Disabling ClassLength here as the migration is likely going to be removed
-# at some point since we don't accept these reports anymore and old data was
-# already migrated
-# rubocop:disable Metrics/ClassLength
-class XCCDFReportMigration
```
akofink, line 11: I guess we'll always have it in our git history if it's ever needed again (not that it should be needed ever again).

config/environments/test.rb
```diff
@@ -13,7 +13,11 @@
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
   # preloads Rails for running tests, you may have to set it to true.
-  config.eager_load = false
+  if ENV['CODECOV_TOKEN']
+    config.eager_load = true
+  else
+    config.eager_load = false
+  end
```
akofink, line 20: I had tried this in the past as well, but it doesn't seem to help :(

akofink: If you add a `puts` statement like the SimpleCov docs say in their Readme under troubleshooting (https://github.com/colszowka/simplecov#troubleshooting), the files are loaded _after_ simplecov is loaded, which is correct. :shrug: :sob:

PR #483, opened 2020-05-25
API v1 namespacing

dLobatog:
This commit does NOT add new functionality. It merely makes a new path
/v1 for the v1 version of the API. The controllers fully inherit the
ones we have by default.

In the future, we should make a v2/ set of controllers, which can
coexist with the v1 version until it's stable enough to deprecate it.


app/controllers/v1/profiles_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# API for Profiles
+module V1
+  class ProfilesController < ::ProfilesController
+  end
```
akofink, line 6: This should be superclassed by a `V1::ApplicationController`, and we should copy everything from existing controllers into these V1 controllers and delete the old controllers. I guess there may still be some common logic between the versions, but any common logic that is slightly different for one version over the other will be difficult to maintain without breaking one of the APIs.

bastilian:
Should we for now have `/api` default to `V1`?

config/routes.rb
```diff
@@ -3,15 +3,31 @@
...
         member do
           get 'tailoring_file'
         end
       end
-      resources :rule_results, only: [:index]
-      resources :systems, only: [:index, :destroy]
-      resources :rules, only: [:index, :show]
+      resources :rule_results, controller: 'v1/rule_results', only: [:index]
+      resources :systems, controller: 'v1/systems', only: [:index, :destroy]
+      resources :rules, controller: 'v1/rules', only: [:index, :show]
```
dLobatog, line 30: Right now these are the only things making the `/api/compliance/resource` endpoints default to v1. We could make them default to v2 at any point in time by changing these routes.

app/controllers/v1/application_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Superclass for all V1 controllers
+module V1
+  class ApplicationController < ::ApplicationController
```
dLobatog, line 5: So far this can share all logic with the application controller we have for GraphQL

akofink: This is acceptable IMO :+1:

app/controllers/v1/profiles_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    def profile
+      @profile ||= Pundit.policy_scope(current_user, resource).find(params[:id])
+    end
+
+    def resource
+      Profile
+    end
+
+    def serializer
+      ProfileSerializer
```
akofink, line 39: I'm surprised this doesn't cause an issue. It also reminds me that the serializers should also be namespaced at some point (we can do it in another PR if you want).

PR #484, opened 2020-05-26
Change development image to RHEL 8 UBI ruby26

dLobatog:
In order to run a dev environment that's much more similar to the one in
production, it's probably a good idea to switch to the base image we are
using already in production.

Since the image is meant to be built using s2i, we can call manually the
assemble script in the Dockerfile.

Additionally, this image is tried and tested for podman if we want to
move to it at some point.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,17 @@
-FROM ruby:2.6
+FROM registry.access.redhat.com/ubi8/ruby-26
 
-WORKDIR /app
+USER root
 
-RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
-      gstreamer1.0-plugins-base gstreamer1.0-tools gstreamer1.0-x libopenscap-dev \
-      postgresql-client
+COPY . /tmp/src
 
-COPY vendor/ ./vendor
-COPY Gemfile* ./
-COPY entrypoint.sh ./
+RUN rm -rf /tmp/src/.git* && \
+    chown -R 1001 /tmp/src && \
+    chgrp -R 0 /tmp/src && \
+    chmod -R g+w /tmp/src
 
-RUN bundle -j4
+USER 1001
 
-ENTRYPOINT ["/app/entrypoint.sh"]
+ENV RAILS_ENV development
+ENV DISABLE_ASSET_COMPILATION true
+RUN /usr/libexec/s2i/assemble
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,16 +1,17 @@
-FROM ruby:2.6
+FROM registry.access.redhat.com/ubi8/ruby-26
 
-WORKDIR /app
+USER root
```
bastilian, line 3: Is this advised somewhere for UBI images?

dLobatog: No, the base image comes as USER 1001 - I only move to root to copy the source into `/tmp/src` , which is the location used by `s2i/assemble` - after that, I go back to 1001

Dockerfile
```diff
@@ -1,16 +1,17 @@
...
 
-RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
-      gstreamer1.0-plugins-base gstreamer1.0-tools gstreamer1.0-x libopenscap-dev \
-      postgresql-client
+COPY . /tmp/src
```
akofink, line 5: We should probably only copy what is needed. For example, I sometimes have SCAP scans or DB backups or other large files in my project root that don't need to be a part of the image.

Dockerfile
dLobatog:
```

mv: cannot move '/tmp/src/Gemfile' to './Gemfile': Permission denied
mv: cannot move '/tmp/src/Gemfile.local.example' to './Gemfile.local.example': Permission denied
mv: cannot move '/tmp/src/Gemfile.lock' to './Gemfile.lock': Permission denied
```
happens if removed

akofink: Using (or not using) root has security implications which don't matter in devel. Red Hat has taken the position that we will ship non-root images because of security, but they are meant to be used primarily in prod. Ideally, we could build the image as root and then run it via the docker-compose under whatever the $UID/$GID env variables are set to so file access is seamless in a development environment.

dLobatog: Again - this doesn't build the image as root nor makes it the default - it's the same as we do in prod. I revert to 1001 a few lines after: 
  - 1. Set root
  - 2. Copy source to /tmp/src - you *must* be root to do it
  - 3. Allow user 1001 to read it
  - 4. Set user back to 1001

PR #487, opened 2020-05-29
Fully implement V1 profiles#index

akofink:
Separated from https://github.com/RedHatInsights/compliance-backend/pull/410 to ease reviewing.

- Adds relationships to API response
- Allows side-loaded includes of all related resources
- Allows searching profiles by `canonical`

### DIFF
--- a/app/controllers/concerns/rendering.rb
+++ b/app/controllers/concerns/rendering.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# Rendering logic for all controllers
+module Rendering
+  extend ActiveSupport::Concern
+
+  included do
+    def render_json(model, **args)
+      opts = ['index'].include?(action_name) ? metadata : {}
+      opts.merge!(include: params[:include].split(',')) if params[:include]
+      render({ json: serializer.new(model, opts) }.merge(args))
+    end
+
+    def render_error(model, **args)
+      render({ json: {
+        errors: model.errors.full_messages
+      }, status: :not_acceptable }.merge(args))
+    end
+
+    private
+
+    def serializer
+      raise NotImplementedError
+    end
+  end
+end
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+# Methods that are related to profile searching
+module ProfileSearching
+  extend ActiveSupport::Concern
+
+  included do
+    scoped_search on: %i[id name ref_id account_id compliance_threshold
+                         external parent_profile_id]
+    scoped_search relation: :hosts, on: :id, rename: :system_ids
+    scoped_search relation: :hosts, on: :name, rename: :system_names
+    scoped_search on: :has_test_results, ext_method: 'test_results?',
+                  only_explicit: true, operators: ['=']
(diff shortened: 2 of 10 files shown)


app/models/concerns/profile_searching.rb
```diff
@@ -0,0 +1,42 @@
...
+    scope :has_test_results, lambda { |has_test_results = true|
+      test_results = TestResult.select(:profile_id).distinct
+      has_test_results && where(id: test_results) || where.not(id: test_results)
+    }
+  end
+
+  class_methods do
+    def canonical?(_filter, _operator, value)
+      profiles = Profile.canonical(ActiveModel::Type::Boolean.new.cast(value))
+      { conditions: profiles.arel.where_sql.gsub(/^where /i, '') }
```
akofink, line 33: This is an attempt at making `scoped_search` tied to our `scope` for this model and search parameter. It seems silly to have the logic in two different places, but this solution is also kind of a hack.

app/models/concerns/profile_searching.rb
```diff
@@ -0,0 +1,42 @@
...
+    scoped_search on: %i[id name ref_id account_id compliance_threshold
+                         external parent_profile_id]
+    scoped_search relation: :hosts, on: :id, rename: :system_ids
+    scoped_search relation: :hosts, on: :name, rename: :system_names
+    scoped_search on: :has_test_results, ext_method: 'test_results?',
+                  only_explicit: true, operators: ['=']
+    scoped_search on: :canonical, ext_method: 'canonical?', only_explicit: true,
+                  operators: ['=']
+
+    scope :canonical, lambda { |canonical = true|
```
akofink, line 17: Notice the added param here: `Profile.canonical` or `Profile.canonical(true)` are equivalent. `Profile.canonical(false)` is the negative case. The same is true of `Profile.external` and `Profile.has_test_results`

app/serializers/parent_profile_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# JSON API serialization for an OpenSCAP Profile
+class ParentProfileSerializer < ProfileSerializer
```
akofink, line 4: This is required as a quirk of the way FastJsonapi works with `belongs_to :parent_profile`

spec/swagger_helper.rb
```diff
@@ -117,11 +117,64 @@
           properties: {
```
akofink, line 117: This ought to always match the parameters returned by the associated serializer. I wonder if we could add an automated test that ensures they match for all API resources.

app/controllers/concerns/rendering.rb
```diff
@@ -0,0 +1,26 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    def render_json(model, **args)
+      opts = ['index'].include?(action_name) ? metadata : {}
+      opts.merge!(include: params[:include].split(',')) if params[:include]
+      render({ json: serializer.new(model, opts) }.merge(args))
+    end
+
+    def render_error(model, **args)
```
dLobatog, line 14: Is this used anywhere?

akofink: Good catch - this one was from create/update, but we'll be using it soon :)

PR #489, opened 2020-06-03
Fix readiness/liveness probes

dLobatog:
Probes are returning 401 on CI already because the v1 namespacing in routes.rb shouldn't include the openAPI endpoint as the openapi controller already has embedded 'v1', 'v2' or whatever version is needed


akofink:
@dLobatog this may be a silly question, but - should we have a functional test for this endpoint?

PR #491, opened 2020-06-03
Add tests for openapi.json endpoints

akofink:
`api/compliance/v1/openapi.json` test works fine.

### DIFF
--- a/test/controllers/v1/openapi_endpoint_test.rb
+++ b/test/controllers/v1/openapi_endpoint_test.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module V1
+  class OpenapiEndpointTest < ActionDispatch::IntegrationTest
+    test 'openapi v1 success' do
+      User.current = nil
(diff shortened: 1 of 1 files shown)


akofink:
Updated to only test v1

akofink:
This one's ready for re-review

test/controllers/v1/openapi_endpoint_test.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module V1
+  class OpenapiEndpointTest < ActionDispatch::IntegrationTest
+    test 'openapi v1 success' do
+      get '/api/compliance/v1/openapi.json'
```
dLobatog, line 8: Can we get an assertion before calling `get` on `User.current = nil` ? That's just to make sure this can get called without authorization or any current user set. I think this test could be rendered useless in the future if by mistake we set the User.current somewhere else.

akofink:
Updated with the User.current == nil assertion

PR #493, opened 2020-06-05
Version API spec to V1, refactor schema into files

akofink:
This should make the API specs much easier to maintain, especially as we
create new API versions. The entire API spec exists under spec/api/vX;
openapi schemas are separated into files depending on resource; utility
functions related to API schemas are separated into their own util
class. Note that this is just a refactoring and does not change the
openapi.json (other than some example ID for some reason).

### DIFF
--- a/spec/api/v1/schemas.rb
+++ b/spec/api/v1/schemas.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require './spec/api/v1/schemas/types'
+require './spec/api/v1/schemas/errors'
+require './spec/api/v1/schemas/metadata'
+require './spec/api/v1/schemas/hosts'
+require './spec/api/v1/schemas/benchmarks'
+require './spec/api/v1/schemas/profiles'
+require './spec/api/v1/schemas/rule_results'
+require './spec/api/v1/schemas/rules'
+
+module Api
+  module V1
+    module Schemas
+      include Types
+      include Errors
+      include Metadata
+      include Hosts
+      include Benchmarks
+      include Profiles
+      include RuleResults
+      include Rules
+
+      SCHEMAS = {
+        uuid: UUID,
+        relationship: RELATIONSHIP,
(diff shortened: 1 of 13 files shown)


spec/api/v1/schemas.rb
```diff
@@ -0,0 +1,38 @@
...
+      include Types
+      include Errors
+      include Metadata
+      include Hosts
+      include Benchmarks
+      include Profiles
+      include RuleResults
+      include Rules
+
+      SCHEMAS = {
```
akofink, line 24: I almost used methods instead of constants, but none of this content is dynamic. We can still call methods from constants, such as `ref_schema('foo')`, as seen in the types schemas. Furthermore, rubocop's method length cop does not apply to constants, and some of these specs are still quite long, well past the method length limit.

spec/api/v1/schemas/profiles.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Schemas
+      module Profiles
```
akofink, line 6: Most of these modules only define the resource schema, such as `PROFILE` here, but we can also create schemas for profile-related things, such as `PROFILE_RELATIONSHIPS` in the future. Right now, they're just defined inline in the `profile_spec.rb` test.

swagger/v1/openapi.json
```diff
@@ -415,7 +415,7 @@
                   "relationships": {
                     "account": {
                       "data": {
-                        "id": "cf8ae1be-2626-4c7b-8134-bda1c78946c6",
+                        "id": "6a888174-88fd-4e80-a100-25550e5f77e6",
```
akofink, line 418: Due to the nature of generated uuids by postgres, autogenerated examples from our fixtures will update uuids like this. I am not sure if we can work around this somehow, perhaps by setting the uuid manually for fixtures. It's a problem to solve for another day :)

PR #495, opened 2020-06-08
Implement profiles#destroy endpoint

akofink:
A user may destroy any profile that is associated to their account, both
internal or external. Destroying a profile also removes associated test
results.

Towards https://projects.engineering.redhat.com/browse/RHICOMPL-595

```json
$ http DELETE https://ci.foo.redhat.com:1337/api/compliance/profiles/e8c1675a-49a0-4ac6-aa60-b58508a5c0d6
HTTP/1.1 202 Accepted
cache-control: no-cache
connection: close
content-type: application/vnd.api+json
date: Mon, 08 Jun 2020 18:26:30 GMT
transfer-encoding: chunked
vary: Origin
x-request-id: be4cec94-709f-4864-a7d5-650cfadeac89
x-runtime: 0.260293

{
    "data": {
        "attributes": {
            "business_objective": null,
            "canonical": false,
            "compliance_threshold": 100.0,
            "compliant_host_count": 0,
            "description": "The HIPAA Security Rule establishes U.S. national standards to protect individuals’\nelectronic personal health information that is created, received, used, or\nmaintained by a covered entity. The Security Rule requires appropriate\nadministrative, physical and technical safeguards to ensure the\nconfidentiality, integrity, and security of electronic protected health\ninformation.\n\nThis profile configures Red Hat Enterprise Linux 7 to the HIPAA Security\nRule identified for securing of electronic protected health information.",
            "external": true,
            "name": "Health Insurance Portability and Accountability Act (HIPAA)",
            "os_major_version": "7",
            "parent_profile_id": "74fc0c7f-2f2d-4e4e-8688-76a8406b57ee",
            "parent_profile_ref_id": "xccdf_org.ssgproject.content_profile_hipaa",
            "ref_id": "xccdf_org.ssgproject.content_profile_hipaa",
            "score": 0,
            "tailored": true,
            "total_host_count": 0
        },
        "id": "e8c1675a-49a0-4ac6-aa60-b58508a5c0d6",
        "relationships": {
            "account": {
                "data": {
                    "id": "8c926d2a-17d6-4398-8895-2652971113c1",
                    "type": "account"
                }
            },
            "benchmark": {
                "data": {
                    "id": "3401d042-81a9-4155-b63a-5a114cfe702e",
                    "type": "benchmark"
                }
            },
            "business_objective": {
                "data": null
            },
            "parent_profile": {
                "data": {
                    "id": "74fc0c7f-2f2d-4e4e-8688-76a8406b57ee",
                    "type": "profile"
                }
            }
        },
        "type": "profile"
    }
}
```

CC @Victoremepunto 

### DIFF
--- a/app/controllers/v1/profiles_controller.rb
+++ b/app/controllers/v1/profiles_controller.rb
@@ -12,6 +12,11 @@ module V1
       render_json profile
     end
 
+    def destroy
+      authorize profile
+      render_json profile.destroy, status: :accepted
+    end
+
     def tailoring_file
       return unless profile.tailored?
 
--- a/test/controllers/v1/profiles_controller_test.rb
+++ b/test/controllers/v1/profiles_controller_test.rb
@@ -11,28 +11,74 @@ module V1
       profiles(:one).update! account: accounts(:test)
     end
 
-    test 'tailoring_file with a canonical profile returns no content' do
-      profiles(:one).update! rules: [rules(:one)]
-      get tailoring_file_v1_profile_url(profiles(:one).id)
-      assert_response :no_content
+    class TailoringFileTest < ProfilesControllerTest
+      test 'tailoring_file with a canonical profile returns no content' do
+        profiles(:one).update! rules: [rules(:one)]
+        get tailoring_file_v1_profile_url(profiles(:one).id)
+        assert_response :no_content
+      end
+
+      test 'tailoring_file with a noncanonical profile matching its parent'\
+        'returns no content' do
+        profiles(:two).update!(rules: [rules(:one)])
+        profiles(:one).update!(parent_profile: profiles(:two),
+                               rules: [rules(:one)])
+        get tailoring_file_v1_profile_url(profiles(:one).id)
+        assert_response :no_content
+      end
+
+      test 'tailoring_file with a noncanonical profile '\
+           'returns tailoring file' do
+        profiles(:two).update!(rules: [rules(:one), rules(:two)])
+        profiles(:one).update!(parent_profile: profiles(:two),
+                               rules: [rules(:one)])
+        get tailoring_file_v1_profile_url(profiles(:one).id)
+        assert_response :success
+        assert_equal Mime[:xml].to_s, @response.content_type
+      end
     end
 
-    test 'tailoring_file with a noncanonical profile matching its parent'\
-      'returns no content' do
-      profiles(:two).update!(rules: [rules(:one)])
-      profiles(:one).update!(parent_profile: profiles(:two),
-                             rules: [rules(:one)])
-      get tailoring_file_v1_profile_url(profiles(:one).id)
-      assert_response :no_content
-    end
+    class DestroyTest < ProfilesControllerTest
+      require 'sidekiq/testing'
+      Sidekiq::Testing.inline!
 
-    test 'tailoring_file with a noncanonical profile returns tailoring file' do
-      profiles(:two).update!(rules: [rules(:one), rules(:two)])
-      profiles(:one).update!(parent_profile: profiles(:two),
-                             rules: [rules(:one)])
-      get tailoring_file_v1_profile_url(profiles(:one).id)
-      assert_response :success
-      assert_equal Mime[:xml].to_s, @response.content_type
+      test 'destroy an existing, accessible profile' do
+        profile_id = profiles(:one).id
+        assert_difference('Profile.count' => -1) do
+          delete profile_path(profile_id)
+        end
+        assert_response :success
+        assert_equal 202, response.status, 'Response should be 202 accepted'
+        assert_equal profile_id, JSON.parse(response.body).dig('data', 'id'),
+                     'Profile ID did not match deleted profile'
+      end
+
+      test 'v1 destroy an existing, accessible profile' do
+        profile_id = profiles(:one).id
+        assert_difference('Profile.count' => -1) do
+          delete v1_profile_path(profile_id)
+        end
+        assert_response :success
+        assert_equal 202, response.status, 'Response should be 202 accepted'
+        assert_equal profile_id, JSON.parse(response.body).dig('data', 'id'),
+                     'Profile ID did not match deleted profile'
+      end
+
+      test 'destroy a non-existant profile' do
+        profile_id = profiles(:one).id
+        profiles(:one).destroy
+        assert_difference('Profile.count' => 0) do
+          delete v1_profile_path(profile_id)
+        end
+        assert_response :not_found
+      end
+
+      test 'destroy an existing, not accessible profile' do
+        assert_difference('Profile.count' => 0) do
+          delete v1_profile_path(profiles(:two).id)
+        end
+        assert_response :not_found
+      end
     end
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,7 +5,7 @@ Rails.application.routes.draw do
     scope "#{prefix}/#{ENV['APP_NAME']}" do
       namespace :v1 do
         resources :benchmarks, only: [:index, :show]
-        resources :profiles, only: [:index, :show] do
+        resources :profiles, only: [:index, :show, :destroy] do
           member do
             get 'tailoring_file'
           end
@@ -15,7 +15,7 @@ Rails.application.routes.draw do
         resources :rules, only: [:index, :show]
       end
       resources :benchmarks, controller: 'v1/benchmarks', only: [:index, :show]
-      resources :profiles, controller: 'v1/profiles', only: [:index, :show] do
+      resources :profiles, controller: 'v1/profiles', only: [:index, :show, :destroy] do
         member do
           get 'tailoring_file'
         end
--- a/spec/integration/profiles_spec.rb
+++ b/spec/integration/profiles_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'swagger_helper'
+require 'sidekiq/testing'
 
 describe 'Profiles API' do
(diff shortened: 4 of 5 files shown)


test/controllers/v1/profiles_controller_test.rb
```diff
@@ -11,28 +11,65 @@ class ProfilesControllerTest < ActionDispatch::IntegrationTest
...
 
-    test 'tailoring_file with a noncanonical profile returns tailoring file' do
-      profiles(:two).update!(rules: [rules(:one), rules(:two)])
-      profiles(:one).update!(parent_profile: profiles(:two),
-                             rules: [rules(:one)])
-      get tailoring_file_v1_profile_url(profiles(:one).id)
-      assert_response :success
-      assert_equal Mime[:xml].to_s, @response.content_type
+    class DestroyTest < ProfilesControllerTest
+      require 'sidekiq/testing'
```
dLobatog, line 42: I tried removing it, there was no issue with it - I think for this to be effective you need to associate the profile to some hosts so that some DeleteHost jobs get actually enqueued

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -11,28 +11,65 @@ class ProfilesControllerTest < ActionDispatch::IntegrationTest
...
+      end
+
+      test 'v1 destroy an existing, accessible profile' do
+        profile_id = profiles(:one).id
+        delete v1_profile_path(profile_id)
+        assert_response :success
+        assert_equal 202, response.status, 'Response should be 202 accepted'
+        assert_equal profile_id, JSON.parse(response.body).dig('data', 'id'),
+                     'Profile ID did not match deleted profile'
+      end
```
dLobatog, line 60: These should be wrapped in an `assert_difference` or something to verify the profile was truly deleted after the action. You may also want to check `DeleteHost.jobs.size` and call `DeleteHost.drain` to ensure the orphaned associated hosts are removed with the Profile - your call (we already check this in the ProfileHost tests), but if you don't check any jobs, I don't think sidekiq/testing is required

app/controllers/v1/profiles_controller.rb
```diff
@@ -12,6 +12,11 @@ def show
       render_json profile
     end
 
+    def destroy
+      authorize profile
+      render_json profile.destroy, status: :accepted
```
dLobatog, line 17: Just to make sure I understand - this is `accepted` because of the ProfileHost relationship, is that right? From what I see, this returns the deleted profile in the body just right

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -11,28 +11,65 @@ class ProfilesControllerTest < ActionDispatch::IntegrationTest
...
     end
 
-    test 'tailoring_file with a noncanonical profile returns tailoring file' do
-      profiles(:two).update!(rules: [rules(:one), rules(:two)])
-      profiles(:one).update!(parent_profile: profiles(:two),
-                             rules: [rules(:one)])
-      get tailoring_file_v1_profile_url(profiles(:one).id)
-      assert_response :success
-      assert_equal Mime[:xml].to_s, @response.content_type
+    class DestroyTest < ProfilesControllerTest
```
dLobatog, line 41: No need for a subclass here, thoughts about using a `context 'destroy'` block?

akofink: context is for rspec - this is minitest. The subclasses are nice IMO and allow us to one day even move them out to their own files when this test gets really big, like `test/controllers/v1/profiles_controller/destroy_test.rb`.

app/controllers/v1/profiles_controller.rb
akofink: This is part of the JSON spec: https://jsonapi.org/format/#crud-deleting-responses

> If a deletion request has been accepted for processing, but the processing has not been completed by the time the server responds, the server MUST return a 202 Accepted status code.

Basically, because we're queuing a job for any part of the deletion, it must return a 202.

test/controllers/v1/profiles_controller_test.rb
akofink: I'll add those checks. Anytime a test queues any job in Sidekiq, `sidekiq/testing` must be required or the server tries to connect to redis and fails.

test/controllers/v1/profiles_controller_test.rb
akofink: Odd - I get errors when I remove it. Are you running the tests on their own or the whole suite?

```
$ d-c exec -e TEST=$TEST reports-consumer bundle exec rake test TEST=test/controllers/v1/profiles_controller_test.rb
Started with run options --seed 31260

DEPRECATION WARNING: 'alias' argument deprecated, prefer aliases: [..] since scoped_search 4.0.0 (called from <class:Rule> at /app/app/models/rule.rb:9)---] 0% Time: 00:00:00,  ETA: ??:??:??
DEPRECATION WARNING: 'alias' argument deprecated, prefer aliases: [..] since scoped_search 4.0.0 (called from <class:Rule> at /app/app/models/rule.rb:10)
DEPRECATION WARNING: 'alias' argument deprecated, prefer aliases: [..] since scoped_search 4.0.0 (called from <class:Benchmark> at /app/app/models/xccdf/benchmark.rb:10)
DEPRECATION WARNING: 'alias' argument deprecated, prefer aliases: [..] since scoped_search 4.0.0 (called from <class:Benchmark> at /app/app/models/xccdf/benchmark.rb:12)
ERROR["test_v1_destroy_an_existing,_accessible_profile", #<Minitest::Reporters::Suite:0x000055c08dd0be18 @name="V1::ProfilesControllerTest::DestroyTest">, 0.367351297987625]
 test_v1_destroy_an_existing,_accessible_profile#V1::ProfilesControllerTest::DestroyTest (0.37s)
Minitest::UnexpectedError:         Redis::CannotConnectError: Error connecting to Redis on 127.0.0.1:6379 (Errno::ECONNREFUSED)
            app/models/concerns/profile_policy_association.rb:40:in `destroy_policy_test_results'
            app/controllers/v1/profiles_controller.rb:17:in `destroy'
            test/controllers/v1/profiles_controller_test.rb:53:in `block in <class:DestroyTest>'

ERROR["test_destroy_a_non-existant_profile", #<Minitest::Reporters::Suite:0x000055c08e2920d0 @name="V1::ProfilesControllerTest::DestroyTest">, 0.38759273797040805]
 test_destroy_a_non-existant_profile#V1::ProfilesControllerTest::DestroyTest (0.39s)
Minitest::UnexpectedError:         Redis::CannotConnectError: Error connecting to Redis on 127.0.0.1:6379 (Errno::ECONNREFUSED)
            app/models/concerns/profile_policy_association.rb:40:in `destroy_policy_test_results'
            test/controllers/v1/profiles_controller_test.rb:62:in `block in <class:DestroyTest>'

ERROR["test_destroy_an_existing,_accessible_profile", #<Minitest::Reporters::Suite:0x000055c08f67ff48 @name="V1::ProfilesControllerTest::DestroyTest">, 0.4133574939914979]
 test_destroy_an_existing,_accessible_profile#V1::ProfilesControllerTest::DestroyTest (0.41s)
Minitest::UnexpectedError:         Redis::CannotConnectError: Error connecting to Redis on 127.0.0.1:6379 (Errno::ECONNREFUSED)
            app/models/concerns/profile_policy_association.rb:40:in `destroy_policy_test_results'
            app/controllers/v1/profiles_controller.rb:17:in `destroy'
            test/controllers/v1/profiles_controller_test.rb:44:in `block in <class:DestroyTest>'

  7/7: [=================================================================================================================================================] 100% Time: 00:00:00, Time: 00:00:00

Finished in 0.57739s
7 tests, 5 assertions, 0 failures, 3 errors, 0 skips
Coverage report generated for RSpec to /app/coverage. 434 / 2903 LOC (14.95%) covered.
```

test/controllers/v1/profiles_controller_test.rb
akofink: So, part of the reason I didn't add this is because the controller has no knowledge of how the host gets deleted - it just calls `destroy` and the logic passes into the model, where there are unit tests to test host deletion. Whether a job is queued or whatever happens shouldn't be tested by the controller IMO. I think it's okay to test that the profile was deleted, though, just not how it was deleted.

akofink:
Updated with `assert_difference` checks. I'm using `Sidekiq::Testing.inline!` here so that we're not having the controller test any intricacies of how the profile or associated records are deleted - we should leave that up to the model tests.

test/controllers/v1/profiles_controller_test.rb
dLobatog: I was running the test on its own

test/controllers/v1/profiles_controller_test.rb
dLobatog: FWIW context is available in minitest because we use https://github.com/thoughtbot/shoulda-context

test/controllers/v1/profiles_controller_test.rb
dLobatog: Agreed - I think I was not getting the error because I did have redis on while testing it

test/controllers/v1/profiles_controller_test.rb
akofink: Right, but this doesn't work with the minitest `test 'foo'` syntax that we're using in all of our controllers. I'm open to changing all of them to `should 'foo'`, though I don't really see the downside of defining classes like this either. Either way, whatever we ultimately decide in the future, we can easily swap between the different syntax.

It seems also that the shoulda-context benefits are mostly the helper methods provided with shoulda-matchers like we're using in our model tests (i.e. `should validate_presence_of(:first_name)`). There are also helpers like this for controllers, but they seem less useful to me.

PR #496, opened 2020-06-09
RHICOMPL-610 - Internal profiles are only shown by default

dLobatog:
Profiles#index now only shows internal profiles by default, which are
non-external, non-canonical profiles.


app/models/concerns/profile_searching.rb
```diff
@@ -24,6 +24,9 @@ module ProfileSearching
     scope :external, lambda { |external = true|
       where(external: external)
     }
+    scope :internal , lambda {
+      external(false).canonical(false)
+    }
```
akofink, line 29: This scope name is misleading IMO - we have `external` - `internal` should be the opposite. What is it exactly you want to return with this scope, user-created profiles, policy-like profiles? I think we should name it something more descriptive.

app/models/concerns/profile_searching.rb
```diff
@@ -24,6 +24,9 @@ module ProfileSearching
     scope :external, lambda { |external = true|
       where(external: external)
     }
+    scope :internal , lambda {
+      external(false).canonical(false)
```
akofink, line 28: The profile pundit policy already limits to canonical = false - this `canonical(false)` change doesn't mean much if we don't update the policy so users may query for canonical profiles in the first place.

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -11,6 +11,29 @@ class ProfilesControllerTest < ActionDispatch::IntegrationTest
       profiles(:one).update! account: accounts(:test)
     end
 
+    context 'index' do
```
akofink, line 14: Somehow we overlooked index controller tests when the PR went in. I'd specifically like a test that shows that a user can still query for all internal and external, canonical and non-canonical profiles (essentially all profiles that are not a part of someone else's account) with this change.

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -11,6 +11,29 @@ class ProfilesControllerTest < ActionDispatch::IntegrationTest
...
+      should 'not contain external or canonical profiles by default' do
+        get v1_profiles_url
+        assert_response :success
+
+        profiles = JSON.parse(response.body)
+        assert_empty profiles['data']
+      end
+
+      should 'only contain internal profiles by default' do
```
akofink, line 23: Why are we using `should` here? It doesn't match any of our other controller tests.

dLobatog: We do use should from and context from shoulda-context in `test/controllers/concerns/metadata_test.rb`, `test/controllers/concerns/authentication_test.rb` and many service and model tests

app/models/concerns/profile_searching.rb
dLobatog: "policy-like profiles" - what the UI shows as policies in the list. I can call it "policy_like" if it's more clear that way

akofink: I'm fine with policy_like.

test/controllers/v1/profiles_controller_test.rb
akofink: Ok, it's not really a big deal, just weird to have mixed syntax in a single test file.

app/models/concerns/profile_searching.rb
akofink: This is solved in the backend by https://github.com/RedHatInsights/compliance-backend/pull/497

PR #497, opened 2020-06-09
Users may view/search for canonical profiles

akofink:
This updates the pundit policy for profiles to allow users to see
canonical profiles when searching or listing profiles

Canonical profiles must be visible to any user in order to pick a parent
profile during policy creation via the API or UI.
Note: Currently the UI does not use this pundit policy to limit the scope of
profiles visible to the user.

Requires frontend changes to add `canonical = false` to profile queries:
- https://github.com/RedHatInsights/compliance-frontend/pull/624
- https://github.com/RedHatInsights/frontend-components/pull/556

### DIFF
--- a/test/controllers/v1/profiles_controller_test.rb
+++ b/test/controllers/v1/profiles_controller_test.rb
@@ -74,10 +74,19 @@ module V1
       end
 
       test 'destroy an existing, not accessible profile' do
+        profiles(:two).update! parent_profile: profiles(:one)
         assert_difference('Profile.count' => 0) do
           delete v1_profile_path(profiles(:two).id)
         end
         assert_response :not_found
+      end
+
+      test 'destroy an existing, accessible profile that is not authorized '\
+           'to be deleted' do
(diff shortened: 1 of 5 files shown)


test/controllers/v1/profiles_controller_test.rb
```diff
@@ -74,11 +74,20 @@ class DestroyTest < ProfilesControllerTest
       end
 
       test 'destroy an existing, not accessible profile' do
+        profiles(:two).update! parent_profile: profiles(:one)
```
akofink, line 77: This now creates two different cases: (1) delete a not accessible profile, one that exists in the DB, but the user is not authorized by pundit to view the policy at all; (2) delete an accessible profile (accessible via show/index) that the user is not authorized to delete (i.e. a canonical profile).

Case 1 returns a 404 as if the profile does not exist for that user. Case 2 returns a 403 forbidden since the user knows the profile exists but is forbidden to delete it.

dLobatog:
Frontend PRs deployed in CI / QA - merging

PR #498, opened 2020-06-10
RHICOMPL-668 - Import os_major/minor on report parsing

dLobatog:
Just a draft for now. The idea is to have these fields available to us so that:

- Supportability can be quickly checked
- Filtering becomes trivial
- Showing the OS doesn't require us waiting a potentially long time for inventory to release changes
- Hosts without an OS can be quickly deemed unsupported

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -3,9 +3,7 @@
 # Host representation in insights compliance backend. Most of the times
 # these hosts will also show up in the insights-platform host inventory.
 class Host < ApplicationRecord
-  scoped_search on: %i[id name account_id os_major os_minor]
-  scoped_search on: :os_major, aliases: [:os_major_release]
-  scoped_search on: :os_minor, aliases: [:os_minor_release]
+  scoped_search on: %i[id name account_id os_major_version os_minor_version]
   scoped_search on: :compliant, ext_method: 'filter_by_compliance',
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -4,6 +4,8 @@
 # these hosts will also show up in the insights-platform host inventory.
 class Host < ApplicationRecord
   scoped_search on: %i[id name account_id os_major os_minor]
+  scoped_search on: :os_major, aliases: [:os_major_release]
+  scoped_search on: :os_minor, aliases: [:os_minor_release]
   scoped_search on: :compliant, ext_method: 'filter_by_compliance',
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -3,7 +3,7 @@
 # Host representation in insights compliance backend. Most of the times
 # these hosts will also show up in the insights-platform host inventory.
 class Host < ApplicationRecord
-  scoped_search on: %i[id name account_id]
+  scoped_search on: %i[id name account_id os_major os_minor]
   scoped_search on: :compliant, ext_method: 'filter_by_compliance',
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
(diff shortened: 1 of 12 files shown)


app/models/host.rb
```diff
@@ -3,7 +3,7 @@
 # Host representation in insights compliance backend. Most of the times
 # these hosts will also show up in the insights-platform host inventory.
 class Host < ApplicationRecord
-  scoped_search on: %i[id name account_id]
+  scoped_search on: %i[id name account_id os_major os_minor]
```
akofink, line 6: Let's strive for consistency with benchmarks and profiles:

```suggestion
  scoped_search on: %i[id name account_id os_major_release os_minor_release]
```

This is especially important for scoped_search because it makes it to our APIs.

app/services/host_inventory_api.rb
```diff
@@ -26,9 +28,29 @@ def inventory_host
...
+    response.body['results'].inject([]) do |acc, host|
+      os_major, os_minor = host['system_profile']['os_release'].split('.')
+      acc << { id: host['id'], os_major: os_major, os_minor: os_minor }
+    end
+  end
+
+  def import_os_releases(ids)
+    os_releases = host_os_releases(ids)
+    Host.import os_releases
+  end
```
akofink, line 52: This looks out of place here, IMO - why should the `HostInventoryApi` know or do anything with our `Host` model?

db/migrate/20200610083045_add_os_major_and_os_minor_to_hosts.rb
```diff
@@ -0,0 +1,6 @@
+class AddOsMajorAndOsMinorToHosts < ActiveRecord::Migration[5.2]
+  def change
+    add_column :hosts, :os_major, :integer
+    add_column :hosts, :os_minor, :integer
```
akofink, line 4: Are these indexed? They should be, don't you think?

app/services/host_inventory_api.rb
```diff
@@ -3,37 +3,54 @@
...
 
+  def system_profile(ids)
+    response = Platform.connection.get(
+      "#{@url}/#{ids.join(',')}/system_profile", { per_page: 50, page: 1 },
+      X_RH_IDENTITY: @b64_identity
+    )
+    JSON.parse(response.body)['results'].inject([]) do |acc, host|
+      os_major, os_minor = host['system_profile']['os_release'].split('.')
+      acc << { id: host['id'], os_major: os_major, os_minor: os_minor }
+    end
```
akofink, line 46: Nitpick: we're only returning id, os_major, os_minor here, which makes sense for what we're using it for today, but the method name makes it seem it returns the entire profile. We may use this in the future for more information on the system profile, such as installed ssg package version, etc.

app/models/host.rb
```diff
@@ -3,7 +3,9 @@
...
+  scoped_search on: :os_major, aliases: [:os_major_release]
+  scoped_search on: :os_minor, aliases: [:os_minor_release]
```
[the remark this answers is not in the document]
dLobatog: I think just `os_major` should suffice as it's what is returned by the API - you asked for `_release` in the name of consistency with Benchmarks, but as far as I can see Benchmarks query is on `os_major_version`

app/services/host_inventory_api.rb
dLobatog: Sure - that makes sense to me, we only care about OS for now

app/models/host.rb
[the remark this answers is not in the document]
akofink: Good catch! Yep, this should be `os_major_version`/`os_miner_version` for consistency with profiles and benchmarks. Sorry for the confusion! The hosts API should also be returning these attributes like that as well to match profiles and benchmarks.

db/migrate/20200610083045_add_os_major_and_os_minor_to_hosts.rb
```diff
@@ -0,0 +1,8 @@
...
```
akofink, line 4: And here, if we just go ahead and name the columns `os_major_version`/`os_minor_version` we won't need the scoped_search renames.

PR #499, opened 2020-06-10
Implement profiles#create endpoint

akofink:
Any params not specified are inherited from the parent profile

Minimal params:
```js
{
	"data": {
		"attributes": {
			"parent_profile_id": "1007b72e-7f0f-486f-911a-5c392c724faa"
		}
	}
}
```

All allowable params:
```js
{
	"data": {
		"attributes": {
			"parent_profile_id": "1007b72e-7f0f-486f-911a-5c392c724faa",
			"name": "My profile",
			"description": "The best profile ever",
			"compliance_threshold": 95.0,
			"business_objective": "LATAM Expansion"
		}
	}
}
```

rswag :( I'm trying to figure out how to pass JSON like parameters, but I haven't yet figured it out (see profiles_spec.rb).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,8 +11,20 @@ class ApplicationController < ActionController::API
   include Search
   include Rendering
 
+  ParamType = ActionController::Parameters
+
   def openapi
     send_file Rails.root.join('swagger/v1/openapi.v3.yaml')
+  end
+
+  def pundit_scope
+    Pundit.policy_scope(current_user, resource)
+  end
+
+  def resource_params
+    params.permit(data: ParamType.map(attributes: ParamType.map))
+    params.require(:data).require(:attributes)
+          .permit(serializer.attributes_to_serialize.keys)
   end
 
   rescue_from Pundit::NotAuthorizedError do
@@ -22,13 +34,19 @@ class ApplicationController < ActionController::API
 
   rescue_from ActiveRecord::RecordNotFound do |error|
     logger.info "#{error.message} (#{error.class})"
-    render json: { errors: 'Resource not found' },
+    render json: { errors: "#{error.model} not found with ID #{error.id}" },
            status: :not_found
   end
 
   rescue_from ActionController::ParameterMissing do |error|
     logger.info "#{error.message} (#{error.class})"
     render json: { errors: "Parameter missing: #{error.message}" },
+           status: :unprocessable_entity
+  end
+
+  rescue_from StrongerParameters::InvalidParameter do |error|
+    logger.info "#{error.message} (#{error.class})"
+    render json: { errors: error.message },
            status: :unprocessable_entity
   end
 end
--- a/app/controllers/v1/profiles_controller.rb
+++ b/app/controllers/v1/profiles_controller.rb
@@ -3,6 +3,11 @@
 module V1
   # API for Profiles
   class ProfilesController < ApplicationController
+    ALLOWED_CREATE_ATTRIBUTES = %i[
+      name description compliance_threshold business_objective
+      parent_profile_id
+    ].freeze
+
     def index
       params[:search] ||= 'external=false and canonical=false'
       render_json scope_search.sort_by(&:score)
@@ -16,6 +21,16 @@ module V1
     def destroy
       authorize profile
       render_json profile.destroy, status: :accepted
+    end
+
+    def create
+      profile = Profile.new(profile_create_attributes).fill_from_parent
+
+      if profile.save
+        render_json profile, status: :created
+      else
+        render_error profile
+      end
     end
 
     def tailoring_file
@@ -34,7 +49,31 @@ module V1
     end
 
     def profile
-      @profile ||= Pundit.policy_scope(current_user, resource).find(params[:id])
+      @profile ||= pundit_scope.find(params[:id])
+    end
+
+    def parent_profile
+      @parent_profile ||= pundit_scope.find(
+        resource_params.require(:parent_profile_id)
+      )
+    end
+
+    def business_objective
+      @business_objective ||= Pundit.policy_scope(
+        current_user, BusinessObjective
+      ).from_title(resource_params[:business_objective])
+    end
+
+    def profile_create_attributes
+      resource_params.to_h.slice(*ALLOWED_CREATE_ATTRIBUTES)
+                     .merge(account_id: current_user.account_id).tap do |attrs|
+        parent_profile
+
+        if business_objective
+          attrs.except! :business_objective
+          attrs[:business_objective_id] = business_objective.id
+        end
+      end
     end
 
     def resource
--- a/app/models/business_objective.rb
+++ b/app/models/business_objective.rb
@@ -8,6 +8,14 @@ class BusinessObjective < ApplicationRecord
   validates :title, presence: true
 
   scope :in_account, lambda { |account_or_account_id|
-    joins(:accounts).where(accounts: { id: account_or_account_id })
+    joins(:accounts).where(accounts: { id: account_or_account_id }).distinct
   }
+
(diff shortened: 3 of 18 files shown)


app/controllers/application_controller.rb
```diff
@@ -11,18 +11,30 @@ class ApplicationController < ActionController::API
...
+
   rescue_from Pundit::NotAuthorizedError do
     render json: { errors: 'You are not authorized to access this action.' },
            status: :forbidden
   end
 
   rescue_from ActiveRecord::RecordNotFound do |error|
     logger.info "#{error.message} (#{error.class})"
-    render json: { errors: 'Resource not found' },
+    render json: { errors: "#{error.model} not found with ID #{error.id}" },
```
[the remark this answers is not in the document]
dLobatog: Makes sense to me

app/controllers/application_controller.rb
```diff
@@ -31,4 +43,10 @@ def openapi
     render json: { errors: "Parameter missing: #{error.message}" },
            status: :unprocessable_entity
   end
+
+  rescue_from StrongerParameters::InvalidParameter do |error|
+    logger.info "#{error.message} (#{error.class})"
+    render json: { errors: error.message },
+           status: :unprocessable_entity
```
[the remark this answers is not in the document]
dLobatog: Very helpful, thanks!

app/controllers/v1/profiles_controller.rb
```diff
@@ -17,6 +22,16 @@ def destroy
       render_json profile.destroy, status: :accepted
     end
 
+    def create
+      profile = Profile.new(profile_create_attributes).fill_from_parent
+
+      if profile.save
+        render_json profile, status: :created
```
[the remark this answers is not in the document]
dLobatog: Sure, 201 makes sense

app/models/business_objective.rb
```diff
@@ -10,4 +10,8 @@ class BusinessObjective < ApplicationRecord
   scope :in_account, lambda { |account_or_account_id|
     joins(:accounts).where(accounts: { id: account_or_account_id })
   }
+
+  def self.from_title(title)
+    find_or_create_by(title: title) if title
```
dLobatog, line 15: `find_or_create_by` has to be by `title` AND `account_id`, otherwise it will match the BO to any BO in any account

dLobatog: In fact the validation above should ideally take that into account...

app/models/profile.rb
```diff
@@ -43,6 +43,16 @@ def from_openscap_parser(op_profile, benchmark_id: nil, account_id: nil)
...
   end
 
+  def fill_from_parent
+    self.ref_id = parent_profile.ref_id
+    self.name ||= parent_profile.name
+    self.description ||= parent_profile.description
+    self.benchmark_id = parent_profile.benchmark_id
+    self.external = false
+
+    self
```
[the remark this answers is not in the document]
dLobatog: External would be false by default and can't be passed through the API via the validated params you added here, but it's probably good to be explicit

app/controllers/v1/profiles_controller.rb
```diff
@@ -18,6 +23,16 @@ def destroy
...
+      else
+        render_error profile
```
dLobatog, line 32: There's a scenario that looks like it could be handled here. 

1. POST contained a `business_objective` attribute "foo" - this BO doesn't exist in the current account
2. BO is created via `business_objective` 
3. Any error happens during `profile.save`, e.g: name already exists 

At this point, the "FOO" business objective will remain in the DB without being assigned to any profile.

app/models/profile.rb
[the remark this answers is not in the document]
dLobatog: I think this is missing a call to `add_rules`, or where else are the rules from the parent added to the profile?

app/models/business_objective.rb
akofink: The BusinessObjective model has no account_id:

```rb
BusinessObjective(id: uuid, title: string, created_at: datetime, updated_at: datetime)
```

The BO <-> Account association happens on the profile with BO <-> Profile <-> Account

app/controllers/v1/profiles_controller.rb
akofink: Good catch!

app/models/business_objective.rb
dLobatog: Oh, good point - it should take into account the account in here, perhaps by searching on the `Pundit.policy_scope` of BOs for the current account? Just ` find_or_create_by` is searching *across all accounts*

spec/api/v1/schemas/types.rb
```diff
@@ -16,6 +16,18 @@ module Types
             type: :string
           }
         }.freeze
+
+        RELATIONSHIP_COLLECTION = {
```
akofink, line 20: i.e. profile rules, profile hosts, profile test results - any belongs_to or has_many relationship

swagger/v1/openapi.json
```diff
@@ -1280,6 +1388,29 @@
           }
         }
       },
+      "profile_relationships": {
```
akofink, line 1391: New schema here for profile relationships

test/models/profile_test.rb
```diff
@@ -233,6 +233,48 @@ class ProfileTest < ActiveSupport::TestCase
...
+    should 'copy rules from the parent profile on save' do
+      profile = Profile.new(parent_profile_id: profiles(:one).id)
+                       .fill_from_parent
+
+      assert_not_equal Set.new(profiles(:one).rules.pluck(:id)),
+                       Set.new(profile.rules.pluck(:id))
+      profile.save
+      assert_equal Set.new(profiles(:one).rules.pluck(:id)),
+                   Set.new(profile.reload.rules.pluck(:id))
+    end
```
akofink, line 260: Ok, @dLobatog, this now copies the rules from the parent profile. I haven't yet implemented tailoring yet - this PR is quite complete as is and already pretty complicated. Perhaps it'd be better to implement tailoring in another PR. For tailoring, I'm thinking we should have a separate API for the profile <-> rules association _and_ allow including relationships in the POST request for profile creation.

1. https://jsonapi.org/format/#crud-creating - notice the `"relationships"` passed in with the POST request
1. https://jsonapi.org/format/#crud-updating-resource-relationships - updating relationships in a PATCH request
1. https://jsonapi.org/format/#crud-updating-relationships - similar thing here, but with just the relationship, not the editable attributes. I think we might hold off on this until V2 and only rely on the PATCH option  for V1

CC @bastilian for your thoughts, as well

dLobatog: Just to confirm we are on the same page:

`The relationship’s value will be replaced with the value specified in this member.` 

That means if you send the PATCH request to update the rules relationship, you override it - it's not additive or anything. I think that sounds good to me, but we need to document it, I don't think it's overly surprising but it can be.

akofink: Right - I think the idea is that you GET a resource, it returns a json document in the schema of a profile, you edit the document with the changes you want, and then PATCH the whole thing back with some thing changed and some things remaining the same.  I guess, one final thing to note is that, we can have default behavior if the relationship isn't provided (assigning parent_profile rules).

For tailoring, you'd GET the parent profile and benchmark - you can use a single request if you `include = benchmark`, copy the rules from that, edit them how you want, and add them to the profile create json before POSTing it to create a new profile with tailored rules.

For documenting, I'll be sure to mention that any provided relationships override all existing values in the relationships param description. Anywhere else we should document it?

akofink:
@dLobatog anything else missing here other than tailoring? I think I've addressed all the comments.

akofink:
Updated with newly registered mime type and swagger docs for profile creation.

PR #500, opened 2020-06-11
[stable] Update stable

akofink:
This contains all the changes on master up to but not including the profile pundit policy change to allow users to see canonical profiles (https://github.com/RedHatInsights/compliance-backend/pull/497).


akofink:
Removed https://github.com/RedHatInsights/compliance-backend/pull/486 from this PR. It will be split into its own stable PR.

PR #502, opened 2020-06-11
Handle RBAC errors appropriately by returning 403

akofink:
Well, this is embarrassing. Fixed, and now with tests!

`Rails.logger.info(...)` returns `true`

Tip: hide whitespace changes for easier reviewing.

https://github.com/RedHatInsights/compliance-backend/pull/503 fixes existing tests on the master branch

### DIFF
--- a/app/services/rbac_api.rb
+++ b/app/services/rbac_api.rb
@@ -9,12 +9,16 @@ class RbacApi
   end
 
   def check_user
-    body = JSON.parse(access_check_response.body)
-    body['data'].find do |access_check|
-      return true if access_check['permission'] == 'compliance:*:*'
+    begin
+      body = JSON.parse(access_check_response.body)
+      body['data'].find do |access_check|
+        return true if access_check['permission'] == 'compliance:*:*'
+      end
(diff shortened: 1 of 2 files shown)


app/services/rbac_api.rb
```diff
@@ -9,12 +9,16 @@ def initialize(b64_identity)
...
+      body['data'].find do |access_check|
+        return true if access_check['permission'] == 'compliance:*:*'
+      end
+    rescue Faraday::ClientError => e
+      Rails.logger.info("#{e.message} #{e.response}")
     end
-  rescue Faraday::ClientError => e
-    Rails.logger.info("#{e.message} #{e.response}")
+
+    false
```
akofink, line 21: Best to use a real begin/rescue/end block instead of the very weirdly behaving in-method rescue/ensure. This makes it super clear what happens if everything falls through.

dLobatog:
:+1:  - I think tests would be green after a rebase, feel free to merge then - thanks @akofink

akofink:
Rebased, and :cactus: tests

PR #505, opened 2020-06-18
Fix for docker-compose

Victoremepunto:
- Add a mandatory environment variable for PostgreSQL container to
 run
 - Remove unnecessary network section
 - Add a note on README to include the upload-service pre-requisite

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ SPANDX_CONFIG=$(pwd)/compliance-backend.js ../insights-proxy/scripts/run.sh
 
 ### Option 3: Docker Development setup
 
+## pre-requisites:
+
+There should be an [upload-service](https://github.com/RedHatInsights/insights-upload) container available in the same network the `rails` container runs.
+
 Bring up the DB, rails, and racecar:
 
 ```shell
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,8 @@ services:
   db:
     image: postgres
     restart: on-failure
+    environment:
+        - POSTGRES_PASSWORD=postgres
     volumes:
       - ./tmp/insights-compliance-db:/var/lib/postgresql/data:z
   rails:
@@ -95,7 +97,3 @@ services:
   redis:
     image: redis:latest
 
-networks:
-  default:
-    external:
-      name: docker_default # should match insights-upload's network name


Victoremepunto:
Not sure about the pre-requisite section, specially since `upload-service` is archived, though there's no mention which service replaces it (I think it's ingress go?)

README.md
```diff
@@ -88,6 +88,10 @@ SPANDX_CONFIG=$(pwd)/compliance-backend.js ../insights-proxy/scripts/run.sh
 
 ### Option 3: Docker Development setup
 
+## pre-requisites:
+
+There should be an [upload-service](https://github.com/RedHatInsights/insights-upload) container available in the same network the `rails` container runs.
```
akofink, line 93: It only really requires access to the same kafka instance that an upload instance is producing on.

docker-compose.yml
```diff
@@ -95,7 +97,3 @@ services:
   redis:
     image: redis:latest
 
-networks:
-  default:
-    external:
-      name: docker_default # should match insights-upload's network name
```
akofink, line 101: Without this defined as an external network, we won't be able to access kafka, defined in its own docker-compose, by its service name (i.e. `kafka`). https://github.com/RedHatInsights/insights-ingress-go/blob/master/docker-compose.yml

docker-compose.yml
```diff
@@ -3,6 +3,8 @@ services:
   db:
     image: postgres
     restart: on-failure
+    environment:
+        - POSTGRES_PASSWORD=postgres
```
akofink, line 7: I'm not sure why it's worked this whole time without this. The docs do say it's mandatory. Does it not work for you?

bastilian: This seems to have only changed recently. If the DB was initialised before that it probably is not needed, but bootstrapping fresh, did also require the password for me. https://github.com/docker-library/postgres/commit/f1bc8782e7e57cc403d0b32c0e24599535859f76

akofink:
@Victoremepunto since you're working on podman now, shall we close this one?

Victoremepunto:
> @Victoremepunto since you're working on podman now, shall we close this one?

I'm not sure we don't need fixes for docker-compose, since the podman scripts won't necessarily replace docker-compose immediately, I was discussing with @bastilian about this subject, and I think it is in our best interest to keep maintaining docker-compose for a while, even if we get to a point where we get a functional podman based alternative.

akofink:
Makes sense. There are some unaddressed comments inline.

docker-compose.yml
akofink: Don't we also need to add this env variable as `'POSTGRESQL_ADMIN_PASSWORD'` to the services which require the DB, like `rails`, `sidekiq`, etc.?

Victoremepunto:
Closed by #527

PR #506, opened 2020-06-19
Fix profiles create/delete request/response schema

dLobatog:
Before this PR, the documentation section of Swagger UI was lacking a
proper example and response. Now it's more accurate to what it is meant
to be.

Before - these fields were just `{}`.

After: 

Screenshot from 2020-06-19 09-13-08
Screenshot from 2020-06-19 09-13-21

### DIFF
--- a/spec/api/v1/schemas.rb
+++ b/spec/api/v1/schemas.rb
@@ -31,6 +31,7 @@ module Api
         links: LINKS,
         benchmark: BENCHMARK,
         profile: PROFILE,
+        profile_create: PROFILE_CREATE,
         profile_relationships: PROFILE_RELATIONSHIPS,
         rule_result: RULE_RESULT,
         rule: RULE
--- a/spec/api/v1/schemas.rb
+++ b/spec/api/v1/schemas.rb
@@ -31,7 +31,6 @@ module Api
         links: LINKS,
         benchmark: BENCHMARK,
         profile: PROFILE,
-        profile_create: PROFILE_CREATE,
         profile_relationships: PROFILE_RELATIONSHIPS,
         rule_result: RULE_RESULT,
         rule: RULE
--- a/spec/api/v1/schemas/profiles.rb
+++ b/spec/api/v1/schemas/profiles.rb
@@ -10,10 +10,11 @@ module Api
 
         PROFILE = {
           type: 'object',
-          required: %w[name parent_profile_id],
+          required: %w[parent_profile_id],
           properties: {
             name: {
               type: 'string',
+              example: 'my custom profile'
             },
             parent_profile_id: {
               type: 'string',
@@ -71,40 +72,6 @@ module Api
             os_major_version: {
               type: 'string',
               example: '7'
-            }
-          }
-        }.freeze
-
-        PROFILE_CREATE = {
-          type: :object,
-          required: %w[name parent_profile_id],
-          properties: {
-            name: {
-              type: 'string',
-            },
-            business_objective: {
-              type: 'string',
-              example: 'APAC Expansion',
-              nullable: true
-            },
-            parent_profile_id: {
-              type: 'string',
-              format: 'uuid',
-              nullable: true,
-              example: '0105a0f0-7379-4897-a891-f95cfb9ddf9c',
-              required: true
-            },
-            description: {
-              type: 'string',
-              nullable: true,
-              example: 'This profile contains rules to ensure standard '\
-              'security baseline\nof a Red Hat Enterprise Linux 7 '\
-              'system. Regardless of your system\'s workload\nall '\
-              'of these checks should pass.'
-            },
-            compliance_threshold: {
-              type: 'number',
-              example: 95.0
             }
           }
         }.freeze
--- a/spec/api/v1/schemas/profiles.rb
+++ b/spec/api/v1/schemas/profiles.rb
@@ -10,8 +10,11 @@ module Api
 
         PROFILE = {
           type: 'object',
-          required: %w[parent_profile_id],
+          required: %w[name parent_profile_id],
           properties: {
+            name: {
+              type: 'string',
+            },
             parent_profile_id: {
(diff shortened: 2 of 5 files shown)


spec/api/v1/schemas/profiles.rb
```diff
@@ -10,8 +10,11 @@ module Profiles
 
         PROFILE = {
           type: 'object',
-          required: %w[parent_profile_id],
+          required: %w[name parent_profile_id],
```
akofink, line 13: name isn't actually required. It will be copied from the parent profile if it's not provided.

spec/api/v1/schemas.rb
```diff
@@ -31,6 +31,7 @@ module Schemas
         links: LINKS,
         benchmark: BENCHMARK,
         profile: PROFILE,
+        profile_create: PROFILE_CREATE,
```
akofink, line 34: The schema accepted for profile creation should be identical to the schema returned by a GET request. Some attributes are not able to be set, but the entire set of attributes should be allowed (and are allowed).

spec/api/v1/schemas/profiles.rb
```diff
@@ -10,8 +10,11 @@ module Profiles
...
           properties: {
+            name: {
+              type: 'string',
+            },
```
akofink, line 17: :laughing: whoops :+1: This should also be `required: false`.

spec/api/v1/schemas/types.rb
```diff
@@ -11,19 +11,27 @@ module Types
...
-            type: :string
+          type: :object,
+          properties: {
+            data: {
+              type: :object,
+              properties: {
+                id: ref_schema('uuid')
+              }
+            },
+            type: { type: :string }
```
akofink, line 22: This isn't correct. The type and id should be on the same level like they were:

```
        "relationships": {
            "account": {
                "data": null
            },
            "benchmark": {
                "data": {
                    "id": "c0e2ce89-fb9a-4e29-a51b-23f2398cde83",
                    "type": "benchmark"
                }
            },
            "hosts": {
                "data": []
            },
            "parent_profile": {
                "data": null
            },
            "rules": {
                "data": [
                    {
                        "id": "02f088c1-0287-4580-aa24-9bcf2514a293",
                        "type": "rule"
                    },
                    {
                        "id": "04112d7a-24b6-4231-94e5-9b634dff44b7",
                        "type": "rule"
                    }
                ]
            }
```

spec/api/v1/schemas/types.rb
```diff
@@ -11,19 +11,27 @@ module Types
...
-              properties: {
-                id: ref_schema('uuid'),
-                type: :string
+          type: :object,
+          properties: {
+            data: {
+              type: :array,
+              items: {
+                properties: {
+                  id: ref_schema('uuid')
```
akofink, line 33: This needs the type still.

swagger/v1/openapi.json
```diff
@@ -459,72 +347,46 @@
         "responses": {
           "201": {
             "description": "creates a profile",
-            "examples": {
```
akofink, line 462: There's a bug in rswag that doesn't generate examples. You're supposed to be able to configure it to turn off `--dry-run`, but I haven't been able to get it to work. Instead, you can run the rswag command directly with something like this:

```sh
d-c exec reports-consumer /usr/local/bin/ruby -I/usr/local/bundle/gems/rspec-core-3.9.2/lib:/usr/local/bundle/gems/rspec-support-3.9.3/lib /usr/local/bundle/gems/rspec-core-3.9.2/exe/rspec --pattern spec/requests/\*\*/\*_spec.rb,\ spec/api/\*\*/\*_spec.rb,\ spec/integration/\*\*/\*_spec.rb --format Rswag::Specs::SwaggerFormatter --order defined

```

spec/integration/profiles_spec.rb
```diff
@@ -75,12 +75,29 @@
...
 
-      parameter name: :data, in: :body, required: true, schema: {
+      parameter name: :data, in: :body, schema: {
         type: :object,
-        data: {
-          type: :object,
-          properties: {
-            attributes: ref_schema('profile')
+        properties: {
+          type: { type: :string, example: 'profile' },
```
akofink, line 81: :+1: Good catch. I feel like there are so many place like this that are easy to forget to put something.

swagger/v1/openapi.json
```diff
@@ -572,87 +450,10 @@
...
-              "application/vnd.api+json": {
-                "errors": "Profile not found with ID invalid"
-              }
-            },
             "content": {}
           },
           "200": {
             "description": "retrieves a profile with included benchmark",
-            "examples": {
```
akofink, line 584: It looks like you might still be running rswag with the --dry-run option on. To autogenerate examples, it has to not be `--dry-run`.

dLobatog: I'm actually just running "rake rswag"  - in fact, these deletions seem to happen on master too. I think since on the refactoring no new `openapi.json` was committed, I do see the example for the 200 but not for the 404, let me check how can I get it back

dLobatog: They were removed here https://github.com/RedHatInsights/compliance-backend/pull/492/files

dLobatog: Oh I see the comment to run `/usr/local/bin/ruby -I/usr/local/bundle/gems/rspec-core-3.9.2/lib:/usr/local/bundle/gems/rspec-support-3.9.3/lib /usr/local/bundle/gems/rspec-core-3.9.2/exe/rspec --pattern spec/requests/\*\*/\*_spec.rb,\ spec/api/\*\*/\*_spec.rb,\ spec/integration/\*\*/\*_spec.rb --format Rswag::Specs::SwaggerFormatter --order defined
` - that does make a difference in the schema, but I don't see the examples in the Swagger UI just yet

PR #507, opened 2020-06-23
Allow rules to be tailored on profiles#create API

akofink:
POST body containing rules relationships:

```js
{
	"data": {
		"attributes": {
			"parent_profile_id": "1007b72e-7f0f-486f-911a-5c392c724faa",
			"name": "My tailored profile",
			"compliance_threshold": 95.0
		},
		"relationships": {
			"rules": {
				"data": [
					{
						"type": "rule"
					},
					{
						"id": "06a19f0e-5c7a-4d54-bc66-e932a96bf954",
						"type": "rule"
					},
					{
						"id": "fcd8ee29-5af4-47bf-9e29-2de12f20fe9a",
						"type": "rule"
					}
				]
			}
		}
	}
}

# Response:

HTTP/1.1 201 Created
cache-control: max-age=0, private, must-revalidate
connection: close
content-type: application/vnd.api+json
date: Tue, 23 Jun 2020 15:42:29 GMT
etag: W/"62bba6ee45343fbc2bbe364beb645a35"
transfer-encoding: chunked
vary: Origin
x-request-id: 31f18b51-105c-48b6-8844-880f335f5162
x-runtime: 0.275060

{
    "data": {
        "attributes": {
            "business_objective": null,
            "canonical": false,
            "compliance_threshold": 95.0,
            "compliant_host_count": 0,
            "description": "Ensures PCI-DSS v3.2.1 related security configuration settings are applied.",
            "external": false,
            "name": "My tailored profile",
            "os_major_version": "8",
            "parent_profile_id": "1007b72e-7f0f-486f-911a-5c392c724faa",
            "parent_profile_ref_id": "xccdf_org.ssgproject.content_profile_pci-dss",
            "ref_id": "xccdf_org.ssgproject.content_profile_pci-dss",
            "score": 0,
            "tailored": true,
            "total_host_count": 0
        },
        "id": "679133ea-bbe0-47ed-ae36-4093d4402b03",
        "relationships": {
            "account": {
                "data": {
                    "id": "8c926d2a-17d6-4398-8895-2652971113c1",
                    "type": "account"
                }
            },
            "benchmark": {
                "data": {
                    "id": "c0e2ce89-fb9a-4e29-a51b-23f2398cde83",
                    "type": "benchmark"
                }
            },
            "hosts": {
                "data": []
            },
            "parent_profile": {
                "data": {
                    "id": "1007b72e-7f0f-486f-911a-5c392c724faa",
                    "type": "profile"
                }
            },
            "rules": {
                "data": [
                    {
                        "id": "06a19f0e-5c7a-4d54-bc66-e932a96bf954",
                        "type": "rule"
                    },
                    {
                        "type": "rule"
                    },
                    {
                        "id": "fcd8ee29-5af4-47bf-9e29-2de12f20fe9a",
                        "type": "rule"
                    }
                ]
            },
            "test_results": {
                "data": []
            }
        },
        "type": "profile"
    }
}
```

Note that the shape of the request and response are the same, other than unspecified attributes or relationships in the request.

On create, if no valid rules are specified, the parent profile rules are added to the new profile. Valid rules are those that exist in the parent profile's benchmark only.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,8 +10,7 @@ class ApplicationController < ActionController::API
   include Pagination
   include Search
   include Rendering
-
-  ParamType = ActionController::Parameters
+  include Parameters
 
   def openapi
     send_file Rails.root.join('swagger/v1/openapi.v3.yaml')
@@ -19,12 +18,6 @@ class ApplicationController < ActionController::API
 
   def pundit_scope
     Pundit.policy_scope(current_user, resource)
-  end
-
-  def resource_params
-    params.permit(data: ParamType.map(attributes: ParamType.map))
-    params.require(:data).require(:attributes)
-          .permit(serializer.attributes_to_serialize.keys)
   end
 
   rescue_from Pundit::NotAuthorizedError do
--- a/app/controllers/concerns/parameters.rb
+++ b/app/controllers/concerns/parameters.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+# Reusable parameter checking for all controllers
+module Parameters
+  extend ActiveSupport::Concern
+
+  ParamType = ActionController::Parameters
+
+  RELATIONSHIP_TYPE = ParamType.map(
+    id: ParamType.string,
+    type: ParamType.string
+  )
+
+  included do
+    def resource_params
+      params.permit(data: ParamType.map(
+        attributes: ParamType.map,
+        relationships: ParamType.map
+      ))
+      params.require(:data).permit(attributes: {}, relationships: {})
+    end
+
+    def resource_attributes
+      resource_params[:attributes]&.permit(
+        serializer.attributes_to_serialize.keys
+      )
+    end
+
+    def resource_relationships
+      resource_params[:relationships]&.permit(relationship_types)
+    end
+
+    private
+
+    def relationship_types
+      serializer.relationships_to_serialize.keys
+                .each_with_object({}) do |relationship, h|
+        h[relationship] = ParamType.map(
+          data: ParamType.array(RELATIONSHIP_TYPE) | RELATIONSHIP_TYPE
+        )
+      end
+    end
+  end
+end
--- a/app/controllers/v1/benchmarks_controller.rb
+++ b/app/controllers/v1/benchmarks_controller.rb
@@ -4,23 +4,21 @@ module V1
   # API for Benchmarks
   class BenchmarksController < ApplicationController
     def index
-      render json: benchmarks
+      render_json benchmarks
     end
 
     def show
-      render json: benchmark
+      render_json benchmark
     end
 
     private
 
     def benchmarks
-      @benchmarks ||= BenchmarkSerializer.new(authorize(scope_search), metadata)
+      @benchmarks ||= authorize(scope_search)
     end
 
     def benchmark
-      @benchmark ||= BenchmarkSerializer.new(
-        authorize(resource.find(params[:id]))
-      )
+      @benchmark ||= authorize(resource.find(params[:id]))
     end
 
     def scope
@@ -29,6 +27,10 @@ module V1
 
     def resource
       Xccdf::Benchmark
+    end
+
+    def serializer
+      BenchmarkSerializer
     end
   end
 end
--- a/app/controllers/v1/profiles_controller.rb
+++ b/app/controllers/v1/profiles_controller.rb
@@ -27,6 +27,7 @@ module V1
       profile = Profile.new(profile_create_attributes).fill_from_parent
 
       if profile.save
+        profile.update_rules(ids: new_rule_ids)
         render_json profile, status: :created
       else
         render_error profile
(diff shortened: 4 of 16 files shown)


app/controllers/application_controller.rb
```diff
@@ -21,12 +20,6 @@ def pundit_scope
     Pundit.policy_scope(current_user, resource)
   end
 
-  def resource_params
-    params.permit(data: ParamType.map(attributes: ParamType.map))
-    params.require(:data).require(:attributes)
-          .permit(serializer.attributes_to_serialize.keys)
-  end
-
```
akofink, line 29: Moved to `Parameters` concern

app/controllers/concerns/parameters.rb
```diff
@@ -0,0 +1,44 @@
...
+# Reusable parameter checking for all controllers
+module Parameters
+  extend ActiveSupport::Concern
+
+  ParamType = ActionController::Parameters
+
+  RELATIONSHIP_TYPE = ParamType.map(
+    id: ParamType.string,
+    type: ParamType.string
+  )
```
akofink, line 12: Type validations thus far are for request shape only (i.e. data, attributes, relationships are all `ParamType.map`) and not individual attribute types with the exception of type/id here which holds for any relationship.

app/controllers/v1/benchmarks_controller.rb
```diff
@@ -30,5 +28,9 @@ def scope
     def resource
       Xccdf::Benchmark
     end
+
+    def serializer
+      BenchmarkSerializer
+    end
```
akofink, line 34: Changes in this file are to allow users to see which rules exist on a benchmark, via its relationships and optional includes, so they can select rules for tailoring a profile.

app/controllers/v1/profiles_controller.rb
```diff
@@ -27,6 +27,7 @@ def create
       profile = Profile.new(profile_create_attributes).fill_from_parent
 
       if profile.save
+        profile.update_rules(ids: new_rule_ids)
```
akofink, line 30: This no longer happens on save automatically.

app/models/profile.rb
```diff
@@ -84,31 +83,37 @@ def in_account(account)
...
+  def update_rules(ids: nil, ref_ids: nil)
+    ids_to_add = rule_ids_to_add(ids, ref_ids)
+
+    ProfileRule.where(rule_id: rule_ids - ids_to_add, profile_id: id)
+               .destroy_all
+
+    ProfileRule.import!(ids_to_add.map do |rule|
+      ProfileRule.new(profile_id: id, rule_id: rule.id)
+    end)
+  end
```
akofink, line 100: Renamed from `add_rules`. This method now intelligently updates the rules on the profile and will remove rules that aren't specified. Also, `ignore: true` was removed from the import, as we don't expect duplicates or any other errors anymore.

config/initializers/wrap_parameters.rb
```diff
@@ -7,7 +7,7 @@
 
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json]
+  wrap_parameters format: []
```
akofink, line 10: Before this change, we would always get an extra param sent to the controller named after the resource (i.e. `params[:profile]`). That doesn't fit with our JSON spec where everything is wrapped in `params[:data]`.

app/serializers/benchmark_serializer.rb
```diff
@@ -4,4 +4,6 @@
 class BenchmarkSerializer
   include FastJsonapi::ObjectSerializer
   attributes :ref_id, :title, :version, :description
+  has_many :rules
+  has_many :profiles
```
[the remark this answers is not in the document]
akofink: This should be limited by the pundit policy, but it's a good thing for us to double check, for sure. It makes sense for a user to want to see all profiles, even non-canonical, associated to a specific benchmark in their account, but this API will most likely be used for showing canonical profiles, so perhaps we should add a default `canonical = true` search to the benchmarks controller. What do you think?

app/models/profile.rb
dLobatog: This is actually pretty useful, I am working on making the frontend Edit Policy button bring up the wizard to allow re-tailoring and this comes in handy

spec/integration/profiles_spec.rb
```diff
@@ -97,6 +100,14 @@
               'of these checks should pass.',
               compliance_threshold: 95.0,
               business_objective: 'APAC Expansion'
+            },
+            relationships: {
```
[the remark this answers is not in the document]
akofink: Added. :+1:

app/serializers/benchmark_serializer.rb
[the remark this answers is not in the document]
akofink: Nice catch, @dLobatog - this was indeed showing profiles from other user accounts. Now I've limited it by pundit policy. We'll have to be careful in the future to do this for relationships in serializers.

akofink:
Updated! I think all comments so far have been addressed.

app/serializers/benchmark_serializer.rb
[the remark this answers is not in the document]
dLobatog: Definitely, top-level objects are properly filtered via pundit easily but serialized relations have to be also filtered

PR #508, opened 2020-06-29
RHICOMPL-748 - Support no os_release in inventory

dLobatog:
Currently, when someone assigns a new system that doesn't exist in compliance to a policy, if the system doesn't have a os_release (which happens often on testing environments), we get the following error:
`NoMethodError: undefined method `split' for nil:NilClass`

This PR prevents that error from happening.

### DIFF
--- a/app/services/host_inventory_api.rb
+++ b/app/services/host_inventory_api.rb
@@ -41,7 +41,7 @@ class HostInventoryAPI
       X_RH_IDENTITY: @b64_identity
     )
     JSON.parse(response.body)['results'].inject([]) do |acc, host|
-      os_major, os_minor = host['system_profile']['os_release'].split('.')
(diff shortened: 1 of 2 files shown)


app/services/host_inventory_api.rb
```diff
@@ -50,6 +50,12 @@ def system_profile(ids)
 
   private
 
+  def find_os_release(system_profile)
+    return [nil, nil] if system_profile['os_release'].blank?
+
+    system_profile['os_release'].split('.')
```
akofink, line 56: How sure are we that `os_release` will exist?

bastilian: Line 54 should prevent that, or?

akofink: Oh yeah, true. :)

PR #509, opened 2020-06-30
RHICOMPL-750 - Remove external profiles after system culling

dLobatog:
When the following happens:

 - A host is registered and assigned to an external policy
 - The host is later on removed via System culling

The host is removed, but the external policy remains. This is
not only problematic because of the unused cruft in the
database, but also if the user tries to create a new Policy now
with the same ref ID, they won't be able to do it (422 error),
and they won't be able to remove the old Policy except through
the API (which is not in prod yet)

### DIFF
--- a/test/models/profile_host_test.rb
+++ b/test/models/profile_host_test.rb
@@ -23,17 +23,31 @@ class ProfileHostTest < ActiveSupport::TestCase
     assert_equal 0, DeleteHost.jobs.size
   end
 
-  test 'destory associated external policies if they have no more hosts' do
+  test 'destroy associated external policies if they have no more hosts' do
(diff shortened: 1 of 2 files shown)


test/models/profile_host_test.rb
```diff
@@ -22,4 +22,18 @@ class ProfileHostTest < ActiveSupport::TestCase
     assert_empty Host.where(id: host.id)
     assert_equal 0, DeleteHost.jobs.size
   end
+
+  test 'destory associated external policies if they have no more hosts' do
```
akofink, line 26:
```suggestion
  test 'destroy associated external policies if they have no more hosts' do
```

PR #510, opened 2020-07-06
Allow updating profiles

akofink:
`PATCH profiles/<uuid>`

```js
{
	"data": {
		"attributes": {
			"name": "My updated tailored profile",
			"compliance_threshold": 90
		},
		"relationships": {
			"rules": {
				"data": [
					{
						"type": "rule"
					}
				]
			}
		}
	}
}
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,6 +20,12 @@ class ApplicationController < ActionController::API
     Pundit.policy_scope(current_user, resource)
   end
 
+  rescue_from ActiveRecord::RecordNotUnique do |error|
+    render json: {
+      errors: "Duplicate record: #{error.message[/Key \(.+\).+\./]}"
+    }, status: :conflict
+  end
+
   rescue_from Pundit::NotAuthorizedError do
     render json: { errors: 'You are not authorized to access this action.' },
            status: :forbidden
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -89,20 +89,27 @@ class Profile < ApplicationRecord
   alias os_major_version major_os_version
 
   def update_rules(ids: nil, ref_ids: nil)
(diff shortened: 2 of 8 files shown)


app/models/profile.rb
```diff
@@ -89,20 +89,27 @@ def major_os_version
   alias os_major_version major_os_version
 
   def update_rules(ids: nil, ref_ids: nil)
```
akofink, line 91: This method wasn't quite right - it wasn't able to add a rule and delete a rule at the same time. There's a difference between rule_ids_to_add (just the additional rules) and new_rules (the complete set of rules the profile should have after this update). We want to delete `rule_ids - new_rules`, not `rule_ids - rule_ids_to_add`.

test/models/profile_test.rb
```diff
@@ -265,6 +265,51 @@ class ProfileTest < ActiveSupport::TestCase
     end
   end
 
+  context 'update_rules' do
```
akofink, line 268: Now with tests! Not sure why we left these off when the method was added.

app/controllers/application_controller.rb
```diff
@@ -20,6 +20,12 @@ def pundit_scope
     Pundit.policy_scope(current_user, resource)
   end
 
+  rescue_from ActiveRecord::RecordNotUnique do |error|
+    render json: {
+      errors: "Duplicate record: #{error.message[/Key \(.+\).+\./]}"
+    }, status: :conflict
+  end
```
akofink, line 27: This is a possibility for any controllers using ActiveRecord `import!`, though we really never expect it for `update_rules` since we provide a specific list of rules, those that don't exist on the profile already and are in the profile's benchmark. Still, this rescue may prove useful to another controller.

akofink:
Thanks for the review!

PR #511, opened 2020-07-06
[RHICOMPL-467] Allow host/profile association via API V1

akofink:
Works the same as the rules relationship:

```js
PATCH api/v1/profiles/{uuid}
{
	"data": {
		"attributes": {
			"name": "My updated tailored profile",
			"compliance_threshold": 90
		},
		"relationships": {
			"rules": {
				"data": [
					{
						"type": "rule"
					},
					{
						"id": "06a19f0e-5c7a-4d54-bc66-e932a96bf954",
						"type": "rule"
					},
					{
						"id": "fcd8ee29-5af4-47bf-9e29-2de12f20fe9a",
						"type": "rule"
					}
				]
			},
			"hosts": {
				"data": [
					{
						"type": "host"
					}
				]
			}
		}
	}
}
```

Also works with POST/create.

Based on changes in https://github.com/RedHatInsights/compliance-backend/pull/510 - I will rebase this PR once 510 is merged.

### DIFF
--- a/app/controllers/v1/profiles_controller.rb
+++ b/app/controllers/v1/profiles_controller.rb
@@ -12,13 +12,16 @@ module V1
       name description compliance_threshold business_objective
     ].freeze
 
+    before_action only: %i[show update] do
+      authorize profile
+    end
+
     def index
       params[:search] ||= 'external=false and canonical=false'
       render_json scope_search.sort_by(&:score)
     end
 
     def show
-      authorize profile
       render_json profile
     end
 
@@ -31,6 +34,7 @@ module V1
       profile = Profile.new(profile_create_attributes).fill_from_parent
 
       if profile.save
+        profile.update_hosts(new_host_ids)
         profile.update_rules(ids: new_rule_ids)
         render_json profile, status: :created
       else
@@ -39,9 +43,8 @@ module V1
     end
 
     def update
-      authorize profile
-
       if profile.update(profile_update_attributes)
+        profile.update_hosts(new_host_ids)
         profile.update_rules(ids: new_rule_ids)
         render_json profile
       else
@@ -77,7 +80,7 @@ module V1
     def business_objective
       @business_objective ||= Pundit.policy_scope(
         current_user, BusinessObjective
-      ).from_title(resource_attributes[:business_objective])
+      ).from_title(resource_attributes&.dig(:business_objective))
     end
 
     def profile_create_attributes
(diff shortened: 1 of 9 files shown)


app/models/concerns/profile_rules.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# Methods that are related to a profile's rules
+module ProfileRules
```
akofink, line 4: This and ProfileHosts modules were separated out of the Profile model due to class length restrictions. Also, there's quite a bit of logic here for ProfileRules; it's nice having it in one place.

app/controllers/v1/profiles_controller.rb
```diff
@@ -27,13 +34,24 @@ def create
...
       end
     end
 
+    def update
+      if profile.update(profile_update_attributes)
+        profile.update_hosts(new_host_ids)
+        profile.update_rules(ids: new_rule_ids)
+        render_json profile
+      else
+        render_error profile
```
dLobatog, line 51: I believe this should be rescuing any `StandardError` in here - e.g, if `profile.update` succeeds but either of `profile.update_hosts` or `profile.update_rules` fail, it's a 500 instead of a captured error that we can return to the user.

app/controllers/v1/profiles_controller.rb
```diff
@@ -8,13 +8,20 @@ class ProfilesController < ApplicationController
       parent_profile_id
     ].freeze
 
+    ALLOWED_UPDATE_ATTRIBUTES = %i[
+      name description compliance_threshold business_objective
+    ].freeze
+
+    before_action only: %i[show update] do
+      authorize profile
```
dLobatog, line 16: Neat!

app/controllers/v1/profiles_controller.rb
akofink: True, but how exactly might `update_hosts` or `update_rules` fail? If we're catching exceptions, we have to figure out which specific exceptions are valid and rescue those errors. For example, `import!` might raise a `ActiveRecord::RecordNotUnique`, though we've implemented it in a way to always avoid that scenario. I've added a global `rescue_from` in the application_controller in this PR to address that: https://github.com/RedHatInsights/compliance-backend/pull/510/files#r451533427 Any other errors?

akofink:
Conflicts should be resolved easier once we merge https://github.com/RedHatInsights/compliance-backend/pull/510/, since this PR is based on update already existing.

PR #515, opened 2020-07-07
Match serializers to openapi spec

akofink:
This revisits all controllers, serializers, specs, schemas, etc. for API V1. So far we have implemented profiles and benchmarks. This matches that work to the other V1 resources.

### DIFF
--- a/app/controllers/v1/rule_results_controller.rb
+++ b/app/controllers/v1/rule_results_controller.rb
@@ -4,10 +4,14 @@ module V1
   # API for RuleResults
   class RuleResultsController < ApplicationController
     def index
-      render json: RuleResultSerializer.new(scope_search, metadata)
+      render_json scope_search
     end
 
     private
+
+    def serializer
+      RuleResultSerializer
+    end
 
     def resource
       RuleResult
--- a/app/controllers/v1/rules_controller.rb
+++ b/app/controllers/v1/rules_controller.rb
@@ -4,7 +4,7 @@ module V1
   # REST API for Rules
   class RulesController < ApplicationController
     def index
-      render json: RuleSerializer.new(scope_search, metadata)
+      render_json scope_search
     end
 
     def show
@@ -16,7 +16,7 @@ module V1
 
       raise ActiveRecord::RecordNotFound if rule.blank?
 
-      render json: RuleSerializer.new(rule)
+      render_json rule
     end
 
     private
@@ -25,12 +25,16 @@ module V1
       Rule
     end
 
+    def serializer
+      RuleSerializer
+    end
+
     def search_by_id
-      ::Pundit.policy_scope(User.current, ::Rule).friendly.find(params[:id])
+      pundit_scope.friendly.find(params[:id])
     end
 
     def search_by_ref_id
-      rule = Rule.latest.where(
+      rule = pundit_scope.latest.where(
         'rules.slug LIKE ?',
         "%#{ActiveRecord::Base.sanitize_sql_like(params[:id])}%"
       )
--- a/app/graphql/types/system.rb
+++ b/app/graphql/types/system.rb
@@ -43,10 +43,7 @@ module Types
     end
 
     def last_scanned(args = {})
-      latest_test_result = ::TestResult.latest.find_by(
-        profile_id: args[:profile_id], host_id: object.id
-      )
-      latest_test_result&.end_time&.iso8601 || 'Never'
+      object.last_scanned(args)
     end
 
     private
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -3,7 +3,7 @@
 # Host representation in insights compliance backend. Most of the times
 # these hosts will also show up in the insights-platform host inventory.
 class Host < ApplicationRecord
-  scoped_search on: %i[id name account_id os_major_version os_minor_version]
+  scoped_search on: %i[name os_major_version os_minor_version]
   scoped_search on: :compliant, ext_method: 'filter_by_compliance',
                 only_explicit: true
   scoped_search on: :compliance_score, ext_method: 'filter_by_compliance_score',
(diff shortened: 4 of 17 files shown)


app/controllers/v1/rule_results_controller.rb
```diff
@@ -4,11 +4,15 @@ module V1
   # API for RuleResults
   class RuleResultsController < ApplicationController
     def index
-      render json: RuleResultSerializer.new(scope_search, metadata)
+      render_json scope_search
```
akofink, line 7: `render_json` in rendering.rb gives automatic side loaded includes (the `include` parameter), metadata, and error handling.

app/controllers/v1/rules_controller.rb
```diff
@@ -25,12 +25,16 @@ def resource
...
+    end
+
     def search_by_id
-      ::Pundit.policy_scope(User.current, ::Rule).friendly.find(params[:id])
+      pundit_scope.friendly.find(params[:id])
     end
 
     def search_by_ref_id
-      rule = Rule.latest.where(
+      rule = pundit_scope.latest.where(
```
akofink, line 37: This change caused a test to fail - I'm quite sure we always want to filter records through the pundit policy. In this case, canonical rules are intended to be returned (canonical rules are those assigned to any canonical profile). The rule pundit policy allows this.

app/graphql/types/system.rb
```diff
@@ -43,10 +43,7 @@ def rules_failed(args = {})
     end
 
     def last_scanned(args = {})
-      latest_test_result = ::TestResult.latest.find_by(
-        profile_id: args[:profile_id], host_id: object.id
-      )
-      latest_test_result&.end_time&.iso8601 || 'Never'
+      object.last_scanned(args)
```
akofink, line 46: Moved to the model for shared use with API V1.

app/models/host.rb
```diff
@@ -3,7 +3,7 @@
 # Host representation in insights compliance backend. Most of the times
 # these hosts will also show up in the insights-platform host inventory.
 class Host < ApplicationRecord
-  scoped_search on: %i[id name account_id os_major_version os_minor_version]
```
akofink, line 6: scoped_search on id isn't something other models allow. I'm not quite sure why it'd be useful when there's a `show` endpoint. There wouldn't ever be a reason to use a 'like' search (`id ~ foo`) on UUIDs :shrug:

app/serializers/host_serializer.rb
```diff
@@ -4,5 +4,12 @@
...
   set_type :host
-  attributes :name, :profiles, :compliant
+  attributes :name, :os_major_version, :os_minor_version, :last_scanned,
+             :rules_passed, :rules_failed
+  attribute :compliant do |host|
+    host.compliant.values.all?
+  end
+
+  has_many :test_results
+  has_many :profiles
```
akofink, line 14: My intention here is to try and match the info returned by the related graphql system type.

app/serializers/rule_serializer.rb
```diff
@@ -3,6 +3,9 @@
 # JSON API serialization for an OpenSCAP Rule
 class RuleSerializer
   include FastJsonapi::ObjectSerializer
-  attributes :created_at, :updated_at, :ref_id, :title, :rationale,
-             :description, :severity, :slug
```
akofink, line 7: created_at/updated_at isn't returned by any other serializers. IMO we could return it for all resources or none.

app/serializers/host_serializer.rb
bastilian: Is there maybe a way that we can abstract the definitions to use in both and keep their responses in sync?

spec/api/v1/schemas/hosts.rb
```diff
@@ -1,23 +1,57 @@
...
+
 module Api
   module V1
     module Schemas
       module Hosts
+        extend Api::V1::Schemas::Util
+
         HOST = {
           type: 'object',
-          required: %w[name account_id],
```
akofink, line 9: account_id is an odd thing to show users IMO. Everything in our API is scoped to the current account.

test/controllers/v1/rules_controller_test.rb
```diff
@@ -41,9 +41,12 @@ class RulesControllerTest < ActionDispatch::IntegrationTest
...
     end
 
-    test 'finds rules not within the user scope but within latest' do
+    test 'finds latest canonical rules' do
+      assert(profiles(:one).canonical?)
+      profiles(:one).rules << rules(:two)
+      assert(rules(:two).canonical?)
       assert_includes(Rule.latest, rules(:two))
-      assert_not_includes(::Pundit.policy_scope(users(:test), ::Rule),
+      assert_not_includes(accounts(:test).profiles.map(&:rules).uniq,
```
akofink, line 49: Here's the failing rule test due to adding the pundit policy for finding canonical rules by slug.

spec/api/v1/schemas/hosts.rb
bastilian: Agreed, we should ask @Victoremepunto to confirm this is not used in any tests QA runs.

app/serializers/host_serializer.rb
akofink: It's a good goal, but I'm not sure how we'd achieve it. For example, graphql allows fields with args, something we can't do with REST. Also, I'm not sure we want to lock graphql to REST because the UI depends on graphql - it's already useful and may be useful in the future to have graphql information available to the UI in some format the doesn't make sense in the REST API.

spec/api/v1/schemas/hosts.rb
Victoremepunto: Doesn't rings a bell, I don't think this is going to impact us. thanks for the heads up.

PR #516, opened 2020-07-07
Update stable

akofink:
Please double check the expected bug fixes are included here.


akofink:
This needs an ACK to merge

PR #526, opened 2020-07-14
Ensure all consumers log all messages received

akofink:
While debugging our consumers recently, I realized that not all messages are logged; sometimes validations are performed first, which makes it difficult to debug what went wrong. Shared logic like logging should be part of the ApplicationConsumer.

This is fine to merge as is, but I may add to it as I continue to debug the consumers.

### DIFF
--- a/app/consumers/application_consumer.rb
+++ b/app/consumers/application_consumer.rb
@@ -2,4 +2,14 @@
 
 # Parent class for all Racecar consumers, contains general logic
 class ApplicationConsumer < Racecar::Consumer
+  def process(message)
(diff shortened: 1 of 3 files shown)


akofink:
Merging! This should help debug the consumers in CI.

Victoremepunto:
:thinking:  I wonder if this will make the logging grow in size like crazy... are the files included in the messages ?

akofink:
@Victoremepunto no, just the URL to the file is included. We already log this message if the metadata contains an ID - this now logs the message even if no ID is present.

PR #527, opened 2020-07-20
Add inventory, ingress, kafka, etc. to devel

akofink:
This PR adds necessary services to the compliance devel setup including ingress, inventory, puptoo, kafka, zookeeper, minio (+createbuckets), and inventory-web.


docker-compose.yml
```diff
@@ -1,10 +1,134 @@
...
+      - REJECTION_TOPIC=platform.upload.validation
+      - LOGLEVEL=INFO
+    depends_on:
+      - kafka
+  minio:
+    image: minio/minio
+    command: server /data
+    volumes:
+      - ~/mnt/data:/data:z
+      - ~/mnt/config:/root/.minio:z
```
bastilian, line 38: It should be possible to run minio without these volumes mounted, like with the postgres service.

akofink: When createbuckets runs in its own container, doesn't it need to share a volume with this minio container? If you remove the volume mounts from both services and run `d-c up`, visit `localhost:9000`, do you see the buckets created?

bastilian: You are right, the do need to share the volume. I may have run the create bucket commands manually.
Can we try to use named volumes though?

akofink: Sure, do you mind if we push that off to another PR? I'm trying to get this working for now to fix uploads in CI.

bastilian: 👍

docker-compose.yml
```diff
@@ -1,10 +1,161 @@
...
+    environment:
+      - REJECTION_TOPIC=platform.upload.validation
+      - LOGLEVEL=INFO
+    depends_on:
+      - kafka
+  minio:
+    image: minio/minio
+    command: server /data
+    volumes:
+      - /data
```
akofink, line 37: @bastilian here you go - it turned out to just work for me :)

akofink:
@bastilian this works for me know end to end (tested with docker-compose, not yet podman-compose). Give it a go and let me know what you think. Thanks!

config/settings/development.yml
```diff
@@ -1,6 +1,7 @@
-host_inventory_url: https://insights-inventory-platform-ci.ask-on-slack-for-the-inventory-devel-url
-rbac_url: https://insights-inventory-platform-ci.ask-on-slack-for-the-rbac-devel-url
-#prometheus_exporter_host: prometheus
-#prometheus_exporter_port: 9394
+host_inventory_url: http://inventory-web:8081
+remediations_url: https://ci.cloud.redhat.com
+rbac_url: https://ci.cloud.redhat.com
```
akofink, line 3: idk how bad this is, but if we want it to work out of the box, this is required.

config/settings/development.yml
```diff
@@ -1,6 +1,7 @@
-host_inventory_url: https://insights-inventory-platform-ci.ask-on-slack-for-the-inventory-devel-url
-rbac_url: https://insights-inventory-platform-ci.ask-on-slack-for-the-rbac-devel-url
-#prometheus_exporter_host: prometheus
-#prometheus_exporter_port: 9394
+host_inventory_url: http://inventory-web:8081
```
akofink, line 1: At least inventory is local now :)

docker-compose.yml
```diff
@@ -1,10 +1,162 @@
+# This docker-compose file will stand up the entire stack including ingress and inventory.
 version: "3"
 services:
+  zookeeper:
+    image: confluentinc/cp-zookeeper
+    environment:
+      - ZOOKEEPER_CLIENT_PORT=32181
+      - ZOOKEEPER_SERVER_ID=1
+      - KAFKA_OPTS='-Dzookeeper.admin.enableServer=false'
```
Victoremepunto, line 9: I had to remove the quotes, otherwise the variable will keep the quotes (at least if using `podman-compose`)

akofink:
We could disable RBAC, but I'm not convinced that it'd give us much since the credentials are shared with inventory and remediations. To get them all working, you need to add basic auth to [platform.rb](https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/platform.rb#L23), like this:

```rb
f.basic_auth 'username', 'password'
```

If we disable RBAC, then I kind of worry about not catching RBAC bugs in development.

Victoremepunto:
> We could disable RBAC, but I'm not convinced that it'd give us much since the credentials are shared with inventory and remediations. To get them all working, you need to add basic auth to [platform.rb](https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/platform.rb#L23), like this:
> 
> ```ruby
> f.basic_auth 'username', 'password'
> ```
> 
> If we disable RBAC, then I kind of worry about not catching RBAC bugs in development.

Can't we just then reploy the RBAC service as well as part of the deployment process ? maybe it doesn't bring any added value from using the existing one from the target environment , but if we want to deploy "everything" locally I think that includes RBAC, we do deploy RBAC during smoke tests fwiw.

akofink:
This PR isn't about deploying everything locally, just those things that are necessary for local development. For example, we need local inventory now because report parsing messages are now produced by the inventory instead of ingress. I can add a commented out variable for disabling RBAC that way it's clear how to do it.

akofink:
@Victoremepunto I updated this based on your requested changes. Good to merge? We can always go back and improve this more.

docker-compose.yml
```diff
@@ -1,10 +1,162 @@
...
+      - kafka
+  minio:
+    image: minio/minio
+    command: server /data
+    volumes:
+      - /data
+    ports:
+      - 9000:9000
+    environment:
+      - MINIO_ACCESS_KEY
```
Victoremepunto, line 42: It wouldn't hurt to note somewhere you need to export this variables to the environment previous to bring the entire thing up. As this is meant for a local development environment, maybe we can just set a default value for both the access and the secret keys ? [Ingress already does this](https://github.com/RedHatInsights/insights-ingress-go/blob/master/.env#L3-L4)

Victoremepunto:
> @Victoremepunto I updated this based on your requested changes. Good to merge? We can always go back and improve this more.

I don't see [the changes we discussed for the quotes in zookeeper](https://github.com/RedHatInsights/insights-ingress-go/blob/master/.env#L3-L4), and I'm also missing yesterday's last minute change for `createbuckets` container to gracefully wait for Minio to be ready. 

Other than that it looks good enough to me :)

docker-compose.yml
bastilian: +1 It would be nice to have a `.env.example` file as well with defaults set and mentioned in the readme to copy or touch a empty `.env`.

docker-compose.yml
```diff
@@ -1,10 +1,162 @@
...
+      - MINIO_SECRET_KEY
+  createbuckets:
+    image: minio/mc
+    depends_on:
+      - minio
+    entrypoint: /bin/sh
+    command: -c '
+      /bin/sleep 5;
```
Victoremepunto, line 50: As discussed off-line, probably we should wait for Minio to be ready in a smarter fashion:

```
until /usr/bin/mc config host add myminio http://minio:9000 $MINIO_ACCESS_KEY $MINIO_SECRET_KEY >/dev/null; do sleep 1; done
```

bastilian:
Saw earlier that we actually do have a `.env` already in the repo, when I added the Minio vars.
That and adding `links:` to services in addition to the `depends_on:` made it work better.
It might just need the links section, which should at least have to wait till a container is up to have the right IP to put into the containers hosts file.

docker-compose.yml
akofink: @Victoremepunto I tried this yesterday many different ways to get it working but I couldn't figure out how to escape a question mark `?` in the shell expression. We need it to check the return code, `$?`. For now, I've just increased the sleep time to 10.

.env.example
```diff
@@ -5,3 +5,5 @@ SETTINGS__PROMETHEUS_EXPORTER_HOST=prometheus.compliance-test.svc
 APPLICATION_TYPE=test
 HOSTNAME=testmachine
 OPENSHIFT_BUILD_NAME=testmachine-test-42
+export MINIO_ACCESS_KEY=BQA2GEXO711FVBVXDWKM
+export MINIO_SECRET_KEY=uvgz3LCwWM3e400cDkQIH/y1Y4xgU4iV91CwFSPC
```
akofink, line 9: Added an .env.example and removed the .env file from the repo. It's already part of the .gitignore.

docker-compose.yml
```diff
@@ -1,10 +1,185 @@
+# This docker-compose file will stand up the entire stack including ingress and inventory.
 version: "3"
 services:
+  zookeeper:
+    image: confluentinc/cp-zookeeper
+    environment:
+      - ZOOKEEPER_CLIENT_PORT=32181
+      - ZOOKEEPER_SERVER_ID=1
+      - KAFKA_OPTS=-Dzookeeper.admin.enableServer=false
```
akofink, line 9: No more quotes here. Sorry for missing this change before.

docker-compose.yml
Victoremepunto: we don't need that, I tried this and it worked for me. can you test this ? 

here's the snippet:
```
    command: -c '
      until /usr/bin/mc config host add myminio http://minio:9000 $MINIO_ACCESS_KEY $MINIO_SECRET_KEY >/dev/null; do sleep 1; done;
      /usr/bin/mc mb myminio/insights-upload-perma;
      /usr/bin/mc mb myminio/insights-upload-rejected;
      /usr/bin/mc policy set download myminio/insights-upload-perma;
      /usr/bin/mc policy set download myminio/insights-upload-rejected;
      exit 0;'
    volumes:

```

README.md
```diff
@@ -86,9 +86,48 @@ using insights-proxy. You may run the proxy using the SPANDX config provided her
 SPANDX_CONFIG=$(pwd)/compliance-backend.js ../insights-proxy/scripts/run.sh
 ```
 
-### Option 3: Docker Development setup
```
akofink, line 89: I've also updated the readme with some more details about this environment.

bastilian:
Awesome! @akofink++!

README.md
Victoremepunto: thanks @akofink

akofink:
Your wait script in createbuckets works as expected for me, @Victoremepunto. :+1: PR is updated

docker-compose.yml
```diff
@@ -1,10 +1,186 @@
...
+  createbuckets:
+    image: minio/mc
+    depends_on:
+      - minio
+    links:
+      - minio
+    entrypoint: /bin/sh
+    command: -c '
+      until /usr/bin/mc config host add myminio http://minio:9000 $MINIO_ACCESS_KEY $MINIO_SECRET_KEY >/dev/null; do sleep 1; done ;
+      /usr/bin/mc config host add myminio http://minio:9000 $MINIO_ACCESS_KEY $MINIO_SECRET_KEY ;
```
Victoremepunto, line 58: I think this line is not needed, the `until` clause should've covered it

akofink: Fixed. Thanks!

config/settings.yml
```diff
@@ -1,3 +1,6 @@
+app_name: 'compliance'
+path_prefix: '/api'
+old_path_prefix: '/r/insights/platform'
```
akofink, line 3: These defaults make sense for all environments. They are still customizable via env variables, like `SETTINGS__APP_NAME` if required.

akofink:
Since the .env file is now not included in the repo, CI was failing. Now, I've updated this PR to use rails settings for variables that we need defined, with defaults in config/settings.yml. Everywhere that was reading those values as raw ENV variables have been updated to use `Settings`.

PR #529, opened 2020-07-28
Ignore scap-security-guide content in the repo

akofink:
The ssg:sync_rhel rake tasks lays down SSG files for RHEL 6, 7, and 8.
We won't ever want to commit them to the repo unless as part of the
tests.


bastilian:
It would make sense to change the import to put the `scap-security-guide` files into the `./tmp` folder.

PR #531, opened 2020-07-29
Increase ingress maxsize to 100MB, up from 10MB, in devel

akofink:
Reports are often around 11MB and fail to upload locally. This matches
what ingress uses in its e2e-deploy template


akofink:
It looks like all deployments to openshift default to 100MB, so we're good there.

PR #532, opened 2020-07-30
Improve parsing performance

akofink:
This change essentially boils down to removing N `selects` for N / 1000 (due to batch size). In order to accurately parse a report, we must determine if the records already exist in the DB. Before, we were using `find_or_initialize_by` for each RuleReference and RuleReferencesRule (37K records just for this table in the RHEL 7 0.1.46 SSG). Now we are executing a where query to determine existing records and only importing new records.

Importing RHEL 7 SSG, version 0.1.46, initial and subsequent imports. Initial import takes about half the time:

```
rake ssg:import DATASTREAM_FILE='./scap-security-guide-0.1.46/ssg-rhel7-ds.xml'

# Before

Importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml at 2020-07-30 04:14:25 UTC
Finished importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml in 86.885234702 seconds.

Importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml at 2020-07-30 04:16:10 UTC
Finished importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml in 7.394536394 seconds.

# After

Importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml at 2020-07-30 04:10:54 UTC
Finished importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml in 44.367159831 seconds.

Importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml at 2020-07-30 04:12:20 UTC
Finished importing ./scap-security-guide-0.1.46/ssg-rhel7-ds.xml in 7.29955355 seconds.
```

There is more room for improvement during parsing. The same thing done here with RuleReferences and RuleReferencesRules, our biggest tables, can be done for Rules as well. 

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -3,4 +3,21 @@
 # Abstract record class to be applied to all models
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  def self.arel_find(collection, batch_size = 1000)
+    base_arel = arel_table[column_names.first].eq('')
+    collection.in_groups_of(batch_size, false).map do |batch|
+      where(batch.inject(base_arel) do |rel, record|
+        rel.or(yield(record))
+      end)
+    end.inject([], &:+)
+  end
+
+  def ==(other)
+    return super if self.class.column_names.include? 'id'
+
+    self.class.column_names.map do |col|
+      send(col) == other.send(col)
+    end.all?
+  end
 end
--- a/app/services/concerns/xccdf/rule_references_rules.rb
+++ b/app/services/concerns/xccdf/rule_references_rules.rb
@@ -6,21 +6,54 @@ module Xccdf
     extend ActiveSupport::Concern
 
     included do
-      def save_rule_references_rules
-        @rule_references_rules ||= @op_rules.flat_map do |op_rule|
-          rule_references_for(op_rule: op_rule).map do |reference|
-            ::RuleReferencesRule
-              .find_or_initialize_by(rule_reference_id: reference.id,
-                                     rule_id: rule_for(ref_id: op_rule.id).id)
-          end
-        end
+      RuleReferencesRuleStruct = Struct.new(:rule_id, :rule_reference_id)
 
-        ::RuleReferencesRule.import!(
-          @rule_references_rules.select(&:new_record?), ignore: true
-        )
+      def save_rule_references_rules
+        @rule_references_rules = new_rule_references_rules +
+                                 existing_rule_references_rules
+
+        ::RuleReferencesRule.import!(new_rule_references_rules)
       end
 
       private
+
+      def op_rule_references_rules
+        @op_rule_references_rules ||= @op_rules.flat_map do |op_rule|
+          rule = rule_for(ref_id: op_rule.id)
+          rule_references_for(op_rule: op_rule).map do |rule_reference|
+            RuleReferencesRuleStruct.new(rule.id, rule_reference.id)
+          end
+        end
+      end
+
+      def existing_rule_references_rules
+        @existing_rule_references_rules ||= ::RuleReferencesRule.find_unique(
+          op_rule_references_rules
+        )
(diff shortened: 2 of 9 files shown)


app/models/application_record.rb
```diff
@@ -3,4 +3,21 @@
...
   self.abstract_class = true
+
+  def self.arel_find(collection, batch_size = 1000)
+    base_arel = arel_table[column_names.first].eq('')
+    collection.in_groups_of(batch_size, false).map do |batch|
+      where(batch.inject(base_arel) do |rel, record|
+        rel.or(yield(record))
+      end)
+    end.inject([], &:+)
+  end
```
akofink, line 14: This replaces an old paradigm that doesn't perform quite as well:

```rb
models.inject(where('1=0')) do |rel, model|
      rel.or(where(foo: model.foo, bar: model.bar))
end
```

By using arel directly instead of AR, we have more room for large queries before hitting a `SystemStackError` when the stack grows too large. The old way allowed up to about 200 records in the collection; the new code allows about 2000 records. I've set the default batch size to 1000 to be safe, and we can reduce it further in the future if it becomes problematic.

app/models/application_record.rb
```diff
@@ -3,4 +3,21 @@
...
+
+  def ==(other)
+    return super if self.class.column_names.include? 'id'
+
+    self.class.column_names.map do |col|
+      send(col) == other.send(col)
+    end.all?
+  end
```
akofink, line 22: This method short circuits for normal tables with an ID field. For those, equality is based on their ids matching. For join tables, such as RuleReferencesRule, there is no ID, so the records are never equal. This method instead compares their column values, which for a join table is simply two IDs (i.e. rule_id and rule_reference_id).

bastilian: Are there any performance penalties when it tries to compare every column value? 
Maybe this could go into a concern and only include it for join tables?

app/services/concerns/xccdf/rule_references_rules.rb
```diff
@@ -7,20 +7,52 @@ module RuleReferencesRules
...
+        ::RuleReferencesRule.import!(new_rule_references_rules)
+      end
+
+      private
+
+      def op_rule_references_rules
+        @op_rule_references_rules ||= @op_rules.flat_map do |op_rule|
+          rule = rule_for(ref_id: op_rule.id)
+          rule_references_for(op_rule: op_rule).map do |rule_reference|
+            OpenStruct.new(rule_id: rule.id,
```
bastilian, line 22: Do we have any more benefit from using a struct, other than being able to use dot notation?

The Ruby docs state that if performance is a concern it should be considered to use `Hash` instead.[1]

[1] https://ruby-doc.org/stdlib-2.5.3/libdoc/ostruct/rdoc/OpenStruct.html#class-OpenStruct-label-Implementation

app/models/application_record.rb
akofink: For most models (those with an 'id' column), this changes nothing. Notice that it short circuits to `super` on line 17.

app/services/concerns/xccdf/rule_references_rules.rb
akofink: It looks like we can use a `Struct` here instead of an `OpenStruct` and not face this performance penalty - I'll update the PR.

The dot notation is useful because the `RuleReferencesRule#find_unique` method takes a model-like collection with the fields required for uniqueness (in this case, `rule_id` and `rule_reference_id`). It allows passing real models or objects that behave like models (respond to the appropriate column methods without the overhead of being a model).

akofink:
Updated to use a `Struct` instead of an `OpenStruct`.

bastilian:
@akofink Awesome! Even more approved now! :)

PR #538, opened 2020-08-04
[stable] [revert] RHICOMPL-698

akofink:
Revert "Fix finding new rule references rules in parsing"
Revert "Improve parsing performance"


akofink:
This apparently did not break stable after all.

PR #540, opened 2020-08-07
[RHICOMPL-742] Allow editing the profile description

bastilian:
Required for https://github.com/RedHatInsights/compliance-frontend/pull/685

### DIFF
--- a/app/graphql/mutations/profile/edit.rb
+++ b/app/graphql/mutations/profile/edit.rb
@@ -7,21 +7,27 @@ module Mutations
       graphql_name 'UpdateProfile'
 
       argument :id, ID, required: true
+      argument :name, String, required: false
+      argument :description, String, required: false
       argument :compliance_threshold, Float, required: false
       argument :business_objective_id, ID, required: false
+
       field :profile, Types::Profile, null: true
 
       def resolve(args = {})
-        profile = Pundit.authorize(
+        profile = authorized_profile(args)
+        profile.update(args.compact)
+        { profile: profile }
+      end
+
+      private
+
+      def authorized_profile(args)
+        Pundit.authorize(
           context[:current_user],
           ::Profile.find(args[:id]),
           :edit?
         )
-        profile.update(
-          compliance_threshold: args[:compliance_threshold],
-          business_objective_id: args[:business_objective_id]
-        )
-        { profile: profile }
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/graphql/mutations/profile/edit.rb
```diff
@@ -7,22 +7,33 @@ class Edit < BaseMutation
...
-          :edit?
-        )
-        profile.update(
+        profile = authorized_profile(args)
+        profile.update!(
+          name: args[:name] || profile.name,
+          description: args[:description] || profile.description,
           compliance_threshold: args[:compliance_threshold],
           business_objective_id: args[:business_objective_id]
         )
```
akofink, line 24: If we call `update!` and pass `nil` for a field (such as `business_objective_id`), it will be updated to `nil` instead of not being updated at all.

bastilian: Yes, and in the case of `business_objective_id` we actually want that.
Not for the compliance_threshold, that is still something to be fixed.

akofink: Not quite - make sure to support the graphql API as written, not just what the UI needs. Both `business_objective_id` and `compliance_threshold` are optional. See the graphql variables I used above in my review.

app/graphql/mutations/profile/edit.rb
```diff
@@ -7,22 +7,33 @@ class Edit < BaseMutation
       graphql_name 'UpdateProfile'
 
       argument :id, ID, required: true
+      argument :name, String, required: false
+      argument :description, String, required: false
       argument :compliance_threshold, Float, required: false
       argument :business_objective_id, ID, required: false
```
akofink, line 13: @bastilian note the required args here

app/graphql/mutations/profile/edit.rb
```diff
@@ -7,22 +7,33 @@ class Edit < BaseMutation
...
-          ::Profile.find(args[:id]),
-          :edit?
-        )
+        profile = authorized_profile(args)
         profile.update(
+          name: args[:name] || profile.name,
+          description: args[:description] || profile.description,
           compliance_threshold: args[:compliance_threshold],
           business_objective_id: args[:business_objective_id]
         )
```
akofink, line 24: I think this should work to only update arguments that are passed in graphql:

```suggestion
        profile.update(args.compact)
```

bastilian: YES! Works like a charm and so much nicer! Thank you!

PR #542, opened 2020-08-12
Allow profiles to differ only by in/external

akofink:

### DIFF
--- a/app/jobs/parse_report_job.rb
+++ b/app/jobs/parse_report_job.rb
@@ -32,7 +32,8 @@ class ParseReportJob
     notify_payload_tracker(:processing, "Job #{jid} is now processing")
     parser.save_all
     notify_payload_tracker(:success, "Job #{jid} has completed successfully")
-  rescue ::MissingIdError, ::WrongFormatError, ::InventoryHostNotFound => e
+  rescue ::MissingIdError, ::WrongFormatError, ::InventoryHostNotFound,
+         ::ActiveRecord::RecordInvalid => e
     error_message = "Cannot parse report: #{e} - #{@msg_value.to_json}"
     notify_payload_tracker(:error, error_message)
     Sidekiq.logger.error(error_message)
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -11,6 +11,7 @@ module ProfilePolicyAssociation
       return self unless external
 
       Profile.includes(:benchmark)
+             .older_than(created_at)
              .find_by(account: account_id, external: false, ref_id: ref_id,
                       benchmarks: { ref_id: benchmark.ref_id })
     end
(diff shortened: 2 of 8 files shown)


app/jobs/parse_report_job.rb
```diff
@@ -32,7 +32,8 @@ def save_all
     notify_payload_tracker(:processing, "Job #{jid} is now processing")
     parser.save_all
     notify_payload_tracker(:success, "Job #{jid} has completed successfully")
-  rescue ::MissingIdError, ::WrongFormatError, ::InventoryHostNotFound => e
+  rescue ::MissingIdError, ::WrongFormatError, ::InventoryHostNotFound,
+         ::ActiveRecord::RecordInvalid => e
```
akofink, line 36: We don't want to push the job back to redis if the record is invalid (i.e. someone uploads the same report twice).

app/models/concerns/profile_policy_association.rb
```diff
@@ -11,6 +11,7 @@ def policy
       return self unless external
 
       Profile.includes(:benchmark)
+             .older_than(created_at)
```
akofink, line 14: Profile A is a policy of profile B only if profile A is older than profile B. This is Mohit's clever workaround.

test/graphql/mutations/associate_profiles_test.rb
```diff
@@ -31,7 +31,8 @@ class AssociateProfilesMutationTest < ActiveSupport::TestCase
       context: { current_user: users(:test) }
     )['data']['associateProfiles']['system']
 
-    assert_equal hosts(:one).reload.profiles, [profiles(:one), profiles(:two)]
+    assert_equal Set.new(hosts(:one).reload.profiles),
+                 Set.new([profiles(:one), profiles(:two)])
```
akofink, line 35: This test sometimes fails if the arrays aren't in the same order. We've had to do this for a lot of our tests so that they don't sometimes fail.

akofink:
Sorry for the delay on this one. It's now passing tests and ready for review! :)

akofink:
@bastilian please review

app/jobs/parse_report_job.rb
bastilian: Can't we be more deterministic and assert before parsing if we already are parsing the incoming report?

test/graphql/mutations/associate_profiles_test.rb
bastilian: This is just an idea, but you could `assert hosts(:one).reload.profiles.includes?(profiles([:one, :two].sample))`

bastilian (approved):
LGTM. My only minor concern is with relying on the `RecordInvalid` error.

test/graphql/mutations/associate_profiles_test.rb
akofink: That would be okay, but it doesn't check that both profiles exist. There's no problem using sets like this that I can think of, and they have no real performance overhead, especially with length 2.

app/jobs/parse_report_job.rb
akofink: We could check for specific failures, like if a TestResult already exists, before trying to save; however, if a validation error is raised in any of the model transactions, we don't want to continue to retry, because the same report upload will just fail again (and we'll try it 3 times before removing the job from redis). By rescuing this exception, we will never retry this kind of errored report upload.

akofink:
Thanks for the review!

PR #528, opened 2020-07-21
Listen for reports on the inventory events consumer

akofink:
This PR removes the reports consumer; Now reports will come in via the `platform.inventory.events` topic.

Signed-off-by: Andrew Kofink <akofink@redhat.com>

PR #543, opened 2020-08-12
Un-memoize instance variables in inventory-consumer

akofink:
When I implemented this, I thought it'd be good to memoize some of these things, especially `report_contents` since it pulls from S3. It turns out that racecar consumers keep a single object around the entire time, so every time we get a message, we're dealing with the exact same object with all instance variables set from the last time.

Introduced in https://github.com/RedHatInsights/compliance-backend/pull/528
Related to https://projects.engineering.redhat.com/browse/RHICOMPL-765

### DIFF
--- a/app/consumers/concerns/report_parsing.rb
+++ b/app/consumers/concerns/report_parsing.rb
@@ -82,7 +82,7 @@ module ReportParsing
     end
 
     def identity
-      @identity ||= IdentityHeader.new(b64_identity)
+      IdentityHeader.new(b64_identity)
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/consumers/concerns/report_parsing.rb
```diff
@@ -82,7 +83,7 @@ def report_contents
     end
 
     def identity
-      @identity ||= IdentityHeader.new(b64_identity)
+      IdentityHeader.new(b64_identity)
```
akofink, line 86: There's no reason to memoize this one at all, since we only use it once.

app/consumers/inventory_events_consumer.rb
```diff
@@ -10,15 +10,26 @@ class InventoryEventsConsumer < ApplicationConsumer
   def process(message)
     super(message)
 
-    if service == 'compliance'
-      produce(parse_report, topic: Settings.platform_kafka_validation_topic)
-    end
+    handle_report_parsing
```
akofink, line 13: Had to move this due to rubocop

PR #546, opened 2020-08-13
Update openapi paths to remove path prefix

akofink:
This currently breaks our spec tests, since it tries to get our API
without the path prefix. More investigation on how to do this with rswag
is needed.


akofink:
Closing this in favor of exploring active and fixed rswag forks.

PR #547, opened 2020-08-17
Create new hosts from inventory on profile update

bastilian:
When assigning a host that is not yet in the compliance database, updating the profile via the API will fail.
This adds a step to `update_hosts` to ensure that hosts are in the database and can be assigned.

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -22,7 +22,43 @@ class Host < ApplicationRecord
   validates :name, presence: true
   validates :account, presence: true
 
+  def in_inventory
+    host_in_inventory(id)
+  end
+
   class << self
+    def host_in_inventory(host_id)
+      ::HostInventoryAPI.new(
+        User.current.account,
+        ::Settings.host_inventory_url,
+        nil
+      ).inventory_host(host_id)
+    end
+
+    def find_or_create_hosts_by_inventory_ids(ids)
+      ids.map { |id| find_or_create_from_inventory(id) }
+    end
+
+    def find_or_create_from_inventory(host_id)
+      existing_hosts = ::Pundit.policy_scope(User.current, self)
+                               .where(id: host_id).first
+
+      existing_hosts || create_from_inventory(host_id)
+    end
+
+    def create_from_inventory(host_id)
+      i_host = host_in_inventory(host_id)
+
+      host = find_or_initialize_by(
+        id: i_host['id'],
+        account_id: User.current.account.id
+      ) do |h|
+        h.name = i_host['display_name']
+      end
+      host.save!
+      host
+    end
+
     def filter_by_compliance(_filter, operator, value)
       ids = Host.includes(:profiles).select do |host|
         host.compliant.values.all?(ActiveModel::Type::Boolean.new.cast(value))
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -70,3 +70,18 @@ unless Rails.env.production?
     end
   end
 end
+
+def mock_platform_api
+  require 'securerandom'
+
+  @url = 'http://localhost'
+  @b64_identity = '1234abcd'
+  @api = HostInventoryAPI.new(@account, @url, @b64_identity)
+  @connection = mock('faraday_connection')
+  @system_profile_response = OpenStruct.new(body: {
+    results: [{
+      id: "MOCK_INVENTORY_HOST_ID_#{SecureRandom.uuid}",
+      system_profile: { os_release: '8.2' } }]
+  }.to_json)
+  Platform.stubs(:connection).returns(@connection)
+end
--- a/app/graphql/mutations/base_mutation.rb
+++ b/app/graphql/mutations/base_mutation.rb
@@ -8,52 +8,6 @@ module Mutations
     end
   end
 
-  # Helpers for inventory service related mutations
-  module InventoryServiceHelper
-    include UserHelper
-
-    def inventory_host(id)
-      ::HostInventoryAPI.new(
-        current_user.account,
-        ::Settings.host_inventory_url,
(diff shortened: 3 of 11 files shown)


app/models/host.rb
```diff
@@ -23,6 +23,38 @@ class Host < ApplicationRecord
...
+      end
+    end
+
+    def host_in_inventory(host_id)
+      ::HostInventoryAPI.new(
+        User.current.account,
+        ::Settings.host_inventory_url,
+        nil
+      ).inventory_host(host_id)
+    end
```
akofink, line 56: All this logic already exists in the [InventoryServiceHelper](https://github.com/RedHatInsights/compliance-backend/blob/master/app/graphql/mutations/base_mutation.rb#L12) as part of the graphql base mutations. We should just pull it out to a common location, like a model concern (app/models/concerns/).

app/models/host.rb
```diff
@@ -22,7 +22,43 @@ class Host < ApplicationRecord
   validates :name, presence: true
   validates :account, presence: true
 
+  def in_inventory
+    ::HostInventoryAPI.new(
+      User.current.account,
+      ::Settings.host_inventory_url,
+      nil
+    ).inventory_host(id)
+  end
```
akofink, line 31: :+1: This seems really useful, and I like how it looks `Host.first.in_inventory`

app/models/host.rb
```diff
@@ -22,7 +22,43 @@ class Host < ApplicationRecord
...
+
+    def find_or_create_hosts_by_inventory_ids(ids)
+      ids.map { |id| find_or_create_from_inventory(id) }
+    end
+
+    def find_or_create_from_inventory(host_id)
+      existing_hosts = ::Pundit.policy_scope(User.current, self)
+                               .where(id: [host_id])
+
+      existing_hosts.first || create_from_inventory(host_id)
```
akofink, line 46: `where(...).first` and then checking for `nil` can just use `find_by(...)` instead

app/models/host.rb
```diff
@@ -22,7 +22,43 @@ class Host < ApplicationRecord
...
+
+    def create_from_inventory(host_id)
+      i_host = host_in_inventory(host_id)
+
+      host = find_or_initialize_by(
+        id: i_host['id'],
+        account_id: User.current.account.id
+      ) do |h|
+        h.name = i_host['display_name']
+      end
```
akofink, line 57: `find_or_initialize_by` is something I see as a bit of a red flag, performance-wise. This method, `create_from_inventory` should just create blindly, without checking if the host already exists, and then pass in only new IDs by using a single `where` query to determine what already exists.

Also, it should arguably be `new_from_inventory` so that we can use `Host.import!` to save all records in one query.

I doubt we will find these performance issues during our own testing, but some users may try to associate >1000 systems to a profile, and that's where we'll see timeouts or similar. For this deployment, I'm okay to forgo performance improvements here - we can create a jira to revisit it.

test/test_helper.rb
```diff
@@ -70,3 +70,18 @@ def parsed_data
...
+def mock_platform_api
+  require 'securerandom'
+
+  @url = 'http://localhost'
+  @b64_identity = '1234abcd'
+  @api = HostInventoryAPI.new(@account, @url, @b64_identity)
+  @connection = mock('faraday_connection')
+  @system_profile_response = OpenStruct.new(body: {
+    results: [{
+      id: "MOCK_INVENTORY_HOST_ID_#{SecureRandom.uuid}",
```
bastilian, line 83: @akofink Do you have an idea how to mock it so that when `Host.host_in_inventory` gets called it just returns a struct with whatever id it gets passed? Mocha doesn't seem to be able to take a block to use arguments passed. :/ Here I'm trying to mock the API, but it could be higher up as well.

akofink: Like this?

```rb
Host.stubs(:host_in_inventory).returns(...)
```

bastilian: Yeah, but it can't take any argument. something like:
```ruby
Host.stubs(:host_in_inventory) do |host_id|
  ...
end
```
So the struct returned has the `host_id` as `id`

bastilian: I tried it with just redefining the method with `def Host.host_in_inventory` in the `setup`s, but that doesn't seem right.

app/models/host.rb
```diff
@@ -22,7 +22,43 @@ class Host < ApplicationRecord
...
   validates :account, presence: true
 
+  def in_inventory
+    host_in_inventory(id)
+  end
+
   class << self
+    def host_in_inventory(host_id)
+      ::HostInventoryAPI.new(
+        User.current.account,
```
akofink, line 32: Even if we don't use this PR, I'd like to point out a couple things that we would have needed to fix.

Here's one - getting the current user from thread local memory like this should be done at the controller level and never at the model level. I know, User.current is used to store it, and User is a model, but the controllers (and graphql) are responsible for managing the state of who the current user is. graphql does not set `User.current` like REST does - it uses `context[:current_user]`, some kind of graphql context. Anyway, I don't know for sure if this would have caused issues, but it's something that worries me about this implementation.

app/models/host.rb
```diff
@@ -22,7 +22,43 @@ class Host < ApplicationRecord
   validates :name, presence: true
   validates :account, presence: true
 
+  def in_inventory
+    host_in_inventory(id)
+  end
```
akofink, line 27: I really liked this at first because of how convenient it was, but I think there may be an issue related to `current_user` I discussed already. Because the HostInventoryAPI requires a user account, I'm not sure we can ever move this sort of thing into the model directly like this.

bastilian:
Closing got fixed in #552

PR #545, opened 2020-08-13
[stable] New report parsing from platform.inventory.events (Host creation API deprecation)

akofink:
:warning: To be merged Tuesday, Aug 18 :warning: in sync with puptoo and inventory changes in QA and **after** the compliance deployment to production.

https://projects.engineering.redhat.com/browse/RHICOMPL-765

:no_smoking:

PR #549, opened 2020-08-19
[stable] New report parsing from platform.inventory.events (Host creation API deprecation)

bastilian:
See https://github.com/RedHatInsights/compliance-backend/pull/545

+ codecov update


Victoremepunto (approved):
Seems okay, though a bit flaky as we discussed with regards to the reports parsing, maybe something slipped through the cracks.

PR #552, opened 2020-08-26
[RHICOMPL-467] Consolidate REST and graphql inventory logic

akofink:
When a profile and host are associated but the host only exists in the
inventory, we must create the host from the inventory in our own
database. We must handle this in both REST and graphql. This change
consolidates the two code paths to use the same InventoryServiceHelper.

### DIFF
--- a/app/controllers/concerns/inventory_service_helper.rb
+++ b/app/controllers/concerns/inventory_service_helper.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+# Module to support controller integrations with the host inventory service
+module InventoryServiceHelper
+  extend ActiveSupport::Concern
+
+  included do
+    def add_inventory_hosts(ids)
+      existing_hosts = ::Pundit.policy_scope(current_user, ::Host)
+                               .where(id: ids)
+      save_hosts(ids - existing_hosts.pluck(:id)) + existing_hosts
+    end
+
+    def save_hosts(ids)
+      ids.map do |id|
+        save_host(id)
+      end
+    end
+
+    def save_host(id)
+      i_host = inventory_host(id)
+      host = ::Host.find_or_initialize_by(
+        id: i_host['id'],
+        account_id: current_user.account.id
+      )
+
+      host.update!(
+        name: i_host['display_name']
+      )
+
+      host
+    end
+
+    def inventory_host(id)
+      ::HostInventoryAPI.new(
+        current_user.account,
+        ::Settings.host_inventory_url,
+        nil # infer identity from account
+      ).inventory_host(id)
+    end
+  end
+end
--- a/app/controllers/v1/profiles_controller.rb
+++ b/app/controllers/v1/profiles_controller.rb
@@ -3,6 +3,8 @@
 module V1
   # API for Profiles
   class ProfilesController < ApplicationController
+    include InventoryServiceHelper
+
     ALLOWED_CREATE_ATTRIBUTES = %i[
       name description compliance_threshold business_objective
       parent_profile_id
@@ -31,11 +33,10 @@ module V1
     end
 
     def create
-      profile = Profile.new(profile_create_attributes).fill_from_parent
+      @profile = Profile.new(profile_create_attributes).fill_from_parent
 
       if profile.save
-        profile.update_hosts(new_host_ids)
-        profile.update_rules(ids: new_rule_ids)
+        update_relationships
         render_json profile, status: :created
       else
         render_error profile
@@ -44,8 +45,7 @@ module V1
 
     def update
       if profile.update(profile_update_attributes)
-        profile.update_hosts(new_host_ids)
-        profile.update_rules(ids: new_rule_ids)
+        update_relationships
         render_json profile
       else
         render_error profile
@@ -62,6 +62,12 @@ module V1
     end
 
     private
+
+    def update_relationships
+      add_inventory_hosts(new_host_ids || [])
+      profile.update_hosts(new_host_ids)
+      profile.update_rules(ids: new_rule_ids)
+    end
 
     def tailoring_filename
       "#{profile.benchmark.ref_id}__#{profile.ref_id}__tailoring.xml"
--- a/app/graphql/mutations/profile/associate_systems.rb
+++ b/app/graphql/mutations/profile/associate_systems.rb
@@ -12,18 +12,13 @@ module Mutations
 
       def resolve(args = {})
         profile = find_profile(args[:id])
-        hosts = find_hosts(args[:system_ids])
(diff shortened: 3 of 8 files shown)


app/controllers/concerns/inventory_service_helper.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+# Module to support controller integrations with the host inventory service
+module InventoryServiceHelper
```
akofink, line 4: This whole module was moved from base_mutation.rb, nearly unchanged except for renaming the `find_hosts` method to `add_inventory_hosts` to be explicit that it's related to the inventory.

app/controllers/concerns/inventory_service_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    def add_inventory_hosts(ids)
+      existing_hosts = ::Pundit.policy_scope(current_user, ::Host)
+                               .where(id: ids)
+      save_hosts(ids - existing_hosts.pluck(:id)) + existing_hosts
```
akofink, line 11: This part is slightly different - we only reach out to inventory if we need to. All hosts passed in are still returned from this method, including those that were already in our DB.

app/controllers/concerns/inventory_service_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+    def save_host(id)
+      i_host = inventory_host(id)
+      host = ::Host.find_or_initialize_by(
+        id: i_host['id'],
+        account_id: current_user.account.id
+      )
+
+      host.update!(
+        name: i_host['display_name']
+      )
```
akofink, line 29: This is one place where we need to update OS info on the host. I'm not doing that in this PR, though.

app/controllers/concerns/inventory_service_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      host
+    end
+
+    def inventory_host(id)
+      ::HostInventoryAPI.new(
+        current_user.account,
```
akofink, line 36: Here's a bit of a trick - `current_user` must be defined in the including class. This is important because with graphql, it uses the `UserHelper` defined in the base_mutations.rb, which uses the graphql `context[:current_user]`. For REST controllers, `current_user` is defined in the `application_controller.rb`.

app/controllers/v1/profiles_controller.rb
```diff
@@ -63,6 +63,12 @@ def tailoring_file
 
     private
 
+    def update_relationships
+      add_inventory_hosts(new_host_ids || [])
+      profile.update_hosts(new_host_ids)
+      profile.update_rules(ids: new_rule_ids)
+    end
+
```
akofink, line 71: This new method is because of rubocop, but also the logic is identical for `create` and `update`, so it makes sense to consolidate them.

app/graphql/mutations/profile/associate_systems.rb
```diff
@@ -12,18 +12,13 @@ class AssociateSystems < BaseMutation
 
       def resolve(args = {})
         profile = find_profile(args[:id])
-        hosts = find_hosts(args[:system_ids])
-        profile_hosts = hosts.map do |host|
-          ProfileHost.new(profile_id: profile.id, host_id: host.id)
-        end
-        ProfileHost.import!(profile_hosts)
+        add_inventory_hosts(args[:system_ids])
+        profile.update_hosts(profile.hosts.pluck(:id) | args[:system_ids])
```
akofink, line 16: Leveraging `profile.update_hosts` here, which does nearly the same thing as the existing code, with the added benefit that it also removes hosts that are not specified. There should be no implications for this mutation used with creating a profile.

app/graphql/mutations/profile/associate_systems.rb
```diff
@@ -12,18 +12,13 @@ class AssociateSystems < BaseMutation
...
         { profile: profile }
       end
 
       include HostHelper
-      include UserHelper
       include ProfileHelper
-      include InventoryServiceHelper
```
akofink, line 26: These two weren't directly used here, so I've removed them.

spec/integration/profiles_spec.rb
```diff
@@ -71,6 +71,11 @@
 
     post 'Create a profile' do
       fixtures :accounts, :profiles, :rules, :hosts, :benchmarks
+
+      before do
+        accounts(:one).hosts = hosts
+      end
```
akofink, line 77: This is the only kind of change needed in our tests. With Pundit enforcement, we must ensure the account used for testing is associated to the hosts used for testing.

bastilian:
There seems to be one issue, when I try to "edit" which is just sending another `associateSystems` mutation with a new set of systemIds, it doesn't work properly to remove systems. Here is a short screen recording just to illustrate. 

Kapture 2020-08-26 at 21 56 21

akofink:
Thanks, @bastilian! I'll check that out. I've just pushed two additional tests for profile create and update with inventory hosts via REST.

app/graphql/mutations/profile/associate_systems.rb
```diff
@@ -12,18 +12,13 @@ class AssociateSystems < BaseMutation
 
       def resolve(args = {})
         profile = find_profile(args[:id])
-        hosts = find_hosts(args[:system_ids])
-        profile_hosts = hosts.map do |host|
-          ProfileHost.new(profile_id: profile.id, host_id: host.id)
-        end
-        ProfileHost.import!(profile_hosts)
+        add_inventory_hosts(args[:system_ids])
+        profile.update_hosts(args[:system_ids])
```
akofink, line 16: @bastilian now we update the profile to the exact set of host IDs passed in, not including any existing hosts. If you pass an empty array, the profile will be unassociated with all hosts.

test/graphql/mutations/associate_systems_test.rb
```diff
@@ -71,4 +71,35 @@ class AssociateSystemsMutationTest < ActiveSupport::TestCase
 
     assert_equal(profiles(:one).hosts.pluck(:id), [NEW_ID])
   end
+
+  test 'removes systems from a profile' do
+    query = <<-GRAPHQL
```
akofink, line 76: New test! This associateSystems mutation actually changes slightly from the way it used to work. The array of systems passed in now will become the systems associated with the profile, no more, no less, regardless of if the profile already has systems associated. Those systems not specified in the mutation will be removed from the profile.

bastilian (approved):
Looks good to me and works as expected! Thank you @akofink!

akofink:
Thanks for the review, @bastilian!

PR #554, opened 2020-08-27
[RHICOMPL-420] Update scoped_search to 4.1.9

akofink:
The newest version of scoped search fixes searching on has_many through
relations, which is required for searching by a reference on a rule,
since there is a join table, rule_references_rules.

Examples of queries that are now fixed (from the IQE suite):
```
/api/compliance/rules?search=reference=\"AU-8(1)\"
/api/compliance/rules?search=identifier=CCE-27012-4
```


bastilian (approved):
Nice! Works for me. Thank you @akofink!

PR #555, opened 2020-08-27
[RHICOMPL-837] Deploy remediations as part of smoke tests

akofink:
Draft until we confirm this is working correctly


akofink:
This is not actually needed. The ssg import task fails gracefully without remediations available.

PR #557, opened 2020-08-28
[RHICOMPL-844] Set sidekiq concurrency to 1 in devel

akofink:
https://projects.engineering.redhat.com/browse/RHICOMPL-844

Debugging with 5 concurrent job workers is tough, and there is no real load locally, so it makes sense to only allow 1 job at a time.


bastilian (approved):
Thank you @akofink!

PR #558, opened 2020-08-31
[RHICOMPL-574] Update host OS info from inventory

akofink:
Now, every time we update our host record from the inventory (3 places
total), we update the host's name and OS information. Updates happen
during report parsing, InventoryHostUpdatedJob, and when we initially
create the host from an existing inventory record.


akofink:
Codecov is broken - we do have coverage on app/services/concerns/xccdf/hosts.rb. See https://projects.engineering.redhat.com/browse/RHICOMPL-847.

bastilian (approved):
Looks good to me! Thanks @akofink!

PR #560, opened 2020-08-31
[RHICOMPL-838] Rescue remediations errors, bubble errors on import

akofink:
This change adds connection errors to the list of rescued errors when
importing remediations. It also ensures that any exceptions that are
raised cause the entire rake task to fail (ssg:import_latest_supported).
It also updates the slack channel for exception notifications to
forum-cloudservices-compliance.

https://projects.engineering.redhat.com/browse/RHICOMPL-838

### DIFF
--- a/app/services/remediations_api.rb
+++ b/app/services/remediations_api.rb
@@ -12,8 +12,8 @@ class RemediationsAPI
     ::Rule.with_profiles.find_in_batches(batch_size: 100) do |rules|
       update_rules(remediations_available(remediations_response(rules)))
     end
-  rescue Faraday::ClientError => e
-    Rails.logger.info("#{e.message} #{e.response}")
+  rescue Faraday::ClientError, Faraday::ConnectionFailed => e
+    Rails.logger.error(e.full_message)
   end
 
   private
--- a/lib/tasks/import_datastream.rake
+++ b/lib/tasks/import_datastream.rake
@@ -56,6 +56,8 @@ namespace :ssg do
         e,
         data: OpenshiftEnvironment.summary
       )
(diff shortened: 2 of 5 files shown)


app/services/remediations_api.rb
```diff
@@ -12,8 +12,8 @@ def import_remediations
     ::Rule.with_profiles.find_in_batches(batch_size: 100) do |rules|
       update_rules(remediations_available(remediations_response(rules)))
     end
-  rescue Faraday::ClientError => e
-    Rails.logger.info("#{e.message} #{e.response}")
+  rescue Faraday::ClientError, Faraday::ConnectionFailed => e
+    Rails.logger.error(e.full_message)
```
akofink, line 16: `e.full_message` has the class name (i.e. Faraday::ConnectionFailed) so we'll know what went wrong. It also includes the backtrace, in case that's useful.

lib/tasks/import_datastream.rake
```diff
@@ -56,6 +56,8 @@ namespace :ssg do
         e,
         data: OpenshiftEnvironment.summary
       )
+      puts "Import failed for #{filename} in #{Time.zone.now - start} seconds."
+      raise e
```
akofink, line 60: Bubble up the error, never swallow it. :champagne:

Victoremepunto:
Wouldn't mind to see a test to ensure the exceptions are logged and raised , other than that, looks great , thanks @akofink

bastilian (approved):
Approved! Thank you @akofink!

akofink:
Thanks for the reviews! I've added a few tests to cover this behavior.

PR #561, opened 2020-08-31
[RHICOMPL-574] sync_with_inventory actually saves host

akofink:
I'm not sure how this ever worked without saving the host to the DB. Now with a test!

https://projects.engineering.redhat.com/browse/RHICOMPL-574


bastilian (approved):
LGTM! Thanks @akofink! I believe originally the task was really only meant to remove hosts from our database, rather than "sync", while culling was not fully implemented yet.

PR #562, opened 2020-09-01
[RHICOMPL-574] Ensure host inventory API uses string hashes

akofink:
JSON always comes in as string keys, not symbols, in hashes from the
inventory API.

To test:

Start with a host that doesn't have OS information in the Compliance DB
```rb
[11] pry(main)> Host.find('73050cfc-52bb-4f8a-87d5-ba2b1b1be4be')
  Host Load (0.4ms)  SELECT  "hosts".* FROM "hosts" WHERE "hosts"."id" = $1 LIMIT $2  [["id", "73050cfc-52bb-4f8a-87d5-ba2b1b1be4be"], ["LIMIT", 1]]
=> #<Host:0x0000564284446e80
 id: "73050cfc-52bb-4f8a-87d5-ba2b1b1be4be",
 name: "rhel7-insights-client.virbr0.akofink-desktop",
 created_at: Tue, 01 Sep 2020 13:54:46 UTC +00:00,
 updated_at: Tue, 01 Sep 2020 15:18:06 UTC +00:00,
 account_id: "a8477b45-60f1-4ccf-8f64-0934ca2e2aab",
 os_major_version: nil,
 os_minor_version: nil>
```

Run sync_with_inventory:
```
podman exec -it  compliance-backend_rails_1 bundle exec rake sync_with_inventory
```

Check that the host has OS information:
```rb
[12] pry(main)> Host.find('73050cfc-52bb-4f8a-87d5-ba2b1b1be4be')
  Host Load (1.2ms)  SELECT  "hosts".* FROM "hosts" WHERE "hosts"."id" = $1 LIMIT $2  [["id", "73050cfc-52bb-4f8a-87d5-ba2b1b1be4be"], ["LIMIT", 1]]
=> #<Host:0x0000564284497240
 id: "73050cfc-52bb-4f8a-87d5-ba2b1b1be4be",
 name: "rhel7-insights-client.virbr0.akofink-desktop",
 created_at: Tue, 01 Sep 2020 13:54:46 UTC +00:00,
 updated_at: Tue, 01 Sep 2020 15:22:43 UTC +00:00,
 account_id: "a8477b45-60f1-4ccf-8f64-0934ca2e2aab",
 os_major_version: 7,
 os_minor_version: 7>
```

https://projects.engineering.redhat.com/browse/RHICOMPL-574

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -56,8 +56,8 @@ class Host < ApplicationRecord
   end
 
   def update_from_inventory_host!(i_host)
-    update!(name: i_host['display_name'],
-            os_major_version: i_host['os_major_version'],
-            os_minor_version: i_host['os_minor_version'])
+    update!({ name: i_host['display_name'],
+              os_major_version: i_host['os_major_version'],
+              os_minor_version: i_host['os_minor_version'] }.compact)
   end
 end
--- a/app/services/host_inventory_api.rb
+++ b/app/services/host_inventory_api.rb
@@ -29,8 +29,8 @@ class HostInventoryAPI
     raise ::InventoryHostNotFound if @inventory_host.blank?
 
     if (os_release = system_profile([host_id]).first)
-      @inventory_host[:os_major_version] = os_release[:os_major_version]
-      @inventory_host[:os_minor_version] = os_release[:os_minor_version]
+      @inventory_host['os_major_version'] = os_release['os_major_version']
+      @inventory_host['os_minor_version'] = os_release['os_minor_version']
     end
     @inventory_host
   end
@@ -42,9 +42,9 @@ class HostInventoryAPI
     )
     JSON.parse(response.body)['results'].inject([]) do |acc, host|
       os_major, os_minor = find_os_release(host['system_profile'])
-      acc << { id: host['id'],
-               os_major_version: os_major,
-               os_minor_version: os_minor }
+      acc << { 'id' => host['id'],
+               'os_major_version' => os_major,
+               'os_minor_version' => os_minor }
     end
   end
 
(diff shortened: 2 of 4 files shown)


vkrizan:
LGTM.

bastilian (approved):
+1 Thank you @akofink! Works as expected.

I did notice an oddity syncing with the inventory, where when the inventory doesn't return any OS information, the `nil`s turn into `0`.

akofink:
That seems pretty bad, actually, since it will show up in the UI. I'm going to fix it and update this PR.

akofink:
Updated! The host will now not be updated at all for those fields which are `nil`. Also added a test to cover that scenario.

akofink:
Nevermind - I need to fix more tests now :see_no_evil:

app/models/host.rb
```diff
@@ -56,8 +56,8 @@ def test_results?(_filter, _operator, value)
   end
 
   def update_from_inventory_host!(i_host)
-    update!(name: i_host['display_name'],
-            os_major_version: i_host['os_major_version'],
-            os_minor_version: i_host['os_minor_version'])
+    update!({ name: i_host['display_name'],
+              os_major_version: i_host['os_major_version'],
+              os_minor_version: i_host['os_minor_version'] }.compact)
```
akofink, line 61: Ok, this change is straightforward. Just remove `nil` values completely.

vkrizan: So the default value in db for them will be 0?

akofink: No, the default value in the DB is `nil`.

bastilian (approved):
Thank you @akofink! Version numbers now stay `nil`, when the inventory reports `nil`!

Tested again setting it to `nil` for a host and verified it stays `nil` after running the sync task.

akofink:
Thanks for the reviews!

PR #563, opened 2020-09-02
[stable][RHICOMPL-574] Ensure host inventory API uses string hashes

akofink:
JSON always comes in as string keys, not symbols, in hashes from the
inventory API.


bastilian (approved):

akofink:
:no_smoking: :tada:

PR #564, opened 2020-09-02
[RHICOMPL-778] Rule updates never happen to existing profiles

akofink:
When parsing a test result upload, rules of existing profiles should
never be updated.

https://projects.engineering.redhat.com/browse/RHICOMPL-778

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -70,10 +70,11 @@ class Profile < ApplicationRecord
       (new_profile = dup).update!(account: account, hosts: [host],
                                   parent_profile: self,
                                   external: external)
+      new_profile.update_rules(ref_ids: rules.pluck(:ref_id))
     else
       new_profile.hosts << host unless new_profile.hosts.include?(host)
     end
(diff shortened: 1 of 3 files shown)


app/models/profile.rb
```diff
@@ -70,10 +70,11 @@ def clone_to(account: nil, host: nil, external: true)
       (new_profile = dup).update!(account: account, hosts: [host],
                                   parent_profile: self,
                                   external: external)
+      new_profile.update_rules(ref_ids: rules.pluck(:ref_id))
     else
       new_profile.hosts << host unless new_profile.hosts.include?(host)
     end
-    new_profile.update_rules(ref_ids: rules.pluck(:ref_id))
+
```
akofink, line 77: This is it. Never update rules on existing profiles. The TestResult will still show all relevant rules, even outside of the associated profile, but the profile rules themselves should never be changed.

bastilian (approved):
LGTM! Thanks @akofink!

PR #566, opened 2020-09-03
[RHICOMPL-875] Docker-compose: Clean up duplicate params and sdin

vkrizan:
Remove duplicate `links` declarations, and leave only `rails` service having SDIN opened.

(RHICOMPL-875)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -107,9 +107,6 @@ services:
       - KAFKA_BOOTSTRAP_SERVERS=kafka:29092
       - INVENTORY_LOG_LEVEL=ERROR
       - KAFKA_SECONDARY_TOPIC_ENABLED=True
-    links:
-      - db
-      - kafka
     depends_on:
       - db
       - kafka
@@ -129,8 +126,6 @@ services:
       - LISTEN_PORT=8081
     ports:
       - 8081:8081
-    links:
-      - db
     depends_on:
       - db
       - kafka
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -236,7 +236,6 @@ services:
   prometheus:
     image: compliance-backend-rails
     tty: true
-    stdin_open: true
     environment:
       - DATABASE_SERVICE_NAME=postgres
       - POSTGRES_SERVICE_HOST=db


Victoremepunto:
Apparently all `links` elements can be removed safely as it is deprecated, and all containers are reachable (not doing anything specific with custom networks also).

https://docs.docker.com/compose/compose-file/#links

docker-compose.yml
```diff
@@ -107,9 +107,6 @@ services:
       - KAFKA_BOOTSTRAP_SERVERS=kafka:29092
       - INVENTORY_LOG_LEVEL=ERROR
       - KAFKA_SECONDARY_TOPIC_ENABLED=True
-    links:
```
bastilian, line 110: `depends_on` does not wait for a service to be up, this can lead to hostnames like `db` and `kafka` to be unreachable.
However `links` does need to wait to have the right IP to link to, that is why there are both. 

We could try to remove the depends_on.

vkrizan: I did not had problem with the dependent runs, only when it is being shut down the order seemed somewhat random. First it shut down kafka and then the zookeeper was stuck waiting for a connection to kafka.

I'll investigate.

vkrizan: The `depends_on` will start the containers in the order. They would have an IP and DNS that can be resolved by the next dependent service. Neither `depends_on` nor `links` wait for the service to be up and running. [Documentation](https://github.com/docker/docker.github.io/blob/master/compose/startup-order.md) suggest to handle connections resiliently in the application, or, some "wait for it" script.

[`links`](https://github.com/docker/docker.github.io/blob/master/compose/compose-file/index.md#links) is a legacy way of connecting containers within their own network. `links` is not needed as the containers are reachable w/o it.

bastilian: `links:` is not a legacy feature, only the `--links` flag[1].`links:` is part of the "user-defined networks" features that are references and recommended to use instead.

When we added the whole stack we added `links:` on purpose[2]. 
It may work properly and fast enough on some systems to get everything up in time, but not on all. 
For example, I still have to separately start inventory services, because they fail to reach kafka, as it isn't ready in time.

"wait for it" scripts would of course prevent any issues, and make sure to start only when all dependencies are reachable as well, but we can't add these for all services, without customising their images.

[1] https://docs.docker.com/compose/compose-file/#links
[2] https://github.com/RedHatInsights/compliance-backend/pull/527#issuecomment-663574839

vkrizan: Sure, I'll revert it.

I do understand it that the `links:` in YAML sets the flag `--link` (for each line) when starting a container. It seems that the links (or `--link` param) was kept there to have the possibility to set an alias name. A hostname alias is not configured anywhere.

From the https://docs.docker.com/compose/networking/#links:
> Links allow you to define extra aliases by which a service is reachable from another service. They are not required to enable services to communicate - by default, any service can reach any other service at that service’s name.

But, I understand that you might have a different version of docker/podman where it can work differently.

vkrizan: Reverted back to just remove duplicate `links` declarations.

akofink (approved):
Seems to work fine under docker or podman for me. @bastilian did you see any breakages related to the removed `links`?

bastilian:
@akofink @vkrizan  as I mentioned in the inline comment, it already is a wonky setup and it requires to start inventory and the inventory-web services separately again, because they do not come up properly with `docker-compose up`. Theses issues were mitigated by adding `links:` definitions and I fear removing them will make the setup even wonkier.

akofink (approved):
Ok, looks good and continues to work as expected. For dev setup items like this, we need to ensure everyone continues to have a working devel setup, of course. @bastilian, thanks for raising the issues you saw. We'll wait for your ACK to merge this one.

vkrizan:
> @akofink @vkrizan as I mentioned in the inline comment, it already is a wonky setup (for me at least) and it requires to start inventory and the inventory-web services separately again, because they do not come up properly with `docker-compose up`. Theses issues were mitigated by adding `links:` definitions and I fear removing them will make the setup even wonkier.

@bastilian Luckily, I've kept the `links` intact. This only removed some that were declared twice.

bastilian (approved):
LGTM! Tested it on a clean setup and everything builds and comes up the same as it did before.

PR #567, opened 2020-09-04
[RHICOMPL-869] Connect to Prometheus only when the hostname is configured

vkrizan:
When developing the Prometheus is most of the time not needed.

(RHICOMPL-869)

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,10 @@ To create hosts in the inventory the `kafka_producer.py` script can be used from
 docker-compose run -e NUM_HOSTS=1000 -e INVENTORY_HOST_ACCOUNT=00001 inventory-web bash -c 'pipenv install --system --dev; python3 ./utils/kafka_producer.py;'
 ```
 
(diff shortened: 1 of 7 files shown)


bastilian:
Awesome @vkrizan! 

It would make sense to add a paragraph to the README to mention when the `prometheus_exporter_host` setting is not given it won't enable prometheus and/or start defaulting to disabling it.

We only ever need this in development when actually working on prometheus export data, so it would make sense to disable it by default.

/cc @akofink

vkrizan:
@bastilian, indeed. I'll update the README. If needed we can create a separate config option to disable it, but I've chosen the easiest one with the existing option.

I've set this as WIP, as there are still connections to prometheus being made when graphql is hit:
```
Prometheus Exporter, failed to send message Connection refused - connect(2) for "localhost" port 9394
```

vkrizan:
Ready for review.

Readme updated. Condition for test env refactored.

PR #572, opened 2020-09-11
[RHICOMPL-661] Add a policy model and migrate profiles

akofink:
This change adds a policy model to match what users create in the UI. It
also migrates some data from profiles to policies, such as business
objective and compliance threshold.

Still to do: add more tests.


Victoremepunto:
> Extremely early PoC. Do not merge.
> 
> Signed-off-by: Andrew Kofink [akofink@redhat.com](mailto:akofink@redhat.com)

That's what the "Draft" status is for ;)

app/controllers/concerns/rendering.rb
```diff
@@ -11,14 +11,24 @@ def render_json(model, **args)
...
 
+    def model_errors(models = [])
+      models = [models].flatten
+      models.flat_map do |model|
+        model.errors.full_messages.map do |error|
+          error[0] = error[0].downcase
+          "#{model.class.name} #{error}"
+        end
+      end
+    end
```
akofink, line 30: Since the V1 profiles controller now creates/updates profiles and policies, we need to be able to show errors of both models. This does not break the old functionality of only passing a single model, via `models = [models].flatten`.

app/controllers/concerns/v1/profiles_controller/attributes.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module V1
+  class ProfilesController
+    # Helper methods related to profile attributes
+    module Attributes
```
akofink, line 6: This new helper is thanks to rubocop's class length cop. It's copied from profiles_controller with the addition of policy-related attributes.

app/controllers/v1/profiles_controller.rb
```diff
@@ -33,18 +25,18 @@ def destroy
...
 
     def create
-      @profile = Profile.new(profile_create_attributes).fill_from_parent
-
-      if profile.save
-        update_relationships
-        render_json profile, status: :created
-      else
-        render_error profile
+      Policy.transaction do
```
akofink, line 28: New: transactions in case any one thing fails

app/controllers/v1/profiles_controller.rb
```diff
@@ -89,28 +90,6 @@ def business_objective
...
-
-    def profile_update_attributes
-      resource_attributes.to_h.slice(*ALLOWED_UPDATE_ATTRIBUTES).tap do |attrs|
-        if business_objective
-          attrs.except! :business_objective
-          attrs[:business_objective_id] = business_objective.id
-        end
-      end
-    end
-
```
akofink, line 113: Moved to `Attributes` mixin.

app/graphql/mutations/profile/associate_systems.rb
```diff
@@ -13,7 +13,7 @@ class AssociateSystems < BaseMutation
       def resolve(args = {})
         profile = find_profile(args[:id])
         add_inventory_hosts(args[:system_ids])
-        profile.update_hosts(args[:system_ids])
+        profile&.policy&.update_hosts(args[:system_ids])
```
akofink, line 16: As a clarification, after this PR, profiles are only assigned hosts when a TestResult is uploaded.

app/graphql/mutations/profile/create.rb
```diff
@@ -18,26 +18,36 @@ class Create < BaseMutation
...
+        policy = ::Policy.new(new_policy_options(args))
+        profile = ::Profile.new(new_profile_options(args)).fill_from_parent
+
+        Policy.transaction do
+          policy.save!
+          profile.update!(policy: policy)
+          profile.update_rules(ref_ids: args[:selected_rule_ref_ids])
+        end
+
         { profile: profile }
```
akofink, line 30: Note that policy information (i.e. business_objective, compliance_threshold, etc.) is communicated via the profile, even though those columns still exist on profiles (for now).

app/graphql/mutations/profile/create.rb
```diff
@@ -18,26 +18,38 @@ class Create < BaseMutation
...
           compliance_threshold: args[:compliance_threshold]
         }
       end
+
+      def new_profile_options(args)
+        {
+          account_id: context[:current_user].account_id,
+          parent_profile_id: args[:clone_from_profile_id]
+        }
+      end
```
akofink, line 52: The more I work between graphql and REST APIs, the more I want to consolidate the logic. I'm intentionally not doing it here mostly because the PR is so big already, but also because we may remove the graphql API completely at some point.

app/graphql/mutations/profile/edit.rb
```diff
@@ -16,7 +19,7 @@ class Edit < BaseMutation
 
       def resolve(args = {})
         profile = authorized_profile(args)
-        profile.update(args.compact)
```
akofink, line 19: There is nothing on the profile itself that is editable by our APIs after this change.

app/models/concerns/profile_policy_association.rb
```diff
@@ -5,43 +5,18 @@ module ProfilePolicyAssociation
...
-                      benchmarks: { ref_id: benchmark.ref_id })
-    end
-
-    def policy_profiles
-      return Profile.none if account_id.nil?
-
-      Profile.includes(:benchmark)
-             .where(account: account_id, ref_id: ref_id,
-                    benchmarks: { ref_id: benchmark.ref_id })
-    end
```
akofink, line 25: These weird methods that pointed to other profiles now point to policies in the DB.

app/models/concerns/profile_policy_association.rb
```diff
@@ -5,43 +5,18 @@ module ProfilePolicyAssociation
...
 
-    def destroy_policy_test_results
-      if Settings.async
-        DestroyProfilesJob.perform_async(policy_profiles.pluck(:id))
-      else
-        DestroyProfilesJob.new.perform(policy_profiles.pluck(:id))
-      end
+    def destroy_empty_policy
+      policy.destroy if policy&.profiles&.empty?
     end
```
akofink, line 20: If you destroy the last profile in a policy, the policy is also destroyed.

app/models/concerns/profile_searching.rb
```diff
@@ -22,7 +24,8 @@ module ProfileSearching
         where.not(parent_profile_id: nil)
     }
     scope :external, lambda { |external = true|
-      where(external: external)
+      external && where(policy_id: nil) ||
+        where.not(policy_id: nil)
     }
```
akofink, line 29: `external` now simply means that no policy association exists for the profile

app/services/copy_profiles_to_policies.rb
```diff
@@ -0,0 +1,29 @@
...
+    def create_policies
+      profiles.find_each do |profile|
+        policy = Policy.create!(Policy.attrs_from(profile: profile))
+        policy.hosts << profile.hosts
+        profile.update!(
+          policy: policy
+        )
+      end
+    end
+  end
```
akofink, line 28: This is a fairly naive implementation, but we have only about 1000 policies in production today.

vkrizan:
@akofink One commit with 42 changes :slightly_smiling_face:, iterative changes are welcome.

akofink:
@vkrizan this one could be split into more pieces, for sure; however, once the models are updated in the migration, a bunch of other things have to change to keep everything working, so I can't make individual commits that keep the build green. Hopefully our review will shed some light on the changes and show that there are many of the same type of change across different files.

app/services/copy_profiles_to_policies.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def create_policies
+      profiles.find_each do |profile|
+        policy = Policy.create!(Policy.attrs_from(profile: profile))
+        policy.hosts << profile.hosts
+        profile.update!(
+          policy: policy
+        )
+      end
```
akofink, line 26: I think there's a bug here. I will write a test to reproduce it and then fix it.

app/models/concerns/profile_policy_association.rb
```diff
@@ -5,43 +5,18 @@ module ProfilePolicyAssociation
...
 
-    def destroy_policy_test_results
-      if Settings.async
-        DestroyProfilesJob.perform_async(policy_profiles.pluck(:id))
-      else
-        DestroyProfilesJob.new.perform(policy_profiles.pluck(:id))
-      end
+    def destroy_empty_policy
+      policy.destroy if policy&.profiles&.empty?
```
vkrizan, line 19: Observation: This could only happen if the current Profile is assigned to a policy and is not saved yet. Policy would have at least the current profile (`self`).

app/models/host.rb
```diff
@@ -14,10 +14,13 @@ class Host < ApplicationRecord
   has_many :rule_results, dependent: :delete_all
   has_many :rules, through: :rule_results, source: :rule
   has_many :profile_hosts, dependent: :destroy
+  has_many :policy_hosts, dependent: :destroy
   has_many :test_results, dependent: :destroy
   include SystemLike
 
   has_many :profiles, through: :profile_hosts, source: :profile
+  has_many :policies, through: :policy_hosts
```
vkrizan, line 22: Observation: Could be
```
has_and_belongs_to_many :policies
```
The same on other side.

app/models/policy.rb
```diff
@@ -0,0 +1,58 @@
...
+    return unless new_host_ids
+
+    policy_hosts.where.not(host_id: new_host_ids).destroy_all
+    PolicyHost.import((new_host_ids - host_ids).map do |host_id|
+      { host_id: host_id, policy_id: id }
+    end)
+  end
+
+  def benchmark
+    return benchmarks.first if benchmarks.uniq.one?
```
vkrizan, line 48: Nitpick: `s/.uniq/.distinct/`

app/models/profile.rb
```diff
@@ -56,20 +51,12 @@ def canonical?
...
-  def destroy_orphaned_business_objective
-    bo_changes = (previous_changes.fetch(:business_objective_id, []) +
-                  changes.fetch(:business_objective_id, [])).compact
-    return if bo_changes.blank?
-
-    BusinessObjective.without_profiles.where(id: bo_changes).destroy_all
-  end
-
-  def clone_to(account: nil, host: nil, external: true)
+  def clone_to(account: nil, host: nil, policy: nil)
```
vkrizan, line 54: Observation: This method is used in XccdfReprtParser, but the policy isn't passed.

vkrizan (commented):
There seems to be a small bug in the migration/copy service. The move of hosts to Policy seems a bit illogical, or at least it does not make sense to be done for this change (see my comment).

I'll do testing next.

db/migrate/20200910153704_create_policies_and_policy_profiles.rb
```diff
@@ -0,0 +1,30 @@
...
+  def change
+    create_table :policies, id: :uuid do |t|
+      t.references :business_objective, foreign_key: true, type: :uuid
+      t.float :compliance_threshold, default: 100
+      t.string :name
+      t.string :description
+      t.references :account, foreign_key: true, type: :uuid
+    end
+
+    create_table :policy_hosts, id: :uuid do |t|
```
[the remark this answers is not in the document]
vkrizan: Talked in person: The policy hosts are necessary for adding hosts when a user creates a new policy, assigning them regardless of their SSG version (latest is assumed). Previous Profile hosts are no longer necessary. For external Profiles the association to host is done through the TestResult.

app/models/concerns/profile_hosts.rb
```diff
@@ -7,14 +7,7 @@ module ProfileHosts
...
-    def update_hosts(new_host_ids)
-      return unless new_host_ids
-
-      profile_hosts.where.not(host_id: new_host_ids).destroy_all
-      ProfileHost.import((new_host_ids - host_ids).map do |host_id|
-        { host_id: host_id, profile_id: id }
-      end)
-    end
+    has_many :policy_hosts, through: :policy
+    has_many :assigned_hosts, through: :policy_hosts, source: :host
```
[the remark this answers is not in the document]
vkrizan: Talked in person: The intermediate model is used to have a validation, additionally to the db unique constraints, to produce better errors. `has_and_belongs_to_many` does some internal conversion to access the intermediate db table, but I think that ORM would not break these without notifying about a breaking change.

Outside of in person discussion: I believe that the `has_many :policy_hosts, through: :policy` is not necessary for `:assigned_hosts` to work.

app/models/concerns/profile_policy_association.rb
vkrizan: (discard my previous comment)

Note, that this can resolve RHICOMPL-1031.

db/migrate/20200910153704_create_policies_and_policy_profiles.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.references :host, foreign_key: true, type: :uuid, null: false
+
+      t.timestamps null: true
+    end
+    add_index(:policy_hosts, %i[policy_id host_id], unique: true)
+
+    add_column :test_results, :supported, :boolean
+    add_reference :profiles, :policy, type: :uuid
+    remove_index :profiles, column: %i[ref_id account_id benchmark_id external], unique: true, name: 'uniqueness'
+    add_index :profiles, %i[ref_id account_id benchmark_id policy_id], unique: true, name: 'uniqueness'
```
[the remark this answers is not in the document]
vkrizan: Please check whether the removal of `external` from the unique index does not break/revert RHICOMPL-652 PR #542

app/controllers/v1/profiles_controller.rb
```diff
@@ -33,18 +25,18 @@ def destroy
...
     def create
-      @profile = Profile.new(profile_create_attributes).fill_from_parent
-
-      if profile.save
-        update_relationships
-        render_json profile, status: :created
-      else
-        render_error profile
+      Policy.transaction do
+        if new_policy.save && new_profile.update(policy: new_policy)
```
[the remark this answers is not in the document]
akofink: It'd have to be `new_profile.create_policy(new_policy.attributes).persisted?`, which then wouldn't be shorter

app/controllers/v1/profiles_controller.rb
```diff
@@ -33,18 +25,18 @@ def destroy
...
+          render_json new_profile, status: :created
+        else
+          render_error [new_profile, new_policy]
+        end
       end
     end
 
     def update
-      if profile.update(profile_update_attributes)
+      if profile.policy.update(policy_update_attributes)
```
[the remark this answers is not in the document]
akofink: I've updated it to give this message when trying to update a profile without a policy:

```
HTTP/1.1 403 Forbidden
{
    "errors": "Editing an external profile is forbidden."
}
```

app/controllers/v1/profiles_controller.rb
```diff
@@ -74,7 +75,7 @@ def tailoring_filename
     end
 
     def profile
-      @profile ||= pundit_scope.find(params[:id])
+      @profile ||= @new_profile || pundit_scope.find(params[:id])
```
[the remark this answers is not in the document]
akofink: This is a little misleading, I guess. `@new_profile` will be `nil` for updates. Only calling the `new_profile` method will set this variable, but once it's set, we want to be sure we use it instead of finding a profile. I initially wrote it so that `new_profile` memoizes `@profile`, but rubocop complains when the memoized variable doesn't match the method name. :shrug:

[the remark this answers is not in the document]
vkrizan: Oh right, the instance variable would be empty. :+1:

app/controllers/v1/profiles_controller.rb
```diff
@@ -4,15 +4,12 @@ module V1
...
-      name description compliance_threshold business_objective
-      parent_profile_id
-    ].freeze
-
-    ALLOWED_UPDATE_ATTRIBUTES = %i[
-      name description compliance_threshold business_objective
-    ].freeze
+    before_action only: %i[update] do
+      error = 'Editing an external profile is forbidden.'
+      render json: { errors: error }, status: :forbidden if profile.policy.nil?
```
vkrizan, line 11: Observation: There could be a common method for rendering an error. The existing `render_error` can also accept String (or an array of Strings).

app/services/copy_profiles_to_policies.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    private
+
+    def profiles
+      Profile.canonical(false)
+             .where(policy: nil)
+    end
+
+    def create_policies
+      profiles.where(external: false).find_each do |profile|
```
vkrizan, line 18: @akofink The creation of policies and assignment of profiles seems incorrect. 

First problem:
With this migration it would be creating a single Policy for each Profile (from the set; non-canonical without a policy). First it gets all profiles that need don't have policy, for each it will create a policy and then it would reassign all other account's profiles that do not have a policy to newly created Policy. Since this processes it in batches, it would create redundant Policies. Some might become orphans (might be deleted by the hook).
This needs a bit more thinking (if it is a real problem). It might group Profiles by account, ref_id and os_major_version.

Second:
Grouping of profiles to a policy happens non-deterministically. I'd suggest to order the profiles by `created_at DESC` to create policies for latest profiles first. Otherwise policies would inherit attributes from a random profile.

Third:
The grouping to a profile happens regardless of an OS major version. The `ref_id` of a policy is found across OS major versions. The Policy should be only for a single OS major version.

akofink:
Schema migration moved to https://github.com/RedHatInsights/compliance-backend/pull/602

app/models/profile.rb
akofink: `nil` policy is the case when `external: true`. So this method is supposed to be used to clone a canonical profile to a user's account, for both external reports and internal.

akofink:
Converting this to draft, since we might break it up further. I've rebased the schema migration and model additions from master, and tests are all passing, but there are still some known (and unknown) issues and more testing on my side to be done. Also, the CopyProfileToPolicies services is going to be handled by RHICOMPL-662 (@vkrizan), and this PR will be rebased with any changes from that work.

Current smoke test run:

```
================================================================================== short test summary info ===================================================================================
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 90 and compliance_score <= 100-compliant = true] - IndexError...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 90 and compliance_score <= 100-compliant = false] - IndexErro...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 70 and compliance_score <= 89-compliant = true] - IndexError:...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 70 and compliance_score <= 89-compliant = false] - IndexError...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 50 and compliance_score <= 69-compliant = true] - IndexError:...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 50 and compliance_score <= 69-compliant = false] - IndexError...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 0 and compliance_score <= 49-compliant = true] - IndexError: ...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_filters[compliance_score >= 0 and compliance_score <= 49-compliant = false] - IndexError:...
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_cert_type_auth - Failed: DID NOT RAISE <class 'requests.exceptions.HTTPError'>
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_success - iqe_rbac_api.exceptions.ApiException: (401)
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_fails - iqe_rbac_api.exceptions.ApiException: (401)
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rest_api.py::test_list_external_profiles - assert []
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rest_api.py::test_retrieve_profile - assert False
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rest_api.py::test_list_all_rule_results - assert []
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rest_api.py::test_profile_search_by_id - IndexError: list index out of range
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rest_api.py::test_associate_systems_policy - AssertionError: assert '3f18b615-7ee7-4297-84fa-0cf29c6f4ff7' in []
===================================================== 16 failed, 72 passed, 1 skipped, 126 deselected, 339 warnings in 370.20s (0:06:10) =====================================================
```

akofink:
@vkrizan I've updated the base branch to RHICOMPL-661, so we can go ahead and merge this and then identify smaller chunks to work on.

app/services/copy_profiles_to_policies.rb
vkrizan: Being handled in PR #610

akofink: Whoops, sorry for including this :see_no_evil:

PR #574, opened 2020-09-14
[RHICOMPL-759] Refactor: Kafka producer to common one

vkrizan:
Refactoring of Kafka producer to allow creation of more producers.

Preparation for RHICOMPL-759.

### DIFF
--- a/app/producers/application_producer.rb
+++ b/app/producers/application_producer.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+# Common Kafka producer client
+# https://www.rubydoc.info/gems/ruby-kafka/Kafka/Producer
+class ApplicationProducer < Kafka::Client
+  BROKERS = [ENV['KAFKAMQ']].compact.freeze
+  CLIENT_ID = 'compliance-payload-tracker-producer'
+  SERVICE = 'compliance'
+  # Define TOPIC in the inherited class.
+  # Example:
+  #   TOPIC = 'platform.payload-status'
+
+  class << self
+    private
+
+    def deliver_message(msg)
+      msg = msg.merge(
+        date: DateTime.now.iso8601,
+        service: SERVICE,
+        source: ENV['APPLICATION_TYPE']
+      )
+      kafka&.deliver_message(msg.to_json, topic: self::TOPIC)
+    end
+
+    def logger
+      Rails.logger
+    end
+
+    def kafka
+      @kafka ||= Kafka.new(BROKERS, client_id: CLIENT_ID) if BROKERS.any?
+    end
+  end
+end
--- a/app/producers/application_producer.rb
+++ b/app/producers/application_producer.rb
@@ -4,7 +4,7 @@
 # https://www.rubydoc.info/gems/ruby-kafka/Kafka/Producer
 class ApplicationProducer < Kafka::Client
   BROKERS = [ENV['KAFKAMQ']].compact.freeze
-  CLIENT_ID = 'compliance-payload-tracker-producer'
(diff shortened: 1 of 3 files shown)


app/producers/application_producer.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# Common Kafka producer client
+# https://www.rubydoc.info/gems/ruby-kafka/Kafka/Producer
+class ApplicationProducer < Kafka::Client
+  BROKERS = [ENV['KAFKAMQ']].compact.freeze
+  CLIENT_ID = 'compliance-payload-tracker-producer'
```
vkrizan, line 7: @akofink Can the `CLIENT_ID` be set to something else, like "compliance-backend"?

app/producers/application_producer.rb
```diff
@@ -0,0 +1,33 @@
...
+  SERVICE = 'compliance'
+  # Define TOPIC in the inherited class.
+  # Example:
+  #   TOPIC = 'platform.payload-status'
```
akofink, line 11: The typical way to do this is to use a method instead of a constant and `raise NotImplementedError` in the parent class. This will enforce descendants implement their own custom version of the method:

```rb
def topic
  raise NotImplementedError.new('topic must be defined')
end
```

app/producers/application_producer.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  class << self
+    private
+
+    def deliver_message(msg)
+      msg = msg.merge(
+        date: DateTime.now.iso8601,
+        service: self::SERVICE,
```
akofink, line 19: No need for the `self::` here. `TOPIC` below won't require it either once it's updated to be a method.

akofink (commented):
Looks good! A few comments inline.

app/producers/application_producer.rb
vkrizan: Changed to common "compliance-backend". Defining per producer isn't necessary, as there are different topics, unless client instances would collide.

test/producers/payload_tracker_test.rb
```diff
@@ -10,17 +10,20 @@ class PayloadTrackerTest < ActiveSupport::TestCase
...
 
   test 'handles delivery issues' do
     kafka = mock('kafka')
-    Kafka.stubs(:new).returns(kafka)
+    ApplicationProducer.stubs(:new).returns(kafka)
     PayloadTracker.stubs(:kafka).returns(kafka)
-    kafka.expects(:deliver_message).raises(Kafka::DeliveryFailed.new(nil, nil))
+    kafka.expects(:deliver_message)
+         .with(anything, topic: 'platform.payload-status')
+         .raises(Kafka::DeliveryFailed.new(nil, nil))
```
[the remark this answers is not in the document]
vkrizan: Reverted. Please check if it is what you had in mind.

app/producers/application_producer.rb
vkrizan: I prefer the use of a constant here. If it would not be defined then it would also error out.

app/producers/application_producer.rb
vkrizan: `self::` dropped.

akofink (commented):
Looks good, thanks! Could you please add the jira ticket number in the commit message?

vkrizan:
Jira ticket is already in commit ec81e685b8e895d11e73d9550bdf8807af5362f1.

akofink:
Please make it of this format in the commit title, not just the body:
>[RHICOMPL-759] Refactor Kafka producer to common one

akofink (approved):

PR #575, opened 2020-09-14
[RHICOMPL-922] Allow setting platform basic auth

akofink:
This is required for local development. In deployed environments, 3Scale
generates the X_RH_IDENTITY header from different auth types (basic,
jwt, etc); locally, we must pass the header directly. For platform
services that are running in CI behind 3Scale (rbac, remediations, etc.),
the header will not work, so basic auth must be set.

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -2,6 +2,9 @@ host_inventory_url: http://inventory-web:8081
 remediations_url: https://ci.cloud.redhat.com
 rbac_url: https://ci.cloud.redhat.com
 # disable_rbac: true
+# platform_basic_auth:
+#   - username
+#   - password
 prometheus_exporter_host: prometheus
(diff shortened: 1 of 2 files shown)


config/settings/development.yml
```diff
@@ -2,6 +2,9 @@ host_inventory_url: http://inventory-web:8081
 remediations_url: https://ci.cloud.redhat.com
 rbac_url: https://ci.cloud.redhat.com
 # disable_rbac: true
+# platform_basic_auth:
+#   - username
+#   - password
```
akofink, line 7: To test this, keep RBAC enabled, and add your basic auth credentials to `config/settings/development.local.yml`.

bastilian (approved):
Looks good to me! Thank you @akofink! Less to keep in a stash! \o/

akofink:
Thanks for the review! :)

PR #576, opened 2020-09-16
Dev: Bind redis to localhost

vkrizan:
Expect redis to be listening on localhost in development environment.
This allows to have redis running locally or in container with bounded
port.


akofink (approved):
Works as expected!

PR #580, opened 2020-09-18
[RHICOMPL-840] Include os_major_version in benchmarks API

akofink:
```json
GET /compliance/api/v1/benchmarks

{
  "data": [
    {
      "id": "8036e3a9-9e97-4658-ac1f-5a0c03daac5d",
      "type": "benchmark",
      "attributes": {   
        "ref_id": "xccdf_org.ssgproject.content_benchmark_RHEL-6",
        "title": "Guide to the Secure Configuration of Red Hat Enterprise Linux 6",
        "version": "0.1.28",
        "description": "This guide presents a catalog of security-relevant configuration settings for Red Hat Enterprise Linux 6. It is a rendering of content structured in the eXtensible Configuration          
Checklist Description Format (XCCDF) in order to support security automation.  The SCAP content is is available in the scap-security-guide package which is developed at http://fedorahosted.org/scap-security-guid
e. Providing system administrators with such guidance informs them how to securely configure systems under their control in a variety of network roles.  Policy makers and baseline creators can use this catalog  
of settings, with its associated references to higher-level security control catalogs, in order to assist them in security baseline creation.  This guide is a catalog, not a checklist, and satisfaction of every 
item is not likely to be possible or sensible in any operational scenario.  However, the XCCDF format enables granular selection and adjustment of settings, and their association with OVAL and OCIL content      
provides an automated checking capability.  Transformations of this document, and its associated automated checking content, are capable of providing baselines that meet a diverse set of policy objectives.      
Some example XCCDF Profiles, which are selections of items that form checklists and can be used as baselines, are available with this guide.  They can be processed, in an automated fashion, with tools that      
support the Security Content Automation Protocol (SCAP).  The DISA STIG for Red Hat Enterprise Linux 6, which provides required settings for US Department of Defense systems, is one example of a baseline        
created from this guidance.",
        "os_major_version": "6"
      },
      "relationships": {  
        "rules": {…},
        "profiles": {…}
      }
    },
    {…},      
    {…}       
  ],
  "meta": {…},
  "links": {…} 
} 
```

### DIFF
--- a/spec/api/v1/schemas/benchmarks.rb
+++ b/spec/api/v1/schemas/benchmarks.rb
@@ -27,6 +27,9 @@ module Api
             },
             description: {
               type: 'string'
+            },
+            os_major_version: {
+              type: 'string'
             }
           }
         }.freeze
--- a/app/serializers/benchmark_serializer.rb
+++ b/app/serializers/benchmark_serializer.rb
@@ -3,7 +3,7 @@
 # JSON API serialization for an OpenSCAP Benchmark
 class BenchmarkSerializer
   include FastJsonapi::ObjectSerializer
-  attributes :ref_id, :title, :version, :description
+  attributes :ref_id, :title, :version, :description, :os_major_version
   has_many :rules
   has_many :profiles do |benchmark|
     Pundit.policy_scope(User.current, Profile).where(benchmark: benchmark)
--- a/swagger/v1/openapi.json
+++ b/swagger/v1/openapi.json
@@ -82,7 +82,8 @@
                       "ref_id": "xccdf_org.ssgproject.content_benchmark_RHEL-7",
                       "title": "Benchmark One",
                       "version": "0.1.45",
-                      "description": "The first benchmark"
+                      "description": "The first benchmark",
+                      "os_major_version": "7"
                     },
                     "relationships": {
                       "rules": {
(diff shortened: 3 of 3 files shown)


spec/api/v1/schemas/benchmarks.rb
```diff
@@ -27,6 +27,9 @@ module Benchmarks
             },
             description: {
               type: 'string'
+            },
+            os_major_version: {
+              type: 'string'
```
vkrizan, line 32: We might reconsider using an integer, for ease of comparison.

vkrizan:
@akofink In the Jira you've mentioned that the REST and GraphQL should be in sync, but I was unable to find the same attribute `os_major_version` on the `Benchmark` (neither in schema nor in type). What did I miss?

vkrizan (approved):
LGTM. Works!

vkrizan:
Would it make sense to have a smoke test for the REST API, checking the returned attributes? It could catch some regressions.

Victoremepunto:
> Would it make sense to have a smoke test for the REST API, checking the returned attributes? It could catch some regressions.

Absolutely, we will add it to our collection

spec/api/v1/schemas/benchmarks.rb
akofink: Yeah, it's not a bad idea, but this info comes from the inventory API which returns the version info as a string; it's not always of the format `"7.6"`, so for now I think it makes sense to match their API and return a string. It also should match the OS info we give for other resources, such as [profiles](https://github.com/RedHatInsights/compliance-backend/blob/master/spec/api/v1/schemas/profiles.rb#L73).

akofink:
@vkrizan thanks for the review! I've reported [RHICOMPL-958](https://issues.redhat.com/browse/RHICOMPL-958) to track the missing info in graphql. The frontend just parses the ref_id (TIL)!

PR #581, opened 2020-09-21
[RHICOMPL-944] Configurable Kafka Producer Topics

vkrizan:
Refactored settings for Kafka topics configuration to section of consumers/subscriptions and producers. Made the consumers to use their respective settings.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -5,6 +5,8 @@ kafka_consumer_topics:
   inventory_events: 'platform.inventory.events'
 kafka_producer_topics:
   upload_validation: 'platform.upload.validation'
+  payload_tracker: 'platform.payload-status'
+  remediation_updates: 'platform.remediation-updates.compliance'
 host_inventory_url: http://insights-inventory.platform-ci.svc.cluster.local:8080
 remediations_url: http://remediations.remediations-ci.svc.cluster.local:9002
 rbac_url: http://rbac.rbac-ci.svc.cluster.local:9002
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,7 +1,6 @@
 app_name: 'compliance'
 path_prefix: '/api'
 old_path_prefix: '/r/insights/platform'
-platform_kafka_topic: 'platform.upload.compliance'
 platform_kafka_inventory_topic: 'platform.inventory.events'
 platform_kafka_validation_topic: 'platform.upload.validation'
 host_inventory_url: http://insights-inventory.platform-ci.svc.cluster.local:8080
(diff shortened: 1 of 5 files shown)


akofink (commented):
@vkrizan Thanks for thinking about this and refactoring the consumers as well; however, I don't think we should change them, since it requires updates to all of our deployment templates. In general, we should try to make the least invasive changes possible to maintain as much stability as possible. Perhaps @Victoremepunto or @bastilian also have some thoughts whether we should refactor these settings or not.

vkrizan:
I haven't found any configuration overriding the default topics.

config/settings.yml
```diff
@@ -1,9 +1,12 @@
 app_name: 'compliance'
 path_prefix: '/api'
 old_path_prefix: '/r/insights/platform'
-platform_kafka_topic: 'platform.upload.compliance'
-platform_kafka_inventory_topic: 'platform.inventory.events'
-platform_kafka_validation_topic: 'platform.upload.validation'
+kafka_subscription_topics:
```
akofink, line 4: Should it be:
```suggestion
kafka_consumer_topics:
```
to match the `kafka_producer_topics`? That seems to make more sense to me when thinking of consumers and producers.

akofink (approved):
Ah, nice! I didn't realize we haven't overridden these settings. In that case, I think we're good! The new settings organization makes a lot of sense to me. :+1: I just had one nitpick on the naming of the consumer settings.

config/settings.yml
vkrizan: Changed.

Victoremepunto:
> @vkrizan Thanks for thinking about this and refactoring the consumers as well; however, I don't think we should change them, since it requires updates to all of our deployment templates. In general, we should try to make the least invasive changes possible to maintain as much stability as possible. Perhaps @Victoremepunto or @bastilian also have some thoughts whether we should refactor these settings or not.

@akofink  Yeh @vkrizan  is right, I didn't see anywhere in E2E where these values are replaced. we usually export a variable to the environment to overwrite those, I don't know how would this affect to that feature (now that we have 1 depth level), I guess we could use `SETTINGS__KAFKA_CONSUMER_TOPICS.INVENTORY_EVENTS` for that ? not sure how the config framework works for nested keys, to be honest :)

akofink:
@Victoremepunto I think for more complex settings like this, we can use a ConfigMap, if needed, rather than env variables.

vkrizan:
@Victoremepunto Each level from settings is divided by `__` (double underscore). So to set one topic, `SETTINGS__KAFKA_CONSUMER_TOPICS__INVENTORY_EVENTS` can be used.

PR #582, opened 2020-09-21
[RHICOMPL-901] Store SSG and RHEL minor release support matrix

vkrizan:
Initial configuration of SSG support matrix.

### DIFF
--- a/config/supported_ssg.yaml
+++ b/config/supported_ssg.yaml
@@ -1,0 +1,260 @@
+---
+revision: '2020-09-24'
+supported:
+  RHEL-6.6:
+    package: scap-security-guide-0.1.18-3.el6
+    version: 0.1.18
+    profiles:
+      C2S:
+      common:
+      CS2:
+      CSCF-RHEL6-MLS:
+      rht-ccp:
+      server:
+      stig-rhel6-server-upstream:
+      test:
+      usgcb-rhel6-server:
+  RHEL-6.7:
+    package: scap-security-guide-0.1.21-3.el6
+    version: 0.1.21
+    profiles:
+      C2S:
+      common:
+      CS2:
+      CSCF-RHEL6-MLS:
+      rht-ccp:
+      server:
+      stig-rhel6-server-upstream:
+      usgcb-rhel6-server:
+  RHEL-6.8:
+    package: scap-security-guide-0.1.28-2.el6
+    version: 0.1.28
+    profiles:
+      C2S:
+      common:
+      CS2:
+      CSCF-RHEL6-MLS:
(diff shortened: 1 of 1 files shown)


config/supported_ssg.yaml
```diff
@@ -0,0 +1,425 @@
...
+      cjis-rhel7-server:
+        policy_version: 
+      common:
+        policy_version: not versioned
+      docker-host:
+        policy_version: 
+      nist-800-171-cui:
+        policy_version: 
+      ospp-rhel7:
+        policy_version:
```
akofink, line 194: We need some way to track the change in the ref_id, like:
```suggestion
      ospp-rhel7:
        old: ospp-rhel7-server
        policy_version: 
```

akofink (commented):
Looks great, just a couple minor comments

config/supported_ssg.yaml
```diff
@@ -0,0 +1,425 @@
+---
+revision: '2020-09-21'
+supported:
+  RHEL-6.6:
+    package: scap-security-guide-0.1.18-3.el6
+    brew_uri: http://download.eng.bos.redhat.com/brewroot/vol/rhel-6/packages/scap-security-guide/0.1.18/3.el6/noarch/scap-security-guide-0.1.18-3.el6.noarch.rpm
+    version: 0.1.18
+    profiles:
+      C2S:
+        policy_version:
```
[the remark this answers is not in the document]
vkrizan: I can drop it. The terminology is based from the SSG spreadsheet.

config/supported_ssg.yaml
vkrizan: Added `old_names`, an array. Includes also those profiles that were merged into it.

config/supported_ssg.yaml
```diff
@@ -0,0 +1,444 @@
...
+      common:
+        policy_version: not versioned
+      docker-host:
+        policy_version:
+      nist-800-171-cui:
+        policy_version:
+      ospp-rhel7:
+        old_names:
+          - ospp-rhel7-server
+          - nist-cl-il-al
```
akofink, line 196: Is this correct? When was this identified by ref_id `nist-cl-il-al`?

akofink (approved):
Ok, I think we're fine to merge this. The exact fields and format may change in the future, but that's not very important for this PR. Thanks, @vkrizan!

config/supported_ssg.yaml
vkrizan: From the sheet: "Also nist-cl-il-al got merged into this one". We can drop the merged ones if needed.

PR #584, opened 2020-09-23
[RHICOMPL-965] Fix remediation import for rules w/o canonical profile

vkrizan:
Generation of `remediation_issue_id`s dependent on having a rule with
a canonical profile.  This expectation has been broken, and this fix
allows the generation of issue ids without needing a canonical profile
for a rule.

### DIFF
--- a/app/jobs/parse_report_job.rb
+++ b/app/jobs/parse_report_job.rb
@@ -64,5 +64,6 @@ class ParseReportJob
     parser.failed_rules
           .includes(:profiles)
           .collect(&:remediation_issue_id)
+          .compact
   end
 end
--- a/app/models/concerns/rule_remediation.rb
+++ b/app/models/concerns/rule_remediation.rb
@@ -5,13 +5,19 @@ module RuleRemediation
   extend ActiveSupport::Concern
 
   def remediation_issue_id
-    "ssg:rhel7|#{short_profile_ref_id}|#{ref_id}"
+    profile_ref = short_profile_ref_id
+    return unless profile_ref
+
+    "ssg:rhel7|#{profile_ref}|#{ref_id}"
   end
 
   private
 
   def short_profile_ref_id
-    profile = profiles.canonical.first
+    # FIXME: Nondetermenistic canonical selection
+    profile = profiles.select(&:canonical?)&.first
+    return unless profile
+
     short_ref_id = profile.ref_id.downcase.split(
       'xccdf_org.ssgproject.content_profile_'
     )[1]
--- a/test/models/rule_test.rb
+++ b/test/models/rule_test.rb
@@ -59,4 +59,19 @@ class RuleTest < ActiveSupport::TestCase
     )
     assert_equal [rules(:one)], Rule.canonical
   end
+
+  test 'rule generates remediation issue id' do
+    rule = rules(:one)
+    profiles(:one).update!(
+      ref_id: 'xccdf_org.ssgproject.content_profile_profile1'
+    )
+    rule.profiles << profiles(:one)
+
+    assert_equal rule.remediation_issue_id, 'ssg:rhel7|profile1|MyStringOne'
+  end
+
+  test 'rule empty remediation issue id for a rule without a profile' do
+    rule = rules(:one)
+    assert_nil rule.remediation_issue_id
+  end
 end
--- a/app/services/remediations_api.rb
+++ b/app/services/remediations_api.rb
@@ -27,7 +27,10 @@ class RemediationsAPI
   end
 
   def build_issues_list(rules)
-    ::Rule.where(id: rules).includes(:profiles).collect(&:remediation_issue_id)
+    ::Rule.where(id: rules)
+          .includes(:profiles)
+          .map(&:remediation_issue_id)
+          .compact
   end
 
   def remediations_available(response)


app/models/concerns/rule_remediation.rb
```diff
@@ -11,10 +11,14 @@ def remediation_issue_id
   private
 
   def short_profile_ref_id
-    profile = profiles.canonical.first
-    short_ref_id = profile.ref_id.downcase.split(
+    profile_ref_ids = profiles.collect(&:ref_id).to_set
```
bastilian, line 14: Why not use a `[0]` at the end and assign only one ref_id as `profile_ref_id`?

If it's `nil` it can raise the error and it would avoid the `.first`.

Or am I missing something?

vkrizan: `Set` isn't addressable. `.first` would work here, but the check (for non-uniqe refs) would not be possible.

bastilian: Why not use `.uniq.first` instead of transforming it to a set? collect should already return an array.

vkrizan: I guess a remnant from Python idioms. The result is the same.

akofink (commented):
I think this code fixes the issue, but the issue isn't about the existence of a canonical profile or not, it's about the existence of a profile at all. NB: Every non-canonical profile has a canonical parent profile. Let me try to convince you, from the master branch:

```rb
# Some rule of a canonical profile
pry(main)> Profile.canonical.first.rules.first.remediation_issue_id
  Profile Load (0.6ms)  SELECT  "profiles".* FROM "profiles" WHERE "profiles"."parent_profile_id" IS NULL ORDER BY "profiles"."id" ASC LIMIT $1  [["LIMIT", 1]]
  Rule Load (0.7ms)  SELECT  "rules".* FROM "rules" INNER JOIN "profile_rules" ON "rules"."id" = "profile_rules"."rule_id" WHERE "profile_rules"."profile_id" = $1 ORDER BY "rules"."id" ASC LIMIT $2  [["profile_id", "06acbca0-c569-4149-9ff3-749a13af464b"], ["LIMIT", 1]]
  Profile Load (0.6ms)  SELECT  "profiles".* FROM "profiles" INNER JOIN "profile_rules" ON "profiles"."id" = "profile_rules"."profile_id" WHERE "profile_rules"."rule_id" = $1 AND "profiles"."parent_profile_id" IS NULL ORDER BY "profiles"."id" ASC LIMIT $2  [["rule_id", "0ad1c2be-b713-4b73-b7f4-9c16f3543bef"], ["LIMIT", 1]]
=> "ssg:rhel7|pci-dss|xccdf_org.ssgproject.content_rule_accounts_password_pam_unix_remember"

# Some rule of a non-canonical profile
pry(main)> Profile.canonical(false).first.rules.first.remediation_issue_id
  Profile Load (0.6ms)  SELECT  "profiles".* FROM "profiles" WHERE "profiles"."parent_profile_id" IS NOT NULL ORDER BY "profiles"."id" ASC LIMIT $1  [["LIMIT", 1]]
  Rule Load (0.8ms)  SELECT  "rules".* FROM "rules" INNER JOIN "profile_rules" ON "rules"."id" = "profile_rules"."rule_id" WHERE "profile_rules"."profile_id" = $1 ORDER BY "rules"."id" ASC LIMIT $2  [["profile_id", "02567441-6a0a-4980-ac09-e8b63f395efc"], ["LIMIT", 1]]
  Profile Load (0.3ms)  SELECT  "profiles".* FROM "profiles" INNER JOIN "profile_rules" ON "profiles"."id" = "profile_rules"."profile_id" WHERE "profile_rules"."rule_id" = $1 AND "profiles"."parent_profile_id" IS NULL ORDER BY "profiles"."id" ASC LIMIT $2  [["rule_id", "0059b96b-1ec0-49f9-966e-6c827a4a1b80"], ["LIMIT", 1]]
=> "ssg:rhel7|rhelh-stig|xccdf_org.ssgproject.content_rule_audit_rules_privileged_commands_postdrop"

# Some rule that has no profile
pry(main)> Rule.where.not(id: ProfileRule.select(:rule_id)).first.remediation_issue_id
  Rule Load (5.0ms)  SELECT  "rules".* FROM "rules" WHERE "rules"."id" NOT IN (SELECT "profile_rules"."rule_id" FROM "profile_rules") ORDER BY "rules"."id" ASC LIMIT $1  [["LIMIT", 1]]
  Profile Load (0.6ms)  SELECT  "profiles".* FROM "profiles" INNER JOIN "profile_rules" ON "profiles"."id" = "profile_rules"."profile_id" WHERE "profile_rules"."rule_id" = $1 AND "profiles"."parent_profile_id" IS NULL ORDER BY "profiles"."id" ASC LIMIT $2  [["rule_id", "0004fa32-9381-42b1-928b-1c62174ca6ea"], ["LIMIT", 1]]
NoMethodError: undefined method `ref_id' for nil:NilClass
from /app/app/models/concerns/rule_remediation.rb:15:in `short_profile_ref_id'
```

The ref_id of the parent profile and the child profile will always be the same. I guess, one edge case that still isn't handled (and is difficult to handle) is a rule that is a member of multiple profiles. By checking `profile_ref_ids.count == 1`, I think we break that case.

For the missing profile case, I think we should make it as clear as possible with something like:

```rb
raise "No profiles associated with rule #{rule.id}" unless profiles.any?
```

test/models/rule_test.rb
```diff
@@ -59,4 +59,25 @@ class RuleTest < ActiveSupport::TestCase
...
+  end
+
+  test 'rule remediation issue id raises for two non-unique profiles' do
+    rule = rules(:one)
+    rule.profiles = [profiles(:one), profiles(:two)]
+
+    assert_raises StandardError do
+      rule.remediation_issue_id
+    end
+  end
```
akofink, line 82: I think we should just grab the first profile like we did before. It's unclear if the remediations are the same between profiles for the same rule, but I believe that's the case. Essentially, this would break something that's working today, even if it's not ideal. RHICOMPL-648 should resolve these sort of issues with our remediations integration.

test/services/remediations_api_test.rb
```diff
@@ -6,6 +6,8 @@
 class RemediationsAPITest < ActiveSupport::TestCase
   setup do
     @rule = rules(:one)
+    # Append non-canonical profile to the rule
+    profiles(:one).update!(parent_profile: profiles(:two))
```
akofink, line 10: This case represents a tailored profile which adds a rule that does not exist in any other profile, which is something that can happen, but it's a more special case of the broader issue of a rule not having a profile.

vkrizan:
@akofink What we should do with the case of a rule not having a profile? Remediation issue id would not be able to generated. We can either return `nil` or raise it and handle it in upper level.

Another problem is what we should do when a rule is assigned to several profiles with different `ref_id`s. A _single_ remediation issue id would not be able to be generated for a rule, unless a specific profile is passed.

app/models/concerns/rule_remediation.rb
akofink, line 14: I guess my 2 cents here, :P favor SQL over Ruby:

```suggestion
    profile_ref_ids = profiles.distinct.pluck(:ref_id)
```

vkrizan: The original implementation would not make an additional query, as profiles are prefetched in upper level(s). ;-)

bastilian: +1

akofink:
@vkrizan for rules without profiles, we can skip those issue IDs because remediations cannot accept them, unfortunately (we'll be fixing this in the future). For rules with multiple profiles, pick the first profile and use that (again, not ideal, but this is how it works today, so let's not break it until we have a better solution).

app/jobs/parse_report_job.rb
```diff
@@ -64,5 +64,6 @@ def remediation_issue_ids
     parser.failed_rules
           .includes(:profiles)
           .collect(&:remediation_issue_id)
+          .compact
```
akofink, line 67: :+1:

akofink (approved):
Thanks, @vkrizan! Works as expected, and thanks for identifying the N+1 issue. For 3840 rules in my local dev environment, generating issue IDs takes 0.75 seconds instead of over 5 seconds.

PR #589, opened 2020-09-28
[RHICOMPL-987] Do not delete orphaned hosts

akofink:
Before this change, deleting a profile with hosts associated may have
deleted the hosts as well if they were not associated with any other
profiles. After this change, we will only remove hosts from our database
during an Inventory culling message in kafka.


bastilian (approved):
Thanks @akofink!

PR #590, opened 2020-09-29
[RHICOMPL-991] Major OS ver for remediation issue id

vkrizan:
Generate remediation issue id's OS part based on its profile/benchmark. Do not hardcode the "rhel7" value, in order to support different major OS versions.

### DIFF
--- a/app/models/concerns/rule_remediation.rb
+++ b/app/models/concerns/rule_remediation.rb
@@ -4,11 +4,15 @@
 module RuleRemediation
   extend ActiveSupport::Concern
 
+  # It is recommended to prefetch profiles with their benchmarks:
+  #
+  #  `.includes(profiles: :benchmark)`
+  #
   def remediation_issue_id
     # FIXME: Nondetermenistic canonical selection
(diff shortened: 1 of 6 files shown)


app/models/concerns/rule_remediation.rb
```diff
@@ -4,11 +4,15 @@
 module RuleRemediation
   extend ActiveSupport::Concern
 
+  # It is recommended to prefecth profiles with their benchmarks:
+  #
+  #  `.includes(profiles: :benchmark)`
+  #
```
akofink, line 10: I'm not personally a huge fan of comments like this; Rails will warn of the N+1 query if you don't use `includes`, but if you want to keep it:

```suggestion
  # It is recommended to prefetch profiles with their benchmarks:
  #
  #  `.includes(profiles: :benchmark)`
  #
```

akofink (approved):
Thanks, @vkrizan!

app/models/concerns/rule_remediation.rb
vkrizan: Noted, and fixed.

PR #591, opened 2020-09-29
[RHICOMPL-994] Rescue inventory server errors on host sync

akofink:


vkrizan (approved):
LGTM.

I wonder if the error messages would better be in stderr.

akofink:
Thanks for the review! I think we intentionally log to stdout due to the way openshift handles logs for rake tasks, but we should review that assumption.

PR #594, opened 2020-09-30
feat(Rules): RHICOMPL-992 expose remeditation issue ID

vkrizan:
on REST and GraphQL API.  UI can consume it to send requests to Remediation service.  It would keep the ID generation in backend, on one place.

### DIFF
--- a/app/controllers/v1/rules_controller.rb
+++ b/app/controllers/v1/rules_controller.rb
@@ -29,6 +29,10 @@ module V1
       RuleSerializer
     end
 
+    def includes
+      [profiles: :benchmark]
+    end
+
     def search_by_id
       pundit_scope.friendly.find(params[:id])
     end
--- a/app/graphql/types/rule.rb
+++ b/app/graphql/types/rule.rb
@@ -9,6 +9,7 @@ module Types
     field :id, ID, null: false
     field :title, String, null: false
     field :ref_id, String, null: false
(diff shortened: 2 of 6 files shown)


app/graphql/types/rule.rb
```diff
@@ -9,6 +9,7 @@ class Rule < Types::BaseObject
     field :id, ID, null: false
     field :title, String, null: false
     field :ref_id, String, null: false
+    field :remediation_issue_id, String, null: true
```
vkrizan, line 12: GraphQL might be slower after this. We might consider adding a preload of related items, if anyone knows the recommended way to do that.

app/controllers/v1/rules_controller.rb
```diff
@@ -29,6 +29,10 @@ def serializer
       RuleSerializer
     end
 
+    def scope_search
+      super.includes(profiles: :benchmark)
+    end
```
akofink, line 34:
```suggestion
    def includes
      [profiles: :benchmark]
    end
```

akofink (approved):
Not sure why we never reviewed+merged this. Looks great :+1:

app/controllers/v1/rules_controller.rb
vkrizan: :+1:

akofink (approved):
At long last! :laughing: Thanks, @vkrizan!

PR #595, opened 2020-10-01
[RHICOMPL-927] REST API for supported RHEL and SSG version

vkrizan:
Mapping of RHEL minor version release to a supported SSG version package available on REST API.

Example of one record:
```
{
  "id": "RHEL-7.3",
  "type": "supported_ssg",
  "attributes": {
    "package": "scap-security-guide-0.1.30-5.el7_3",
    "version": "0.1.30",
    "os_major_version": "7",
    "os_minor_version": "3",
    "profiles": [
      "C2S",
      "cjis-rhel7-server",
      "common",
      "nist-cl-il-al",
      "ospp-rhel7-server",
      "pci-dss",
      "rht-ccp",
      "standard",
      "stig-rhel7-server-gui-upstream",
      "stig-rhel7-server-upstream",
      "stig-rhel7-workstation-upstream"
    ]
  }
}
```

### DIFF
--- a/app/controllers/v1/supported_ssgs_controller.rb
+++ b/app/controllers/v1/supported_ssgs_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+module V1
+  # API for Supported SSGs mapped to RHEL minor versions
+  class SupportedSsgsController < ApplicationController
+    def index
+      render_json supported_ssgs
+    end
+
+    private
+
+    def metadata(opts = {})
+      opts[:total] ||= supported_ssgs.count
+      {
+        meta: {
+          total: opts[:total],
+          revision: resource.revision
+        }
+      }
+    end
+
+    def resource
+      SupportedSsg
+    end
+
+    def serializer
+      SupportedSsgSerializer
+    end
+
+    def supported_ssgs
+      @supported_ssgs ||= resource.all
+    end
+  end
+end
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -38,7 +38,8 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
     end
 
     def raw_supported
-      YAML.load_file(self::SUPPORTED_FILE)
+      # cached for the whole runtime
+      @raw_supported ||= YAML.load_file(self::SUPPORTED_FILE)
     end
   end
 end
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -30,7 +30,6 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
 
     def map_attributes(values)
       attrs = values.slice(*members.map(&:to_s))
-      attrs['profiles'] = attrs['profiles']&.keys
       attrs.symbolize_keys
     end
 
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+# rubocop:disable Metrics/BlockLength
+SupportedSsg = Struct.new(:id, :package, :version, :profiles,
+                          :os_major_version, :os_minor_version,
+                          keyword_init: true) do
+  self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
+
+  class << self
+    def all
+      raw_supported['supported'].map do |rhel, values|
+        major, minor = os_version(rhel)
+
+        new(
+          id: rhel,
+          os_major_version: major,
+          os_minor_version: minor,
+          **map_attributes(values)
+        )
+      end
+    end
+
+    def revision
+      raw_supported['revision']
+    end
+
+    private
+
+    def map_attributes(values)
+      attrs = values.slice(*members.map(&:to_s))
+      attrs['profiles'] = attrs['profiles']&.keys
+      attrs.symbolize_keys
+    end
+
+    def os_version(rhel)
+      rhel.scan(/(\d+)\.(\d+)$/)[0]
+    end
+
+    def raw_supported
+      YAML.load_file(self::SUPPORTED_FILE)
+    end
+  end
+end
+# rubocop:enable Metrics/BlockLength
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,28 +3,26 @@
 Rails.application.routes.draw do
   def draw_routes(prefix)
     scope "#{prefix}/#{Settings.app_name}" do
-      namespace :v1 do
-        resources :benchmarks, only: [:index, :show]
-        resources :business_objectives, only: [:index, :show]
-        resources :profiles do
-          member do
-            get 'tailoring_file'
+      concern :rest_api_v1 do
+        scope module: 'v1' do
+          resources :benchmarks, only: [:index, :show]
+          resources :business_objectives, only: [:index, :show]
+          resources :profiles do
+            member do
+              get 'tailoring_file'
+            end
           end
-        end
-        resources :rule_results, only: [:index]
-        resources :systems, only: [:index, :show, :destroy]
-        resources :rules, only: [:index, :show]
-      end
-      resources :benchmarks, controller: 'v1/benchmarks', only: [:index, :show]
-      resources :business_objectives, controller: 'v1/business_objectives', only: [:index, :show]
-      resources :profiles, controller: 'v1/profiles' do
-        member do
-          get 'tailoring_file'
+          resources :rule_results, only: [:index]
+          resources :systems, only: [:index, :show, :destroy]
+          resources :rules, only: [:index, :show]
         end
       end
-      resources :rule_results, controller: 'v1/rule_results', only: [:index]
-      resources :systems, controller: 'v1/systems', only: [:index, :show, :destroy]
-      resources :rules, controller: 'v1/rules', only: [:index, :show]
+
+      concerns :rest_api_v1
+      scope 'v1', as: 'v1' do
+        concerns :rest_api_v1
+      end
+
       mount Rswag::Api::Engine => '/',
         as: "#{prefix}/#{Settings.app_name}/rswag_api"
       mount Rswag::Ui::Engine => '/',
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,7 @@ Rails.application.routes.draw do
           resources :rule_results, only: [:index]
           resources :systems, only: [:index, :show, :destroy]
           resources :rules, only: [:index, :show]
+          resources :supported_ssgs, only: [:index]
         end
       end
 
--- a/spec/integration/supported_ssgs_spec.rb
+++ b/spec/integration/supported_ssgs_spec.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'swagger_helper'
+
+describe 'SupportedSsgs API' do
+  fixtures :accounts
+
+  path "#{Settings.path_prefix}/#{Settings.app_name}/supported_ssgs" do
+    get 'List all supported SSGs' do
+      tags 'supported_ssg'
+      description 'List all supported SSGs mapped to RHEL minor version'
+      operationId 'ListSupportedSsgs'
+
+      content_types
+      auth_header
+
+      include_param
+
+      response '200', 'lists all supported_ssgs requested' do
+        let(:'X-RH-IDENTITY') { encoded_header(accounts(:test)) }
+        let(:include) { '' } # work around buggy rswag
+        schema type: :object,
+               properties: {
+                 meta: ref_schema('metadata'),
+                 data: {
+                   type: :array,
+                   items: {
+                     properties: {
+                       type: { type: :string },
+                       id: { type: :string },
+                       attributes: ref_schema('supported_ssg')
+                     }
+                   }
+                 }
+               }
+
+        after { |e| autogenerate_examples(e) }
+
+        run_test!
+      end
+    end
+  end
+end
--- a/swagger/v1/openapi.json
+++ b/swagger/v1/openapi.json
@@ -1922,6 +1922,460 @@
         }
       }
     },
+    "/api/compliance/supported_ssgs": {
+      "get": {
+        "summary": "List all supported SSGs",
+        "tags": [
+          "supported_ssg"
+        ],
+        "description": "List all supported SSGs mapped to RHEL minor version",
+        "operationId": "ListSupportedSsgs",
+        "parameters": [
+          {
+            "name": "X-RH-IDENTITY",
+            "in": "header",
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "include",
+            "in": "query",
+            "type": "string",
+            "required": false,
+            "schema": {
+              "type": "string"
+            },
+            "description": "A comma seperated list of resources to include in the response"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "lists all supported_ssgs requested",
+            "examples": {
+              "application/vnd.api+json": {
+                "data": [
+                  {
+                    "id": "RHEL-6.6",
+                    "type": "supported_ssg",
+                    "attributes": {
+                      "package": "scap-security-guide-0.1.18-3.el6",
+                      "version": "0.1.18",
+                      "os_major_version": "6",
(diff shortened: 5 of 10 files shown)


app/controllers/v1/supported_ssgs_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      rhel.scan(/(\d+)\.(\d+)$/)[0]
+    end
+
+    def raw_supported
+      @raw_supported ||= YAML.load_file(supported_file)
+    end
+
+    def supported_file
+      Rails.root.join('config/supported_ssg.yaml')
+    end
```
akofink, line 65: Looking great, @vkrizan! One initial thing I see here: I think it'd be good to write a pseudo model or service that reads this config and defines the struct, rather than putting it all in this controller.

vkrizan: Any advise where to introduce it? In services or models (there are only db ones)?

akofink: It's fine to put it with the models IMO. The only thing to leave off is the inheritance from `ApplicationRecord`.

vkrizan: Done, and ready for review.

app/controllers/v1/supported_ssgs_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+module V1
+  # API for Supported SSGs mapped to RHEL minor versions
+  class SupportedSsgsController < ApplicationController
+    def index
+      render_json scope_search
```
akofink, line 9: It's a bit weird to use `scope_search` here since `SupportedSsg` isn't backed by the DB, but I think we may be able to still define scoped searches against it, potentially filtering by OS minor version or something in the future, but I've never tried to use it with a model that isn't backed by ActiveRecord.

app/controllers/v1/supported_ssgs_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def metadata(opts = {})
+      opts[:total] ||= scope_search.count
+      {
+        meta: {
+          total: opts[:total],
+          revision: resource.revision
+        }
+      }
+    end
```
akofink, line 22: I like the idea of adding the revision in the metadata; it seems appropriate. :+1: It'd be good to get search, pagination, and links working like the other endpoints as well. Pagination isn't really necessary, but I'd push for it just for the sake of consistency.

app/models/supported_ssg.rb
```diff
@@ -0,0 +1,48 @@
...
+      rhel.scan(/(\d+)\.(\d+)$/)[0]
+    end
+
+    def raw_supported
+      YAML.load_file(supported_file)
+    end
+
+    def supported_file
+      Rails.root.join('config/supported_ssg.yaml')
+    end
```
akofink, line 45: This arguably should be a constant.

app/models/supported_ssg.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+# rubocop:disable Metrics/BlockLength
+SupportedSsg = Struct.new(:id, :package, :version, :profiles,
+                          :os_major_version, :os_minor_version,
+                          keyword_init: true) do
```
akofink, line 8: Wow, neat :) TIL structs take blocks

swagger/v1/openapi.json
```diff
@@ -2535,6 +2535,37 @@
             "$ref": "#/components/schemas/relationship_collection"
           }
         }
+      },
```
akofink, line 2538: Regenerate this file (with automatic examples) like this:

```
podman exec -it compliance-backend_rails_1 /usr/local/bin/ruby -I/usr/local/bundle/gems/rspec-core-3.9.2/lib:/usr/local/bundle/gems/rspec-support-3.9.3/lib /usr/local/bundle/gems/rspec-core-3.9.2/exe/rspec --pattern spec/requests/\*\*/\*_spec.rb,\ spec/api/\*\*/\*_spec.rb,\ spec/integration/\*\*/\*_spec.rb --format Rswag::Specs::SwaggerFormatter --order defined
```

Sorry, it's pretty hacky; something we need to improve. Last time I looked into it, there were some issues with rswag preventing a simpler command, but I can't remember the details.

test/controllers/v1/supported_ssgs_controller_test.rb
```diff
@@ -0,0 +1,19 @@
...
+  class SupportedSsgsControllerTest < ActionDispatch::IntegrationTest
+    setup do
+      SupportedSsgsController.any_instance.stubs(:authenticate_user)
+    end
+
+    context 'index' do
+      should 'lists all supported SSGs' do
+        get v1_supported_ssgs_url
+
+        assert_response :success
```
akofink, line 15: Nit: Maybe also check a record matches the yaml?

akofink (commented):
Looks great! A few comments inline. Also:

I don't see the path for this resource in the openapi.json output:

```js
GET api/compliance/v1/openapi.json

{
  "openapi": "3.0.3",
  "info": {…},
  "paths": {   
    "/api/compliance/benchmarks": {…},
    "/api/compliance/benchmarks/{id}": {…},
    "/api/compliance/business_objectives": {…},
    "/api/compliance/business_objectives/{id}": {…},
    "/api/compliance/profiles": {…},
    "/api/compliance/profiles/{id}": {…},
    "/api/compliance/rule_results": {…},
    "/api/compliance/rules": {…},
    "/api/compliance/rules/{id}": {…},
    "/api/compliance/systems": {…},
    "/api/compliance/systems/{id}": {…}
  },
  "components": {…}
}
```

You will need to write a spec file for this new resource, like `spec/integration/profiles_spec.rb`, to complete the OpenAPI v3 generation.

app/controllers/v1/supported_ssgs_controller.rb
vkrizan: There is definitely room for improvement. There wasn't a way to provide an API endpoint without a db resource. 

The `scope_search` definition is/was needed for dependent methods. I can check if it is still required, as there is `metadata` override.

app/controllers/v1/supported_ssgs_controller.rb
vkrizan: We might introduce search and pagination when we enhance the framework to work with non-db resources. I've played with it a bit, and it did not fit the current state (e.g. the `.count` only counted paginated items ...). It might be a good idea to reconsider the use of a db model with a sync job.

app/models/supported_ssg.rb
```diff
@@ -0,0 +1,48 @@
...
+      attrs['profiles'] = attrs['profiles']&.keys
+      attrs.symbolize_keys
+    end
+
+    def os_version(rhel)
+      rhel.scan(/(\d+)\.(\d+)$/)[0]
+    end
+
+    def raw_supported
+      YAML.load_file(supported_file)
```
bastilian, line 40: Could this be cached to reduce I/O?

vkrizan: I cannot think of a simple way to do it. This is a classmethod, which would cache it for the duration of the app. Would you suggest a better way?

The file is access once for a general data retrieval. The API makes another call to get the revision. We can return the revision along side the collection to reduce the access. I believe that two I/O reads of the YAML are acceptable.

bastilian: Are these two calls per request or per boot?

If it is per request it would make sense to cache it with `@raw_supported ||= ...`.

vkrizan: Two calls per request.

As I've mentioned `@raw_supported ||=` would store it for the whole duration of the app, disallowing to replace supported SSGs in runtime (via configmap).

bastilian: Having it cached for the duration of the app should be fine. The versions won't change often.
Too many request could cause instances to lock up because of I/O issues.

app/models/supported_ssg.rb
vkrizan: Done.

swagger/v1/openapi.json
vkrizan: I've added the spec and run the command.

`bundle exec rspec +args` worked for me. We might add a rake task or improve `rswag:specs:swaggerize`.

test/controllers/v1/supported_ssgs_controller_test.rb
vkrizan: That would require to copy the algorithm over :slightly_smiling_face: . The spec also checks the schema.

app/models/supported_ssg.rb
vkrizan: Do I/O issues happen often on static files in deployed application? 

I don't mind adding the memoization. It looked to me as premature optimization.

vkrizan: How about `Rails.cache` with invalidation?

vkrizan: @akofink What do you think about the memoization for the whole runtime? It should be good enough for production, as change of a configmap would require a restart/re-deployment.

app/controllers/v1/supported_ssgs_controller.rb
vkrizan: Removed the use of `scope_search`.

app/models/supported_ssg.rb
akofink: I think there might be a way to parse and cache this file with [ruby config](https://github.com/rubyconfig/config) `Settings`.

vkrizan: I've added the simpler memoization. Not sure how the config would help here. The Rails.cache could, but it is configured differently on each environment (null cache by default in dev).

app/models/supported_ssg.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def revision
+      raw_supported['revision']
+    end
+
+    private
+
+    def map_attributes(values)
+      attrs = values.slice(*members.map(&:to_s))
+      attrs['profiles'] = attrs['profiles']&.keys
```
akofink, line 33: We will need to be able to get at `old_names` as well; probably not in the API, but certainly in other backend code. This logic to remove certain information from the config to serve in the API can happen in the serializer directly.

config/routes.rb
```diff
@@ -3,28 +3,27 @@
...
   def draw_routes(prefix)
     scope "#{prefix}/#{Settings.app_name}" do
-      namespace :v1 do
-        resources :benchmarks, only: [:index, :show]
-        resources :business_objectives, only: [:index, :show]
-        resources :profiles do
-          member do
-            get 'tailoring_file'
+      concern :rest_api_v1 do
+        scope module: 'v1' do
```
akofink, line 7: I hope this doesn't come back to bite us. Is this some newer way to define namespaces, or?

spec/integration/supported_ssgs_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      content_types
+      auth_header
+      pagination_params
+      search_params
+
+      include_param
+
+      response '200', 'lists all supported_ssgs requested' do
+        let(:'X-RH-IDENTITY') { encoded_header(accounts(:test)) }
+        let(:include) { '' } # work around buggy rswag
```
akofink, line 23: Not needed for this API

spec/integration/supported_ssgs_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  path "#{Settings.path_prefix}/#{Settings.app_name}/supported_ssgs" do
+    get 'List all supported SSGs' do
+      tags 'supported_ssg'
+      description 'List all supported SSGs mapped to RHEL minor version'
+      operationId 'ListSupportedSsgs'
+
+      content_types
+      auth_header
+      pagination_params
+      search_params
```
akofink, line 17: Not needed for this endpoint

akofink (commented):
Looks great! Just a couple more small things, and we're good to merge.

config/routes.rb
vkrizan: Newer way is to use `namespace :v1`, but it affects link generators (`_url` methods) for endpoints outside of the `/v1` path.

akofink: So what's the difference exactly? Today we have (i.e.) `profiles_url` and `v1_profiles_url`.

vkrizan: The `scope module: 'v1'` is necessary to use `V1` module controllers. The previous `namespace :v1` did set the path (`/v1`) as well as the module (`V1`). It is not possible to use it for the concern, as it could not be set on toplevel path.

`profiles_url` points to `/profiles` and `v1_profiles_url` points to `/v1/profiles`, as before. With `namespace` there would not be the `profiles_url`.

app/models/supported_ssg.rb
vkrizan: Should I add it to the API as well, or a new (non-exposed) property would be fine?

spec/integration/supported_ssgs_spec.rb
vkrizan: It is, unfortunately (`URI::InvalidURIError`) :confused:

app/models/supported_ssg.rb
akofink: I don't think it's necessary for the API.

config/routes.rb
akofink: Ok, after looking at this some more, this seems like the right way to do it :) Sorry for the noise! It's much better now with the concern.

app/models/supported_ssg.rb
vkrizan: Kept the profiles structure of YAML in the model. Moved transformation to serializer.

spec/integration/supported_ssgs_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  path "#{Settings.path_prefix}/#{Settings.app_name}/supported_ssgs" do
+    get 'List all supported SSGs' do
+      tags 'supported_ssg'
+      description 'List all supported SSGs mapped to RHEL minor version'
+      operationId 'ListSupportedSsgs'
+
+      content_types
+      auth_header
+
+      include_param
```
akofink, line 17: This param is not used, but we accept it. I guess it's okay to keep it documented, despite this resource not having any relationships. What do you think? It's this `include` param from the JSON spec: https://jsonapi.org/format/#fetching-includes

app/models/supported_ssg.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def revision
+      raw_supported['revision']
+    end
+
+    private
+
+    def map_attributes(values)
+      attrs = values.slice(*members.map(&:to_s))
```
akofink, line 32: Slightly easier to use (but optional):

```suggestion
      attrs = values.slice(*members.map(&:to_sym))
```

app/models/supported_ssg.rb
```diff
@@ -0,0 +1,46 @@
...
+
+require 'yaml'
+
+# rubocop:disable Metrics/BlockLength
+SupportedSsg = Struct.new(:id, :package, :version, :profiles,
+                          :os_major_version, :os_minor_version,
+                          keyword_init: true) do
+  self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
+
+  class << self
```
akofink, line 11: For the next part of this work, we will need to index into this collection by RHEL minor version and get the supported SSG version. We will need to implement `[]` or `find`. We can add it now or later, up to you.

akofink (approved):
Thanks, @vkrizan! I noticed a few more things, but they're not blockers.

app/models/supported_ssg.rb
vkrizan: `members` are symbols already. 
```
> SupportedSsg.members.map(&:to_sym)
=> [:id, :package, :version, :profiles, :os_major_version, :os_minor_version]
```

The keys in `values` are strings, therefore the conversion. Symbolizing is done after sanitation.

app/models/supported_ssg.rb
vkrizan: Yes, I have that in mind.

spec/integration/supported_ssgs_spec.rb
vkrizan: Oh, I thought the include is for the attributes and not related resources.

vkrizan: I'll keep it. :slightly_smiling_face:

PR #596, opened 2020-10-02
[RHICOMPL-751] Autoreconnect to postgress on PG::UnableToSend

vkrizan:
Raw Draft

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -27,6 +27,18 @@ class Profile < ApplicationRecord
   after_rollback :destroy_orphaned_business_objective
 
   class << self
+    def find_by_sql(*args, **kwargs)
+      super(*args, **kwargs)
+    rescue ActiveRecord::StatementInvalid => e
+      if e.message =~ /^PG::UnableToSend:/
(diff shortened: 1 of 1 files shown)


app/models/profile.rb
```diff
@@ -27,6 +27,18 @@ class Profile < ApplicationRecord
   after_rollback :destroy_orphaned_business_objective
 
   class << self
+    def find_by_sql(*args, **kwargs)
+      super(*args, **kwargs)
+    rescue ActiveRecord::StatementInvalid => e
+      if e.message =~ /^PG::UnableToSend:/
```
vkrizan, line 33: This should use `e.cause` instead.

akofink:
This one's pretty stale. Shall we close it?

vkrizan:
@akofink I'd like to finish this at some point. Sorry wanted to do that last Friday.

vkrizan:
I've tried some approaches. It is possible to inject connection resets into the ORM, however it is not ideal. We should rather invest time to create a health endpoint that would report if the connection drops to let OpenShift to restart the container.

PR #597, opened 2020-10-05
[RHICOMPL-544] Validate OS major version of uploaded report

vkrizan:
Raise `OSVersionMismatch` if the major OS version does not match benchmark's.

Example error:
```
ERROR: Cannot parse report: OS major version (7) does not match with benchmark xccdf_org.ssgproject.content_benchmark_RHEL-8 (8)
```

### DIFF
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -4,6 +4,8 @@
 class MissingIdError < StandardError; end
 # Error to raise if the format of the report is wrong
 class WrongFormatError < StandardError; end
+# Error to raise if the OS version does not match benchmark's
+class OSVersionMismatch < StandardError; end
 
 # Takes in a path to an Xccdf file, returns all kinds of properties about it
 # and saves it in our database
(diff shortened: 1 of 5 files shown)


app/services/xccdf_report_parser.rb
```diff
@@ -33,9 +35,21 @@ def valid_message_format?(message)
...
+            " benchmark #{benchmark.ref_id} (#{benchmark.os_major_version})"
+    end
+    # rubocop:enable Style/GuardClause
+  end
+
   def save_all
     Host.transaction do
       save_all_benchmark_info
+      save_all_host_info
+      check_os_version
```
akofink, line 52: We could move this to the top and check `@op_benchmark.id` instead of `@benchmark.ref_id` to fail faster in this situation.

akofink (approved):
Looks good, and works as expected. Thanks, @vkrizan! I have just one potential improvement.

app/services/xccdf_report_parser.rb
vkrizan: Not sure how the `id` would help here, but I'm worried that the `os_major_version` parsing is only available on our `Benchmark` model.

akofink: `@op_benchmark.id` is actually the benchmark's `ref_id`. It's called `id` in the SCAP data model (kind of confusing, I know). It's a good point that the OS parsing is only in the Benchmark model, but it wouldn't be difficult to make the regex a constant, like `REF_ID_OS_PATTERN = /(?<=RHEL-)\d/` and then use `@op_benchmark.id[Xccdf::Benchmark::REF_ID_OS_PATTERN]` here.

vkrizan: Taking the regex out of the models would do tight coupling, which I'd rather avoid. Better way would be to initialize it into `Benchmark`, do the OS check, and then store the benchmark.

akofink: Works for me

vkrizan: It just need to do `save_host` prior to check.

PR #600, opened 2020-10-12
[RHICOMPL-948] Sync SSGs out of supported matrix 

vkrizan:
Syncs uptream SSGs from the supported matrix mapped to RHEL minor version.

Note, that downstream might backport rules to the older versions of SSGs. Currently this app handles/syncs SSGs from upstream, therefore it needs to store both versions.  Downstream version for checking supportability and upstream for syncing benchmarks.

### DIFF
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -3,7 +3,7 @@
 require 'yaml'
 
 # rubocop:disable Metrics/BlockLength
-SupportedSsg = Struct.new(:id, :package, :version, :profiles,
+SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
                           :os_major_version, :os_minor_version,
                           keyword_init: true) do
   self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -26,6 +26,13 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
       raw_supported['revision']
     end
 
+    # Collection of supported SSGs that have ZIP available upstream
+    def available_upstream
+      all.reject do |ssg|
+        ssg.upstream_version&.upcase == 'N/A'
+      end
+    end
+
     private
 
     def map_attributes(values)
--- a/config/supported_ssg.yaml
+++ b/config/supported_ssg.yaml
@@ -4,6 +4,7 @@ supported:
   RHEL-6.6:
     package: scap-security-guide-0.1.18-3.el6
     version: 0.1.18
+    upstream_version: N/A
     profiles:
       C2S:
       common:
@@ -17,6 +18,7 @@ supported:
   RHEL-6.7:
     package: scap-security-guide-0.1.21-3.el6
     version: 0.1.21
+    upstream_version: N/A
     profiles:
       C2S:
       common:
@@ -74,6 +76,7 @@ supported:
   RHEL-7.1:
     package: scap-security-guide-0.1.19-2.el7
     version: 0.1.19
+    upstream_version: N/A
     profiles:
       rht-ccp:
   RHEL-7.2:
--- a/config/supported_ssg.yaml
+++ b/config/supported_ssg.yaml
@@ -206,7 +206,9 @@ supported:
           - stig-rhel7-disa
   RHEL-7.9:
     package: scap-security-guide-0.1.49-13.el7
+    # RHEL 7 with CIS backported to 0.1.49 in downstream
     version: 0.1.49
+    upstream_version: 0.1.50
     profiles:
       anssi_nt28_enhanced:
       anssi_nt28_high:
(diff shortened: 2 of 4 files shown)


app/models/supported_ssg.rb
```diff
@@ -3,10 +3,11 @@
 require 'yaml'
 
 # rubocop:disable Metrics/BlockLength
-SupportedSsg = Struct.new(:id, :package, :version, :profiles,
+SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
                           :os_major_version, :os_minor_version,
                           keyword_init: true) do
   self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
+  self::OLDEST_UPSTREAM_VERSION = '0.1.24'
```
akofink, line 10: We could mark the versions that don't have any upstream counterpart in the config itself to avoid using this constant, perhaps something like `upstream_version: ''` or `upstream_version: 'N/A'`, `upstream_version: 0.0.0` :man_shrugging: What do you think?

lib/tasks/import_datastream.rake
```diff
@@ -29,9 +29,8 @@ namespace :ssg do
     # DATASTREAM_FILENAMES from openscap_parser's ssg:sync
     begin
       Rails.cache.delete('latest_supported_benchmarks')
-      ENV['DATASTREAMS'] = ::Xccdf::Benchmark::
-        LATEST_SUPPORTED_VERSIONS.map do |ref_id, version|
```
akofink, line 33: Could we get rid of this constant altogether and just use the latest version per major OS in the config?

akofink (commented):
Looks great! Just a couple of questions.

akofink:
Also, sorry for the delayed review. GitHub does not notify when a PR moves from draft to ready; please feel free to ping me on PRs when they're ready in the future.

akofink:
@vkrizan we're finally ready to merge this one. There are still some outstanding comments to address.

app/models/supported_ssg.rb
vkrizan: Great suggestion.

lib/tasks/import_datastream.rake
vkrizan: Good question, but a separate concern. Captured in RHICOMPL-1121.

app/models/supported_ssg.rb
```diff
@@ -26,6 +26,13 @@ def revision
       raw_supported['revision']
     end
 
+    # Collection of supported SSGs that have ZIP available upstream
+    def available_upstream
+      all.reject do |ssg|
+        ssg.upstream_version&.upcase == 'N/A'
+      end
```
akofink, line 33: I really would have just preferred nil here. It makes the code and the SSG support config nicer, too

config/supported_ssg.yaml
```diff
@@ -4,6 +4,7 @@ supported:
   RHEL-6.6:
     package: scap-security-guide-0.1.18-3.el6
     version: 0.1.18
+    upstream_version: N/A
```
akofink, line 7: There's really no need to add these if we just assume `nil` means the upstream_version is the same as the downstream_version.

akofink (approved):
Looks good, thanks! Just one remaining nitpick :)

config/supported_ssg.yaml
vkrizan: The problem is that `nil` is means `version` (when upstream and downstream version matches). We'll get rid of this once we start downloading downstream pkgs.

akofink: Sorry, I don't follow. `nil` would mean the upstream and downstream versions are the same, like you said.

vkrizan: I guess there is a misunderstanding. What was your initial concern around using "N/A", which means there is no upstream package (ZIP)?

akofink: Since it's not part of the original config given to us by OpenSCAP, we could have just assumed a `nil` or missing upstream_version meant that the downstream and upstream version were the same. Using 'N/A', we have to go add that to the config every time it's updated (every RHEL release). There's really no reason to use a special string ('N/A') instead of nil/missing value, but there are good reasons against it - mostly because it simplifies the code but also because we only have to modify the records that have different upstream versions rather than all the records.

vkrizan: The `upstream_version` is defined only when it differs from downstream version. The `upstream_version` was set by us in this PR. If it is not defined (missing) or set to nil the subsequent code will pick value `version`. Note, that not defined is the same as `nil`, because of a default value in the `Struct`. The "N/A" has a different purpose.

I agree that it is not very ideal. We could set the `upstream_version` everywhere, and proclaim `nil` as the new "N/A". I don't have more information on the upstream vs. downstream versions, other than on those few defined here.

akofink: > The "N/A" has a different purpose.
Oh, what does 'N/A' mean? That the version doesn't exist at all upstream?

vkrizan: Yes, that the version does not have a ZIP available to download from upstream.
Based on your suggestion from https://github.com/RedHatInsights/compliance-backend/pull/600#discussion_r504970115

What did I miss? :smile:

akofink: Ok, makes sense. Sorry for the confusion! :)

PR #602, opened 2020-10-14
[RHICOMPL-1040] Create policies, policy_hosts tables

akofink:
Shuffle indices, and add profile-policy reference

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_08_04_012906) do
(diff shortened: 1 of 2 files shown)


akofink:
Updated to remove unique index changes

db/schema.rb
```diff
@@ -165,6 +187,7 @@
     t.uuid "host_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "supported"
```
vkrizan, line 190: This one should be removed. Not part of the migration.

akofink: Updated! Sorry :)

vkrizan (approved):

PR #604, opened 2020-10-14
[RHICOMPL-1042] Add policy related non-breaking model changes

akofink:
Adapted all non-breaking model changes from https://github.com/RedHatInsights/compliance-backend/pull/572

Smoke test result run locally (3 failures related to rbac that also exist on master):

```
3 failed, 85 passed, 1 skipped, 126 deselected, 333 warnings in 329.08s (0:05:29)
```

### DIFF
--- a/app/models/concerns/profile_hosts.rb
+++ b/app/models/concerns/profile_hosts.rb
@@ -5,7 +5,10 @@ module ProfileHosts
   extend ActiveSupport::Concern
 
   included do
+    has_many :test_result_hosts, through: :test_results, source: :host
     has_many :profile_hosts, dependent: :destroy
+    has_many :policy_hosts, through: :policy_object
+    has_many :assigned_hosts, through: :policy_hosts, source: :host
     has_many :hosts, through: :profile_hosts, source: :host
 
     def update_hosts(new_host_ids)
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+# Compliance policy
+class Policy < ApplicationRecord
+  DEFAULT_COMPLIANCE_THRESHOLD = 100.0
+  PROFILE_ATTRS = %w[name description account_id compliance_threshold
+                     business_objective_id].freeze
+
+  has_many :profiles, dependent: :destroy, inverse_of: :policy_object
+  has_many :benchmarks, through: :profiles
+  has_many :test_results, through: :profiles, dependent: :destroy
+
+  has_many :policy_hosts, dependent: :destroy
+  has_many :hosts, through: :policy_hosts, source: :host
+
+  belongs_to :business_objective, optional: true
+  belongs_to :account
+
+  validates :compliance_threshold, numericality: true
+  validates :account, presence: true
+  validates :name, presence: true, uniqueness: true
+
+  def self.attrs_from(profile:)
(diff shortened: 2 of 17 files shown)


app/models/concerns/profile_hosts.rb
```diff
@@ -5,7 +5,10 @@ module ProfileHosts
   extend ActiveSupport::Concern
 
   included do
+    has_many :test_result_hosts, through: :test_results, source: :host
```
akofink, line 8: This will be the new way to get hosts after we remove ProfileHost

app/models/policy.rb
```diff
@@ -0,0 +1,58 @@
...
+  belongs_to :business_objective, optional: true
+  belongs_to :account
+
+  validates :compliance_threshold, numericality: true
+  validates :account, presence: true
+  validates :name, presence: true, uniqueness: true
+
+  after_destroy :destroy_orphaned_business_objective
+  after_update :destroy_orphaned_business_objective
+  after_rollback :destroy_orphaned_business_objective
```
vkrizan, line 25: Observation: These hooks might start removing b. objectives that are assigned to a Profile.

app/models/profile.rb
```diff
@@ -64,12 +67,12 @@ def destroy_orphaned_business_objective
...
   end
 
-  def clone_to(account: nil, host: nil, external: true)
+  def clone_to(account: nil, host: nil, external: true, policy: nil)
     new_profile = in_account(account)
     if new_profile.nil?
       (new_profile = dup).update!(account: account, hosts: [host],
                                   parent_profile: self,
-                                  external: external)
+                                  external: external, policy_object: policy)
```
vkrizan, line 75: This is a new addition that should not be in these set of changes.

app/models/policy.rb
akofink: Yes, we want that. If a BO exists on only one profile and the profile is destroyed, the BO should also be destroyed

vkrizan: Yes, but not until the Policy is used. It just might get confusing if the BOs will start disappearing when testing/playing with it. Feel free to keep it.

akofink: I'll remove it for now.

vkrizan (approved):

app/models/policy.rb
akofink: Ok, updated. The tests for this call the method directly instead of relying on after_*, which I've removed for now.

app/models/profile.rb
```diff
@@ -18,6 +18,9 @@ class Profile < ApplicationRecord
   validates :ref_id, uniqueness: {
     scope: %i[account_id benchmark_id external]
   }, presence: true
+  validates :ref_id, uniqueness: {
+    scope: %i[account_id benchmark_id policy_id]
```
vkrizan, line 22: @akofink Wouldn't this uniqueness create a problem when there is an existing `external: true` profile and a user wants to create a new internal profile for the same benchmark? I'll try to test this.

vkrizan: Confirmed, this is a breaking change.

I'm getting:
```
Response not successful: Received status code 422
#<ActiveRecord::RecordInvalid: Validation failed: Ref has already been taken>
```

akofink: Good catch!

PR #606, opened 2020-10-15
[RHICOMPL-1042] Profile is not yet uniq by a policy

vkrizan:
Having a profile unique by a policy before the policy model is fully
used breaks creation of new internal profiles that already have an
`external: true` variant.

Response with the constraint:
```
Response not successful: Received status code 422
<ActiveRecord::RecordInvalid: Validation failed: Ref has already been taken>
```

References: https://github.com/RedHatInsights/compliance-backend/pull/604#discussion_r505419339

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -18,9 +18,6 @@ class Profile < ApplicationRecord
   validates :ref_id, uniqueness: {
     scope: %i[account_id benchmark_id external]
   }, presence: true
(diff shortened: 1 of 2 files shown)


akofink (approved):
Thanks, and nice catch! Apparently smoke tests do not cover this.

akofink:
There's a test in profile_test.rb that needs to be removed for this validation, and then the tests will pass.

PR #610, opened 2020-10-15
[RHICOMPL-662] Migration of old profile data into new policy records

vkrizan:
Creates a service for copying profiles into Policies.
The service creates new Policy models for Profiles that are internal and
that do not have a Policy.  It copies some of the attributes from the
profile.  It assigns the same host set from the internal profile.

The first pass for migration of profile data to policies is mostly intended for testing the migration.
Production should be migrated only when the whole feature is complete.

**IMPORTANT**

We first need to decide whether to do a simpler migration (this PR) or migrate it by using a virtual policy.

Original virtual policy lookup for external profiles:
```
Profile.includes(:benchmark)
        .older_than(created_at)
        .find_by(account: account_id, external: false, ref_id: ref_id,
                benchmarks: { ref_id: benchmark.ref_id })
```

### DIFF
--- a/app/services/copy_profiles_to_policies.rb
+++ b/app/services/copy_profiles_to_policies.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# A service class to copy Profile data to new Policy objects
+class CopyProfilesToPolicies
+  class << self
+    delegate :run!, to: :new
+  end
+
+  def run!
+    create_policies
+  end
+
+  private
+
+  def profiles
+    Profile.canonical(false)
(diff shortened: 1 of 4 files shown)


akofink (approved):
Seems to work as expected. Should we merge it to the RHICOMPL-661 branch?

vkrizan:
@akofink I'd rather have it merged into master, if we agree on this simplified version of migration. That way QE can start verifying it.
We can do a second merge into RHICOMPL-661 branch. How does that sounds?

akofink:
Sounds good! :)

PR #611, opened 2020-10-19
[RHICOMPL-1048] Allow rendering multiple model errors

akofink:
This is in support of RHICOMPL-661 and is a non-breaking change. It will allow rendering the errors of multiple models instead of just one. The ProfilesController will require this when creating multiple different records in one HTTP request.

### DIFF
--- a/app/controllers/concerns/rendering.rb
+++ b/app/controllers/concerns/rendering.rb
@@ -11,13 +11,23 @@ module Rendering
       render({ json: serializer.new(model, opts) }.merge(args))
     end
 
-    def render_error(model, **args)
+    def render_error(models, **args)
       render({ json: {
-        errors: model.errors.full_messages
+        errors: model_errors(models)
       }, status: :not_acceptable }.merge(args))
     end
 
     private
+
+    def model_errors(models = [])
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/rendering.rb
```diff
@@ -11,14 +11,24 @@ def render_json(model, **args)
       render({ json: serializer.new(model, opts) }.merge(args))
     end
 
-    def render_error(model, **args)
+    def render_error(models, **args)
```
vkrizan, line 14: Observation: This could accept a string or list of strings of error messages.

... so we would prevent rendering errors like
```
render json: { errors: error }, status: :forbidden 
```

Rationale: If the the method of rendering error changes, then all places have to change with it.

vkrizan (approved):
LGTM. Thank you for splitting it.

app/controllers/concerns/rendering.rb
akofink: We could add that, and this method could use that method, I guess. For this PR, though, I don't want to update all the places where we do manual rendered error messages.

PR #613, opened 2020-10-21
[RHICOMPL-1049] Profile uniqueness in a policy

vkrizan:
To support the new Policy model, the db unique constraint must be changed to include `policy_id`. Policies can have the same profiles (ref_id and benchmark), but need a different rule set.

It adds a validation, to ensure that a policy has only one profile for an SSG (ref_id and benchmark) regardless of an `external` boolean field value.

To keep the previous uniqueness by `external` field, there is a validation added for profiles that are outside of a policy. This is an intermid solution, before the Policy model is fully used. All internal profiles are expected to have a policy. Note, that the `external` field will be deprecated soon.

/cc @Victoremepunto to check the assumptions/conditions (ty!)

### DIFF
--- a/db/migrate/20201021083148_change_uniq_on_profiles.rb
+++ b/db/migrate/20201021083148_change_uniq_on_profiles.rb
@@ -1,0 +1,15 @@
+class ChangeUniqOnProfiles < ActiveRecord::Migration[5.2]
+  def up
+    remove_index :profiles, name: 'uniqueness'
+    add_index :profiles,
+              %i[ref_id account_id benchmark_id external policy_id],
+              unique: true, name: 'uniqueness'
+  end
+
+  def down
+    remove_index :profiles, name: 'uniqueness'
+    add_index :profiles,
+              %i[ref_id account_id benchmark_id external],
+              unique: true, name: 'uniqueness'
+  end
+end
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -18,6 +18,10 @@ class Profile < ApplicationRecord
   validates :ref_id, uniqueness: {
     scope: %i[account_id benchmark_id external policy_id]
   }, presence: true
+  validates :ref_id, uniqueness: {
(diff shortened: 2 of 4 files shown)


vkrizan:
Smoke tests:
```
=================================== short test summary info ===================================
FAILED lib64/python3.6/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_cert_type_auth
FAILED lib64/python3.6/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_success - ...
FAILED lib64/python3.6/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_fails - iq...
============= 3 failed, 85 passed, 1 skipped, 126 deselected in 322.73s (0:05:22) =============
```

db/migrate/20201021083148_change_uniq_on_profiles.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeUniqOnProfiles < ActiveRecord::Migration[5.2]
+  def change
+    remove_index :profiles, name: 'uniqueness'
+    add_index :profiles,
+              %i[ref_id account_id benchmark_id external policy_id],
+              unique: true, name: 'uniqueness'
+  end
```
akofink, line 7: We should write separate up/down migrations for this because `remove_index` only by name doesn't have a way to go back `down`.

akofink (approved):
Looks great, thanks @vkrizan! Just one small comment about the migration itself.

db/migrate/20201021083148_change_uniq_on_profiles.rb
vkrizan: I'll add it. Reverting it once "external" profiles are present in Policies would not be possible if there would be a duplicate profile outside of policy.

akofink: Good point!

PR #614, opened 2020-10-21
[RHICOMPL-1042] Use BO id on Profile test setup

vkrizan:
As the business_objective is going to be delegated to Policy from a Profile.


akofink (approved):
Thanks!

PR #616, opened 2020-10-21
[RHICOMPL-1051] Fix policy association during report parse

akofink:
Before this change, uploads of a different SSG version would not associate the newly created profile to the policy. To test:

1. Create a policy in the UI, verify 1 policy and 1 associated external=false profile is created
3. Assign a host to the policy
2. Upload a report for that host against a different SSG version than the original profile
3. Check that a new profile is created and that it's associated to the policy (and external=true)

### DIFF
--- a/app/services/concerns/xccdf/hosts.rb
+++ b/app/services/concerns/xccdf/hosts.rb
@@ -28,6 +28,11 @@ module Xccdf
       )
     end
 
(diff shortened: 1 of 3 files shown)


akofink:
Fixed in https://github.com/RedHatInsights/compliance-backend/pull/620

PR #617, opened 2020-10-22
[RHICOMPL-1049] Fix external Profile uniqueness outside of a policy

vkrizan:
The external uniqueness for profiles outside of a policy has to include the validation condition also in the duplicate/relation lookup.

Ensure that profile data migration validates policy assignment.

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -23,7 +23,8 @@ class Profile < ApplicationRecord
     message: 'must be unique in a policy'
   }, if: :policy_id
   validates :external, uniqueness: {
-    scope: %i[ref_id account_id benchmark_id]
+    scope: %i[ref_id account_id benchmark_id],
+    conditions: -> { where(policy_id: nil) }
   }, unless: :policy_id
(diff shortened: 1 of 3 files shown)


app/models/profile.rb
```diff
@@ -23,7 +23,8 @@ class Profile < ApplicationRecord
     message: 'must be unique in a policy'
   }, if: :policy_id
   validates :external, uniqueness: {
-    scope: %i[ref_id account_id benchmark_id]
+    scope: %i[ref_id account_id benchmark_id],
+    conditions: -> { where(policy_id: nil) }
   }, unless: :policy_id
```
akofink, line 28: Do we still need the `unless`?

akofink (approved):
The validation makes sense, just one comment. Thanks!

app/models/profile.rb
vkrizan: Yes. There is probably a possibility to merge those two uniq constraints into one by extending lookup conditions.

PR #618, opened 2020-10-22
[RHICOMPL-661] Cleanup base and sync with master

vkrizan:
Fixup for commit 172644a. Cherry-picks of new commits from master.

Before preparing PR for final merge of RHICOMPL-661 into master, rebase the local branch first:
```
$ git rebase -i $(git merge-base master RHICOMPL-661)
$ git rebase -i master
# drop commits that are already present in master
```


vkrizan:
@akofink What do you think about this approach? Clean the branch from stuff that is already in master, prior to creation of a final PR for merging it to master.

app/controllers/concerns/rendering.rb
```diff
@@ -11,24 +11,14 @@ def render_json(model, **args)
       render({ json: serializer.new(model, opts) }.merge(args))
     end
 
-    def render_error(models, **args)
+    def render_error(model, **args)
```
vkrizan, line 14: Reverting change to accept changes from master, PR #611

app/models/profile.rb
```diff
@@ -13,9 +13,6 @@ class Profile < ApplicationRecord
   belongs_to :benchmark, class_name: 'Xccdf::Benchmark'
   belongs_to :parent_profile, class_name: 'Profile', optional: true
 
-  validates :ref_id, uniqueness: {
```
vkrizan, line 16: Incorrect addition. Uniqueness already handled in master, PR #613.

test/models/profile_test.rb
```diff
@@ -10,7 +10,7 @@ class ProfileTest < ActiveSupport::TestCase
   should have_many(:hosts).through(:test_results)
   should belong_to(:policy_object).optional
   should validate_uniqueness_of(:ref_id)
-    .scoped_to(%i[account_id benchmark_id policy_id])
+    .scoped_to(%i[account_id benchmark_id external])
```
vkrizan, line 13: Revert for a clean rebase with master.

vkrizan:
We would need to cherry-pick changes from current master here. And then with the last rebase, drop the duplicate commits. Merge commits do not help much with the rebases.

vkrizan:
There would be a fail of unit test
```
FAIL["test_create_with_a_found_parent_profile_id_but_existing_ref_id_in_the_account"
```
which awaits PR #615

app/models/profile.rb
akofink: This is the large uniqueness condition that matches the DB constraint. It's not incorrect, just duplicated by #613, right?

akofink (commented):
Looks good. Thanks for putting this together!

akofink (approved):

app/models/profile.rb
vkrizan: Yes, and no. It was missing the condition `if: :policy_id`.

PR #620, opened 2020-10-22
[RHICOMPL-1051] Lookup policy on profile clone

vkrizan:
Lookup policy based on the host(s). Assign newly created profile into the found policy.

For Profile.clone_to it first finds the policy of the passed account.
The policy lookup up starts with the internal profile.

The internal profile:
* must have a policy that have the host assigned
* must be part of the account
* must match the ref_id of the (canonical) profile being cloned
* must match the benchmark ref_id of the profile being cloned

The last benchmark condition is there to ensure that the internal profile is
for the same major OS version.

The policy object is then retrieved from the internal profile.

The second step of the clone_to is to find an existing profile on that
policy.  The existing profile has to match the benchmark id.

### DIFF
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -12,15 +12,18 @@ module ProfilePolicyAssociation
     delegate :business_objective, :business_objective_id, :update_hosts,
              to: :policy_object, allow_nil: true
 
-    def old_policy
-      return if canonical?
-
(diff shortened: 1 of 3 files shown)


akofink (approved):
Seems to work as expected. Thanks, @vkrizan!

vkrizan:
@akofink I'm going to merge this. I think it would be better to lookup the policy outside of `clone_to`. Hopefully we can clean that up in the finish phase of the EPIC.

PR #621, opened 2020-10-23
[RHICOMPL-1054] Hosts API returns assigned and test_result profiles

akofink:
This is how this API behaved before we removed the ProfileHost
relationship. It's not super useful IMO, but it keeps the API
expectations unbroken.


vkrizan (approved):
LGTM.

Note, that the `has_many :profiles, through: :policies, source: :profiles` on a `Host` can be redundant. I'm not sure about GraphQL though.

PR #622, opened 2020-10-23
[RHICOMPL-661] Merge RHICOMPL-661 to master 

akofink:
RHICOMPL-661, RHICOMPL-662, RHICOMPL-1048, RHICOMPL-1042, RHICOMPL-1049, RHICOMPL-1051, RHICOMPL-1054


akofink:
Local smoke results (maybe some pollution of data resulting in failure of test_retrieve_profile):

```
================================================================================== short test summary info ===================================================================================
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_cert_type_auth - Failed: DID NOT RAISE <class 'requests.exceptions.HTTPError'>
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_success - iqe_rbac_api.exceptions.ApiException: (401)
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rbac.py::test_rbac_fails - iqe_rbac_api.exceptions.ApiException: (401)
FAILED .iqe_venv/lib64/python3.8/site-packages/iqe_compliance/tests/test_rest_api.py::test_retrieve_profile - assert False
===================================================== 4 failed, 84 passed, 1 skipped, 126 deselected, 340 warnings in 446.97s (0:07:26) ======================================================
```

akofink:
ACK from me :) I can't actually approve it because I opened the PR

app/graphql/types/profile.rb
```diff
@@ -67,14 +67,19 @@ class Profile < Types::BaseObject
 
     field :major_os_version, String, null: false
 
+    def policy
+      object.old_policy
```
vkrizan, line 71: Non- blocker: For the UI to leverage the Policy association this should return
```
object.policy_object&.profiles&.external(false)&.first
```
(in short)

app/graphql/types/profile.rb
```diff
@@ -87,8 +92,17 @@ def last_scanned(args = {})
...
 
+    def hosts
+      ::CollectionLoader.for(policy_or_report.class, :hosts)
+                        .load(policy_or_report)
+    end
+
     private
 
+    def policy_or_report
+      object.policy_object || object
```
vkrizan, line 103: Observation: I'm not sure if this would fail if the `policy_object` is found, as it is not type of a `Profile`.

app/models/xccdf/benchmark.rb
```diff
@@ -91,6 +91,10 @@ def latest_supported
       end
     end
 
+    def latest_ssg
```
vkrizan, line 94: Not a blocker/Nitpick: this is not used yet.

app/services/duplicate_profile_resolver.rb
```diff
@@ -54,8 +54,9 @@ def grouped_nonunique_profile_tuples
 
     # rubocop:disable Rails/SkipsModelValidations
     def migrate_profile_hosts(existing_p, duplicate_p)
-      duplicate_p.profile_hosts.where.not(host: existing_p.hosts)
-                 .update_all(profile_id: existing_p.id)
+      duplicate_p.profile_hosts.where.not(
```
vkrizan, line 57: Observation: We no longer use `profile_hosts`.

vkrizan (approved):
Added some comments. They all seems to be a non-blocker for the merge.

app/graphql/types/profile.rb
vkrizan: It's only used for internal lookups.

app/graphql/types/profile.rb
akofink: that should be equivalent to `old_policy`, no?

app/services/duplicate_profile_resolver.rb
akofink: This resolver will not be used. It was a single time migration before we added uniqueness constraints in the DB. We can remove the service completely without affecting existing workflows.

app/graphql/types/profile.rb
vkrizan: Yes, with some exceptions.

PR #623, opened 2020-10-26
[RHICOMPL-1062] GraphQL to use profile policy association

vkrizan:
The GraphQL Profile.policy will return the initial profile of the associated policy.

See also: https://github.com/RedHatInsights/compliance-backend/pull/622#discussion_r510953270

### DIFF
--- a/test/graphql/queries/profile_query_test.rb
+++ b/test/graphql/queries/profile_query_test.rb
@@ -5,7 +5,7 @@ require 'test_helper'
 class ProfileQueryTest < ActiveSupport::TestCase
   setup do
     users(:test).update account: accounts(:test)
-    profiles(:one).update account: accounts(:test), hosts: [hosts(:one)]
+    profiles(:one).update account: accounts(:test)
   end
 
   test 'query profile owned by the user' do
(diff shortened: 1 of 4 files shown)


vkrizan:
Two smoke test failures (need to check):
```
FAILED lib64/python3.6/site-packages/iqe_compliance/tests/test_graphql.py::test_graphql_create_internal_policy - IndexError:...
FAILED lib64/python3.6/site-packages/iqe_compliance/tests/test_rest_api.py::test_retrieve_profile - assert False
```

test/graphql/queries/profile_query_test.rb
```diff
@@ -5,7 +5,7 @@
 class ProfileQueryTest < ActiveSupport::TestCase
   setup do
     users(:test).update account: accounts(:test)
-    profiles(:one).update account: accounts(:test), hosts: [hosts(:one)]
+    profiles(:one).update account: accounts(:test)
```
vkrizan, line 8: Hosts are associated via policy. The previous case would associate hosts with a profile through a test result.

vkrizan:
Passing the initial profile in a place of a policy would work only partially. It would be missing a name a description from the policy.

vkrizan:
Reopening. The name and desc. will be handled separately, via RHICOMPL-1035.

bastilian (approved):
LGTM! Thanks @vkrizan!

PR #624, opened 2020-10-26
[RHICOMPL-1035] GraphQL: Pseudo-policy should inherit name and desc

vkrizan:
The pseudo-policy returned on the GraphQL for a profile should inherint
name and description from the policy.
The database records of profiles are keeping the original names and
descriptions (from their parent profiles).
This allows to use the `Profile.name` as policy/profile type, and
`Profile.policy.name` as the policy title (user defined).

Note, that REST API returns name and description of profiles from their
policies.  The policy/profile type on the REST API can be retrieved by
requesting canonical profiles based using the `parent_profile_id`.

### DIFF
--- a/test/graphql/queries/profile_query_test.rb
+++ b/test/graphql/queries/profile_query_test.rb
@@ -91,10 +91,10 @@ class ProfileQueryTest < ActiveSupport::TestCase
 
     assert_equal profiles(:one).id,
                  result['data']['profile']['policy']['id']
-    assert_equal profiles(:one).name,
-                 result['data']['profile']['policy']['name']
     assert_equal profiles(:one).ref_id,
                  result['data']['profile']['policy']['refId']
+    assert_equal policies(:one).name,
+                 result['data']['profile']['policy']['name']
   end
 
   test 'query all profiles' do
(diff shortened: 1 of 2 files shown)


test/graphql/queries/profile_query_test.rb
```diff
@@ -91,10 +91,10 @@ class ProfileQueryTest < ActiveSupport::TestCase
 
     assert_equal profiles(:one).id,
                  result['data']['profile']['policy']['id']
-    assert_equal profiles(:one).name,
-                 result['data']['profile']['policy']['name']
     assert_equal profiles(:one).ref_id,
                  result['data']['profile']['policy']['refId']
+    assert_equal policies(:one).name,
```
bastilian, line 96: Would it make sense to also `update!` the name and description just to verify that the updated values are actually used?

bastilian (approved):
LGTM! Thank you @vkrizan!

test/graphql/queries/profile_query_test.rb
vkrizan: I guess no need, as the policy has name "policy1" and profile has "profile1". Or, did I missed something?

PR #625, opened 2020-10-26
[RHICOMPL-1053] Return policy profiles on GraphQL

vkrizan:
Policy profiles are to be queried on a profile's policy.
All policy profiles are returned including the internal/initial profile.

Recommended way of querying:
```
  query Profile($anyPolicyProfileId: String!){
      profile(id: $anyPolicyProfileId) {
          policy {
              profiles {
		  id
                  ssgVersion  # to be added
                  rules {
                      id
                  }
              }
          }
      }
  }
```

Note, that there can be a case when an internal profile would be the same
as its policy (since there is no Policy GraphQL type, yet) and would be
also found in the list of policy profiles.

### DIFF
--- a/app/graphql/schema.graphql
+++ b/app/graphql/schema.graphql
@@ -102,6 +102,7 @@ type Profile implements Node & RulesPreload {
   majorOsVersion: String!
   name: String!
   policy: Profile
+  profiles: [Profile!]
   refId: String!
   rules(
     """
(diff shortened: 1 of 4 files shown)


bastilian:
@vkrizan Will the `ssg_version` field be added in this PR or in a separate one?

vkrizan:
> @vkrizan Will the `ssg_version` field be added in this PR or in a separate one?

@bastilian Separately, via RHICOMPL-1060.

bastilian (approved):
Thank you @vkrizan!

PR #626, opened 2020-10-27
[RHICOMPL-1063] Return all system's profiles on GraphQL

vkrizan:
GraphQL will return all system profiles, consisting of assigned profiles (through a policy) and profiles from test results. 
This should fix showing failed rules on external reports.

Redundant `Host.profiles` association is being dropped in favor of `associated_profiles`, `test_result_profiles` and combined
`all_profiles`.

### DIFF
--- a/app/serializers/host_serializer.rb
+++ b/app/serializers/host_serializer.rb
@@ -11,7 +11,10 @@ class HostSerializer
   end
 
   has_many :test_results
+
+  # rubocop:disable Style/SymbolProc
   has_many :profiles do |host|
-    (host.assigned_profiles + host.test_result_profiles).uniq
+    host.all_profiles
   end
+  # rubocop:enable Style/SymbolProc
 end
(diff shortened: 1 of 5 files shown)


app/serializers/host_serializer.rb
```diff
@@ -11,7 +11,10 @@ class HostSerializer
   end
 
   has_many :test_results
+
+  # rubocop:disable Style/SymbolProc
```
vkrizan, line 15: Unfortunately if the symbol proc `&:all_profiles` is used, the framework passes some options to it as an argument.

akofink: Yeah, I couldn't figure out how to use symbol to proc here either :(

akofink (approved):
Looks great! Thanks for unifying the logic.

app/serializers/host_serializer.rb
vkrizan: If the method have had at least one argument, then it would work. But I'm not sure what's the logic (why) behind the framework sending some `options` to it.

PR #627, opened 2020-10-27
[RHICOMPL-1060] Expose ssgVersion for the Profile GraphQL type

vkrizan:
A shortcut to retrieving benchmark's version.

### DIFF
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -34,6 +34,7 @@ module Types
     end
     field :hosts, [::Types::System], null: true
     field :benchmark, ::Types::Benchmark, null: true
+    field :ssg_version, String, null: false
     field :business_objective, ::Types::BusinessObjective, null: true
     field :business_objective_id, ID, null: true
     field :total_host_count, Int, null: false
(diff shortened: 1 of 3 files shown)


bastilian (approved):
Thank you @vkrizan!

app/graphql/types/profile.rb
```diff
@@ -70,6 +71,12 @@ class Profile < Types::BaseObject
 
     field :major_os_version, String, null: false
 
+    def ssg_version
+      ::RecordLoader.for(::Xccdf::Benchmark)
+                    .load(object.benchmark_id)
+                    .then(&:version)
+    end
```
akofink, line 78: Nice! This will also be needed in the REST API - perhaps we could put it in the Profile model itself?

akofink (approved):
Thanks, @vkrizan! Just one optional comment :)

app/graphql/types/profile.rb
vkrizan: The attribute on the REST will be added via  RHICOMPL-1034.

An attribute on the Profile model would make sense. Do you know how to pre-load benchmark for all profiles? Should we do a default scope that would `include(:benchmark)`?

akofink: It may make sense to use active record directly with something like:

```rb
belongs_to :ssg_version, through: :benchmark
```

idk if that actually works, but it should automatically include the associated benchmark. Otherwise, like you say, we'll need to `include(:benchmark)`.

vkrizan: I guess that would only work if the `:ssg_version` was an association to some other model.

I had in mind this:
```
class Profile
  default_scope { includes(:benchmark) }
end
```

akofink: `default_scope` is a bit... scary tbh. I'd rather avoid it and instead include benchmark manually in the controller layer or create a named scope to include the benchmark where it's actually needed (i.e. `with_benchmarks` or similar).

PR #629, opened 2020-10-29
[RHICOMPL-1069] Host scope search by policy

vkrizan:
Search supports lookup either by policy id or by a profile id that the host is assigned to through its policy.

This would allow SystemsTable to query the systems on any of the policy profiles.

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -11,6 +11,8 @@ class Host < ApplicationRecord
   scoped_search on: :has_test_results, ext_method: 'test_results?',
                 only_explicit: true, operators: ['=']
   scoped_search relation: :test_results, on: :profile_id
+  scoped_search on: :policy_id, ext_method: 'filter_by_policy',
+                only_explicit: true, operators: ['=']
   has_many :rule_results, dependent: :delete_all
   has_many :rules, through: :rule_results, source: :rule
   has_many :profile_hosts, dependent: :destroy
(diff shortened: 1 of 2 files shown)


app/models/host.rb
```diff
@@ -11,6 +11,8 @@ class Host < ApplicationRecord
   scoped_search on: :has_test_results, ext_method: 'test_results?',
                 only_explicit: true, operators: ['=']
   scoped_search relation: :test_results, on: :profile_id
+  scoped_search on: :policy_id, ext_method: 'filter_by_policy',
+                only_explicit: true, operators: ['=']
```
akofink, line 15: Does `scoped_search relation: :policy` not work here?

vkrizan: It should work, but it would be only limited to lookup by a policy id. Note, that the policy id is not exposed in the APIs. Therefore, there is also a lookup by any of the profile ids that are assigned to the policy.

akofink: Oh! That makes sense :) Thanks for the clarification

vkrizan:
The change on the frontend would be here:
https://github.com/RedHatInsights/compliance-frontend/blob/51d970d1fbe54ade433d81404509404f59e3c043/src/SmartComponents/SystemsTable/SystemsTable.js#L176

akofink (approved):
Thanks, @vkrizan! Works as expected.

PR #631, opened 2020-10-29
[RHICOMPL-1073] Add org_id to generated account identity header

akofink:
This affects any jobs that use an account number to generate an identity
header, and it also affects development when generating an identity to
use with our (and other service's) local APIs without 3Scale.

This was a change


akofink:
I actually don't think this is necessary after all.

PR #633, opened 2020-11-02
[RHICOMPL-1074] Internal/main policy profile deletes whole policy via callback

vkrizan:
**Option 1** (via callback)

Removal of an internal/main profile (`external=false`) of a policy would
remove its policy with all dependent profiles (`external=true`).

The callback is used to guarantee data sanity, eg. when a benchmark is destroyed.

Pros:
* no change needed for UI
* other SSG profiles (with tailoring) on a policy could be removed individually

Cons:
* more than one (same-type) REST resource would be removed
* IQE would need to know which profiles are part of which policy (RHICOMPL-1078)
* main policy profile/SSG could not be removed individually without removing all policy profiles
* this behavior would likely change if the v2 API would be created with real policy type/resource

### DIFF
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -5,6 +5,7 @@ module ProfilePolicyAssociation
   extend ActiveSupport::Concern
 
   included do
+    after_destroy :destroy_policy_with_internal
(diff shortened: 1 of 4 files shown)


akofink (approved):
Thanks, @vkrizan! Seems to work as expected.

PR #634, opened 2020-11-02
[RHICOMPL-1070] Allow unsetting the business objective via GQL

akofink:
via GraphQL. Sending null as the business_objective_id will now unset
the business objective from the profile's policy.

### DIFF
--- a/app/graphql/mutations/profile/edit.rb
+++ b/app/graphql/mutations/profile/edit.rb
@@ -19,7 +19,7 @@ module Mutations
 
       def resolve(args = {})
         profile = authorized_profile(args)
-        profile.policy_object.update(args.compact.slice(*POLICY_ATTRIBUTES))
+        profile.policy_object.update(args.slice(*POLICY_ATTRIBUTES))
         { profile: profile }
       end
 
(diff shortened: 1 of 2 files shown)


bastilian (approved):
Nice! Thank you @akofink!

app/graphql/mutations/profile/edit.rb
```diff
@@ -19,7 +19,7 @@ class Edit < BaseMutation
 
       def resolve(args = {})
         profile = authorized_profile(args)
-        profile.policy_object.update(args.compact.slice(*POLICY_ATTRIBUTES))
+        profile.policy_object.update(args.slice(*POLICY_ATTRIBUTES))
```
vkrizan, line 22: The `.compact` was there even before. This bug is present since RHICOMPL-742.

vkrizan (approved):
Thanks. LGTM.

app/graphql/mutations/profile/edit.rb
akofink: Yeah, I'm really not sure why we never caught this before :man_shrugging:

PR #636, opened 2020-11-04
[RHICOMPL-1036] has_policy_test_results filter on Profile

vkrizan:
`has_policy_test_results` should look up profiles which policy has results, meaning that at least on of policy profile has a test result.

### DIFF
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -15,6 +15,9 @@ module ProfileSearching
                   rename: :test_result_system_names
     scoped_search on: :has_test_results, ext_method: 'test_results?',
                   only_explicit: true, operators: ['=']
+    scoped_search on: :has_policy_test_results,
+                  ext_method: 'policy_test_results?',
+                  only_explicit: true, operators: ['=']
     scoped_search on: :canonical, ext_method: 'canonical?', only_explicit: true,
                   operators: ['=']
     scoped_search on: :has_policy, ext_method: 'policy_object_search',
@@ -38,6 +41,21 @@ module ProfileSearching
       test_results = TestResult.select(:profile_id).distinct
       has_test_results && where(id: test_results) || where.not(id: test_results)
     }
+    scope :has_policy_test_results, lambda { |has_policy_test_results = true|
+      # the use of default scope is to bypass Pundit
+      # and avoid the situation with bind parameters
+      # that must have a different order in the resulting query
+      # of a scoped_search
+      with_policy_test_results = default_scoped.joins(
+        policy_object: :test_results
+      )
+
+      if has_policy_test_results
+        where(id: with_policy_test_results)
+      else
+        where.not(id: with_policy_test_results)
+      end
+    }
     scope :os_major_version, lambda { |major, equals = true|
       where(benchmark: Xccdf::Benchmark.os_major_version(major, equals))
     }
@@ -57,6 +75,12 @@ module ProfileSearching
     def test_results?(_filter, _operator, value)
       has_test_results = ActiveModel::Type::Boolean.new.cast(value)
       profiles = Profile.has_test_results(has_test_results)
+      { conditions: profiles.arel.where_sql.gsub(/^where /i, '') }
+    end
+
+    def policy_test_results?(_filter, _operator, value)
+      has_test_results = ActiveModel::Type::Boolean.new.cast(value)
+      profiles = Profile.has_policy_test_results(has_test_results)
       { conditions: profiles.arel.where_sql.gsub(/^where /i, '') }
     end
 
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -245,6 +245,69 @@ class ProfileTest < ActiveSupport::TestCase
                         profiles(:one))
   end
 
+  context 'has_policy_test_results filter' do
+    setup do
+      test_results(:one).update(profile: profiles(:one), host: hosts(:one))
+      profiles(:two).test_results.destroy_all
+    end
+
+    should 'find a policy profile if it has a test result' do
+      profiles(:one).update!(policy_object: policies(:one))
+      assert profiles(:one).test_results.present?
+      assert profiles(:two).test_results.empty?
+
+      assert_includes(Profile.search_for('has_policy_test_results = true'),
+                      profiles(:one))
+      assert_not_includes(Profile.search_for('has_policy_test_results = true'),
+                          profiles(:two))
+      assert_includes(Profile.search_for('has_policy_test_results = false'),
+                      profiles(:two))
+      assert_not_includes(Profile.search_for('has_policy_test_results = false'),
+                          profiles(:one))
+    end
+
+    should 'find a policy profile if it has a test result on a scope change' do
+      profiles(:one).update!(policy_object: policies(:one))
+      profiles(:two).update!(account: accounts(:one))
+
+      assert profiles(:one).test_results.present?
+      assert profiles(:two).test_results.empty?
+
+      Profile.where(account: accounts(:one)).scoping do
+        assert_includes(Profile.search_for('has_policy_test_results = true'),
+                        profiles(:one))
+        assert_not_includes(
+          Profile.search_for('has_policy_test_results = true'),
+          profiles(:two)
+        )
+        assert_includes(Profile.search_for('has_policy_test_results = false'),
+                        profiles(:two))
+        assert_not_includes(
+          Profile.search_for('has_policy_test_results = false'),
+          profiles(:one)
+        )
+      end
+    end
+
+    should 'find all policy profiles if one has a test result' do
+      profiles(:one).update!(policy_object: policies(:one),
+                             external: true)
+      profiles(:two).update!(policy_object: policies(:one),
+                             external: false)
+
+      assert profiles(:one).test_results.present?
+      assert profiles(:two).test_results.empty?
+      assert_includes(Profile.search_for('has_policy_test_results = true'),
+                      profiles(:two))
+      assert_not_includes(Profile.search_for('has_policy_test_results = false'),
+                          profiles(:two))
+      assert_includes(Profile.search_for('has_policy_test_results = true'),
+                      profiles(:one))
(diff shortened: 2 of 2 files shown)


app/models/concerns/profile_searching.rb
```diff
@@ -35,8 +35,22 @@ module ProfileSearching
...
+      # that must have a different order in the resulting query
+      # of a scoped_search
+      policy_test_results = default_scoped.joins(policy_object: :test_results)
+
+      internal = arel_table[:external].eq(false).and(
+        arel_table[:id].in(policy_test_results.select(:id).arel)
+      )
+
+      external = arel_table[:external].eq(true).and(
+        arel_table[:id].in(TestResult.default_scoped.select(:profile_id).arel)
```
akofink, line 49: This `default_scoped` should not be necessary

app/models/concerns/profile_searching.rb
```diff
@@ -35,8 +35,22 @@ module ProfileSearching
       where(external: external)
     }
     scope :has_test_results, lambda { |has_test_results = true|
-      test_results = TestResult.select(:profile_id).distinct
-      has_test_results && where(id: test_results) || where.not(id: test_results)
+      # the use of default scope is to bypass Pundit
+      # and avoid the situation with bind parameters
+      # that must have a different order in the resulting query
+      # of a scoped_search
+      policy_test_results = default_scoped.joins(policy_object: :test_results)
```
akofink, line 42:
```suggestion
      policy_test_results = policy_object.test_results
```

app/models/concerns/profile_searching.rb
vkrizan: For the sake of completeness and stability lets keep it.

app/models/concerns/profile_searching.rb
```diff
@@ -35,8 +35,22 @@ module ProfileSearching
...
+
+      internal = arel_table[:external].eq(false).and(
+        arel_table[:id].in(policy_test_results.select(:id).arel)
+      )
```
akofink, line 46: I'm not sure this is quite right (notice the change from id to profile_id below). Also, it could be written like this:

```suggestion
      internal = external(false).where(id: policy_test_results.select(:profile_id))
```

app/models/concerns/profile_searching.rb
vkrizan: `policy_object` is on an instance, not on a class:
```
[1] pry(main)> Profile.policy_object
NoMethodError: undefined method `policy_object' for Profile (call 'Profile.connection' to establish a connection):Class
```

app/models/concerns/profile_searching.rb
akofink: Pundit does not affect models randomly. When you request `TestResult` directly like this, no pundit code is run at all.

app/models/concerns/profile_searching.rb
akofink: Oh, right. Good point! lol

app/models/concerns/profile_searching.rb
```diff
@@ -35,8 +35,22 @@ module ProfileSearching
       where(external: external)
     }
     scope :has_test_results, lambda { |has_test_results = true|
-      test_results = TestResult.select(:profile_id).distinct
-      has_test_results && where(id: test_results) || where.not(id: test_results)
+      # the use of default scope is to bypass Pundit
+      # and avoid the situation with bind parameters
```
akofink, line 39: Maybe we could keep the existing behavior so we don't break existing API usage and just define a new search that is a bit more clear about what it returns, like `has_policy_test_results` or similar.

app/models/concerns/profile_searching.rb
vkrizan: You can translate it into:
```
Profile.joins(policy_object: :test_results).select(:id)
# or
Profile.joins(policy_object: :test_results).pluck(:id)
```
It leverages the INNER JOIN. It should yield profile ids have a policy and their policy profiles have a test result.

Here is the subquery:
```
SELECT "profiles"."id"
FROM "profiles"
INNER JOIN "policies" ON "policies"."id" = "profiles"."policy_id"
INNER JOIN "profiles" "profiles_policies_join" ON "profiles_policies_join"."policy_id" = "policies"."id"
INNER JOIN "test_results" ON "test_results"."profile_id" = "profiles_policies_join"."id"
```

app/models/concerns/profile_searching.rb
```diff
@@ -35,8 +35,22 @@ module ProfileSearching
...
+
+      internal = arel_table[:external].eq(false).and(
+        arel_table[:id].in(policy_test_results.select(:id).arel)
+      )
+
+      external = arel_table[:external].eq(true).and(
+        arel_table[:id].in(TestResult.default_scoped.select(:profile_id).arel)
+      )
+
+      cond = internal.or(external)
```
akofink, line 52: I think we can use the old `has_test_results` here instead of `external`.

app/models/concerns/profile_searching.rb
akofink: Ok, I see. Thanks! It seems correct.

app/models/concerns/profile_searching.rb
vkrizan: `TestResult` no longer used directly.

app/models/concerns/profile_searching.rb
vkrizan: I've completely dropped the `external` condition.

app/models/concerns/profile_searching.rb
vkrizan: Changed to add a new filter method instead of changing the existing one.

app/models/concerns/profile_searching.rb
```diff
@@ -38,6 +41,19 @@ module ProfileSearching
       test_results = TestResult.select(:profile_id).distinct
       has_test_results && where(id: test_results) || where.not(id: test_results)
     }
+    scope :has_policy_test_results, lambda { |has_test_results = true|
+      # the use of default scope is to bypass Pundit
+      # and avoid the situation with bind parameters
+      # that must have a different order in the resulting query
+      # of a scoped_search
+      policy_test_results = default_scoped.joins(policy_object: :test_results)
```
akofink, line 49: Nit: this variable makes it look like a collection of test results instead of a collection of profiles. `with_policy_test_results` or similar would be more appropriate.

app/models/concerns/profile_searching.rb
```diff
@@ -38,6 +41,19 @@ module ProfileSearching
       test_results = TestResult.select(:profile_id).distinct
       has_test_results && where(id: test_results) || where.not(id: test_results)
     }
+    scope :has_policy_test_results, lambda { |has_test_results = true|
```
akofink, line 44: nit: `s/has_test_results/has_policy_test_results/`

app/models/concerns/profile_searching.rb
```diff
@@ -60,6 +76,12 @@ def test_results?(_filter, _operator, value)
       { conditions: profiles.arel.where_sql.gsub(/^where /i, '') }
     end
 
+    def policy_test_results?(_filter, _operator, value)
+      has_test_results = ActiveModel::Type::Boolean.new.cast(value)
```
akofink, line 80: Same nitpick here: `s/has_test_results/has_policy_test_results/`

app/models/concerns/profile_searching.rb
akofink, line 49: Instead of `default_scoped` with a comment explaining it, why not use `Profile`?

```suggestion
      policy_test_results = Profile.joins(policy_object: :test_results)
```

akofink (approved):
Works as expected, thanks! Just a few more minor comments.

app/models/concerns/profile_searching.rb
vkrizan: It would still have to be `Profile.default_scope`. (Try it out with `.scoping` wrapper)

akofink: Ok, I see now. :/ That is frustrating. The `default_scoped` thing is kind of a red flag because it makes me think we're exposing records to the user that they shouldn't be able to see, even though we're not in this case; it should be fine since we're plugging it into `where` below. Still, I've looked for ways to get rid of it without success :(
Thanks for considering my ideas regardless!

vkrizan: It makes the line longer. :-)

akofink: Very astute! ;) Readability is far more important to me than comments, and variable naming plays a huge role in that. `policy_test_results` may well become a relationship on profile one day (i.e. `has_many :policy_test_results, through: :policy_object, source: :test_results`).

vkrizan: I've tried my best to pass the parameters from the Pundit-scoped case to the scoped_search, but it has to reorder (reset) used placeholder binds. I can only think of a way to compile those placeholders right into the WHERE clause for the scoped_search.

The only think that can happen here is if a policy would have two profiles, each in different account. Then it would call the policy as having a test result regardless if it is coming from the profile belonging to current account or not. Technically that is a correct result.

akofink: Yes, but the `where` returned has the pundit scope applied, so it should filter out any profiles that are outside of the current account.

vkrizan: Exactly, and that is the problem. ;-)

Check out this example
```
Profile.where(id: "fef2c74b-ef23-4dcf-b2a7-58a7a47ccdb3").arel.where_sql.gsub(/^where /i, '')
=> "\"profiles\".\"id\" = $1
```
If the `$1` gets into final where query of the scope_search, which also has the Pundit in effect, then it would do a wrong substitution.

app/models/concerns/profile_searching.rb
vkrizan: That shadows the scope. ;-)

app/models/concerns/profile_searching.rb
vkrizan: Changed.

PR #637, opened 2020-11-04
[RHICOMPL-841] Prevent creating policies with duplicate types

akofink:
A new validation that happens only on create that ensures no other policy exists in the same account with the same policy type. Policy type is currently just the associated profiles' ref_ids. To test:

1. Create a RHEL 7 HIPAA policy in the UI
2. Import another RHEL 7 benchmark with a different SSG version
3. Create another policy via the REST API and select the new benchmark's HIPAA canonical profile as the parent profile

Before:
The API allows the action, and the policies UI displays two HIPAA policies

After:
The API detects a duplicate policy type and errors

### DIFF
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -12,6 +12,20 @@ module ProfilePolicyAssociation
                                optional: true, inverse_of: :profiles
     delegate :business_objective, :business_objective_id, :update_hosts,
              to: :policy_object, allow_nil: true
+    validate :no_duplicate_policy_types, on: :create
+
+    def no_duplicate_policy_types
+      return if canonical? || external
+
+      error_msg = 'must be unique. Another policy with '\
+                  'this policy type already exists.'
+      profile = Profile.includes(:benchmark).find_by(
+        ref_id: ref_id, account: account, external: false,
+        benchmarks: { ref_id: benchmark.ref_id }
+      )
+
+      errors.add(:policy_type, error_msg) if profile
+    end
 
     # Lookup up internal profile that has the host(s)
     # assinged to a policy an which matches the ref_id
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -59,6 +59,18 @@ class ProfileTest < ActiveSupport::TestCase
     end
   end
 
+  test 'uniqness of policy profiles by policy type' do
+    (bm = benchmarks(:one).dup).update!(version: '123')
+    profiles(:one).update!(policy_id: policies(:one).id,
+                           parent_profile: profiles(:two))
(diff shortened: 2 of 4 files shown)


test/services/copy_profiles_to_policies_test.rb
```diff
@@ -45,7 +45,7 @@ class CopyProfilesToPoliciesTest < ActiveSupport::TestCase
 
   test 'copies two similar internal profiles with to two new policies' do
     first = profiles(:one)
-    (bm = benchmarks(:one).dup).update!(version: '0.1.46')
+    (bm = benchmarks(:one).dup).update!(version: '0.1.46', ref_id: 'bm1')
```
akofink, line 48: These two test updates were the only breakages where a test was creating two policies of the same policy type in the same account.

test/models/profile_test.rb
```diff
@@ -59,6 +59,16 @@ class ProfileTest < ActiveSupport::TestCase
...
 
+  test 'uniqness of policy profiles by policy type' do
+    profiles(:one).update!(policy_id: policies(:one).id)
+    assert profiles(:one).policy_id
+    assert_not profiles(:one).external
+
+    assert_raises ActiveRecord::RecordInvalid do
+      profiles(:one).dup.update!(policy_id: policies(:two).id)
+    end
+  end
```
akofink, line 70: This test isn't quite right. WIP

akofink: Ok, should be good now.

vkrizan (approved):
LGMT.

One thing to consider is that the previous uniqueness was on benchmark UUID and not on it's ref_id (`ref_id, account_id, benchmark_id, external`). This does not affect creation of new profiles, however the existing data in the db might not be sane. We might want to plan some data cleanup (possibly with user actions) if we want to have clean data.

akofink:
Thanks for reviewing!

For existing duplicate policies, no cleanup is necessary; we can just let users delete their old policies over time. It's the reason why this isn't a DB constraint or a validation on update (just create).

app/models/concerns/profile_policy_association.rb
```diff
@@ -12,6 +12,20 @@ module ProfilePolicyAssociation
...
+      return if canonical? || external
+
+      error_msg = 'must be unique. Another policy with '\
+                  'this policy type already exists.'
+      profile = Profile.includes(:benchmark).find_by(
+        ref_id: ref_id, account: account, external: false,
+        benchmarks: { ref_id: benchmark.ref_id }
+      )
+
+      errors.add(:policy_type, error_msg) if profile
```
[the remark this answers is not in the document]
akofink: Yeah. It seems to be OK. I'm also not sure if it will cause an issue, but if it does, we can change it to `:base` in the future.

PR #638, opened 2020-11-05
Remove dead code in Xccdf report parser

vkrizan:


vkrizan:
I guess the whole `Xccdf::Profiles` could go away.

akofink (approved):
Thanks. Why is it draft? I think we should keep these concerns, otherwise we'll have to put them all together in (i.e.) Xccdf::Util or something, and rubocop will start complaining about class length, etc.

vkrizan:
@akofink I've changed it to draft, as the whole concern can go away. But I haven't verified it yet.

akofink:
If you remove the concern, `save_profiles` still has to go somewhere. I'd like to keep the concern since it matches the pattern used by the rest of the parsing code.

vkrizan:
@akofink Yes. I've missed the aliased name being used.

akofink (approved):
Thanks, @vkrizan!

PR #640, opened 2020-11-06
[RHICOMPL-1083] Host policy_id filter finds external profile via results 

vkrizan:
This fixes a bug for the frontend, where the policy_id is used to
retrieve hosts for an external profile.


akofink (approved):
Thanks, @vkrizan! Fixes the issue.

PR #641, opened 2020-11-06
feat(RHICOMPL-1086): Delete old test results within a policy

akofink:
Old test results that are within the same policy for a host will now be
removed when new test results are imported for that host. The same will
happen outside of a policy (i.e. with external reports). The old
behavior would not clean up test reports across SSG versions, but this
new version does.

### DIFF
--- a/test/services/concerns/xccdf/test_result_test.rb
+++ b/test/services/concerns/xccdf/test_result_test.rb
@@ -7,7 +7,7 @@ class TestResultTest < ActiveSupport::TestCase
   class Mock
     include Xccdf::TestResult
 
-    attr_accessor :host, :host_profile, :op_test_result
+    attr_accessor :host, :host_profile, :op_test_result, :test_result
   end
 
   setup do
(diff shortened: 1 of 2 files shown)


test/services/concerns/xccdf/test_result_test.rb
```diff
@@ -38,4 +38,46 @@ class Mock
...
+                         profile: profiles(:one),
+                         end_time: @end_time - 3.minutes)
+
+      TestResult.create!(host: hosts(:one),
+                         profile: profiles(:two),
+                         end_time: @end_time - 8.minutes)
+    end
+
+    assert_difference('TestResult.count' => -1) do
+      @mock.save_test_result
```
vkrizan, line 60: Can you please also check which test result is kept?

vkrizan (approved):
LGTM. Thanks!

test/services/concerns/xccdf/test_result_test.rb
akofink: Updated to check the remaining test result is the correct one.

PR #642, opened 2020-11-06
feat(RHICOMPL-959): Add policy_type attribute to profiles

akofink:
For graphql and REST

Test with https://github.com/RedHatInsights/compliance-frontend/pull/822

### DIFF
--- a/app/graphql/schema.graphql
+++ b/app/graphql/schema.graphql
@@ -102,6 +102,7 @@ type Profile implements Node & RulesPreload {
   majorOsVersion: String!
   name: String!
   policy: Profile
+  policyType: String!
   profiles: [Profile!]
   refId: String!
   rules(
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -50,6 +50,10 @@ class Profile < ApplicationRecord
     end
   end
 
+  def policy_type
+    (parent_profile || self).name
+  end
+
   def fill_from_parent
     self.ref_id = parent_profile.ref_id
     self.benchmark_id = parent_profile.benchmark_id
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -70,6 +70,7 @@ module Types
     field :compliant_host_count, Int, null: false
 
     field :major_os_version, String, null: false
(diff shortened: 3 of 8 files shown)


app/graphql/schema.graphql
```diff
@@ -102,6 +102,7 @@ type Profile implements Node & RulesPreload {
   majorOsVersion: String!
   name: String!
   policy: Profile
+  policyType: String!
```
vkrizan, line 105: Food for thought: If we expose it as `type` in GQL, consumer can request it with:
```
profile(id: $id) {
  policy {
    type
  }
}
```

That gets us closer to v2 and could ease the migration.

app/models/profile.rb
```diff
@@ -50,6 +50,10 @@ def from_openscap_parser(op_profile, benchmark_id: nil, account_id: nil)
     end
   end
 
+  def policy_type
+    (parent_profile || self).name
```
vkrizan, line 54: Observation: parent profile might want to be preloaded (`.includes(:parent_profile)`) in REST/GQL on index/collection queries.

vkrizan (approved):
Thanks!

app/graphql/schema.graphql
akofink: I could see us implementing it that way, but today, the `policy` in graphql is `null` for profiles without a policy, even though they technically also have a policy type. Having this top level attribute adds parity with the REST API's profile resource and helps handle external reports without any additional logic in the frontend.

As an aside: V2 is in regards to a new REST API, not graphql. IMO we can expose real policy IDs and other policy attributes if it makes sense here before we do the same in a V2 REST API. The most ideal case IMO would be to switch the UI to use the REST API and move away from using GraphQL entirely, though we haven't started using REST in any of our frontend components.

app/models/profile.rb
akofink: I was looking for the `N+1` query message but wasn't seeing it. It could be included during some other step, like while checking the profile is or isn't canonical. There are some `N+1` messages on certain endpoints, though, so we should track them down at some point as tech debt.

PR #643, opened 2020-11-06
feat(RHICOMPL-1078): Expose policy profile ID on profile

akofink:
For both graphql and REST. The policy profile is the initial profile on
a policy. Profiles without policies will return their own ID as the
policy profile ID to support 'external policies'.

### DIFF
--- a/app/models/concerns/profile_policy_association.rb
+++ b/app/models/concerns/profile_policy_association.rb
@@ -14,6 +14,14 @@ module ProfilePolicyAssociation
              to: :policy_object, allow_nil: true
     validate :no_duplicate_policy_types, on: :create
 
+    def policy_profile
+      policy_object&.initial_profile
+    end
+
+    def policy_profile_id
+      policy_profile&.id
+    end
+
     def no_duplicate_policy_types
       return if canonical? || external
 
--- a/app/serializers/profile_serializer.rb
+++ b/app/serializers/profile_serializer.rb
@@ -13,7 +13,8 @@ class ProfileSerializer
   end
   has_many :test_results
   attributes :ref_id, :score, :parent_profile_id,
-             :external, :compliance_threshold, :os_major_version
+             :external, :compliance_threshold, :os_major_version,
+             :policy_profile_id
   attribute :parent_profile_ref_id do |profile|
     profile.parent_profile&.ref_id
   end
--- a/spec/api/v1/schemas/profiles.rb
+++ b/spec/api/v1/schemas/profiles.rb
@@ -65,6 +65,10 @@ module Api
               type: 'boolean',
               example: false
             },
+            policy_profile_id: {
+              type: 'uuid',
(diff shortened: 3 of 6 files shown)


app/graphql/schema.graphql
```diff
@@ -102,6 +102,7 @@ type Profile implements Node & RulesPreload {
   majorOsVersion: String!
   name: String!
   policy: Profile
+  policyProfileId: ID
```
vkrizan, line 105: Why are we exposing what is already exposed via 
```
policy {
  id
}
```

app/models/concerns/profile_policy_association.rb
```diff
@@ -14,6 +14,14 @@ module ProfilePolicyAssociation
              to: :policy_object, allow_nil: true
     validate :no_duplicate_policy_types, on: :create
 
+    def policy_profile
+      policy_object&.initial_profile || self
```
vkrizan, line 18: What's the rationale on returning `self`? It does not seem to align with previous policy usage on GQL and neither with the new ways. There might be a case of an unexpected behavior.

app/graphql/schema.graphql
akofink: Oh, sorry - I forgot that this was the profile ID :thinking: It's named poorly IMO if we are ever going to expose the actual policy object via graphql in the future (`initialProfileId` or similar would be more fitting). I added this field to both GraphQL and REST mostly for parity between the two profile endpoints, but I can remove it.

app/models/concerns/profile_policy_association.rb
akofink: For external policies, the single profile involved is the policy itself. The UI shows a similar logic - every report has some policy, be it external or internal; in the same way, this attribute will always return some profile ID. I'm willing to change it for consistency to return `nil` for profiles without a policy.

app/graphql/schema.graphql
akofink: Actually, I think it still makes sense as an attribute, since `policy` is `nil` for profiles without a policy, but we still have to report their policy type (even though it's the same as their name) in the UI. This attribute will allow us to not have any special frontend logic for displaying the policy type of external reports.

akofink:
Updated! Thanks for the review! :)

app/graphql/schema.graphql
akofink: Removed from GraphQL.

app/models/concerns/profile_policy_association.rb
vkrizan: External profiles previously did not get `policy` on the GQL (only if they where created after an internal profile creation). The problem of the OR was that it would return `self` if `initial_profile` yields `nil`. A policy not having an initial profile would already a bad state.

Thank you for the change.

vkrizan (approved):
Thank you Andrew!

PR #644, opened 2020-11-09
[RHICOMPL-1090] Hosts for external profile from test results on REST

vkrizan:
REST v1 for profiles would return hosts for an external profile, if the
host uploaded a report.

### DIFF
--- a/app/serializers/profile_serializer.rb
+++ b/app/serializers/profile_serializer.rb
@@ -9,7 +9,12 @@ class ProfileSerializer
   belongs_to :parent_profile, record_type: :profile
   has_many :rules
   has_many :hosts do |profile|
-    profile.policy_object&.hosts || []
+    if profile.policy_object
+      profile.policy_object.hosts
+    else
+      # external profile
+      profile.test_result_hosts
+    end
   end
   has_many :test_results
   attributes :ref_id, :score, :parent_profile_id,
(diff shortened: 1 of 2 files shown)


app/serializers/profile_serializer.rb
```diff
@@ -9,7 +9,12 @@ class ProfileSerializer
   belongs_to :parent_profile, record_type: :profile
   has_many :rules
   has_many :hosts do |profile|
-    profile.policy_object&.hosts || []
+    if profile.policy_object
+      profile.policy_object.hosts
+    else
+      # external profile
```
akofink, line 15: I'm in general not a fan of comments because they can actually make the code more confusing. If you want to have a comment here, let's at least make it as accurate as possible. The `external` attribute on a profile makes it confusing IMO; I think saying "external policy" instead makes more sense since we're using 'external' as the UI uses it here.

```suggestion
      # external policy
```

akofink (approved):
Looks great, thanks!

app/serializers/profile_serializer.rb
vkrizan: I've added it deliberately, so we can remove it with the external support removal. "external policy" seems more confusing as it works with a profile, and technically there is no such thing as an "external policy".

I'll try to limit adding comments.

PR #645, opened 2020-11-09
[RHICOMPL-1093] Fix orphaned BO removal on policy deletion

akofink:
Also ensure profile business objective associations are nullified on
profile to policy copy.

### DIFF
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -55,9 +55,8 @@ class Policy < ApplicationRecord
 
   def destroy_orphaned_business_objective
     bo_changes = (previous_changes.fetch(:business_objective_id, []) +
-                  changes.fetch(:business_objective_id, [])).compact
-    return if bo_changes.blank?
-
+                  changes.fetch(:business_objective_id, []) +
+                  [business_objective_id]).compact
     BusinessObjective.without_policies.where(id: bo_changes).destroy_all
   end
 end
--- a/app/services/copy_profiles_to_policies.rb
+++ b/app/services/copy_profiles_to_policies.rb
@@ -22,7 +22,7 @@ class CopyProfilesToPolicies
   def create_policies
     profiles.where(external: false).find_each do |profile|
       policy = Policy.create!(Policy.attrs_from(profile: profile))
-      profile.update!(policy_id: policy.id)
+      profile.update!(policy_id: policy.id, business_objective_id: nil)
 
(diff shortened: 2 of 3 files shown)


app/models/policy.rb
```diff
@@ -55,9 +55,8 @@ def initial_profile
 
   def destroy_orphaned_business_objective
     bo_changes = (previous_changes.fetch(:business_objective_id, []) +
-                  changes.fetch(:business_objective_id, [])).compact
-    return if bo_changes.blank?
```
akofink, line 59: When you destroy a policy, this `bo_changes` is empty; that's the crux of the bug. Now we include the `business_objective_id` in the collection as well.

app/services/copy_profiles_to_policies.rb
```diff
@@ -22,7 +22,7 @@ def profiles
   def create_policies
     profiles.where(external: false).find_each do |profile|
       policy = Policy.create!(Policy.attrs_from(profile: profile))
-      profile.update!(policy_id: policy.id)
+      profile.update!(policy_id: policy.id, business_objective_id: nil)
```
akofink, line 25: Since we haven't deployed this yet, it makes sense to just unassign the BO from the profile so we don't face issues with dangling foreign keys.

vkrizan: Haven't we deployed it to stage already?

vkrizan (approved):
LGTM.

Any unit test? :slightly_smiling_face:

app/services/copy_profiles_to_policies.rb
akofink: Yes, but with a prod dump DB just to test the profile to policy migration. It will need to be re-run there against the normal stage DB.

akofink:
@vkrizan there are unit tests for this; I'll investigate why at least one of them wasn't failing and modify it to fail without this change.

test/models/policy_test.rb
```diff
@@ -80,7 +80,7 @@ class PolicyTest < ActiveSupport::TestCase
 
     should 'destroy business objectives without policies on destroy' do
       assert_difference('BusinessObjective.count' => -1) do
-        policies(:one).destroy
+        Policy.where(id: policies(:one)).destroy_all
```
akofink, line 83: @vkrizan crazy, but `destroy_all` has different behavior than `destroy` when it comes to the `changes` and `previous_changes` on the model. :see_no_evil: This change makes the test fail against master.

PR #647, opened 2020-11-10
[RHICOMPL-1027] Filter System profiles by policyId on GraphQL

vkrizan:
First, it introduces scope `in_policy` on Profile for querying profiles under a policy, by providing:
* the exact profile id
* any policy profile id (returns/scopes all policy profiles)
* policy id (returns/scopes all policy profiles)

Adds possibility to custom scoping of Host.all_profiles by returning an `ActiveRecord::Relation`.

And finally it adds an argument `policyId` on field `profiles` of `Systems` GraphQL type.
Provided `policyId` would filter system's assigned profiles and profiles from test results.
The behavior of filtering is the same as for the scope `policy_bunch` (on `Profile` model).

Review Recommendation: go by commits.

### DIFF
--- a/app/graphql/types/system.rb
+++ b/app/graphql/types/system.rb
@@ -11,7 +11,10 @@ module Types
     field :name, String, null: false
     field :os_major_version, Int, null: true
     field :os_minor_version, Int, null: true
-    field :profiles, [::Types::Profile], null: true
+    field :profiles, [::Types::Profile], null: true do
+      argument :policy_id, ID, 'Filter results by policy or profile ID',
+               required: false
+    end
     field :rules_passed, Int, null: false do
       argument :profile_id, String, 'Filter results by profile ID',
                required: false
@@ -25,9 +28,11 @@ module Types
                required: false
     end
 
-    def profiles
+    def profiles(policy_id: nil)
       context_parent
-      object.all_profiles
+      all_profiles = object.all_profiles
+      all_profiles = all_profiles.in_policy(policy_id) if policy_id
+      all_profiles
     end
 
     def rules_passed(args = {})
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -59,6 +59,24 @@ module ProfileSearching
     scope :os_major_version, lambda { |major, equals = true|
       where(benchmark: Xccdf::Benchmark.os_major_version(major, equals))
     }
+    scope :in_policy, lambda { |policy_or_profile_id|
+      return none unless UUID.validate(policy_or_profile_id)
+
+      policy_cond = { policy_id: policy_or_profile_id }
+      profile_cond = {
+        policy_object: {
+          profiles_policies: {
+            id: policy_or_profile_id
+          }
+        }
+      }
+
+      search = left_outer_joins(policy_object: :profiles)
+      search.where(id: policy_or_profile_id)
+            .or(search.where(policy_cond))
+            .or(search.where(profile_cond))
+            .distinct
+    }
   end
 
   class_methods do
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -30,7 +30,9 @@ class Host < ApplicationRecord
   validates :account, presence: true
 
   def all_profiles
-    (assigned_profiles + test_result_profiles).uniq
+    Profile.where(id: assigned_profiles)
+           .or(Profile.where(id: test_result_profiles))
+           .distinct
   end
 
   class << self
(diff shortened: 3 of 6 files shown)


app/models/host.rb
```diff
@@ -30,7 +30,11 @@ class Host < ApplicationRecord
   validates :account, presence: true
 
   def all_profiles
-    (assigned_profiles + test_result_profiles).uniq
+    [assigned_profiles, test_result_profiles].flat_map do |profiles|
+      scoped = profiles
+      scoped = yield scoped if block_given?
+      scoped.nil? ? profiles : scoped
+    end.uniq
```
akofink, line 37: Could we simplify this a bit? I think it'd be great to have this return an `ActiveRecord::Relation` so that we can use it like this: `Host.first.all_profiles.in_policy(policy_or_profile_id)`. I find the block here pretty confusing; it doesn't yield a single profile but the same collection passed into the flat_map (i.e. `all_profiles`). It's admittedly an idiom, but it's usually used in creating new objects, like `Config.new do |config| ...`, rather than working with collections of objects.

akofink (commented):
Looking good! I have a few initial comments inline.

app/models/concerns/profile_searching.rb
```diff
@@ -59,6 +59,24 @@ module ProfileSearching
     scope :os_major_version, lambda { |major, equals = true|
       where(benchmark: Xccdf::Benchmark.os_major_version(major, equals))
     }
+    scope :policy_bunch, lambda { |policy_or_profile_id|
+      return none unless UUID.validate(policy_or_profile_id)
```
akofink, line 63: It's fine to validate here, but it'd be preferable to validate fields like this in the API layer (either GraphQL, REST, or both). REST uses https://github.com/zendesk/stronger_parameters to validate the shape of the request (i.e. `{ data: { attributes: { ... }}}`) and return some 4XX response.

app/graphql/types/system.rb
```diff
@@ -11,7 +11,10 @@ class System < Types::BaseObject
     field :name, String, null: false
     field :os_major_version, Int, null: true
     field :os_minor_version, Int, null: true
-    field :profiles, [::Types::Profile], null: true
+    field :profiles, [::Types::Profile], null: true do
+      argument :policy_id, ID, 'Filter results by policy or profile ID',
+               required: false
+    end
```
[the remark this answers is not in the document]
vkrizan: I'm not aware of filtering by more than one policy. When it would be needed, we might add additonal `policy_ids`.

app/models/concerns/profile_searching.rb
```diff
@@ -59,6 +59,22 @@ module ProfileSearching
     scope :os_major_version, lambda { |major, equals = true|
       where(benchmark: Xccdf::Benchmark.os_major_version(major, equals))
     }
+    scope :policy_bunch, lambda { |policy_or_profile_id|
```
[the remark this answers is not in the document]
vkrizan: :+1:

app/models/host.rb
vkrizan: For it to return `ActiveRecord::Relation` it needs to do a UNION. UNION is not builtin in Rails AFAIK. Even if it was, the joins done by `in_policy` (was `policy_bunch`) scope would need to be applied for each UNION sub query. Which leads to the logic of two yields. I'd be happy to have it return a Relation. Could you suggest changes?

It yields twice because there are two collections/relations. I find it straightforward and effective. Do you have some improvement in mind, considering that the method cannot return single Relation?

app/models/concerns/profile_searching.rb
vkrizan: The validation there is to prevent database error, the same way as the builtin `where` does.
```
> Profile.where(id: 'bogus')
  Profile Load (1.9ms)  SELECT "profiles".* FROM "profiles" WHERE "profiles"."id" = $1  [["id", nil]]
=> []
```

There is a (thin) difference between a model relation scope and a controller. I agree on returning 4XX on incorrect values in controllers for most cases. Here we are focusing on internal and not much on the controller. The type checking should already have happen with the `ID` GraphQL type.

app/models/host.rb
akofink: Doesn't this work?

```suggestion
    Profile.where(id: assigned_profiles).or(Profile.where(id: test_result_profiles))
```

vkrizan: That ... might ... actually ... work! Thanks!

vkrizan: Changed.

akofink (approved):
Looks great! Thanks @vkrizan!

PR #648, opened 2020-11-10
Refactor: remove unused V1::ApplicationController

vkrizan:
increases coverage

### DIFF
--- a/app/controllers/v1/application_controller.rb
+++ b/app/controllers/v1/application_controller.rb
@@ -1,7 +1,0 @@
-# frozen_string_literal: true
-
-# Superclass for all V1 controllers
-module V1
-  class ApplicationController < ::ApplicationController
-  end
-end


akofink (approved):
We're probably going to need this when we implement the V2 API. I'm not that concerned about coverage, especially since it Coveralls misreports coverage for a few files that are completely covered (like `xccdf/hosts.rb`). But if you insist, I really don't feel strongly.

vkrizan:
@akofink I don't feel strongly about it either. Just saw it mentioned on coverage reports.

It did not have any methods, just inherited everything from the global `ApplicationController`. Unless this method-less base class is needed to (auto)load all V1 controllers, it should be safe to remove. We can re-add it when there would be a need for shared methods on V1 controllers. 

The unit test failure makes me cringe a bit:
```
=== FAIL["test_query_attributes_with_profileId_arguments", #<Minitest::Reporters::Suite:0x000055ee48c95158 @name="SystemQueryTest">, 5.49422598000001]
 test_query_attributes_with_profileId_arguments#SystemQueryTest (5.49s)
        Expected: 1
          Actual: 0
        test/graphql/queries/system_query_test.rb:97:in `block in <class:SystemQueryTest>'
``

akofink:
The test failure is odd, especially since it seems to be gone now lol. I also didn't see any test failure locally. Technically, any API::V1 controller used to use this class as the superclass, but now it just uses the normal ApplicationController. This is due to the way Ruby tries to resolve constants inside namespaces. It looks in the namespace first (i.e. `V1`) and then the next higher level. To force using the highest level, you can use this syntax: `::ApplicationController` which you've probably seen in a few places.

Anyway, it seems this PR is :green_circle:

akofink:
Also, somehow codecov reports that the coverage will decrease, not increase :thinking: So, IMO we really ought to just close this PR.

PR #651, opened 2020-11-11
fix(RHICOMPL-1097): invalidate latest_supported after ssg import

akofink:

### DIFF
--- a/lib/tasks/import_datastream.rake
+++ b/lib/tasks/import_datastream.rake
@@ -8,7 +8,6 @@ namespace :ssg do
   task import_rhel: [:environment, 'ssg:sync_rhel'] do
     # DATASTREAM_FILENAMES from openscap_parser's ssg:sync_rhel
     begin
-      Rails.cache.delete('latest_supported_benchmarks')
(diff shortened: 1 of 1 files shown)


lib/tasks/import_datastream.rake
```diff
@@ -22,23 +21,22 @@ namespace :ssg do
         data: OpenshiftEnvironment.summary
       )
     end
+    Rails.cache.delete('latest_supported_benchmarks')
   end
 
   desc 'Update compliance DB with the supported SCAP Security Guide versions'
   task import_rhel_supported: [:environment] do
     # DATASTREAM_FILENAMES from openscap_parser's ssg:sync
-    begin
```
akofink, line 30: I removed this block, since there was no `rescue` as @bastilian pointed out.

vkrizan (approved):
LGTM. Thanks!

lib/tasks/import_datastream.rake
vkrizan: Not sure why rubocop did not complained. :smile:

PR #656, opened 2020-11-13
fix(Test result deletion): RHICOMPL-1109 keeps policy profile

akofink:
The test we had for this case relied on `external=true` rather than the
policy association existing, thus not catching the bug. I've updated the
test and the code to delete the profile only if no policy exists; for
internal policies, deleting the test results for the initial profile
deletes test results for the entire policy's profiles.

### DIFF
--- a/app/graphql/mutations/test_result/delete.rb
+++ b/app/graphql/mutations/test_result/delete.rb
@@ -14,8 +14,9 @@ module Mutations
 
       def resolve(args = {})
         profile = scoped_profiles.find(args[:profile_id])
-        test_results = scoped_test_results(args).destroy_all
-        profile.destroy! if profile.external
+        profiles = profile.external ? [profile] : profile.policy_object.profiles
+        test_results = scoped_test_results(profile_id: profiles).destroy_all
+        profile.destroy! unless profile.policy_id
 
         { profile: profile, test_results: test_results }
       end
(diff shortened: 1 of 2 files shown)


app/graphql/mutations/test_result/delete.rb
```diff
@@ -14,8 +14,10 @@ class Delete < BaseMutation
 
       def resolve(args = {})
         profile = scoped_profiles.find(args[:profile_id])
-        test_results = scoped_test_results(args).destroy_all
-        profile.destroy! if profile.external
+        policy_profiles = profile.policy_object&.profiles || [profile]
+        test_results = scoped_test_results(profile_id: policy_profiles)
```
vkrizan, line 18: This creates a behavior split with profile removal. Profile removal deletes all policy profiles *only* for initial policy profile (`external: false`), whilst this removes policy's test results if *any* of the policy profile id is passed.

I'd suggest to have the same behavior as on policy profile removal: removing all policy test results only when initial profile id is passed.

app/graphql/mutations/test_result/delete.rb
```diff
@@ -14,8 +14,10 @@ class Delete < BaseMutation
...
+                       .destroy_all
+        profile.destroy! unless profile.policy_id
```
vkrizan, line 20: Observation: Only this line seems to fix the described bug.

app/graphql/mutations/test_result/delete.rb
akofink: Thanks for catching this! I agree it should behave consistently with profile deletion. Updated and added a test for this scenario.

app/graphql/mutations/test_result/delete.rb
akofink: Unfortunately, the list of test results to delete isn't necessarily correct when clicking the 'Delete report' button in the UI for profiles that are part of a policy; thus the other changes to find test results outside of the given profile ID for profiles in a policy.

vkrizan (approved):
Thanks! LGTM.

PR #658, opened 2020-11-16
feat: RHICOMPL-724 Mark test results as (un)supported

akofink:
For each test result that we save, we will compare the host's OS to the
supported SSG matrix and appropriately mark the test result as supported
or not.

Old test results will be shown as supported (see DB migration default).
Reports associated to a host with valid OS information and a matching SSG
version in the report's profile will be marked as supported.

For hosts without OS information or for mismatched OS/SSG, the test
result is marked as unsupported.

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -50,6 +50,10 @@ class Profile < ApplicationRecord
     end
   end
 
+  def ssg_version
+    benchmark.version
+  end
+
   def policy_type
     (parent_profile || self).name
   end
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -9,6 +9,16 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
   self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
 
   class << self
+    def supported?(ssg_version:, os_major_version:, os_minor_version:)
+      ssg_version == ssg_for_os(os_major_version, os_minor_version)
+    end
+
+    def ssg_for_os(os_major_version, os_minor_version)
+      raw_supported.dig('supported',
+                        "RHEL-#{os_major_version}.#{os_minor_version}",
+                        'version')
+    end
+
     def all
       raw_supported['supported'].map do |rhel, values|
         major, minor = os_version(rhel)
--- a/db/migrate/20201116160531_add_supported_to_test_results.rb
+++ b/db/migrate/20201116160531_add_supported_to_test_results.rb
@@ -1,0 +1,6 @@
+class AddSupportedToTestResults < ActiveRecord::Migration[5.2]
+  def change
+    add_column :test_results, :supported, :boolean, default: true
+    add_index :test_results, :supported
+  end
+end
--- a/test/models/supported_ssg_test.rb
+++ b/test/models/supported_ssg_test.rb
@@ -31,4 +31,35 @@ class SupportedSsgTest < ActiveSupport::TestCase
       assert_equal in_upstream.count, 3
     end
   end
+
+  context 'supported?' do
+    should 'return true for OS/SSG matches' do
+      SupportedSsg.stubs(:ssg_for_os).returns('1.2.3')
+      assert SupportedSsg.supported?(ssg_version: '1.2.3',
+                                     os_major_version: '7',
+                                     os_minor_version: '4')
+      assert_not SupportedSsg.supported?(ssg_version: '1.2.4',
+                                         os_major_version: '7',
+                                         os_minor_version: '4')
+    end
+  end
+
+  context 'ssg_for_os' do
+    should 'return true for OS/SSG matches' do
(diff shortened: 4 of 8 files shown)


app/models/profile.rb
```diff
@@ -50,6 +50,10 @@ def from_openscap_parser(op_profile, benchmark_id: nil, account_id: nil)
     end
   end
 
+  def ssg_version
+    benchmark.version
```
vkrizan, line 54: Observation: We use similar attribute in API(s). Mind to refactor?

app/models/supported_ssg.rb
```diff
@@ -9,6 +9,14 @@
   self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
 
   class << self
+    def supported?(ssg_version:, os_version:)
```
vkrizan, line 12: The OS version has to be constructed by the consumer, e.g. `"#{@host.os_major_version}.#{@host.os_minor_version}"`.
How about passing major and minor OS version with args and have the OS version, or specifically the key, constructed in this class?

app/models/supported_ssg.rb
```diff
@@ -9,6 +9,14 @@
...
+      ssg_version == ssg_for_os(os_version)
+    end
+
+    def ssg_for_os(os_version)
+      raw_supported.dig('supported', "RHEL-#{os_version}", 'version')
```
vkrizan, line 17: Observation: So far there was no expectation on the key to include "RHEL-".

vkrizan (approved):
Thank you.

We should ensure that we have OS version parsing on one place. Today it is simple to do, but the future is always unknown. :slightly_smiling_face:

db/migrate/20201116160531_add_supported_to_test_results.rb
```diff
@@ -0,0 +1,6 @@
+class AddSupportedToTestResults < ActiveRecord::Migration[5.2]
+  def change
+    add_column :test_results, :supported, :boolean, default: true
+    add_index :test_results, :supported
```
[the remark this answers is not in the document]
akofink: Because we will filter by supported test results in nearly every query after Feature 3 is complete.

app/models/supported_ssg.rb
akofink: Not sure I understand; The keys in the config are of the format `RHEL-X.Y`, as agreed upon with the OpenSCAP team. I specifically wanted to be able to look up the supported SSG by OS; that's why it's a dictionary with OS versions as keys and not an array.

app/models/supported_ssg.rb
akofink: I'm fine with it either way tbh. The inventory returns major and minor together, but we split it into major and minor when we store it in our DB and obviously have to display it as X.Y most places it's displayed. IMO it'd be overkill at this point to write some service class that splits or joins version numbers. It's unclear to me why the SupportedSsg class would be the place to put this logic.

vkrizan: Throughout the code base we should use `os_major_version` and `os_minor_version`. `SupportedSsg` constructs the key to access the data, that's it's own thing. The key can have any format, and might not necessarily match the joined OS version.

We do not necessarily have to create a separate service that would handle OS versions, at least not now.  I guess the project Cyndi and/or enhancements in Inventory could help with this concern.

app/models/supported_ssg.rb
vkrizan: Sorry, I've meant it more as a note (or something to think about) rather than a concern. Food for thought: There are other ways how to retrieve the information out of already parsed/processed values (`all`).

akofink: `all` is useless for lookup by OS (but useful for listing all versions, of course). You have to search the array to find what you're looking for. Again, the whole reason we defined the yaml with OS versions as keys to an object is so we can do lookups by OS version.

app/models/profile.rb
akofink: This was as simple as removing the ssg_version in the profile graphql type :)

app/models/supported_ssg.rb
akofink: Updated to split into major/minor.

vkrizan: Thanks :+1: 

You could have done the same all the way to `ssg_for_os` method (as it is public). :wink:

test/models/supported_ssg_test.rb
```diff
@@ -31,4 +31,35 @@ class SupportedSsgTest < ActiveSupport::TestCase
       assert_equal in_upstream.count, 3
     end
   end
+
```
akofink, line 34: Also just added some tests to cover the new methods in this class.

vkrizan:
:+1:

app/models/supported_ssg.rb
akofink: `ssg_for_os` is now also split into major/minor

akofink:
Thanks for the review(s), @vkrizan! :)

PR #660, opened 2020-11-18
fix(Profile): RHICOMPL-1077 RHICOMPL-1116 RHICOMPL-1115 GraphQL to return policy name and desc

vkrizan:
Initial (main) policy profile should return name and the description
from the policy on GraphQL, which are user defined.
Profile and pseudo policy (profile) would return same name and the
description on GraphQL, if they are the same objects.

This will fix the problem of caching on UI, where the same GQL objects
(identified by id) had a different values depending on context.
It should also improve and fix filtering and displaying in UI.

### DIFF
--- a/app/graphql/types/concerns/profile_pseudo_policy.rb
+++ b/app/graphql/types/concerns/profile_pseudo_policy.rb
@@ -5,18 +5,19 @@ module Types
   module ProfilePseudoPolicy
     extend ActiveSupport::Concern
 
+    def name
+      object == policy ? object.policy_object.name : object.name
+    end
+
+    def description
+      object == policy ? object.policy_object.description : object.description
+    end
+
     # Pseudo policy with a Profile type
-    # inheriting most of the attributes form the policy.
     def policy
       return if object.canonical? || object.policy_object.nil?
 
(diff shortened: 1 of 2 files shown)


app/graphql/types/concerns/profile_pseudo_policy.rb
```diff
@@ -5,18 +5,21 @@ module Types
   module ProfilePseudoPolicy
     extend ActiveSupport::Concern
 
+    # Use policy name for a policy profile
+    def name
+      object == policy ? object.policy_object.name : object.name
+    end
+
+    # Use policy name for a policy profile
```
akofink, line 13: Gotta say, I really hate comments that describe what the code is doing when it's obvious. Even more so when they describe what the code isn't doing.

```suggestion
    # Use policy description for a policy profile
```

akofink (approved):
Looks good, just one comment about a comment :)

app/graphql/types/concerns/profile_pseudo_policy.rb
vkrizan: My bad. :smile:

PR #662, opened 2020-11-18
feat(scoring): RHICOMPL-1085 report accurate scores for a policy

akofink:
For external policies, the score is the same as before, just considering
the single profile. For internal policies, the score must be an
aggregate of its profiles.

WIP for tests and testing with the frontend PR


app/models/policy.rb
```diff
@@ -44,6 +45,18 @@ def update_hosts(new_host_ids)
     end)
   end
 
+  def compliant?(host)
+    score(host: host) >= compliance_threshold
+  end
+
+  def score(host: nil)
```
vkrizan, line 52: Observation: This method is copied from `ProfileScoring`. Mind to have one copy of it in a concern?

app/models/policy.rb
```diff
@@ -44,6 +45,18 @@ def update_hosts(new_host_ids)
...
+    results = test_results.latest
+    results = results.where(host: host) if host
```
vkrizan, line 54: Resulting query:
```
SELECT "test_results".*
FROM "test_results"
INNER JOIN "profiles" ON "test_results"."profile_id" = "profiles"."id"
JOIN (
    SELECT "test_results"."profile_id", "test_results"."host_id", MAX(end_time) as end_time
    FROM "test_results" 
    INNER JOIN "profiles" ON "test_results"."profile_id" = "profiles"."id"
    WHERE "profiles"."policy_id" = 'UUID'
    GROUP BY "test_results"."profile_id", "test_results"."host_id"
) 
as tr on test_results.profile_id = tr.profile_id
    AND test_results.host_id = tr.host_id 
    AND test_results.end_time = tr.end_time
WHERE "profiles"."policy_id" = $1
```

Seems to be correct.

app/graphql/types/concerns/profile_scoring.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module Types
+  # Methods related to profile scoring
+  module ProfileScoring
+    extend ActiveSupport::Concern
+
+    def compliant_host_count
+      ::CollectionLoader.for(
+        object.class,
+        object.policy_id ? :policy_test_result_hosts : :test_result_hosts
```
vkrizan, line 11: Food for thought:
Where the result is dependent on a policy assignment, we could introduce a separate fields. For example `policyCompliantHostCount`. Each profile then can report it's own compliance and counts.

app/graphql/types/profile.rb
```diff
@@ -3,12 +3,14 @@
 require_relative 'interfaces/rules_preload'
 require_relative 'concerns/test_results'
 require_relative 'concerns/profile_pseudo_policy'
+require_relative 'concerns/profile_scoring'
 
 module Types
   # Definition of the Profile type in GraphQL
   class Profile < Types::BaseObject
```
vkrizan, line 10: There are several Profile only fields reporting scores and compliance:
* score
* compliant
* rules_passed and rules_failed
* last_scanned

I guess they should stay profile-based except the `last_scanned`. WDYT?

vkrizan (approved):
Other than the repeated code, it looks good. 
Thanks.

app/graphql/types/concerns/profile_scoring.rb
```diff
@@ -0,0 +1,29 @@
...
+    def total_host_count
+      ::CollectionLoader.for(policy_or_report.class, :hosts)
+                        .load(policy_or_report).then(&:count)
+    end
+
+    def test_result_host_count
+      ::CollectionLoader.for(
+        object.class,
+        object.policy_id ? :policy_test_result_hosts : :test_result_hosts
+      ).load(object).then(&:count)
```
vkrizan, line 26: This does not count hosts distinctively. I had a host reporting under with SSGs (before there was a test result auto-remove) and now it reports as two hosts. I'll never get 100%, unless I clean it manually. 

I guess we should not depend on the test-result auto-remove, and always count it distinctively. There might be the same problem with the compliant host count (if I did have the system compliant for both reporting SSGs).

app/models/policy.rb
akofink: Added another `ProfilePolicyScoring` module.

app/graphql/types/profile.rb
akofink: Right, they are all (except `compliant`) technically on TestResult, but they can exist on profile within the context of a single host as the latest TestResult for that host and profile, which is what it means today. I noticed the `score` attribute has an _optional_ `system_id` argument :thinking: seems that should be required.

app/graphql/types/concerns/profile_scoring.rb
akofink: TIL we can add `-> { distinct }` to our relationships. I'll be adding it to `test_result_hosts` and `policy_test_result_hosts`.

akofink: Ok, so adding `distinct` here fixes the host total counts, but I'm not sure how we should handle the score in the UI. It looks like we're computing an average of all SSG versions, which is definitely _not_ what we want to do. The following host has two reports from different SSG versions, with scores of 95 and 45:

image

Also, the number of compliant systems needs to be fixed for this scenario still.

akofink: Looking at the screenshot above, I realized it's not listing my other system! That system has a report for this policy but was unassigned from it. I'll need to fix it. Here I've re-assigned the second host to the policy:

image

akofink:
Thanks for the reviews, @vkrizan! Merging this with some remaining issues that we will fix later.

PR #664, opened 2020-11-18
fix(graphql): RHICOMPL-1096 Systems -> profiles returns policies

akofink:
When querying GraphQL for systems { profiles { ... } }, we used to
return all profiles, including test_result_profiles and
assigned_profiles. Now, we return only initial profiles of any assigned
policies and any external policy test result profiles; this is
essentially the complete policy list, internal and external.

### DIFF
--- a/app/graphql/types/system.rb
+++ b/app/graphql/types/system.rb
@@ -30,7 +30,7 @@ module Types
 
     def profiles(policy_id: nil)
       context_parent
-      all_profiles = object.all_profiles
+      all_profiles = object.policy_profiles
       all_profiles = all_profiles.in_policy(policy_id) if policy_id
       all_profiles
     end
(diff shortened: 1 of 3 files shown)


app/graphql/types/system.rb
```diff
@@ -30,7 +30,7 @@ class System < Types::BaseObject
 
     def profiles(policy_id: nil)
       context_parent
-      all_profiles = object.all_profiles
+      all_profiles = object.policy_profiles
```
vkrizan, line 33: We cannot restrict it to only one set of profiles without breaking the API and UI. The API here has to report all system's profile, including `external: true` policy profiles that the system is reporting under. Each of the returned profile caries its own result of a report (passed/failed rules). That's by design.

vkrizan: Here is one of the reason: https://github.com/RedHatInsights/compliance-frontend/pull/832#discussion_r526327011

akofink:
Closing in favor of RHICOMPL-1139

PR #666, opened 2020-11-20
fix(Profile): RHICOMPL-1140 policy name and description on GQL

vkrizan:
Aligns GraphQL with REST API in regards to returning name and
description of profiles that are under a policy.  Both now take the
fields from its policy.

This fixes frontend Systems table and details, where reported
profiles and rules are listed.  On System details cards will
list correct user-defined name, and the rules table will show
correct value on the Policy column.

To get the original profile name consumer can request `policyType`
(or retrieve it from a corresponding canonical profile).

### DIFF
--- a/app/graphql/types/concerns/profile_pseudo_policy.rb
+++ b/app/graphql/types/concerns/profile_pseudo_policy.rb
@@ -6,11 +6,11 @@ module Types
     extend ActiveSupport::Concern
 
     def name
-      object == policy ? object.policy_object.name : object.name
+      object.policy_object&.name || object.name
     end
 
     def description
-      object == policy ? object.policy_object.description : object.description
+      object.policy_object&.description || object.description
     end
 
     # Pseudo policy with a Profile type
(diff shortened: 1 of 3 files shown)


akofink (approved):
Looks good, and it works as expected.

There's one weird edge case relating to what happens when you remove a reporting host from a policy. In that case, the system details page still shows the policy and not the profile. It also still uses the policy's threshold, etc. I think that's fine for now; we can fix it if requested, but it kind of makes sense that it would show the policy (telling the user: you uploaded a report for this policy at some point, but the policy list in the systems table and the policy details don't actually contain the system).

Thanks, @vkrizan!

vkrizan:
@akofink Seems like a valid issue. Mind to capture it? I guess it could be under a family of RHICOMPL-1139 (contextual system profiles). Thank you.

akofink:
Yes, it seems related to RHICOMPL-1139. I think we can report this if it's brought up by a user or PM since it's an edge case that is handled just fine here; we're just not clear if we're handling it the way PM intended :)

PR #668, opened 2020-11-23
fix(ApplicationController): RHICOMPL-1129 handle validation errors

vkrizan:
Errors from validation coming from `ActiveRecord::RecordInvalid` are now rendered on all controller endpoints.

Example:
```
{"errors":["Profile policy type must be unique. Another policy with this policy type already exists."]}
```

This also fixes error `Unexpected token < in JSON at position 0` in the frontend happening if a user tried to create the same type of a policy as an existing policy for a major OS version (e.g. when the latest benchmark changed).

Here is the result in frontend with this change:
Screenshot_2020-11-23 Compliance

`render_error` refactored to accept string messages.
Error messages in the response JSON are now always an array.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,6 +37,15 @@ class ApplicationController < ActionController::API
                  status: :not_found
   end
 
+  rescue_from ActiveRecord::RecordInvalid do |error|
+    logger.info "#{error.message} (#{error.class})"
+    if error.record
+      render_model_errors(error.record)
+    else
+      render_error(error.message)
+    end
+  end
+
   rescue_from ActionController::ParameterMissing do |error|
     logger.info "#{error.message} (#{error.class})"
     render_error "Parameter missing: #{error.message}",
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,31 +22,30 @@ class ApplicationController < ActionController::API
   end
 
   rescue_from ActiveRecord::RecordNotUnique do |error|
-    render json: {
-      errors: "Duplicate record: #{error.message[/Key \(.+\).+\./]}"
-    }, status: :conflict
+    render_error "Duplicate record: #{error.message[/Key \(.+\).+\./]}",
+                 status: :conflict
   end
 
   rescue_from Pundit::NotAuthorizedError do
-    render json: { errors: 'You are not authorized to access this action.' },
-           status: :forbidden
+    render_error 'You are not authorized to access this action.',
+                 status: :forbidden
   end
 
   rescue_from ActiveRecord::RecordNotFound do |error|
     logger.info "#{error.message} (#{error.class})"
-    render json: { errors: "#{error.model} not found with ID #{error.id}" },
-           status: :not_found
+    render_error "#{error.model} not found with ID #{error.id}",
+                 status: :not_found
   end
 
   rescue_from ActionController::ParameterMissing do |error|
     logger.info "#{error.message} (#{error.class})"
-    render json: { errors: "Parameter missing: #{error.message}" },
-           status: :unprocessable_entity
+    render_error "Parameter missing: #{error.message}",
+                 status: :unprocessable_entity
   end
 
   rescue_from StrongerParameters::InvalidParameter do |error|
     logger.info "#{error.message} (#{error.class})"
-    render json: { errors: error.message },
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/rendering.rb
```diff
@@ -11,10 +11,15 @@ def render_json(model, **args)
       render({ json: serializer.new(model, opts) }.merge(args))
     end
 
-    def render_error(models, **args)
+    def render_error(messages_or_models, status: :not_acceptable, **opts)
+      messages = [messages_or_models].flatten
+      messages = messages.flat_map do |msg_or_model|
+        msg_or_model.respond_to?(:errors) ? model_errors(msg_or_model) : msg_or_model
+      end
```
akofink: I'm not a huge fan of type checking like this. We could avoid it by simply using keyword arguments:

```suggestion
def render_error(messages: [], models: [], status: :not_acceptable, **args)
  messages ||= model_errors(models)
  ...
end
```

Also, `model_errors` is only used here, I think. So if you decide not to use it for some reason, we should remove it.

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -592,21 +592,21 @@ class UpdateTest < ProfilesControllerTest
...
-                     JSON.parse(response.body).dig('errors')
+                     JSON.parse(response.body).dig('errors')[0]
       end
 
       test 'update with invalid data' do
         patch profile_path(@profile.id), params: params('foo')
         assert_response :unprocessable_entity
         assert_match 'data must be a hash',
-                     JSON.parse(response.body).dig('errors')
+                     JSON.parse(response.body).dig('errors')[0]
```
akofink, line 617: `dig` takes integer indices, too

```rb
JSON.parse(response.body).dig('errors', 0)
```

app/controllers/application_controller.rb
```diff
@@ -37,6 +37,11 @@ def pundit_scope
                  status: :not_found
   end
 
+  rescue_from ActiveRecord::RecordInvalid do |error|
+    logger.info "#{error.message} (#{error.class})"
+    render_error(error.record || error.message)
```
akofink, line 42: What is the case when `record` is `nil`?

akofink (commented):
Looks great, @vkrizan! Just a few small comments.

test/controllers/v1/profiles_controller_test.rb
vkrizan: :+1:

app/controllers/application_controller.rb
vkrizan: I don't know when it happens, however the `record` can be nil. Here is the implemenation: https://github.com/rails/rails/blob/v5.2.4.3/activerecord/lib/active_record/validations.rb#L24

app/controllers/concerns/rendering.rb
vkrizan: I wanted to avoid changing the signature, as it is already being used for model errors. This is a duck-typing (checking an "interface").

akofink: Duck typing is allowing different types to respond to the same message/method without explicit type checks. Like if you had a Table class and a Cat class and both respond to `number_of_legs`. When you write `respond_to?` that's an explicit type check which defeats the purpose of duck typing (the purpose being to hide complexity and share code).

This method is only used in the profiles controller, so it would be trivial to change the signature.

vkrizan: I've reworked it by splitting it to a different method. Check out first commit (c192fef).

`responds_to?` is a feature/interface check (like the `number_of_legs` from your example). I don't want to dispute your interpretation of duck-typing. Similar principle is for example used in Rails. See [`acts_like?` implementation](https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/core_ext/object/acts_like.rb) for inspiration.

I guess your concern was more to the fact that there were possible features checked. One that if it had `errors` and second if it hadn't.

(Technically the check was incomplete, as the subsequent method read `full_messages` out of `errors`.)

akofink (approved):
Thanks, @vkrizan!

app/controllers/application_controller.rb
vkrizan: Note, that it had to be changed to use `render_model_errors` if the record is set.

PR #670, opened 2020-11-26
feat(System): RHICOMPL-1139 test result profiles and policies on APIs 

vkrizan:
Extends System GraphQL type:
* `testResultProfiles` 
* `policies` (initial policy profiles out of assigned profiles, where `external=false`)

The GraphQL fields have the option to filter by `policyId` argument, which filters by any policy profile ID. On `testResultProfiles` the `policyId` filter works even if the host/system is disassociated from its policy.

Note that there is an original `profiles` on both APIs, that returns combined result of test result profiles and associated result. It is kept for compatibility. ~~Changing it into `testResultProfiles` might be up for a question.~~ (Edit: having only test result profiles on `profiles` field would break the API, as it won't include assigned profiles if the host system have not had a report).

Rationale:
The consumer can pick any type of context with regards to profiles. For reports the interest is focused on test result profiles, regardless of (dis)association to a policy. For the context of describing system's policies the focus is only on associated policies. There are two field so both contexts can be requested in the same query.

Absence of external policies is deliberate, as their support is soon to be deprecated.

### DIFF
--- a/app/graphql/types/concerns/system_profiles.rb
+++ b/app/graphql/types/concerns/system_profiles.rb
@@ -15,6 +15,12 @@ module Types
       scope_profiles(object.test_result_profiles, policy_id)
     end
 
+    def policies(policy_id: nil)
+      context_parent
+      policy_profiles = object.assigned_profiles.external(false)
+      scope_profiles(policy_profiles, policy_id)
+    end
+
     private
 
     def scope_profiles(profiles, policy_id)
--- a/app/graphql/types/concerns/system_profiles.rb
+++ b/app/graphql/types/concerns/system_profiles.rb
@@ -7,9 +7,18 @@ module Types
 
     def profiles(policy_id: nil)
       context_parent
-      all_profiles = object.all_profiles
-      all_profiles = all_profiles.in_policy(policy_id) if policy_id
-      all_profiles
+      scope_profiles(object.all_profiles, policy_id)
+    end
+
+    def test_result_profiles(policy_id: nil)
+      context_parent
+      scope_profiles(object.test_result_profiles, policy_id)
+    end
+
+    private
+
+    def scope_profiles(profiles, policy_id)
+      policy_id ? profiles.in_policy(policy_id) : profiles
     end
   end
 end
--- a/app/graphql/types/concerns/system_profiles.rb
+++ b/app/graphql/types/concerns/system_profiles.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Types
+  # Methods related to system profiles and policies
+  module SystemProfiles
+    extend ActiveSupport::Concern
+
+    def profiles(policy_id: nil)
+      context_parent
+      all_profiles = object.all_profiles
+      all_profiles = all_profiles.in_policy(policy_id) if policy_id
+      all_profiles
+    end
+  end
+end
--- a/app/graphql/types/system.rb
+++ b/app/graphql/types/system.rb
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 
+require_relative 'concerns/system_profiles'
+
 module Types
   # Definition of the System GraphQL type
   class System < Types::BaseObject
+    include SystemProfiles
+
     model_class ::Host
     graphql_name 'System'
     description 'A System registered in Insights Compliance'
@@ -26,13 +30,6 @@ module Types
     field :last_scanned, String, null: true do
       argument :profile_id, String, 'Filter results by profile ID',
                required: false
-    end
-
-    def profiles(policy_id: nil)
-      context_parent
-      all_profiles = object.all_profiles
-      all_profiles = all_profiles.in_policy(policy_id) if policy_id
-      all_profiles
     end
 
     def rules_passed(args = {})
--- a/app/graphql/types/system.rb
+++ b/app/graphql/types/system.rb
@@ -19,6 +19,11 @@ module Types
       argument :policy_id, ID, 'Filter results by policy or profile ID',
                required: false
     end
+    field :test_result_profiles, [::Types::Profile], null: true do
+      argument :policy_id, ID,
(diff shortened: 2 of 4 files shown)


app/serializers/host_serializer.rb
```diff
@@ -18,4 +18,8 @@ class HostSerializer
     host.all_profiles
   end
   # rubocop:enable Style/SymbolProc
+
+  has_many :policies do |host|
+    host.assigned_profiles.external(false)
+  end
```
akofink, line 24: So far, we haven't polluted the REST API with this concept that a policy is its initial profile, etc. I'd like to keep it that way. If this is returning profiles, it should be named appropriately. IMO it should be as explicit as possible, like `assigned_internal_profiles` or similar.

I'm curious why we're even changing the REST API - I expected this PR to be GraphQL only.

app/graphql/types/concerns/system_profiles.rb
```diff
@@ -15,6 +15,12 @@ def test_result_profiles(policy_id: nil)
       scope_profiles(object.test_result_profiles, policy_id)
     end
 
+    def policies(policy_id: nil)
+      context_parent
+      policy_profiles = object.assigned_profiles.external(false)
+      scope_profiles(policy_profiles, policy_id)
+    end
```
akofink, line 22: Same comment here wrt naming. Let's be as explicit as possible. Naming it `policies` only increases the confusion surrounding an already confusing topic. There is really no detriment to having descriptive attribute names in our APIs. `assigned_internal_profiles` or similar would communicate better what is being returned.

app/serializers/host_serializer.rb
```diff
@@ -11,6 +11,7 @@ class HostSerializer
   end
 
   has_many :test_results
+  has_many :test_result_profiles
```
akofink, line 14: Again, not sure why we're changing the REST API. This addition makes sense, of course, I just don't know why it should be added with this change.

app/graphql/types/system.rb
```diff
@@ -19,6 +19,10 @@ class System < Types::BaseObject
       argument :policy_id, ID, 'Filter results by policy or profile ID',
                required: false
     end
+    field :test_result_profiles, [::Types::Profile], null: true do
+      argument :policy_id, ID, 'Filter results by policy or profile ID',
+               required: false
```
akofink, line 24: Maybe we should mention that this filters based on test results and not on which hosts are assigned to a policy:

`'Filter results tested against a policy or profile ID'` or similar.

akofink (commented):
Looks great! The graphql stuff just has a minor comment. The REST related changes shouldn't be included with this PR IMO; the Jira specifically mentions GraphQL as a way to support the UI. I think we have some phase 5 work to ensure the REST API maintains parity with the capabilities provided in the UI.

app/serializers/host_serializer.rb
vkrizan: I thought that the parity between APIs was needed. I can take the REST API changes out.

app/graphql/types/concerns/system_profiles.rb
vkrizan: Naming is wrt `Profile.policy`. The name "policies" is deliberate, in a sense that we can change the underlying type keeping the subset of `Profile` type policy-related fields.

`assigned_internal_profiles` would not be accurate. These has to return external profiles/policies until we drop the external support completely. Only "profiles" is accurate from the name. There is no such thing as assigned profile, but an assigned policy. "Internal" adds more confusion to it TBH, as everything would be/is internal.

Let me think of some other name... We need remember that we are in the v1 sphere (even if the GQL isn't versioned) where the real policy does not exist.

vkrizan: Think about it on the abstract level where the first policy profile represents *the* policy.

akofink: I understand what the idea is here, I just don't think the naming accurately reflects what's being returned. One day, we may introduce actual policies to the GraphQL API, and we'll be in a pickle for what to name them. Since we already have the concept of policies on a profile in our GraphQL API, I guess it's okay here since it matches.

vkrizan: Introduction of actual policies on the same field would have very minimal impact. Consumers, including our fronted, would not need many changes. The choice for this name is by design, specifically for this situation.

vkrizan: My alternative name would be `policy_profiles`.

app/graphql/types/concerns/system_profiles.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_result_profiles(policy_id: nil)
+      context_parent
+      scope_profiles(object.test_result_profiles, policy_id)
+    end
+
+    def policies(policy_id: nil)
+      context_parent
+      policy_profiles = object.assigned_profiles.external(false)
```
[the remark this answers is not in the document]
vkrizan: I'm going to deal with this on frontend side, as a band-aid until January.

app/serializers/host_serializer.rb
vkrizan: REST API changes dropped.

app/serializers/host_serializer.rb
vkrizan: REST API changes dropped.

app/graphql/types/system.rb
vkrizan: Changed.

app/graphql/types/concerns/system_profiles.rb
akofink: Ok, that's a fair point. I guess the only thing that will change will be the `id` returned with the policy if we ever actually start returning policy objects.

akofink (approved):
Thanks, @vkrizan!

PR #672, opened 2020-12-02
feat(graphql): RHICOMPL-1114 RHICOMPL-681 supported and scoring

akofink:
Changes to support https://github.com/RedHatInsights/compliance-frontend/pull/845

RHICOMPL-1114: Expose supported information in GraphQL
RHICOMPL-681: Only score supported reports

### DIFF
--- a/app/graphql/types/concerns/profile_scoring.rb
+++ b/app/graphql/types/concerns/profile_scoring.rb
@@ -8,7 +8,7 @@ module Types
     def unsupported_host_count
       ::CollectionLoader.for(
         object.class,
-        object.policy_id ? :policy_test_results : :test_results
+        test_result_method
       ).load(object).then do |test_results|
         test_results.latest.supported(false).count
       end
@@ -17,9 +17,10 @@ module Types
     def compliant_host_count
       ::CollectionLoader.for(
         object.class,
-        object.policy_id ? :policy_test_result_hosts : :test_result_hosts
-      ).load(object).then do |hosts|
-        hosts.count { |host| object.compliant?(host) }
+        test_result_method
+      ).load(object).then do |test_results|
+        Host.where(id: test_results.latest.supported.select(:host_id))
+            .count { |host| object.compliant?(host) }
       end
     end
 
@@ -31,8 +32,16 @@ module Types
     def test_result_host_count
       ::CollectionLoader.for(
         object.class,
-        object.policy_id ? :policy_test_result_hosts : :test_result_hosts
-      ).load(object).then(&:count)
+        test_result_method
+      ).load(object).then do |test_results|
+        Host.where(id: test_results.latest.supported.select(:host_id)).count
+      end
+    end
+
+    private
+
+    def test_result_method
+      object.policy_id ? :policy_test_results : :test_results
     end
   end
 end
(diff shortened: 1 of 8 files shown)


app/graphql/types/concerns/profile_scoring.rb
```diff
@@ -31,8 +32,16 @@ def total_host_count
     def test_result_host_count
       ::CollectionLoader.for(
         object.class,
-        object.policy_id ? :policy_test_result_hosts : :test_result_hosts
-      ).load(object).then(&:count)
+        test_result_method
+      ).load(object).then do |test_results|
+        Host.where(id: test_results.latest.supported.select(:host_id)).count
```
vkrizan, line 37: I'm not sure whether not counting unsupported here is correct. Logically I'd expect `testResultHostCount` return number of host from all latest test results, regardless of supportability. Consumer can construct the number by subtracting `unsupportedHostCount`.

akofink: I guess this gets to @bastilian's point about handling things in the backend as much as possible. What do you think here, @bastilian?

The way I see it, `compliantHostCount` and `totalHostCount` should never include unsupported hosts, even from the backend, since that's what the jira means by 'Report scoring is based only on systems using supported SSG versions'. When we implement this same thing in the REST API, it will also need to report counts without unsupported systems included, no UI logic involved.

vkrizan: It would redefine meaning of "total". I see this as a possible API breakage. `supportedHostCount` would be more accurate.

akofink: Exactly. Total now refers only to systems with supported uploads - that's the feature.

If we add `supportedHostCount`, we'd have to update all frontend uses of `totalHostCount` and change it to `supportedHostCount`, at which point we could remove `totalHostCount` completely (and you could claim we would break the API by doing so).

I don't think your point is invalid (especially wrt explicit naming), I just think that we are supposed to change the meaning of 'total' for this feature when it comes to policy scoring.

app/graphql/types/concerns/profile_scoring.rb
```diff
@@ -5,12 +5,22 @@ module Types
...
+
     def compliant_host_count
       ::CollectionLoader.for(
         object.class,
-        object.policy_id ? :policy_test_result_hosts : :test_result_hosts
-      ).load(object).then do |hosts|
-        hosts.count { |host| object.compliant?(host) }
+        test_result_method
+      ).load(object).then do |test_results|
+        Host.where(id: test_results.latest.supported.select(:host_id))
```
bastilian, line 22: Could it not be `test_results.hosts...` or is that not possible due to GraphQL?

app/graphql/types/concerns/profile_scoring.rb
vkrizan: Oh, apologies, this is `testResultHostCount` not the `totalHostCount`. Sorry, it was too many changes to review today.

vkrizan (approved):
Thank you.

app/graphql/types/concerns/profile_scoring.rb
akofink: I also didn't realize that :)

app/graphql/types/concerns/profile_scoring.rb
akofink: We're joining multiple test results and multiple hosts. What you're thinking of maybe is `test_result.hosts`; it'd be like `test_results.map(&:hosts)` here, which would create N queries.

PR #675, opened 2020-12-04
feat(Benchmark): RHICOMPL-1121 latest supported from SSG matrix 

vkrizan:
Return latest supported benchmarks, one each per OS major version. Benchmarks are matched from the SSG supported matrix, picked by latest minor version.

Drops caching of latest supported benchmarks, as it refactors it into a single SQL query within a scope.

Please note, that due to the use of upstream SSG packages, the benchmarks are being matched with the upstream version.

Review suggestion: go by commits.

RHICOMPL-1121

### DIFF
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -7,6 +7,13 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
                           :os_major_version, :os_minor_version,
                           keyword_init: true) do
   self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
+  OS_NAME = 'RHEL'
+  self::OS_NAME = OS_NAME
+
+  def ref_id
+    'xccdf_org.ssgproject' \
+    ".content_benchmark_#{OS_NAME}-#{os_major_version}"
+  end
 
   class << self
     def supported?(ssg_version:, os_major_version:, os_minor_version:)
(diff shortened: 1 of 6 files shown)


app/models/supported_ssg.rb
```diff
@@ -7,16 +7,25 @@
                           :os_major_version, :os_minor_version,
                           keyword_init: true) do
   self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
+  OS_NAME = 'RHEL'
+  self::OS_NAME = OS_NAME
```
vkrizan, line 11: Since this is not a proper class, this constant is created on a class and on an instance. Please suggest an alternative, if there is one.

app/models/xccdf/benchmark.rb
```diff
@@ -20,17 +20,19 @@ class Benchmark < ApplicationRecord
...
-      'xccdf_org.ssgproject.content_benchmark_RHEL-7': '0.1.50',
-      'xccdf_org.ssgproject.content_benchmark_RHEL-8': '0.1.50'
-    }.freeze
-
     scope :os_major_version, lambda { |major, equals = true|
       where(os_major_version_query(major, equals))
     }
 
+    scope :latest_supported, lambda {
+      SupportedSsg.latest_per_os_major.inject(none) do |supported, ssg|
```
akofink, line 28: TIL `none`

akofink (approved):
Nice work! Thanks, @vkrizan!

vkrizan:
> Using Struct.new with a block is pretty slick, but maybe it's time to define this as a proper class.

Yeah, I thought the same. We can refactor it post-hoc.

app/models/xccdf/benchmark.rb
```diff
@@ -20,17 +20,19 @@ class Benchmark < ApplicationRecord
...
+        supported.or(
+          where(ref_id: ssg.ref_id,
+                version: ssg.upstream_version || ssg.version)
+        )
+      end
```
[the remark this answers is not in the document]
vkrizan: I try to do refactoring along the requested changes. It could be preferred to have them separately. With our processes that gets more difficult.

app/models/supported_ssg.rb
vkrizan: Yeah, I thought the same. We can refactor it post-hoc.

PR #677, opened 2020-12-07
feat(REST): RHICOMPL-1001 Count only supported results

akofink:
This aligns the REST API counts to the GraphQL counts for profile test
results. These attributes are used by the Insights Dashboard and must
match what GraphQL returns to make the dashboard consistent with the
Compliance reports page.

Test with https://github.com/RedHatInsights/insights-dashboard/pull/212


app/models/concerns/profile_scoring.rb
```diff
@@ -4,6 +4,23 @@
 module ProfileScoring
   include ProfilePolicyScoring
 
+  def total_host_count
+    Host.where(id: assigned_hosts).or(Host.where(id: hosts)).count
```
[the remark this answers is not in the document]
akofink: Nice catch! I've added this case to the existing graphql profile query test, and it fails with this code as expected and passes with the new code.

test/graphql/queries/profile_query_test.rb
```diff
@@ -274,7 +274,7 @@ class ProfileQueryTest < ActiveSupport::TestCase
     (host3 = hosts(:two).dup).update!(name: 'host3')
     policies(:one).update(compliance_threshold: 95,
                           account: accounts(:test),
-                          hosts: [hosts(:one), hosts(:two), host3])
+                          hosts: [hosts(:one), host3])
```
akofink, line 277: Creating the scenario where we have test results for a host but it's not assigned to the policy.

test/graphql/queries/profile_query_test.rb
```diff
@@ -285,8 +285,12 @@ class ProfileQueryTest < ActiveSupport::TestCase
     profile1_result = result['data']['allProfiles'].find do |h|
       h['id'] == profiles(:one).id
     end
+    profile2_result = result['data']['allProfiles'].find do |h|
+      h['id'] == profiles(:two).id
+    end
     assert_equal policies(:one).name, profile1_result['name']
-    assert_equal 3, profile1_result['totalHostCount']
+    assert_equal 2, profile1_result['totalHostCount']
+    assert_equal 2, profile2_result['totalHostCount']
```
akofink, line 293: And, as expected, no matter which profile queried for `totalHostCount`, the result is the same.

vkrizan (approved):
Nice. Thank you.

PR #678, opened 2020-12-09
feat(Host): RHICOMPL-1138 Search by policy results

akofink:
Add a scoped search for filtering hosts to those with test results for a
given profile or policy ID. This filter will be used on the report
details page for the systems table to show hosts with reports for a
policy, even if the host is not currently assigned to the policy.

Test with https://github.com/RedHatInsights/compliance-frontend/pull/865


vkrizan (approved):
Works like a charm. Thank you.

I like the refactoring separation. It made reviewing it much quicker. :+1:

PR #679, opened 2020-12-09
fix(compose): ingress port use

vkrizan:
Ingress uses port 3000, and second port 8080 as metrics port.
The env variable INGRESS_PORT seems to be undocumented (not read?) and
its value conflicts with the metrics port.

Instead, it maps the port 3000 to be accessible via 8080 from outside.

See also Ingress README:
https://github.com/RedHatInsights/insights-ingress-go/blob/27a929894e645ddae9afb055495e4e823098ab17/README.md

and dev compose:
https://github.com/RedHatInsights/insights-ingress-go/blob/61330931ee5f684f54293be049b05d1648cca7ca/development/local-dev-start.yml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,14 +76,13 @@ services:
       - INGRESS_VALIDTOPICS=testareno,advisor,buckit,compliance #if you test a different topic, add it here
       - INGRESS_INVENTORYURL=inventory-web:8081/api/inventory/v1/hosts
       - OPENSHIFT_BUILD_COMMIT=woopwoop
-      - INGRESS_PORT=8080
       - INGRESS_MINIODEV=true
       - INGRESS_MINIOACCESSKEY=$MINIO_ACCESS_KEY
       - INGRESS_MINIOSECRETKEY=$MINIO_SECRET_KEY
       - INGRESS_MINIOENDPOINT=minio:9000
       - INGRESS_MAXSIZE=104857600 # 100 MB
     ports:
-      - 8080:8080
+      - 8080:3000
     depends_on:
       - kafka
     links:


vkrizan:
The env vars should be `INGRESS_WEB_PORT` and `INGRESS_METRICS_PORT`.

docker-compose.yml
```diff
@@ -76,14 +76,13 @@ services:
...
       - OPENSHIFT_BUILD_COMMIT=woopwoop
-      - INGRESS_PORT=8080
       - INGRESS_MINIODEV=true
       - INGRESS_MINIOACCESSKEY=$MINIO_ACCESS_KEY
       - INGRESS_MINIOSECRETKEY=$MINIO_SECRET_KEY
       - INGRESS_MINIOENDPOINT=minio:9000
       - INGRESS_MAXSIZE=104857600 # 100 MB
     ports:
-      - 8080:8080
+      - 8080:3000
```
akofink, line 85: But this conflicts with our backend port. We don't care about metrics from ingress. Did this start failing for you with your recent image pull we were discussing yesterday?

akofink:
Are you saying the env variables INGRESS_WEB_PORT, INGRESS_METRICS_PORT should be named that way in your opinion, or that they are the new env variables with this new image? It doesn't seem to work for me:

```
podman run -it -e INGRESS_WEB_PORT=12345 quay.io/cloudservices/insights-ingress
open : no such file or directory
{"@timestamp":"2020-12-09T14:51:00.706Z","@version":1,"Web Port":3000,"app":"ingress","caller":"main.main","levelname":"info","message":"Starting Service","source_host":"06cc0ebff1cf"}{"@timestamp":"2020-12-09T14:51:00.706Z","@version":1,"Metrics Port":8080,"app":"ingress","caller":"main.main","levelname":"info","message":"Starting Service","source_host":"06cc0ebff1cf"}
```

docker-compose.yml
akofink: Maybe you meant `3000:8080` ?

akofink:
BTW, this definitely broke in the last 6 weeks. Here's an image from 6 weeks ago that works with the INGRESS_PORT env var:

```
podman run -it -e INGRESS_PORT=8080 quay.io/cloudservices/insights-ingress 
{"@timestamp":"2020-12-09T14:48:38.854Z","@version":1,"app":"ingress","caller":"main.main","levelname":"info","message":"Starting Service","port":8080,"source_host":"ed6c5babed33"}^C{"@timestamp":"2020-12-09T14:48:48.794Z","@version":1,"app":"ingress","caller":"main.main","levelname":"info","message":"Everything has shut down, goodbye","source_host":"ed6c5babed33"}%
```

akofink:
@SteveHNH could you please shed some light here on what changed and which env variable we can use now instead of `INGRESS_PORT`?

docker-compose.yml
vkrizan: This should be correct. The order is `hostPort:containerPort`.
```
hostPort = 8080
containerPort = 3000
```

Or, are you saying that this conflicts within a pod?

vkrizan:
@akofink I'm not 100% sure what's the correct way, but I guess they should not be separated by `_` (underscore) from the camel cased names. So, the corrected ones would be `INGRESS_WEBPORT` and `INGRESS_METRICSPORT`.
This can shed some more light: https://github.com/spf13/viper#working-with-environment-variables

akofink:
That works:

```
podman run -it -e INGRESS_WEBPORT=12345 quay.io/cloudservices/insights-ingress 
open : no such file or directory
{"@timestamp":"2020-12-09T15:19:58.697Z","@version":1,"Web Port":12345,"app":"ingress","caller":"main.main","levelname":"info","message":"Starting Service","source_host":"973b6aa77869"}{"@timestamp":"2020-12-09T15:19:58.697Z","@version":1,"Metrics Port":8080,"app":"ingress","caller":"main.main","levelname":"info","message":"Starting Service","source_host":"973b6aa77869"}
```

docker-compose.yml
akofink: No, sorry, you're right. This is silly that's it's opposite as the related `-p` flag when running a one-off container. https://docs.docker.com/compose/compose-file/#ports

akofink (approved):
Thanks, @vkrizan!

PR #681, opened 2020-12-11
feat(parsing): RHICOMPL-722 Stop parsing unknown benchmarks

akofink:
This changes the xccdf report parser to not save unknown benchmarks (those that are either not yet imported or have an SSG version that matches no supported SSG version). If a benchmark does not already exist, the report will fail to parse with an appropriate message logged. The only way to import benchmarks into Compliance after this change will be to run the SSG import job.

### DIFF
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -28,6 +28,10 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
       )
     end
 
+    def versions
+      all.map(&:version).uniq
+    end
+
     def all
       raw_supported['supported'].map do |rhel, values|
         major, minor = os_version(rhel)
--- a/app/services/concerns/xccdf/hosts.rb
+++ b/app/services/concerns/xccdf/hosts.rb
@@ -35,12 +35,12 @@ module Xccdf
     private
 
     def test_result_profile
-      @test_result_profile ||= ::Profile.canonical.find_or_initialize_by(
+      @test_result_profile ||= ::Profile.canonical.create_with(
+        name: @test_result_file.test_result.profile_id
+      ).find_or_initialize_by(
         ref_id: @test_result_file.test_result.profile_id,
         benchmark: benchmark
-      ).tap do |profile|
-        profile.name ||= @test_result_file.test_result.profile_id
-      end
+      )
     end
 
     def inventory_host
--- a/app/services/concerns/xccdf/hosts.rb
+++ b/app/services/concerns/xccdf/hosts.rb
@@ -35,14 +35,12 @@ module Xccdf
     private
 
     def test_result_profile
-      ::Profile.canonical
-               .where(ref_id: @test_result_file.test_result.profile_id,
-                      benchmark: benchmark).first ||
-        ::Profile.find_or_initialize_by(
-          ref_id: @test_result_file.test_result.profile_id,
-          name: @test_result_file.test_result.profile_id,
-          benchmark_id: benchmark.id
-        )
+      @test_result_profile ||= ::Profile.canonical.find_or_initialize_by(
+        ref_id: @test_result_file.test_result.profile_id,
+        benchmark: benchmark
+      ).tap do |profile|
+        profile.name ||= @test_result_file.test_result.profile_id
+      end
     end
 
     def inventory_host
--- a/app/services/concerns/xccdf/rule_results.rb
+++ b/app/services/concerns/xccdf/rule_results.rb
@@ -33,11 +33,15 @@ module Xccdf
       end
     end
 
+    def test_result_rules_unknown
+      selected_op_rule_results.map(&:id) - rule_ids.keys
+    end
+
     private
 
     def rule_ids
       @rule_ids ||= Rule.where(
-        benchmark: @benchmark, ref_id: selected_op_rule_results.map(&:id)
+        benchmark: benchmark, ref_id: selected_op_rule_results.map(&:id)
       ).pluck(:ref_id, :id).to_h
     end
   end
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -64,6 +64,10 @@ class XccdfReportParser
     # rubocop:enable Style/GuardClause
   end
 
+  def save_missing_supported_benchmark
+    save_all_benchmark_info if SupportedSsg.versions.include?(benchmark.version)
+  end
+
   def check_for_missing_benchmark
     # rubocop:disable Style/GuardClause
     unless benchmark.persisted?
@@ -108,6 +112,7 @@ class XccdfReportParser
       save_host
       check_os_version
       check_for_external_reports unless Settings.features.parse_external_reports
+      save_missing_supported_benchmark
       check_for_missing_benchmark_info
       save_all_test_result_info
     end
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -8,6 +8,12 @@ class WrongFormatError < StandardError; end
 class OSVersionMismatch < StandardError; end
 # Error to raise if the incoming report belongs to no known policy
 class ExternalReportError < StandardError; end
+# Error to raise if the incoming report belongs to no known benchmark
+class UnknownBenchmarkError < StandardError; end
+# Error to raise if the incoming report belongs to no known profile
+class UnknownProfileError < StandardError; end
+# Error to raise if the incoming report contains an unknown rule
+class UnknownRuleError < StandardError; end
 
 # Takes in a path to an Xccdf file, returns all kinds of properties about it
 # and saves it in our database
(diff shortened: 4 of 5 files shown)


app/services/xccdf_report_parser.rb
```diff
@@ -51,21 +58,65 @@ def check_for_external_reports
...
-            "for account #{@account.account_number}."
+            "No policy found matching benchmark #{benchmark.ref_id}. "\
+            "#{parse_failure_message}"
     end
     # rubocop:enable Style/GuardClause
   end
 
+  def check_for_missing_benchmark
+    # rubocop:disable Style/GuardClause
+    unless benchmark.persisted?
```
vkrizan, line 69: If this is the only condition, then it is too narrow. It has to bypass it for cases when the benchmark isn't imported yet but it is still on the supported matrix. This bypass has to exist, until all necessary downstream SSGs are imported.

If there would be another upstream vs. downstream version mismatch (backport) then it would not allow to import these benchmarks (unless I've missed something).

app/services/xccdf_report_parser.rb
```diff
@@ -51,21 +58,65 @@ def check_for_external_reports
...
+      raise UnknownBenchmarkError,
+            "No benchmark found matching ref_id #{benchmark.ref_id} and "\
+            "SSG #{benchmark.version}. #{parse_failure_message}"
+    end
+    # rubocop:enable Style/GuardClause
+  end
+
+  def check_for_missing_test_result_profile
+    # rubocop:disable Style/GuardClause
+    unless test_result_profile.persisted?
```
vkrizan, line 79: Related to `test_result_profile`:
* There seems to be a bug within the second lookup
```
::Profile.find_or_initialize_by(
  ref_id: @test_result_file.test_result.profile_id,
  name: @test_result_file.test_result.profile_id,
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  benchmark_id: benchmark.id
)
```
* there could be a memoization
* the method is private (which probably explains the absent memoization)

app/services/concerns/xccdf/rule_results.rb
```diff
@@ -33,11 +33,15 @@ def selected_op_rule_results
       end
     end
 
+    def test_result_rules_missing
+      selected_op_rule_results.map(&:id) - rule_ids.keys
```
vkrizan, line 37: This rather seems to check extra rules present in the test result rather than missing rules in a test result.

app/services/xccdf_report_parser.rb
```diff
@@ -51,21 +58,65 @@ def check_for_external_reports
...
+            "No profile found matching ref_id #{test_result_profile.ref_id} "\
+            "in benchmark #{benchmark.ref_id} with SSG "\
+            "#{benchmark.version}. #{parse_failure_message}"
+    end
+    # rubocop:enable Style/GuardClause
+  end
+
+  def check_for_missing_rules
+    # rubocop:disable Style/GuardClause
+    if test_result_rules_missing.any?
```
vkrizan, line 90: Can you please elaborate on this check?

If this checks whether the test result contains more rules that there are in the stored benchmark, then this should not fail but rather set is as unsupported or skip the check at all. It this should somehow compare rule sets of between test result and stored benchmark it would likely need a different approach.

vkrizan (commented):
Thank you for a quick PR.

Most of the checks seem too strict regarding the cases that it has to accept. Can you please verify my comments?

app/services/xccdf_report_parser.rb
akofink: Hmm, that's a good point. Perhaps we need to import all downstream versions, in addition to any upstream version required. So we'd import RHEL 7 0.1.49 and 0.1.50. The whole point with this change is to only import benchmarks during the ssg import task and never during report parsing.

app/services/xccdf_report_parser.rb
akofink: - Wow! How long has that been there? lol It's definitely incorrect and probably should not include `name` at all. I'll submit a separate PR with that change to make sure nothing breaks.
- Any memoization here is OK and beneficial
- The XccdfReportParser included the Xccdf::Hosts module, so it has access to all private methods included in it. Why use private at all, then? Well, it's mostly just trying to communicate intention at that point, similar to Python's "private" methods (`_foo`). I guess if we're going to use it outside the module, we should make the method public.

app/services/concerns/xccdf/rule_results.rb
akofink: Well, think about the failure conditions: if there are extra rules present in the test result, we can't parse it because our benchmark doesn't know those rules. If there are missing rules in the test result, that's OK because some rules may not be applicable or selected. We can still ingest the report in that case, and we shouldn't expect errors in later parsing steps (i.e. RuleResult saving against a non-existent Rule record).

app/services/xccdf_report_parser.rb
akofink: See https://github.com/RedHatInsights/compliance-backend/pull/681#discussion_r542400864

app/services/xccdf_report_parser.rb
akofink: @vkrizan ok, so I did some more investigation of `test_result_profile`, and this code doesn't cause any unintended side effects. This test_result_profile is only ever used during cloning to create a profile in the user's account. As you can see, the method attempts to find the profile by benchmark and profile ref_id, and then creates a new object if not found. We can essentially replace `find_or_initialize_by` with `new`. It's kind of weird, for sure, but it isn't causing any issues.

app/services/concerns/xccdf/rule_results.rb
vkrizan: Can it filter the subset of rules from those that it already knows, and report it as unsupported? It would create a new scenario.

Would you mind renaming it to something like `test_result_extra_rules`?

app/services/xccdf_report_parser.rb
akofink: Sorry, I don't understand. We import all the versions listed in the support matrix during the import_ssg job, not just the latest supported versions for each major OS. Which versions are missing assuming we also import upstream versions?

app/services/concerns/xccdf/rule_results.rb
akofink: The name of the method matches the error message displayed:

```
UnknownRuleError: The following rules are missing from profile xccdf_org.ssgproject.content_profile_standard in benchmark xccdf_org.ssgproject.content_benchmark_RHEL-8 with SSG
 0.1.40:                                                                                                 
xccdf_org.ssgproject.content_rule_sshd_set_keepalive
...
xccdf_org.ssgproject.content_rule_grub2_nousb_argument.rule
Report for profile xccdf_org.ssgproject.content_profile_standard against lenovolobato.lobatolan.home of account 0 could not be parsed.
```

I will rename it to `test_result_rules_unknown` to match the error class rather than the message.

app/services/xccdf_report_parser.rb
akofink: Ok, I see what you mean now. There are some versions that don't exist at all upstream. I'll add some logic to keep importing those versions if they don't already exist.

akofink: Updated

app/services/xccdf_report_parser.rb
akofink: Updated in a separate commit.

app/services/concerns/xccdf/hosts.rb
```diff
@@ -35,14 +35,12 @@ def external_report?
...
-        ::Profile.find_or_initialize_by(
-          ref_id: @test_result_file.test_result.profile_id,
-          name: @test_result_file.test_result.profile_id,
-          benchmark_id: benchmark.id
-        )
+      @test_result_profile ||= ::Profile.canonical.find_or_initialize_by(
+        ref_id: @test_result_file.test_result.profile_id,
+        benchmark: benchmark
+      ).tap do |profile|
+        profile.name ||= @test_result_file.test_result.profile_id
```
vkrizan, line 42: Nitpick: Could have used `create_with(name: @test_result_file.test_result.profile_id)` (before `find_or_initialize_by`) to avoid setting name on found profile.

app/services/xccdf_report_parser.rb
```diff
@@ -67,9 +67,13 @@ def check_for_external_reports
   def check_for_missing_benchmark
     # rubocop:disable Style/GuardClause
     unless benchmark.persisted?
-      raise UnknownBenchmarkError,
-            "No benchmark found matching ref_id #{benchmark.ref_id} and "\
-            "SSG #{benchmark.version}. #{parse_failure_message}"
+      if SupportedSsg.versions.include?(benchmark.version)
+        save_all_benchmark_info
```
vkrizan, line 71: Observation: Unexpected side-effect on a check method. I'd suggest to move this (with the condition) before doing the group of checks.

app/models/supported_ssg.rb
```diff
@@ -28,6 +28,10 @@ def ssg_for_os(os_major_version, os_minor_version)
       )
     end
 
+    def versions
+      all.map(&:version)
```
vkrizan, line 32: Nitpick: `.uniq` ?

vkrizan (approved):
Thank you Andrew. Looks good.

Do we have means to report some of the added errors to us (on Slack) when they occur? We'd be able to track any issues that customers can stumble, as there is no other messaging.

app/services/xccdf_report_parser.rb
akofink: Updated

akofink:
The only means of tracking errors are in the sidekiq logs. It's a long standing issue that has yet to be addressed.

PR #684, opened 2020-12-16
fix(profiles migration): RHICOMPL-1045 policy_id foreign key

akofink:
Add a foreign key to profiles.policy_id. This means that a policy cannot
be removed from the database while some profile references it.

This also adds a validation for the same case so we can detect and
report a nice error for it, if it ever happens.

### DIFF
--- a/db/migrate/20201216190051_add_policy_id_foreign_key_on_profiles.rb
+++ b/db/migrate/20201216190051_add_policy_id_foreign_key_on_profiles.rb
@@ -1,0 +1,5 @@
+class AddPolicyIdForeignKeyOnProfiles < ActiveRecord::Migration[5.2]
+  def change
+    add_foreign_key :profiles, :policies
+  end
+end
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -68,7 +68,16 @@ class ProfileTest < ActiveSupport::TestCase
 
     p = profiles(:one).dup
     assert_not p.update(policy_id: policies(:two).id, benchmark: bm)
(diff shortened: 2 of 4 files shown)


test/models/profile_test.rb
```diff
@@ -68,7 +68,16 @@ class ProfileTest < ActiveSupport::TestCase
 
     p = profiles(:one).dup
     assert_not p.update(policy_id: policies(:two).id, benchmark: bm)
-    assert  p.errors.full_messages.join['Policy type must be unique']
+    assert p.errors.full_messages.join['Policy type must be unique']
```
akofink, line 71: I'm not sure why rubocop never complained about this before, but there was an extra space here.

db/migrate/20201216190051_add_policy_id_foreign_key_on_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class AddPolicyIdForeignKeyOnProfiles < ActiveRecord::Migration[5.2]
+  def change
+    add_foreign_key :profiles, :policies
```
vkrizan, line 3: We would likely need to test this on stage with real data. There might be dangling records that could break the migration.

vkrizan (approved):
Thanks.

db/migrate/20201216190051_add_policy_id_foreign_key_on_profiles.rb
akofink: There shouldn't be, but yes, there's no guarantee. Getting prod snapshots in stage is such a hassle.. but I guess we'll do it again.

PR #686, opened 2020-12-17
fix(docker-compose): Remove conflicting ports

akofink:
This essentially reverts #679 but uses the new env variables. I did not
properly test #679 at the time, but it breaks podman-compose, which runs all
containers in the same pod. That means ports cannot conflict internally
or externally because they are mapped from the pod, not each container.


vkrizan (approved):
Sorry about that. I had concerns about the port conflicts within one networking space.

Thank you for fixing this.

PR #687, opened 2020-12-17
fix(coverage): RHICOMPL-847 Correct coverage

akofink:
Coverage has been misreported for quite some time without us knowing
why. The codecov documentation doesn't say anything about this 'rails'
option we were passing in. It was doing some weird things with some
files, like xccdf/hosts.rb not being covered at all, despite having
tests that cover it. Now it seems to be working as expected and shows
valid coverage deficiencies.

The new coverage number without the false negatives is 98.72% reported
from a local test run. Hopefully the codecov service will report a
similar result.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -2,22 +2,19 @@
 
 unless Rails.env.production?
   require 'simplecov'
-  SimpleCov.start 'rails' do
-    # don't test the schema!
-    # https://github.com/rmosolgo/graphql-ruby/blob/master/guides/schema/
-    # testing.md#dont-test-the-schema
-    add_filter 'db/schema.rb'
-    # these just contain the description of the attributes of the
-    # API responses, no code (https://github.com/Netflix/fast_jsonapi)
-    add_filter 'app/serializers'
-    # Skip testing empty module definition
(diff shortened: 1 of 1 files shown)


test/test_helper.rb
```diff
@@ -2,22 +2,19 @@
...
-    # these just contain the description of the attributes of the
-    # API responses, no code (https://github.com/Netflix/fast_jsonapi)
-    add_filter 'app/serializers'
-    # Skip testing empty module definition
-    add_filter 'app/models/xccdf.rb'
-    add_filter 'app/services/xccdf_report.rb'
-    # empty class that has to be there for prometheus integration
-    add_filter 'lib/'
+  SimpleCov.start do
+    add_filter 'config'
```
vkrizan, line 6: It could be beneficial to see coverage on initializers.

vkrizan (approved):
Thanks for investigating this!

test/test_helper.rb
akofink: Unfortunately, it doesn't count initializers even without this filter. All I see in `config` in the coverage report is `routes.rb`. The initializers must already run before the test_helper is required.

PR #689, opened 2020-12-21
feat(Profile): RHICOMPL-1034 ssg_version attribute

akofink:
Add ssg_version attribute to the profiles REST API and enable searching
by ssg_version in both graphql and REST.

### DIFF
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -25,7 +25,12 @@ module ProfileSearching
     scoped_search on: :os_major_version, ext_method: 'os_major_version_search',
                   only_explicit: true, operators: ['=', '!='],
                   validator: ScopedSearch::Validators::INTEGER
+    scoped_search on: :ssg_version, ext_method: 'ssg_version_search',
+                  only_explicit: true, operators: ['=', '!=']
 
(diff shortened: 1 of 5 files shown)


app/models/concerns/profile_searching.rb
```diff
@@ -107,5 +113,17 @@ def os_major_version_search(_filter, operator, value)
       benchmarks = Xccdf::Benchmark.os_major_version(value, operator == '=')
       { conditions: benchmark_id.in(benchmarks.pluck(:id)).to_sql }
     end
+
+    def ssg_version_search(_filter, operator, value)
+      benchmark_id = Profile.arel_table[:benchmark_id]
+      benchmarks = operator == '=' &&
+                   Xccdf::Benchmark.where(version: value) ||
```
vkrizan, line 120: I'm not fully sure, I suspect that within conditions it will query for data to evaluate the condition. If that's true, it would query all fields (as by default), in addition to creating a sub query.

vkrizan (approved):
Works as expected. Thanks!

app/models/concerns/profile_searching.rb
akofink: In this case, the condition below uses `to_sql`, and this is just a setup for `benchmarks.pluck(:id)` below.

PR #690, opened 2020-12-22
refactor(Profile): RHICOMPL-999 remove BO and threshold columns

akofink:
Remove business_objective and compliance_threshold columns from the
profiles table.


vkrizan (approved):
Looks and works correctly. Thank you!

PR #693, opened 2021-01-05
fix(parsing): RHICOMPL-933 Identify policies with changing ref_ids

akofink:
Submitting as a draft for transparency, but the work is still in a rough state at the moment and does not function. :)

Between certain versions of RHEL and SSG, profile ref_ids change. This
prevents identifying a policy by the existing profile ref_ids. Instead,
we must find equivalent ref_ids to search by on the policy. Note that
new ref_ids may be renamed to allow a new profile to take the old
ref_id, so it's important we lookup profiles by ref_id, SSG version, and
benchmark (i.e. OS major version).


akofink:
Abandoning this work for now to design a better solution.

PR #694, opened 2021-01-06
[stable] Sync with master for 2021-01-06

vkrizan:
RHICOMPL-999
RHICOMPL-1034
RHICOMPL-1045
RHICOMPL-1057


akofink (approved):
Thanks, @vkrizan!

PR #695, opened 2021-01-06
fix(report parse): RHICOMPL-1295 Find Policy by test_result_profile ref_id

akofink:
This fixes an oversight where policies are found only by account and
host association while never checking ref_id. This changes it so
policies are matched on exact ref_id of the uploaded test result
profile.

### DIFF
--- a/test/services/xccdf_report_parser_test.rb
+++ b/test/services/xccdf_report_parser_test.rb
@@ -372,13 +372,32 @@ class XccdfReportParserTest < ActiveSupport::TestCase
     end
   end
 
-  context 'missing policy' do
-    should 'raise an error and halt parsing (external report)' do
+  context 'finding an associated policy' do
+    should 'raise an error with no policy found (external report)' do
       @report_parser.stubs(:external_report?).returns(true)
       @report_parser.save_host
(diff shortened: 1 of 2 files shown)


test/services/xccdf_report_parser_test.rb
```diff
@@ -372,13 +372,32 @@ class TestParser < ::XccdfReportParser
...
+    should 'find a policy by hosts, account, and ref_id' do
+      @report_parser.stubs(:external_report?)
+      @report_parser.save_host
+      profiles(:one).update!(
+        ref_id: 'xccdf_org.ssgproject.content_profile_standard', hosts: []
+      )
+      policies(:one).hosts = [@report_parser.host]
+      policies(:two).hosts = [@report_parser.host]
+      Profile.any_instance.expects(:clone_to).with(policy: nil,
+                                                   account: accounts(:test))
```
akofink, line 393: This expectation fails without the fix.

vkrizan (approved):
LGTM. Thank you!

vkrizan:
There are issues with data created between the existence of this bug. Check Jira for further details.

PR #697, opened 2021-01-07
fix(data): RHICOMPL-1301 remove incorrect profiles from policies

akofink:
A bug introduced in RHICOMPL-1295 caused some profiles to be incorrectly
associated to policies. The initial profile on a policy essentially
determines the policy's policy type. If any profile ref_id does not
match this initial profile's ref_id, it does not belong on the policy
and will be removed along with any test results.

### DIFF
--- a/app/services/incorrect_profile_remover.rb
+++ b/app/services/incorrect_profile_remover.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# A class to remove profiles on policies that do not have a matching ref_id
+class IncorrectProfileRemover
+  def self.run!
+    counts = Policy.includes(:profiles).find_each.flat_map do |policy|
+      next unless policy.profiles.length > 1
+
+      find_incorrect_profiles(policy).destroy_all.count
+    end.compact
+
+    Logger.new(STDOUT).info "#{counts.sum} profiles removed from policies with"\
+                            ' mismatched ref_ids'
+  end
+
+  def self.find_incorrect_profiles(policy)
+    Profile.where(policy_object: policy)
+           .where.not(ref_id: policy.initial_profile.ref_id)
+  end
+end
--- a/db/migrate/20210107151954_remove_incorrect_profiles_from_policies.rb
+++ b/db/migrate/20210107151954_remove_incorrect_profiles_from_policies.rb
@@ -1,0 +1,5 @@
+class RemoveIncorrectProfilesFromPolicies < ActiveRecord::Migration[5.2]
+  def change
+    IncorrectProfileRemover.run!
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_12_22_134736) do
+ActiveRecord::Schema.define(version: 2021_01_07_151954) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
--- a/test/services/incorrect_profile_remover_test.rb
+++ b/test/services/incorrect_profile_remover_test.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class IncorrectProfileRemoverTest < ActiveSupport::TestCase
+  setup do
+    profiles(:one).update!(policy_object: policies(:one),
+                           account: accounts(:test))
+    profiles(:two).update!(policy_object: policies(:two),
+                           account: accounts(:test))
+  end
+
+  test 'removes profiles with a mismatched ref_id' do
+    profiles(:one).dup.update!(ref_id: 'foo', external: true)
+    assert_equal 2, policies(:one).profiles.count
+    assert_difference('Profile.count' => -1) do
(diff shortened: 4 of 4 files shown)


app/services/incorrect_profile_remover.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# A class to remove profiles on policies that do not have a matching ref_id
+class IncorrectProfileRemover
+  def self.run!
+    Policy.includes(:profiles).find_each do |policy|
```
vkrizan, line 6: Observation: The `.includes(:profiles)` is redundant, as those prefetched values/models are not used.

Could the prefetched values be used to do heuristics to improve run on a bigger data set? For example adding a condition `policy.profiles.count > 1` to prevent `policy.initial_profile.ref_id` doing a query for a single-profile policy.

db/migrate/20210107151954_remove_incorrect_profiles_from_policies.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveIncorrectProfilesFromPolicies < ActiveRecord::Migration[5.2]
+  def change
+    IncorrectProfileRemover.run!
```
vkrizan, line 3: Can you please log how many of rouge profiles have been removed?

app/services/incorrect_profile_remover.rb
akofink: Removed the `includes`. It seems like a good idea to limit the run to policies with more than one profile, but the query is not so straightforward; because of that, I think it kind of falls under the category of premature optimization. There are currently less than 1500 policies today.

app/services/incorrect_profile_remover.rb
```diff
@@ -0,0 +1,15 @@
...
+# A class to remove profiles on policies that do not have a matching ref_id
+class IncorrectProfileRemover
+  def self.run!
+    counts = Policy.find_each.flat_map do |policy|
+      Profile.where(policy_object: policy)
+             .where.not(ref_id: policy.initial_profile.ref_id)
+             .destroy_all.count
+    end
+
+    Rails.logger.info "#{counts.sum} profiles removed from policies with "\
```
vkrizan, line 12: For some reason this did not logged anything on `db:migrate`. Mind to change it to return the sum and then use logging in the migration (using `say` or `say_with_time`)?

vkrizan (approved):

app/services/incorrect_profile_remover.rb
vkrizan: My preference is to keep the migration within database and short as possible. That is using minimal queries as necessary (i.e. subqueries).

app/services/incorrect_profile_remover.rb
akofink: This logs to `log/development.log`, but I can make it log to STDOUT. I'm not sure what the behavior would be in production, since all logs are set to STDOUT there.

app/services/incorrect_profile_remover.rb
akofink: The thing you suggested is not a SQL query, though: `policy.profiles.count > 1`. I took it as illustration only, but I don't know how to limit the policy list in a single query by that condition. Do you have a recommendation?

akofink: Sorry, I see what you mean now. I'll keep the `includes` and add the `policy.profiles.count > 1` check. That _shouldn't_ add any other queries. :+1: thanks for the recommendation!

app/services/incorrect_profile_remover.rb
vkrizan: Sorry that I'm nitpicking, why not log outside of the service or force the logging to stdout only in wrapped block:
```
# in migration
do_log_to_stdout_wrapper_if_it_exists_in_rails do
   IncorrectProfileRemover.run!
end
```
This way the logging inside of the service is preserved as configured if used outside of migration.

app/services/incorrect_profile_remover.rb
vkrizan: I've meant it as a general rule for writing/executing migrations. My first idea was to use Ruby to limit number of queries made. There are ways to have it all in db, but that would likely break from the ORM and it is not necessary to do. :wink:

vkrizan:
Thanks @akofink for the changes. Seems to work and look good. It logs, even on rollback (:smile:). We'd likely need to test it on prod-like data before deployment to be sure.

akofink:
NB: there is no real rollback. It's an idempotent (one-way) data migration, so the `down` does the same thing as `up`.

akofink:
Even testing this on prod-like data will not be sufficient to detect any issues with it. We should try and get an RDS snapshot right before deploying so we can roll back if anything goes amiss.

app/services/incorrect_profile_remover.rb
akofink: If we want to change logging for all our database migrations, we can apparently set `ActiveRecord::Base.logger = Logger.new(STDOUT)`, but again, I believe migrations (and everything else) log to stdout in production already. These data migration service classes should always log to STDOUT - there's no reason why we'd want it to log anywhere else. IMO we really don't need some overly-architected logging solution here; this will only run once, after all.

PR #698, opened 2021-01-07
feat(external policies): RHICOMPL-1151 remove them all

akofink:
Remove all external policies from the perspective of the UI. This means
removing all profiles that are not associated to a policy that have the
external attribute set to true. Note: canonical profiles are marked as
external=false and are not affected by this migration.

### DIFF
--- a/db/migrate/20210107161848_remove_external_policies.rb
+++ b/db/migrate/20210107161848_remove_external_policies.rb
@@ -1,0 +1,5 @@
+class RemoveExternalPolicies < ActiveRecord::Migration[5.2]
+  def change
+    ExternalPolicyRemover.run!
+  end
+end
--- a/test/services/external_policy_remover_test.rb
+++ b/test/services/external_policy_remover_test.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+# A class to test removal of external policies
+class ExternalPolicyRemoverTest < ActiveSupport::TestCase
+  test 'does nothing if no external policies exist' do
+    assert_empty Profile.external.where(policy_id: nil)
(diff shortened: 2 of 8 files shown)


test/services/external_policy_remover_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+  end
+
+  test 'removes all external policies' do
+    profiles(:one).dup.update!(external: true, account: accounts(:test))
+    profiles(:two).dup.update!(external: true, policy_object: policies(:one),
+                               account: accounts(:test))
+    assert_equal 1, Profile.external.where(policy_id: nil).count
+    assert_difference('Profile.count' => -1) do
+      ExternalPolicyRemover.run!
```
vkrizan, line 20: Suggestion, to add test results and rules to those profiles and check whether they are deleted. It could uncover a regression, as there is a luck of foreign keys.

vkrizan (approved):
Thanks. Looks great.

db/migrate/20210107161848_remove_external_policies.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveExternalPolicies < ActiveRecord::Migration[5.2]
+  def change
+    ExternalPolicyRemover.run!
```
vkrizan, line 3: Also a suggestion (from rouge profile removal), to log how many of external profiles it has deleted and if possible for how many accounts. This would be good statistic to report.

test/services/external_policy_remover_test.rb
akofink: IMO this case should be covered in the profile model tests; I'll ensure they exist there.

PR #700, opened 2021-01-12
chore(frameworks): RHICOMPL-1312 drop use of ActionCable

vkrizan:
ActionCable is unused.

Note, that the dependency sill exists and would likely be cleaned out in near future.


akofink (approved):
Thanks!

PR #702, opened 2021-01-13
chore(Rails6) RHICOMPL-1317: Upgrade to Rails 6.0

akofink:
See the commits for details. The changes were fairly straightforward and seem to not have broken any tests. I've only done minor manual testing; it needs further verification.


sthirugn:
wow this is exciting!

akofink:
The tests run great locally. It seems like there's some issue with either the Jenkinsfile or one of our migrations; it may take some time for me to return to this issue to investigate due to competing priorities.

akofink:
Ok, so a slight misunderstanding with Cyndi which affects this work: we are adding a new schema, not a whole new database; therefore, this work is not necessary to connect our models to the inventory schema. We should revisit it at some point, however, since this PR demonstrates 99% of a working solution. The last thing to fix here is to adjust the Jenkinsfile so tests pass in CI.

akofink:
Closing this for now. We'll keep the branch around for reference when we do finally upgrade.

PR #703, opened 2021-01-14
feat(SupportedSsg): RHICOMPL-1320 more than one for OS minor

vkrizan:
Allow more than one SSG version/package for OS minor release.  The
pseudo-id has been changed to include package.

### DIFF
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -17,15 +17,18 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
 
   class << self
     def supported?(ssg_version:, os_major_version:, os_minor_version:)
-      ssg_version == ssg_for_os(os_major_version, os_minor_version)
+      ssg_versions_for_os(os_major_version, os_minor_version)
+        .include?(ssg_version)
     end
 
-    def ssg_for_os(os_major_version, os_minor_version)
-      raw_supported.dig(
-        'supported',
-        "#{self::OS_NAME}-#{os_major_version}.#{os_minor_version}",
-        'version'
-      )
+    def ssg_versions_for_os(os_major_version, os_minor_version)
+      os_major_version = os_major_version.to_s
+      os_minor_version = os_minor_version.to_s
+
+      all.select do |ssg|
+        ssg.os_major_version == os_major_version &&
+          ssg.os_minor_version == os_minor_version
+      end.map(&:version)
     end
 
     def versions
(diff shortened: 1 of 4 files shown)


app/models/supported_ssg.rb
```diff
@@ -58,14 +59,15 @@ def available_upstream
...
       end
     end
 
     private
 
-    def map_attributes(values)
-      attrs = values.slice(*members.map(&:to_s))
+    def map_attributes(rhel, raw_attrs)
+      attrs = raw_attrs.slice(*members.map(&:to_s))
+      attrs[:id] = "#{rhel}:#{raw_attrs['package']}"
```
akofink, line 70: :+1: I think this makes sense

app/models/supported_ssg.rb
```diff
@@ -17,31 +17,32 @@ def ref_id
...
 
-    def ssg_for_os(os_major_version, os_minor_version)
+    def ssg_versions_for_os(os_major_version, os_minor_version)
       raw_supported.dig(
         'supported',
-        "#{self::OS_NAME}-#{os_major_version}.#{os_minor_version}",
-        'version'
-      )
+        "#{self::OS_NAME}-#{os_major_version}.#{os_minor_version}"
+      )&.map { |raw_ssg| raw_ssg['version'] }
```
akofink, line 28: Nit: We could remove the need for safe operators here and anywhere else we use this in the future.

```suggestion
      )&.map { |raw_ssg| raw_ssg['version'] } || []
```

akofink (commented):
Looking good, @vkrizan. Seems straightforward and low impact. :+1:

app/models/supported_ssg.rb
vkrizan: I thought the same. I wanted to convert it to use non-raw attributes, but then got hold off because of the conversion from int to str.

vkrizan: Changed to use parsed objects. It'd allow to add more search if we'd add a `supported` flag.

app/models/supported_ssg.rb
```diff
@@ -17,31 +17,35 @@ def ref_id
...
-        'version'
-      )
+    def ssg_versions_for_os(os_major_version, os_minor_version)
+      os_major_version = os_major_version.to_s
+      os_minor_version = os_minor_version.to_s
+
+      all.select do |ssg|
+        ssg.os_major_version == os_major_version &&
+          ssg.os_minor_version == os_minor_version
+      end.map(&:version)
```
akofink, line 31: I really, really wish we could keep from searching through the list like this. I'd much rather have some unique key we can use for lookups, but I can't think of a better solution at the moment :/

akofink (approved):
Thanks, @vkrizan! Seems to work as expected

app/models/supported_ssg.rb
vkrizan: The list isn't that big to search on. And this way it fits into other usage with parsed data. Hashing also takes some cycles, but I get your point.

PR #704, opened 2021-01-15
revert: RHICOMPL-1301 remove incorrect profiles from policies

akofink:
We will need to create a new migration when we fix this. This migration
is now just a no-op but must stay in place to keep deployed schemas
versions consistent.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_01_07_161848) do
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_01_07_161848) do
+ActiveRecord::Schema.define(version: 2021_01_12_221033) do
```
akofink, line 13: This schema version was not correct on `master` due to an incorrect rebase.

sthirugn (approved):

PR #706, opened 2021-01-18
feat(rake): RHICOMPL-1330 debug SQL queries

vkrizan:
Creates rake task `db:debug` that would print out all subsequent SQL
queries made by other rake tasks in sequence.


bastilian (approved):
LGTM! Thanks @vkrizan!

PR #707, opened 2021-01-18
chore(migrations): RHICOMPL-1301 activate remove incorrect profiles from policies

vkrizan:
This partially reverts commit 220bea2f7acdac477d7f94ce2ed0fb88dd235467.

This would be used for testing on stage and later on updated
accordingly.


bastilian (approved):
LGTM! Thank you @vkrizan!

PR #710, opened 2021-01-19
fix: RHICOMPL-1335 high volume removal of Policy/Profile

vkrizan:
Intermediate tables and child tables of high volume should be deleted
right away without invoking destroy callbacks.  An examples is a removal
of a Policy that has a few Profiles but has a high volume of hosts
attached.  Note, that the hosts are attached via intermediate tables
(PolicyHost and ProfileHost) and via TestResults.

When there is a high removal of Policies/Profiles the destroy callbacks
are called for each intermediate/connecting record slowing down the
delete operation.  The destroy callbacks should only be run from the
other hand, from hosts side in this case.

Hence the change from `dependent: destroy` to `dependent: delete_all`.
It might be possible to keep the destroy if there is a possibility to
know the direction in the callback.

Drops destroy_orphaned_external_profiles hooks, as external profiles are
(being) removed.

### DIFF
--- a/app/models/concerns/profile_hosts.rb
+++ b/app/models/concerns/profile_hosts.rb
@@ -9,7 +9,7 @@ module ProfileHosts
     has_many :test_result_hosts, -> { distinct },
              through: :test_results, source: :host
     has_many :rule_results, through: :test_results
-    has_many :profile_hosts, dependent: :destroy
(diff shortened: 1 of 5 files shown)


akofink (approved):
Seems ok to me since the related models have no callbacks of their own.

PR #714, opened 2021-01-20
fix(Account): RHICOMPL-1342 Add auth_type to fake_identity_header

akofink:
This is a new field within the identity that the Inventory service is
now checking for and fails if it does not exist.


akofink:
It seems this change may not be necessary after all.

vkrizan (approved):
Thanks for taking care of this. Looks good.

PR #719, opened 2021-01-21
feat(cert-auth): RHICOMPL-1080 Validate cert-auth

akofink:
For cert-auth, the identity.system.cn must match the system's owner_id
in the system_profile, otherwise the request is unauthorized. In this
implementation, we simply pass cert-auth identity headers through to
inventory, and the response is filtered by identity.system.cn matching
owner_id in the system_profile. If nothing is returned, the identity
header is invalid for that host.

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -68,7 +68,7 @@ module Authentication
 
   def valid_cert_auth?
     valid_cert_endpoint? && HostInventoryAPI.new(
-      nil, Settings.host_inventory_url, raw_identity_header
+      b64_identity: raw_identity_header
     ).hosts.dig('results').present?
   rescue Faraday::Error => e
     Rails.logger.error(e.full_message)
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -52,22 +52,36 @@ module Authentication
 
   def rbac_allowed?
     return true if ActiveModel::Type::Boolean.new.cast(Settings.disable_rbac)
-    return true if skip_rbac_for_cert_based_auth?
+    return valid_cert_auth? if identity_header.cert_based?
 
-    @rbac_api ||= ::RbacApi.new(request.headers['X-RH-IDENTITY'])
+    @rbac_api ||= ::RbacApi.new(raw_identity_header)
     @rbac_api.check_user
   end
 
   private
 
-  def skip_rbac_for_cert_based_auth?
-    identity_header.cert_based? &&
-      ALLOWED_CERT_BASED_RBAC_ACTIONS.include?(controller: controller_name,
-                                               action: action_name)
+  def valid_cert_endpoint?
+    ALLOWED_CERT_BASED_RBAC_ACTIONS.include?(
+      controller: controller_name, action: action_name
+    )
+  end
+
+  def valid_cert_auth?
+    valid_cert_endpoint? && HostInventoryAPI.new(
+      nil, Settings.host_inventory_url, raw_identity_header
+    ).hosts.dig('results').present?
+  rescue Faraday::Error => e
+    Rails.logger.error(e.full_message)
+
+    false
+  end
+
+  def raw_identity_header
+    request.headers['X-RH-IDENTITY']
   end
 
   def identity_header
-    @identity_header ||= IdentityHeader.new(request.headers['X-RH-IDENTITY'])
+    @identity_header ||= IdentityHeader.new(raw_identity_header)
   end
 
   def account_from_header
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -67,7 +67,7 @@ module Authentication
   end
 
   def valid_cert_auth?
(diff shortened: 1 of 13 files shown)


app/services/host_inventory_api.rb
```diff
@@ -50,6 +50,12 @@ def system_profile(ids)
     end
   end
 
+  def hosts
+    JSON.parse(Platform.connection.get(
+      @url, {}, X_RH_IDENTITY: @b64_identity
```
vkrizan, line 55: Nit: Can this use named attributes, at it is not immediately obvious on what is `{}` for?

app/controllers/concerns/authentication.rb
```diff
@@ -52,22 +52,32 @@ def user
...
+  end
+
+  def valid_cert_auth?
+    valid_cert_endpoint? && HostInventoryAPI.new(
+      nil, Settings.host_inventory_url, raw_identity_header
+    ).hosts.dig('results').present?
+  end
+
+  def raw_identity_header
+    request.headers['X-RH-IDENTITY']
```
vkrizan, line 76: Observation: the same code is used for @rbac_api:
```
@rbac_api ||= ::RbacApi.new(request.headers['X-RH-IDENTITY'])
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Mind to refactor?

app/controllers/concerns/authentication.rb
```diff
@@ -52,22 +52,32 @@ def user
...
+  def valid_cert_endpoint?
+    ALLOWED_CERT_BASED_RBAC_ACTIONS.include?(
+      controller: controller_name, action: action_name
+    )
+  end
+
+  def valid_cert_auth?
+    valid_cert_endpoint? && HostInventoryAPI.new(
+      nil, Settings.host_inventory_url, raw_identity_header
+    ).hosts.dig('results').present?
```
vkrizan, line 72: If I understand this correctly, we are using Inventory API for checking the auth. Does that need to use an API that'd likely return big chunk of data to just check whether it returns something? Is there a shorter option, e.g. using lighter endpoint or `HEAD`?

How are non-200 HTTP states handled? Wouldn't it crash on 401 form Inventory with 500 on our side?

vkrizan (commented):
Make sense. Can you please check behavior on non-200 HTTP?

app/controllers/concerns/authentication.rb
```diff
@@ -52,22 +52,32 @@ def user
...
-                                               action: action_name)
+  def valid_cert_endpoint?
+    ALLOWED_CERT_BASED_RBAC_ACTIONS.include?(
+      controller: controller_name, action: action_name
+    )
+  end
+
+  def valid_cert_auth?
+    valid_cert_endpoint? && HostInventoryAPI.new(
+      nil, Settings.host_inventory_url, raw_identity_header
```
vkrizan, line 71: Observation: The `Settings.host_inventory_url` is repeated for every use of `HostInventoryAPI`. Can it be set as default on inside the class? Note, that the class already hardcodes path, which isn't configurable.

akofink:
I will work on the refactorings. How do you feel about rescuing `Faraday::ClientError` and `Faraday::ServerError` in the application_controller?

app/controllers/concerns/authentication.rb
akofink: It doesn't appear HEAD returns any useful information about the host list, unfortunately.

vkrizan: Can you please elaborate why it has to request all hosts from Inventory when it does not need any of those for our auth?

vkrizan (approved):
Looks good. The `Faraday::Error` should cover all cases, if we want to hide incorrect responses.

app/controllers/concerns/authentication.rb
akofink: Note that this isn't actually requesting all hosts. Instead, when the cert-auth identity header of 'system' type is sent to inventory, only the hosts that have a matching `owner_id` to the `system.cn` field of the identity are returned. I agree, it's not ideal, but Inventory hasn't provided a better way of validating the identity.

vkrizan: Oh, nice! :+1:  Sorry, I've haven't understood it the first time from the description.

akofink:
Let's wait to merge this until we collaborate with the Inventory team on when they are planning to release their changes. Thanks!

app/services/host_inventory_api.rb
```diff
@@ -5,23 +5,22 @@
...
 
   ERRORS = [InventoryHostNotFound].freeze
 
-  def initialize(account, url, b64_identity)
+  def initialize(account: nil, url: Settings.host_inventory_url,
+                 b64_identity: account&.b64_identity)
     @url = "#{URI.parse(url)}#{Settings.path_prefix}/inventory/v1/hosts"
     @account = account
-    @b64_identity = b64_identity || @account.b64_identity
+    @b64_identity = b64_identity
```
akofink, line 18: This introduced a regression. See https://github.com/RedHatInsights/compliance-backend/pull/727

PR #720, opened 2021-01-22
chore(devel): Update DB schema to match migrations

akofink:
This only affects places where we use `db:schema:load` (currently nowhere afaik).


vkrizan (approved):
Thanks. Merging right way...

PR #722, opened 2021-01-25
feat(Cyndi): RHICOMPL-1374 create cyndi schema in compose

akofink:
This PR just creates the inventory schema for cyndi and creates a view that comes directly from the local inventory database (using [dblink](https://www.postgresql.org/docs/9.3/contrib-dblink-function.html)). We won't need to run any kind of app-connector here to keep it in sync, since it's just a view of the inventory database. Thanks to beav for the help figuring this out :) This shouldn't break any existing devel workflows. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,12 +98,34 @@ First, let's install all dependencies and initialize the database.
 ```shell
 bundle install
 bundle exec rake db:create db:migrate
+bundle exec rake db:test:prepare # for test DB setup
 ```
 
 Once you have database initialized, you might want to import SSG policies:
 
 ```shell
 bundle exec rake ssg:import_rhel_supported
+```
+
+#### Project Cyndi
+
+The compliance project integrates with project cyndi. For local development, a database view is created, built from the inventory database which runs alongside the compliance database. The Cyndi hosts view exists within an inventory schema within the compliance database.
+
+```shell
+bundle exec rails db < db/cyndi_setup.sql # syndicated (cyndi) hosts from inventory
+RAILS_ENV=test bundle exec rails db < db/cyndi_setup.sql # cyndi for test DB
+```
+
+You can verify everything worked as expected within psql for compliance_dev and
+compliance_test databases:
+
+```
+compliance_dev=# \dv inventory.
+          List of relations
+  Schema   | Name  | Type |  Owner
+-----------+-------+------+----------
+ inventory | hosts | view | insights
+(1 row)
 ```
 
 #### Kafka Consumers (XCCDF report consumers)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -141,7 +141,12 @@ services:
   build-rails-db:
     image: compliance-backend-rails
     entrypoint: ''
-    command: 'bundle exec rake db:create db:migrate'
+    command: sh -c '
+      bundle exec rake db:create db:migrate;
+      bundle exec rake db:test:prepare;
+      bundle exec rails db < db/cyndi_setup.sql;
+      RAILS_ENV=test bundle exec rails db < db/cyndi_setup.sql;
+      '
     volumes:
       - .:/app:z
     depends_on:
--- a/db/cyndi_setup.sql
+++ b/db/cyndi_setup.sql
@@ -1,0 +1,32 @@
+--
+-- Intializes a Cyndi-like inventory schema and creates a view copied from a
+-- local inventory DB
+--
+
+CREATE SCHEMA IF NOT EXISTS inventory;
+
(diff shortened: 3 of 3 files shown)


cyndi_setup.sql
```diff
@@ -0,0 +1,32 @@
...
+    id,
+    account,
+    display_name,
+    created_on as created,
+    modified_on as updated,
+    stale_timestamp,
+    stale_timestamp + INTERVAL '1' DAY * 7 AS stale_warning_timestamp,
+    stale_timestamp + INTERVAL '1' DAY * 14 AS culled_timestamp,
+    tags,
+    system_profile_facts as system_profile
```
akofink, line 22: This must match the view that cyndi defines. We unfortunately don't have any say in the column names (`account` is an unfortunate choice, as you'll see in https://github.com/RedHatInsights/compliance-backend/pull/723)

cyndi_setup.sql
```diff
@@ -0,0 +1,32 @@
+--
+-- Intializes a Cyndi-like inventory schema and creates a view copied from a
+-- local inventory DB
+--
+
+CREATE SCHEMA IF NOT EXISTS inventory;
+
+CREATE EXTENSION IF NOT EXISTS dblink;
```
akofink, line 8: This is required, just for development, to define a view across databases.

cyndi_setup.sql
```diff
@@ -0,0 +1,32 @@
+--
+-- Intializes a Cyndi-like inventory schema and creates a view copied from a
+-- local inventory DB
```
vkrizan, line 3: Would it be possible to move this sql file to `db/`?

docker-compose.yml
```diff
@@ -141,7 +141,12 @@ services:
   build-rails-db:
     image: compliance-backend-rails
     entrypoint: ''
-    command: 'bundle exec rake db:create db:migrate'
+    command: sh -c '
+      bundle exec rake db:create db:migrate;
+      bundle exec rake db:test:prepare;
+      bundle exec rails db < cyndi_setup.sql;
```
vkrizan, line 147: Is this setup targeted *only* for development? If not, can we move it to a migration?

docker-compose.yml
```diff
@@ -141,7 +141,12 @@ services:
...
+      RAILS_ENV=test bundle exec rails db < cyndi_setup.sql;
```
vkrizan, line 148: Can you please document these steps in the README? Thank you.

vkrizan (approved):
Make sense. Thanks!

cyndi_setup.sql
```diff
@@ -0,0 +1,32 @@
...
+
+CREATE SCHEMA IF NOT EXISTS inventory;
+
+CREATE EXTENSION IF NOT EXISTS dblink;
+
+-- This view should be queried instead
+CREATE OR REPLACE VIEW inventory.hosts AS
+SELECT
+    id,
+    account,
```
vkrizan, line 14: Xref https://github.com/RedHatInsights/compliance-backend/pull/723#discussion_r564421641
Can it be aliased? `account AS account_number`

docker-compose.yml
akofink: Yes, this is only for development. In other environments, we create the schema but don't create the view itself.

cyndi_setup.sql
akofink: @vkrizan we cannot alias account as account_number, as I've described because we don't have control over this view in other environments.

vkrizan: Fair enough. Can we gain control of the view?

akofink: All apps use the same app-connector which writes to the underlying DB and in some cases sets up the view itself. Only in devel do we get to create the view ourselves, unfortunately. Not having control of the view is a feature of Cyndi to ensure consistency across apps.

akofink:
@vkrizan please check that the readme updates are acceptable.

README.md
```diff
@@ -106,6 +107,27 @@ Once you have database initialized, you might want to import SSG policies:
 bundle exec rake ssg:import_rhel_supported
 ```
 
+### Project Cyndi
```
vkrizan, line 110: Can you please add one more heading level, as after this is the continuation of an Option. Or, move it a bit down but copy the init part above.

Text looks good (with internal hyperlinking). Thanks. :-)

akofink: Yeah :/ I can remove the link if you think we should. I'll fix the header, too.

akofink: Link removed, header level increased one. Thanks for the review!

PR #723, opened 2021-01-25
feat(cyndi): RHICOMPL-1285 hosts come from cyndi

akofink:
- Update our testing to split test and devel environments wrt the cyndi schema. In devel, it uses the local inventory. In test, it uses the hosts_v1_1 table within the inventory schema
- Make the host model read only. Any code that modifies or creates hosts is altered or removed.
- Switch the host model from our own internal hosts table to the cyndi hosts view, and remove our existing hosts table along with any host-related foreign keys (i.e. policy_host), since FK relationships are not allowed across schemas.
- Add new host fixtures within the hosts_v1_1 cyndi hosts table for use with tests
- Change the host.account relation to use `account_number` due to a conflicting `account` column in the cyndi hosts view
- Add cleanup_db rake task to ensure host relationships are (eventually) cleaned up when a host is removed from the cyndi view
- Add an engineering_collector prometheus collector to be able to count how many dangling host relationships exist

NB: HostInventoryApi cannot be completely removed, since we must still use it to validate cert-auth identities; however, it is not used anywhere else after this change.

I've tried to split this PR into separate commits, but the nature of the changes make it difficult. Please let me know if you see any room for improvement.


app/models/account.rb
```diff
@@ -3,7 +3,10 @@
 # Represents a Insights account. An account can be composed of many users
 class Account < ApplicationRecord
   has_many :users, dependent: :nullify
-  has_many :hosts, dependent: :nullify
+  # rubocop:disable Rails/HasManyOrHasOneDependent
+  has_many :hosts, foreign_key: :account, primary_key: :account_number,
+                   inverse_of: :account_object
+  # rubocop:enable Rails/HasManyOrHasOneDependent
```
akofink, line 9: `dependent: nil` does not disable rubocop here :(

app/models/concerns/system_like.rb
```diff
@@ -9,7 +9,9 @@ module SystemLike
     scoped_search on: %i[id name]
     has_many :rule_results, dependent: :destroy
     has_many :rules, through: :rule_results, source: :rule
-    belongs_to :account, optional: true
+    belongs_to :account_object, optional: true, foreign_key: :account,
+                                primary_key: :account_number,
+                                inverse_of: :hosts, class_name: 'Account'
```
akofink, line 14: The column name in the cyndi view is `account`, so we can't have a method/relationship named the same, unfortunately.

app/models/host.rb
```diff
@@ -17,8 +20,13 @@ class Host < ApplicationRecord
   has_many :policies, through: :policy_hosts
   has_many :assigned_profiles, through: :policies, source: :profiles
 
-  validates :name, presence: true
-  validates :account, presence: true
+  def readonly?
+    true
+  end
```
akofink, line 25: This is good for detecting issues with our tests, but also it should allow active_record to warn of deletion/update/insert events rather than postgres (can't modify records from a view).

app/models/host.rb
```diff
@@ -17,8 +20,13 @@ class Host < ApplicationRecord
...
+
+  def name
+    display_name
+  end
```
akofink, line 29: Translation from what we called `name` to the view, which uses `display_name`. This change alone may actually solve the display_name issues we've had with the client.

db/migrate/20210125214324_drop_hosts.rb
```diff
@@ -0,0 +1,23 @@
...
+      t.datetime "created_at", null: false
+      t.datetime "updated_at", null: false
+      t.uuid "account_id"
+      t.integer "os_major_version"
+      t.integer "os_minor_version"
+      t.index ["account_id"], name: "index_hosts_on_account_id"
+      t.index ["name"], name: "index_hosts_on_name"
+      t.index ["os_major_version"], name: "index_hosts_on_os_major_version"
+      t.index ["os_minor_version"], name: "index_hosts_on_os_minor_version"
+    end
```
akofink, line 21: Unfortunately, because the `up` removes the foreign key, we can't just add it back here because there is likely existing policy_host objects which will raise an error when the hosts table is empty.

db/schema.rb
```diff
@@ -10,9 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_01_07_161848) do
+ActiveRecord::Schema.define(version: 2021_01_25_214324) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "dblink"
```
akofink, line 16: :thinking: this _shouldn't_ be here except in development, but db/schema.rb is really only used for development and testing, so I think it's ok.

app/models/concerns/profile_searching.rb
```diff
@@ -8,10 +8,11 @@ module ProfileSearching
...
                          external parent_profile_id]
     scoped_search relation: :assigned_hosts, on: :id, rename: :system_ids
-    scoped_search relation: :assigned_hosts, on: :name, rename: :system_names
+    scoped_search relation: :assigned_hosts, on: :display_name,
+                  rename: :system_names
     scoped_search relation: :test_result_hosts, on: :id,
                   rename: :test_result_system_ids
-    scoped_search relation: :test_result_hosts, on: :name,
+    scoped_search relation: :test_result_hosts, on: :display_name,
                   rename: :test_result_system_names
```
akofink, line 16: This should fix [RHICOMPL-1111](https://issues.redhat.com/browse/RHICOMPL-1111).

app/models/host.rb
```diff
@@ -17,18 +20,29 @@ class Host < ApplicationRecord
...
-  validates :name, presence: true
-  validates :account, presence: true
+  def readonly?
+    true
+  end
+
+  def destroy; end
+
+  def delete; end
+
```
akofink, line 30: `readonly?` does nothing for destroy/delete

app/models/concerns/profile_searching.rb
sthirugn: this is awesome!

app/policies/host_policy.rb
```diff
@@ -21,7 +21,9 @@ def destroy?
   # Only show hosts in our user account
   class Scope < ::ApplicationPolicy::Scope
     def resolve
-      only_matching_account
+      return scope.where('1=0') if user.account_number.blank?
```
vkrizan, line 24: Nit: `scope.none` should work, replacing `where('1=0')`.

app/consumers/inventory_events_consumer.rb
```diff
@@ -12,13 +12,6 @@ def process(message)
 
     handle_report_parsing
 
-    case @msg_value['type']
-    when 'delete'
-      DeleteHost.perform_async(@msg_value)
-    when 'updated'
-      InventoryHostUpdatedJob.perform_async(@msg_value)
-    end
```
vkrizan, line 20: :scissors: Nice!

db/migrate/20210125214324_drop_hosts.rb
vkrizan: Wish we had removed policy_hosts :wink:

db/schema.rb
vkrizan: The assumption of `schema.rb` not being used in production could likely change, esp. when we decide to clean up old migration files. Keeping the extension should not do any harm, I guess.

vkrizan (commented):
Great job! Looks great. I'll give a spin in my local dev.

I wish it would be possible to workaround the naming collisions.

app/models/concerns/profile_searching.rb
sthirugn: will this search with `name` and `display_name`?

akofink: `name` was a field we were saving in our own database. afaik, there is no `name` field in inventory, only `display_name`.

app/models/concerns/system_like.rb
akofink: We cannot change the DB view provided to us by Cyndi in any way, even column aliases. One option would be to create a `hosts` view within our own schema that points to the cyndi hosts view and renames fields. A view of a view... seems not so great to me, but it's a possibility. IMO it's not a huge problem within our code to account for this (pun intended :) )

akofink: In fact, it makes way more sense to use `account_number` instead of `id`, and ideally, we will be able to completely remove the accounts table if, for every related record, we store the `account_number` directly.

app/consumers/inventory_events_consumer.rb
```diff
@@ -12,13 +12,6 @@ def process(message)
 
     handle_report_parsing
 
-    case @msg_value['type']
-    when 'delete'
-      DeleteHost.perform_async(@msg_value)
```
vkrizan, line 17: Even-though it no longer would have our custom Host records, it would still have host assignments. We should find a way how to ensure that host assignments are removed/cleaned after a host is removed.

app/models/business_objective.rb
```diff
@@ -6,6 +6,8 @@ class BusinessObjective < ApplicationRecord
   has_many :profiles, through: :policies
   has_many :accounts, through: :policies
 
+  delegate :account_number, to: :account
```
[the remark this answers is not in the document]
akofink: There's really no reason that we should store account and reference related objects by its UUID `id`. Eventually, I'd like to remove the accounts table and store `account_number` directly on our models that require it. No, not all our models are directly related to an account; it's only hosts, profiles, and policies.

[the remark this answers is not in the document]
akofink: One more point I'd like to raise here: if we remove the accounts table altogether, the cyndi hosts view with the `account` field won't be nearly as problematic, and we can remove any references to `account_object`.

app/policies/application_policy.rb
```diff
@@ -40,7 +40,7 @@ def destroy?
   end
 
   def match_account?
-    record.account_id == user.account_id
+    record.account_number == user.account_number
```
akofink, line 43: To respond to your premature optimization comment, @vkrizan, this app-wide pundit policy calls `account_number` on some unknown model type. It's duck typing! That means that any model that pundit authorizes must respond to `account_number`. Sometimes `record` is a Host, so we can't just use `account` because of the unfortunate column name in the view. Furthermore, `account_id` is a 'magic' rails attribute that is now not available on a host (it's now `account_object_id`). We could, of course, define that method on the host to keep this working, but both solutions seem similar in terms of complexity/impact to me. I'm making the change this way looking forward to when we remove the accounts table altogether.

app/policies/application_policy.rb
```diff
@@ -53,12 +53,6 @@ def initialize(user, scope)
       @scope = scope
     end
 
-    def only_matching_account
-      return scope.where('1=0') if user.account_id.blank?
-
-      scope.where(account_id: user.account_id)
-    end
```
akofink, line 60: This was only used in the host policy.

README.md
```diff
@@ -112,8 +112,8 @@ bundle exec rake ssg:import_rhel_supported
 The compliance project integrates with project cyndi. For local development, a database view is created, built from the inventory database which runs alongside the compliance database. The Cyndi hosts view exists within an inventory schema within the compliance database.
 
 ```shell
-bundle exec rails db < db/cyndi_setup.sql # syndicated (cyndi) hosts from inventory
-RAILS_ENV=test bundle exec rails db < db/cyndi_setup.sql # cyndi for test DB
+bundle exec rails db < db/cyndi_setup_devel.sql # syndicated (cyndi) hosts from inventory
+RAILS_ENV=test bundle exec rails db < db/cyndi_setup_test.sql # cyndi for test DB
```
akofink, line 116: @vkrizan check out this solution to testing. In devel, the view comes from the inventory table. In test environments (local or CI), it comes from a local table like the Cyndi onboarding specifies. This seems to work great. I still have some work to go through the tests and remove any places that try to add/update hosts, but I'm excited about this working solution. :)

db/cyndi_setup_test.sql
```diff
@@ -0,0 +1,40 @@
+--
```
akofink, line 1: Going to pull these changes into a separate commit, but here's the cyndi setup for testing.

db/cyndi_setup_test.sql
```diff
@@ -0,0 +1,40 @@
...
+    stale_timestamp + INTERVAL '1' DAY * 14 AS culled_timestamp,
+    tags,
+    system_profile
+FROM inventory.hosts_v1_1;
+
+--
+-- These are the host seeds for use with testing locally and in CI
+--
+INSERT INTO inventory.hosts_v1_1 VALUES ('5e65ac3f-8b44-4f60-9e99-abdafb31740c', '00001', 'MyStringone', '[{"namespace": "insights-client", "key": "env", "value": "prod"}]', '2020-07-21 05:35:53.682554+00', '2020-07-21 05:35:53.682554+00', '2030-07-21 05:35:53.682554+00', '{"sap_system": true}');
+INSERT INTO inventory.hosts_v1_1 VALUES ('328fb4c0-42fc-0139-06c0-6e70056f34f5', '00002', 'MyStringtwo', '[{"namespace": "insights-client", "key": "env", "value": "prod"}]', '2020-07-21 05:35:53.682554+00', '2020-07-21 05:35:53.682554+00', '2030-07-21 05:35:53.682554+00', '{"sap_system": true}');
```
akofink, line 40: These two hosts match the old hosts fixtures that I removed.

app/models/business_objective.rb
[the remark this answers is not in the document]
vkrizan: Requiring every table to have the association to an account might be able to shorten the distance, e.g. for checking whether the object belongs to an account. It could be a nightmare have it synced correctly and kept coherent with connected models. Seems like your targeting to have some kind of a single table inheritance for an Account. Full removal of an Account object might seem premature. In some form it could be kept, e.g. when there would be another source/view for Account data (like Hosts within Cyndi).

Technically only three models need an Account association: Hosts, Policies, and Business Objectives. Other models have associations to their parent. Therefore, it does not seems to make sense to have `account_number` (or other type of an account identifier) to be on all models.

app/policies/application_policy.rb
vkrizan: For top level models that are exposed via API the account number delegate make sense. For their children, not so much. See https://github.com/RedHatInsights/compliance-backend/pull/723#discussion_r565915627 

I'd be happy to discuss pros and cons of some of the approach at some time.

README.md
vkrizan: :+1: Thanks!

app/models/business_objective.rb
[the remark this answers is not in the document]
akofink: I'm not suggesting to have it on all models, just the three that have direct account associations, `hosts`, `policies`, and `profiles` (business_objectives have an account through policy). In any case, we can debate removal of the accounts table later. This PR does not remove it :)

app/services/host_inventory_api.rb
```diff
@@ -6,47 +6,13 @@
 # Interact with the Insights Host Inventory. Usually HTTP calls
 # are all that's needed.
 class HostInventoryApi
-  # Error to raise if the inventory host could not be found
-  class InventoryHostNotFound < StandardError; end
-
-  ERRORS = [InventoryHostNotFound].freeze
```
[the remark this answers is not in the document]
akofink: @vkrizan When this PR is working end-to-end, I'll mark it as ready to review. Until then, it's definitely still a WIP and likely broken. Feel free to review the code, but you may want to wait to test it until it's 'ready to review'.

Jenkinsfile
```diff
@@ -26,8 +26,10 @@ def runStages() {
         }
 
         gitUtils.stageWithContext("Prepare-db", shortenURL = false) {
-            migrateStatus = sh(script: "bundle exec rake db:migrate --trace", returnStatus: true)
-            sh "bundle exec rake db:test:prepare"
+            migrateStatus = sh(script: "bundle exec rake db:test:prepare --trace", returnStatus: true)
+            container("postgresql") {
+              sh("psql compliance < db/cyndi_setup_test.sql")
+            }
```
akofink, line 32: @vkrizan it really wasn't a big issue once we figured out how to tell Jenkins to use a different container :P Thanks to @Victoremepunto for helping figure this out.

db/migrate/20210125214324_drop_hosts.rb
akofink: Are you suggesting a change, @vkrizan? I'm not sure how we could remove policy_hosts.

app/consumers/inventory_events_consumer.rb
akofink: I've added the DeleteHost job back for now. I'm not sure of a better way to solve this other than running a periodic DB cleanup job that looks for dangling references. IMO we kind of need to add such a job regardless, since it's possible that we could miss kafka messages.

app/jobs/delete_host.rb
```diff
@@ -3,15 +3,12 @@
...
-      RuleResult.where(host_id: message['id']).delete_all
-      rescue_not_found do
-        Sidekiq.logger.info("Destroying host #{message['id']}")
-        Host.find_by!(id: message['id']).destroy
+      Sidekiq.logger.info("Deleting related records for host #{message['id']}")
+      [RuleResult, TestResult, ProfileHost, PolicyHost].each do |model|
+        model.where(host_id: message['id']).delete_all
       end
     end
   end
```
akofink, line 14: This PR no longer removes the DeleteHost job. It's repurposed to remove any host relationships. In addition, I've added a rake task that we can either schedule to run, say, once per week (CronJob) or have it run as part of the deployment (Job).

lib/prometheus/engineering_collector.rb
```diff
@@ -0,0 +1,61 @@
...
+
+unless defined? Rails
+  require File.expand_path('../../config/environment', __dir__)
+end
+
+# rubocop:disable Metrics/MethodLength
+# rubocop:disable Metrics/AbcSize
+# Collects stats relevent for the engineers from the database
+# when /metrics is called. This runs directly on prometheus_exporter.
+class EngineeringCollector < PrometheusExporter::Server::TypeCollector
```
akofink, line 11: :see_no_evil: I took the opportunity to add an engineering collector for prometheus. We should use this for any info that we'd like to collect about the health/integrity of our data set. These first metrics are all related to detecting records that reference some missing (i.e. deleted) host ID.

lib/tasks/cleanup_db.rake
```diff
@@ -0,0 +1,25 @@
...
+
+  puts 'Beginning cleanup_db.'
+
+  num_deleted = Account.where.not(
+    account_number: Host.select(:account)
+  ).where.not(
+    id: Profile.where.not(account_id: nil).select(:account_id)
+  ).where.not(
+    id: Policy.select(:account_id)
+  ).delete_all
```
akofink, line 15: This task is tested for all cases, both positive and negative, in its associated test. These conditions must all hold for the account to be removed (no hosts & no profiles & no policies).

app/models/business_objective.rb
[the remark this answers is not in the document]
akofink: Perhaps not directly related, but I just realized our accounts table does not currently validate that account_number is unique. This could potentially cause only subsets of an account's data to be returned if there are duplicate account objects in our DB sharing account_numbers.

Jenkinsfile
```diff
@@ -47,7 +49,7 @@ def runStages() {
         execSmokeTest (
             ocDeployerBuilderPath: "compliance/compliance-backend",
             ocDeployerComponentPath: "compliance",
-            ocDeployerServiceSets: "compliance,ingress,inventory,platform-mq,rbac",
+            ocDeployerServiceSets: "compliance,ingress,inventory,platform-mq,rbac,cyndi",
```
akofink, line 52: @Victoremepunto I've added the cyndi service set here for the smoke tests.

vkrizan (approved):
Great work! Works as expected.

akofink:
Thank you for the reviews, @vkrizan! I'm still in the process of setting up the cyndi views in all our deployed environments, so for now, I'll mark this PR as a draft to prevent merging.

akofink:
Rebased with the ProfileHosts removal work.

PR #726, opened 2021-01-28
fix(smoke): RHICOMPL-1379 grant compliance_user permissions

akofink:
This is required after running migrations because migrations run as the
db super user, but the application runs with compliance_user.


vkrizan (approved):
Looks reasonable. Thanks.

akofink:
This errors as is. https://github.com/RedHatInsights/s2i-insights-compliance/pull/25 should allow us to run `rails db` like this.

```
+ POSTGRESQL_USER=postgres

+ bundle exec rails db

Couldn't find database client: psql. Check your $PATH and try again.
```

akofink:
Closing, as it's not needed because we will use the postgres superuser in pre-prod environments.

PR #727, opened 2021-01-28
fix(HostInventoryApi): RHICOMPL-1383 handles nil b64_identity

akofink:
This is a regression introduced with RHICOMPL-1080. The HostInventoryApi
gets a nil b64_identity during report parse and must use the
fake_identity from the account.

See https://github.com/RedHatInsights/compliance-backend/pull/719/files#r566446351


sthirugn (approved):
lgtm

vkrizan (approved):
Make sense. Merging right away.

PR #724, opened 2021-01-27
[stable] Align stable to master

akofink:
RHICOMPL-1374
RHICOMPL-1080
RHICOMPL-1371

:no_smoking:

PR #728, opened 2021-01-29
[stable] Align stable with master

vkrizan:
RHICOMPL-1374
RHICOMPL-1080
RHICOMPL-1371
RHICOMPL-1383

Continuation after #724


Victoremepunto:
The only test that fails , as expected , is the `cert-based authentication` test - we're overriding the test and merging this now, as we cannot introduce the `owner_id` on Inventory's hosts system profile column as required, further testing will be conducted in QA.

vkrizan:
Fixed stable branch to match master with the commits. Stable branch was set to dd66f6a.

PR #731, opened 2021-02-04
chore: update db setup on dev to use db:setup

vkrizan:
When using the docker-compose, it would first check the db version and
run db:migrate or db:setup.  The check needs to be in place, because the
db:setup (precisely db:schema:load) would drop the data.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -142,7 +142,11 @@ services:
     image: compliance-backend-rails
     entrypoint: ''
     command: sh -c '
-      bundle exec rake db:create db:migrate;
+      if bundle exec rake db:version 2> /dev/null; then
+        bundle exec rake db:migrate;
+      else
+        bundle exec rake db:setup;
+      fi;
       bundle exec rake db:test:prepare;
       bundle exec rails db < db/cyndi_setup.sql;
       RAILS_ENV=test bundle exec rails db < db/cyndi_setup.sql;
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ First, let's install all dependencies and initialize the database.
 
 ```shell
 bundle install
-bundle exec rake db:create db:migrate
+bundle exec rake db:setup # invoke only on first setup!
 bundle exec rake db:test:prepare # for test DB setup
 ```
 


docker-compose.yml
```diff
@@ -142,7 +142,11 @@ services:
     image: compliance-backend-rails
     entrypoint: ''
     command: sh -c '
-      bundle exec rake db:create db:migrate;
+      bundle exec rake db:version 2> /dev/null;
+      if [[ $? -eq 0 ]]; then
+        bundle exec rake db:migrate;
+        else bundle exec rake db:setup;
+      fi;
```
akofink, line 149: This doesn't seem to work correctly. I had an existing DB and brought down and back up the compose services; my DB was cleared. The logs show some `sh` error:

```
$ podman logs -f compliancebackend_build-rails-db_1                                                                                           
2021-02-04T16:13:53.458Z 2 TID-gtfxt6d7y INFO: GitLab reliable fetch activated!                                                                                                                                    
Current version: 20210201233645                                                                                                                                                                                    
sh: 1: [[: not found                                                                                                                                                                                               
2021-02-04T16:13:55.051Z 19 TID-gqs57y1xr INFO: GitLab reliable fetch activated!                                                                                                                                   
Database 'compliance_dev' already exists                                                                                                                                                                           
Database 'compliance_test' already exists                                                                                                                                                                          
-- enable_extension("pgcrypto")                                                                                                                                                                                    
   -> 0.0260s                                                                                                                                                                                                      
-- enable_extension("plpgsql")                                                                                                                                                                                     
   -> 0.0016s                                                                                                                                                                                                      
-- create_table("accounts", {:id=>:uuid, :default=>#<Proc:0x000055ec3b183760@/app/db/schema.rb:19 (lambda)>, :force=>:cascade})                                                                                    
   -> 0.0446s                                                                                                                                                                                                      
-- create_table("benchmarks", {:id=>:uuid, :default=>#<Proc:0x000055ec3b1866b8@/app/db/schema.rb:26 (lambda)>, :force=>:cascade})                                                                                  
   -> 0.0415s                                                                                                                                                                                                      
-- create_table("business_objectives", {:id=>:uuid, :default=>#<Proc:0x000055ec3b18de68@/app/db/schema.rb:36 (lambda)>, :force=>:cascade})
...
```

vkrizan: Oh, yes. I thought we use bash (or that sh is a symlink to it). Will change.

akofink: This seems to work correctly:

```suggestion
      if bundle exec rake db:version; then
        bundle exec rake db:migrate;
        else bundle exec rake db:setup;
      fi;
```

vkrizan: Check/try now...

akofink (approved):
Thanks! Works as expected.

PR #734, opened 2021-02-08
feat(logging): RHICOMPL-1217 introduce audit logging 

vkrizan:
Introduction of a middleware and utility to add entries to an audit log.  By default it logs them to STDOUT, but eventually it can be tied into CloudWatch. Log entries are formatted into JSON, split by a new-line.

Originally posted at https://github.com/RedHatInsights/insights-api-common-rails/pull/217

TODO:
- [X] Capture Rails requests
- [x] Add tests
- [ ] Find a way to have a shortcut to the logger
- [x] Support within a Sidekiq

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/lib/audit_log/audit_log.rb
+++ b/lib/audit_log/audit_log.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+require_relative 'audit_log/formatter'
+require_relative 'audit_log/middleware'
+
+# Common space for Insights API stuff
+module Insights
+  module API
+    module Common
+      # Audit Logger into selected logger, but primarily to CloudWatch
+      class AuditLog
+        class << self
+          def logger
+            @logger ||= init_logger
+          end
+
+          def logger=(logger)
+            @logger = init_logger(logger)
+          end
+
+          def setup(logger = nil)
+            self.logger = logger
+            self
+          end
+
+          def with_account(account_number)
+            original = Thread.current[:audit_account_number]
+            Thread.current[:audit_account_number] = account_number
+            return unless block_given?
+
+            begin
+              yield
+            ensure
+              Thread.current[:audit_account_number] = original
+            end
+          end
+
+          private
+
+          def init_logger(logger = nil)
+            logger ||= Logger.new($stdout)
+            logger.level = Logger::INFO
+            logger.formatter = Formatter.new
+            logger
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/audit_log/audit_log/middleware.rb
+++ b/lib/audit_log/audit_log/middleware.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+
+# Common space for Insights API stuff
+module Insights
+  module API
+    module Common
+      class AuditLog
+        # Request events listener, capturing for evidence
+        class Middleware
+          attr_reader :logger, :evidence, :request, :status
+
+          def initialize(app)
+            @app = app
+            @logger = AuditLog.logger
+            @subscribers = []
+            @evidence = {}
+          end
+
+          def call(env)
+            subscribe
+            @request = ActionDispatch::Request.new(env)
+            @app.call(env).tap do |status, _headers, _body|
+              @status = status
+              response_finished
+            end
+          ensure
+            unsubscribe
+          end
+
+          private
+
+          def response_finished
+            payload = {
+              controller: evidence[:controller],
+              remote_ip: request.remote_ip,
+              message: generate_message
+            }
+            log(payload)
+          end
+
+          def generate_message
+            status_label = Rack::Utils::HTTP_STATUS_CODES[status]
+            parts = [
+              "#{request.method} #{request.original_fullpath}" \
+              " -> #{status} #{status_label}",
+              unpermitted_params_msg,
+              halted_cb_msg
+            ]
+            parts.compact.join('; ')
+          end
+
+          def unpermitted_params_msg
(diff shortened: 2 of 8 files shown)


akofink (commented):
Looking good! I've already reviewed this on https://github.com/RedHatInsights/insights-api-common-rails/pull/217. Thanks, @vkrizan!

lib/audit_log/audit_log/middleware.rb
```diff
@@ -0,0 +1,116 @@
+# frozen_string_literal: true
+
+# Common space for Insights API stuff
+module Insights
+  module API
+    module Common
+      class AuditLog
+        # Request events listener, capturing for evidence
+        class Middleware
```
akofink, line 9: This represents the top level object, the middleware itself. If I understand correctly, it's intended to use this like:

```rb
config.middleware.use Insights::API::Common::AuditLog::Middleware
```

If that's the case, we need to ensure all the `require`s start from here so things like `AuditLog.logger` are already defined properly.

test/lib/audit_log/middleware_test.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+require 'audit_log/audit_log'
```
akofink, line 3: I suppose this is the `require` intended to be used within `application.rb`?

akofink (commented):
Looking great, @vkrizan! Just a couple small comments about naming, paths, etc., but I was able to enable the middleware without much fuss locally. Here's what a log look like:

```
{"@timestamp":"2021-02-09T23:05:15.780582Z","hostname":"dfa3fe94f424","pid":1,"thread_id":"2b0cf4a3925c","level":"warning","transaction_id":"57ba8762f3b44f8b08a3","message":"GET /api/compliance/v1/systems/435c3df2-4e81-4986-8462-8a9ba930c512 -> 404 Not Found","controller":"V1::SystemsController#show","remote_ip":"127.0.0.1"}
```

Are you intending to enable the logger in this PR? It'd certainly help to review if it was enabled. I'd love if we could get this going to `log/audit.log` in `Rails.env == 'development'` for debugging purposes (instead of mixed into STDOUT and `log/development.log` with all the rest).

lib/audit_log/audit_log.rb
```diff
@@ -0,0 +1,50 @@
...
+
+          def with_account(account_number)
+            original = Thread.current[:audit_account_number]
+            Thread.current[:audit_account_number] = account_number
+            return unless block_given?
+
+            begin
+              yield
+            ensure
+              Thread.current[:audit_account_number] = original
```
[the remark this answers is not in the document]
vkrizan: We can always go back and change it. :wink:

lib/audit_log/audit_log.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require_relative 'audit_log/formatter'
+require_relative 'audit_log/middleware'
+
+# Common space for Insights API stuff
+module Insights
+  module API
```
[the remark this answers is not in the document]
vkrizan: Are you sure about that? There are Rails stuff named [`API` in `ActionController`](https://github.com/rails/rails/blob/v5.2.4.3/actionpack/lib/action_controller/api.rb#L89). If there are upper cased letters together then it concats them, not split them.

Test:
```
[1] pry(main)> 'API'.underscore
=> "api"
```
I can certainly move it to the appropriate path. I don't see other benefits other than the autoload, which can be neglected.

lib/audit_log/audit_log/middleware.rb
vkrizan: Yes, the usage is correct.

The logger is setup with `Insights::API::Common::AuditLog.setup(Logger.new)`. Require has no effect on it, only the first use of the logger. The default one can be overridden by the mentioned setup.

vkrizan:
It'd certainly help to review if it was enabled.

@akofink 

> Are you intending to enable the logger in this PR?

No. Please treat it as a library.

>  I'd love if we could get this going to log/audit.log in Rails.env == 'development' for debugging purposes (instead of mixed into STDOUT and log/development.log with all the rest).

Sure, you can use setup:
```
Insights::API::Common::AuditLog.setup(Logger.new('log/audit.log'))
```

test/lib/audit_log/middleware_test.rb
vkrizan: Yes.

lib/audit_log/audit_log.rb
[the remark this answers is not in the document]
akofink: Yes, it's only for Rails autoloaded constants. It's not necessary to match this format if we manually `require` the correct file, which is how this code will ultimately be used. Also, +1 to matching the insights-api-common-rails. Just for the sake of argument :trollface: 

```rb
[21] pry(main)> 'api'.camelize
=> "Api"
[22] pry(main)> 'a_p_i'.camelize
=> "API"
```

This was important for our rails classes like `HostInventoryApi` which we expect to be autoloaded without a `require`.

akofink (approved):
Thanks, @vkrizan! This is an exciting addition to the application.

PR #735, opened 2021-02-08
feat(Host): RHICOMPL-1397 add has_policy attribute

akofink:
Since we added a search term, it makes sense to also add an attribute so users can know what to expect when using the search.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/controllers/concerns/rendering.rb
+++ b/app/controllers/concerns/rendering.rb
@@ -6,9 +6,8 @@ module Rendering
 
   included do
     def render_json(model, **args)
-      opts = ['index'].include?(action_name) ? metadata : {}
-      opts.merge!(include: params[:include].split(',')) if params[:include]
-      render({ json: serializer.new(model, opts) }.merge(args))
+      model = model.includes(includes) if index? && includes
+      render({ json: serializer.new(model, serializer_opts) }.merge(args))
     end
 
     def render_error(messages, status: :not_acceptable, **opts)
@@ -25,6 +24,17 @@ module Rendering
 
     private
 
+    def serializer_opts
+      opts = index? ? metadata : {}
+      opts.merge!(include: params[:include].split(',')) if params[:include]
+
+      opts
+    end
+
+    def index?
+      ['index'].include?(action_name)
+    end
+
     def model_errors(models = [])
       models = [models].flatten
       models.flat_map do |model|
@@ -38,5 +48,7 @@ module Rendering
     def serializer
       raise NotImplementedError
     end
+
+    def includes; end
   end
 end
--- a/app/controllers/v1/systems_controller.rb
+++ b/app/controllers/v1/systems_controller.rb
@@ -25,5 +25,9 @@ module V1
     def resource
       Host
     end
+
+    def includes
+      %i[policy_hosts]
+    end
   end
 end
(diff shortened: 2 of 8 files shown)


app/models/host.rb
```diff
@@ -25,6 +25,11 @@ def readonly?
   alias destroy save
   alias delete save
 
+  def policy_hosts?
+    policy_hosts.any?
```
vkrizan, line 29: Observation: Preload policy hosts?

vkrizan (commented):
Code looks good so far. Will test in a bit...

Personally, I don't like adding single-purpose flags to API, especially when the information can be retrieved by other/general means. What is the difference between `has_policy` and checking profiles (relationship)?

akofink:
Unfortunately, the profiles relationship on a host is `all_profiles`, which includes test result profiles as well as policy profiles. We'd need to add another relationship (or two), like `test_result_profiles` and `policy_profiles`.

app/controllers/concerns/rendering.rb
```diff
@@ -6,9 +6,8 @@ module Rendering
 
   included do
     def render_json(model, **args)
-      opts = ['index'].include?(action_name) ? metadata : {}
-      opts.merge!(include: params[:include].split(',')) if params[:include]
-      render({ json: serializer.new(model, opts) }.merge(args))
+      model = model.includes(includes) if index? && includes
+      render({ json: serializer.new(model, serializer_opts) }.merge(args))
```
akofink, line 10: Rubocop forced a refactor of this method; it was getting pretty complex.

app/controllers/concerns/rendering.rb
```diff
@@ -38,5 +48,7 @@ def model_errors(models = [])
     def serializer
       raise NotImplementedError
     end
+
+    def includes; end
```
akofink, line 52: @vkrizan I've added a new feature to this rendering concern which allows controllers to specify any includes to perform prior to serialization.

app/controllers/v1/systems_controller.rb
```diff
@@ -25,5 +25,9 @@ def serializer
     def resource
       Host
     end
+
+    def includes
+      %i[policy_hosts]
+    end
```
akofink, line 31: And here is the systems_controller using the new includes rendering feature.

akofink:
Ok, @vkrizan, please let me know what you think. I've opted to add includes as a part of our rendering code so we can more easily correct N+1 issues in the future. I've only included `policy_hosts` to the hosts API for now, but we can revisit this and other controllers to easily add more `includes`.

app/controllers/concerns/rendering.rb
vkrizan: :+1:

vkrizan (approved):
Works as expected.

> Unfortunately, the profiles relationship on a host is all_profiles, which includes test result profiles as well as policy profiles. We'd need to add another relationship (or two), like test_result_profiles and policy_profiles.

There is nothing stopping us to add those relationships to REST API. We already have them in the GraphQL.

PR #740, opened 2021-02-09
fix(Host): RHICOMPL-1415 loose filter search

vkrizan:
Search/filter value without field specified is searching through all
available search fields.  There were some search search fields set that
no longer have a db field representative.  Those would need an external
method lookup.

/cc @Victoremepunto 

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/models/concerns/host_searching.rb
+++ b/app/models/concerns/host_searching.rb
@@ -5,7 +5,7 @@ module HostSearching
   extend ActiveSupport::Concern
 
   included do
-    scoped_search on: %i[display_name os_major_version os_minor_version]
+    scoped_search on: %i[id display_name]
     scoped_search on: :display_name, rename: :name
     scoped_search on: :compliant, ext_method: 'filter_by_compliance',
                   only_explicit: true
--- a/app/models/concerns/system_like.rb
+++ b/app/models/concerns/system_like.rb
@@ -6,7 +6,6 @@ module SystemLike
   extend ActiveSupport::Concern
 
   included do
-    scoped_search on: %i[id name]
     has_many :rule_results, dependent: :destroy
     has_many :rules, through: :rule_results, source: :rule
     belongs_to :account_object, optional: true, foreign_key: :account,
--- a/test/models/host_test.rb
+++ b/test/models/host_test.rb
@@ -17,6 +17,11 @@ class HostTest < ActiveSupport::TestCase
     end
   end
 
+  test 'loose filter search' do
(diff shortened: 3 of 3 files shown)


app/models/concerns/host_searching.rb
```diff
@@ -5,7 +5,7 @@ module HostSearching
   extend ActiveSupport::Concern
 
   included do
-    scoped_search on: %i[display_name os_major_version os_minor_version]
+    scoped_search on: %i[id display_name]
     scoped_search on: :display_name, rename: :name
     scoped_search on: :compliant, ext_method: 'filter_by_compliance',
                   only_explicit: true
```
akofink, line 11: `only_explicit` is here for fields which we don't want to search through without their keyword

app/models/concerns/host_searching.rb
```diff
@@ -5,7 +5,7 @@ module HostSearching
   extend ActiveSupport::Concern
 
   included do
-    scoped_search on: %i[display_name os_major_version os_minor_version]
+    scoped_search on: %i[id display_name]
```
akofink, line 8: We have frontend tasks that will require these search fields. Here's one way to query them with ActiveRecord:

```rb
Host.where('system_profile @> ?', {operating_system: {major: 7}}.to_json)
```

akofink (changes requested):
As you were saying yesterday, this alters the API unnecessarily. It's also required for an upcoming feature in the frontend to filter the policy wizard by OS. Please adjust the search to use what's in the host's system_profile.

app/models/concerns/host_searching.rb
vkrizan: Is your suggestion to move those os lookups to `only_explicit`?

vkrizan:
@akofink please check the follow up task RHICOMPL-1417.

akofink:
Working on this as https://github.com/RedHatInsights/compliance-backend/pull/741 (still needs tests)

app/models/concerns/host_searching.rb
akofink: Pretty much everything should be `only_explicit` IMO. In this case, we have to continue supporting `display_name`/`name` search implicitly.

PR #744, opened 2021-02-11
refactor: platform_basic_auth to be settable via env

vkrizan:
Changes platform_basic_auth config option to a Hash (from an array) to
be able to be settable via env variables.
Note, that this option is only being used in dev env.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/services/platform.rb
+++ b/app/services/platform.rb
@@ -4,7 +4,8 @@ require 'faraday'
 
 # Methods related to connecting to other platform services
 module Platform
-  BASIC_AUTH = Settings.platform_basic_auth
+  BASIC_AUTH = [Settings.platform_basic_auth_username,
+                Settings.platform_basic_auth_password].freeze
   RETRY_OPTIONS = {
     max: 3,
     interval: 0.05,
@@ -22,7 +23,7 @@ module Platform
       f.request :retry, RETRY_OPTIONS
       f.adapter Faraday.default_adapter # this must be the last middleware
       f.ssl[:verify] = Rails.env.production?
-      f.basic_auth(*BASIC_AUTH) if BASIC_AUTH
+      f.basic_auth(*BASIC_AUTH) if BASIC_AUTH[0].present?
     end
     faraday
   end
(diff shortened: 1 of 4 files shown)


config/settings/development.yml
```diff
@@ -2,9 +2,9 @@ host_inventory_url: http://inventory-web:8081
 remediations_url: https://ci.cloud.redhat.com
 rbac_url: https://ci.cloud.redhat.com
 # disable_rbac: true
-# platform_basic_auth:
-#   - username
-#   - password
+platform_basic_auth:
+  username:
+  password:
```
akofink, line 7: This should be commented out by default. Also, all the currently commented out settings are examples, so it'd be good to match that with something like:

```
# platform_basic_auth:
#   username: myuser
#   password: changeme
```

app/services/platform.rb
```diff
@@ -22,7 +22,9 @@ def self.connection
       f.request :retry, RETRY_OPTIONS
       f.adapter Faraday.default_adapter # this must be the last middleware
       f.ssl[:verify] = Rails.env.production?
-      f.basic_auth(*BASIC_AUTH) if BASIC_AUTH
+      if BASIC_AUTH && BASIC_AUTH[:username].present?
+        f.basic_auth(BASIC_AUTH[:username], BASIC_AUTH[:password])
+      end
```
akofink, line 27: Mind adding a test?

akofink (commented):
Makes sense, thanks. Just a couple small things

config/settings/development.yml
vkrizan: Sure I can comment it out. There is no difference. I believe that the above is self-explanatory.

app/services/platform.rb
vkrizan: There wasn't one before, but there might be one now.

vkrizan:
Commented. Test added. And decided to split them into two options (instead of one-depth hash).

akofink (approved):
Thanks!

PR #750, opened 2021-02-18
 fix(auth): RHICOMPL-1439 ensure current user reset 

vkrizan:
Changes `before_action` to `around_action` for `authenticate_user` to add the ensure block for resetting the current user (thread local).

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -12,7 +12,7 @@ module Authentication
   ].freeze
 
   included do
-    before_action :authenticate_user
+    around_action :authenticate_user
   end
 
   def authenticate_user
@@ -25,6 +25,9 @@ module Authentication
     return if performed?
 
     User.current = user
+    yield
+  ensure
+    User.current = nil
   end
 
   def current_user
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -16,14 +16,15 @@ module Authentication
   end
 
   def authenticate_user
+    if identity_header.blank?
+      return unauthenticated(error: 'Error parsing the X-RH-IDENTITY header')
+    end
     return unauthenticated unless identity_header.valid?
     return unauthorized unless rbac_allowed?
 
     return if performed?
 
     User.current = user
-  rescue JSON::ParserError, NoMethodError
-    unauthenticated error: 'Error parsing the X-RH-IDENTITY header'
   end
 
   def current_user
--- a/app/models/identity_header.rb
+++ b/app/models/identity_header.rb
@@ -10,14 +10,21 @@ class IdentityHeader
     @b64_identity = b64_identity
   end
 
+  def blank?
+    @b64_identity.blank?
+  end
+
   def valid?
     identity.present? && entitled?
   end
 
   def content
-    return @content if @content.present?
-
-    @content = JSON.parse(Base64.decode64(@b64_identity))
+    @content ||=
+      begin
+        JSON.parse(Base64.decode64(@b64_identity))
+      rescue JSON::ParserError
+        {}
+      end
   end
 
   def identity
--- a/test/controllers/concerns/authentication_test.rb
+++ b/test/controllers/concerns/authentication_test.rb
@@ -3,28 +3,47 @@
 require 'test_helper'
 
 # This class tests a "dummy" controller for authentication.
-class AuthenticationTest < ActionDispatch::IntegrationTest
-  setup do
-    User.current = nil
-    ApplicationController.any_instance.stubs(:index).returns('Response Body')
-    Rails.application.routes.draw do
-      root 'application#index'
+class AuthenticationTest < ActionController::TestCase
+  # rubocop:disable Rails/ApplicationController
+  class AuthenticatedMockController < ActionController::Base
+    include Authentication
+
+    def index
+      user = User.current
+      return false unless user
+
+      render plain: user.account_number
+    end
+
+    def raising
+      raise 'Error'
+    end
+  end
+  # rubocop:enable Rails/ApplicationController
+
+  MockedRoutes = ActionDispatch::Routing::RouteSet.new
+  MockedRoutes.draw do
+    ActiveSupport::Deprecation.silence do
+      get ':controller(/:action)'
     end
   end
 
-  teardown do
-    Rails.application.reload_routes!
+  setup do
+    @routes = MockedRoutes
+    @controller = AuthenticatedMockController.new
   end
 
   context 'unauthorized access' do
     should 'rh-identity header not found' do
-      get '/'
+      process_test
       assert_response :unauthorized
+      assert_not User.current, 'current user must be reset after request'
     end
 
     should 'error parsing rh-identity' do
-      get '/', headers: { 'X-RH-IDENTITY': 'this should be a hash' }
+      process_test(headers: { 'X-RH-IDENTITY': 'this should be a hash' })
       assert_response :unauthorized
+      assert_not User.current, 'current user must be reset after request'
     end
 
     should 'missing entitlement info completely' do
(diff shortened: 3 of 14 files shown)


app/models/identity_header.rb
```diff
@@ -10,8 +10,14 @@ def initialize(b64_identity)
...
 
+  def blank?
+    @b64_identity.blank?
+  end
+
   def valid?
     identity.present? && entitled?
+  rescue JSON::ParserError, NoMethodError
+    false
   end
```
akofink, line 21: This might not work as you expect. I try to completely avoid returns from a method-level rescue because it's so confusing. Please add a `begin` block.

app/models/identity_header.rb
```diff
@@ -10,8 +10,14 @@ def initialize(b64_identity)
     @b64_identity = b64_identity
   end
 
+  def blank?
+    @b64_identity.blank?
+  end
+
   def valid?
     identity.present? && entitled?
+  rescue JSON::ParserError, NoMethodError
```
akofink, line 19: IMO this should be rescued in the `content` method where parsing occurs so that any call to `content`, `identity`, etc. will be rescued.

app/controllers/concerns/authentication.rb
```diff
@@ -12,18 +12,22 @@ module Authentication
   ].freeze
 
   included do
-    before_action :authenticate_user
+    around_action :authenticate_user
```
akofink, line 15: I think it would simplify our code and make it easier to understand if we split this into separate `before_action` and `after_action`. i.e. `clear_user`

app/controllers/concerns/authentication.rb
```diff
@@ -12,18 +12,22 @@ module Authentication
...
     return unauthorized unless rbac_allowed?
 
     return if performed?
 
     User.current = user
-  rescue JSON::ParserError, NoMethodError
-    unauthenticated error: 'Error parsing the X-RH-IDENTITY header'
+    yield
+  ensure
+    User.current = nil
```
akofink, line 30: It might make sense to implement a `User.clear!`

test/controllers/concerns/authentication_test.rb
```diff
@@ -3,23 +3,48 @@
...
+        raise 'Error'
+      end
+    end
+
+    Object.const_set(:AuthenticatedMockController, AuthenticatedMockController)
+    # rubocop:disable Rails/HttpPositionalArguments
+    Rails.application.routes.draw do
+      root 'authenticated_mock#index'
+      get '/raising', to: 'authenticated_mock#raising'
+    end
```
akofink, line 28: It really shouldn't be necessary to mess with the routes. You can instantiate a controller directly (i.e. `c = AuthenticatedMockController.new`) and then call the actions directly (i.e. `c.index`), and the before/after action callbacks should work as expected.

test/controllers/concerns/authentication_test.rb
```diff
@@ -223,122 +284,15 @@ class AuthenticationTest < ActionDispatch::IntegrationTest
...
+                                 .stubs(:valid_cert_endpoint?)
+                                 .returns(true)
       HostInventoryApi.any_instance.expects(:hosts)
                       .raises(Faraday::Error.new(''))
       RbacApi.expects(:new).never
-      get profiles_url, headers: { 'X-RH-IDENTITY': encoded_header }
-      assert_response :forbidden
-    end
-
-    should 'allow access to profiles#index' do
```
akofink, line 233: Please do not remove these tests. They test cert auth

akofink (commented):
Looking good, thanks @vkrizan! Comments inline.

test/controllers/concerns/authentication_test.rb
vkrizan: They've been moved to profile_controller_test.

akofink: Ah! sorry, I didn't realize that. They're all related to cert auth and are not specific to any one controller, though. IMO they belong in `authentication_test` since that's the code they're meant to test.

app/controllers/concerns/authentication.rb
vkrizan: I've experimented with that. The difference is that `after_action` isn't triggered if the action isn't run (e.g. on auth failure or similar).

test/controllers/concerns/authentication_test.rb
vkrizan: I'm not sure about that, but can try to. I've seen some process methods in Rails. I've spent too much time on this mocking...

app/controllers/concerns/authentication.rb
akofink: Isn't that OK? If auth fails, the current user would not be set.

test/controllers/concerns/authentication_test.rb
akofink: This isn't quite what you're looking for, but I found an example in the rendering_test: https://github.com/RedHatInsights/compliance-backend/blob/master/test/controllers/concerns/rendering_test.rb#L8

app/controllers/concerns/authentication.rb
vkrizan: Here I want to be perfectly sure that it would be cleared out. There might be other callbacks in the registry that can stop the action execution.

app/models/identity_header.rb
vkrizan: Can you please elaborate on the unexpected behavior? The `begin` is implicit for the method block.

app/models/identity_header.rb
vkrizan: Yeah, I was thinking about it... `content` could be also covered. Some other issues (like nil content) can bubble up. Lets see if  I can find something better.

app/controllers/concerns/authentication.rb
vkrizan: The name could be `User.clear_current!`. This concern seems to be the only exclusive place for changing its value, so the assignment should be fine.

test/controllers/concerns/authentication_test.rb
vkrizan: These are specific for the profiles controller, to test the integration of authentication within that controller, hence the move.

app/models/identity_header.rb
vkrizan: Moved to `content` with tests added. `NoMethodError` is no longer needed.

test/controllers/concerns/authentication_test.rb
vkrizan: Still had to set up routes, but now via `@routes` and changed it to `ActionController::TestCase` with less magic. Unfortunately there isn't any headers passing, so those had to be added to a mocked request. Tests run much quicker with these changes.

If you still want to change it, feel free to suggest a working change. I've spent way too much time on tests compared to the actual backend change. :no_mouth:

app/controllers/concerns/authentication.rb
akofink: Yes, I agree that's more clear ;)

app/models/identity_header.rb
akofink: We've just been bitten by this in the past, even within the authentication logic. I can't see why this won't work - I think we ran into issues in the past with explicit `return` statements because the order of execution is a bit strange:

```rb
def foo(r=false)
  raise 'no!' if r
  puts 'one'
  1
rescue
  puts 'two'
  2
ensure
  puts 'three'
  3
end

[11] pry(main)> foo
one
three
=> 1

[12] pry(main)> foo(true)
two
three
=> 2
```

I think we're good here :+1: Sorry for the concern.

akofink (approved):
Thanks, @vkrizan!

test/controllers/concerns/authentication_test.rb
akofink: No worries :) I guess there isn't much benefit from continuing to iterate on these tests.

PR #753, opened 2021-02-23
fix(openapi): RHICOMPL-1454 set a basePath

akofink:
This fixes the API docs so that users can play with our API from within
the rswag-ui at /docs/api/compliance

This fixes one of the many issues raised in RHICOMPL-826. Unfortunately, there are still quite a lot of issues with our openapi.json still.

rswag-ui_fixed

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/spec/api/v1/openapi.rb
+++ b/spec/api/v1/openapi.rb
@@ -15,6 +15,7 @@ module Api
         {
           openapi: '3.0.3',
           info: info,
+          basePath: "#{Settings.path_prefix}/#{Settings.app_name}",
           paths: {},
           components: {
             schemas: SCHEMAS
--- a/spec/integration/profiles_spec.rb
+++ b/spec/integration/profiles_spec.rb
@@ -10,7 +10,7 @@ describe 'Profiles API' do
     policies(:one).update! account: accounts(:one)
   end
 
-  path "#{Settings.path_prefix}/#{Settings.app_name}/profiles" do
+  path '/profiles' do
     get 'List all profiles' do
       tags 'profile'
       description 'Lists all profiles requested'
@@ -175,7 +175,7 @@ describe 'Profiles API' do
     end
   end
 
-  path "#{Settings.path_prefix}/#{Settings.app_name}/profiles/{id}" do
+  path '/profiles/{id}' do
     get 'Retrieve a profile' do
       tags 'profile'
       description 'Retrieves data for a profile'
--- a/spec/integration/systems_spec.rb
+++ b/spec/integration/systems_spec.rb
@@ -5,7 +5,7 @@ require 'swagger_helper'
 describe 'Systems API' do
   fixtures :accounts
 
-  path "#{Settings.path_prefix}/#{Settings.app_name}/systems" do
+  path '/systems' do
     get 'List all hosts' do
       tags 'host'
       description 'Lists all hosts requested'
@@ -45,7 +45,7 @@ describe 'Systems API' do
     end
   end
 
-  path "#{Settings.path_prefix}/#{Settings.app_name}/systems/{id}" do
+  path '/systems/{id}' do
     get 'Retrieve a system' do
       tags 'host'
       description 'Lists all hosts requested'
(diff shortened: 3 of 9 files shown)


spec/api/v1/openapi.rb
```diff
@@ -15,6 +15,7 @@ def doc
         {
           openapi: '3.0.3',
           info: info,
+          basePath: "#{Settings.path_prefix}/#{Settings.app_name}",
```
vkrizan, line 18: Shouldn't the base path include the `v1/` or at least have that included within individual specs?

spec/integration/profiles_spec.rb
```diff
@@ -10,7 +10,7 @@
     policies(:one).update! account: accounts(:one)
   end
 
-  path "#{Settings.path_prefix}/#{Settings.app_name}/profiles" do
+  path '/profiles' do
```
vkrizan, line 13: It seems to request from root, ignoring the base path, for me.
```
http://localhost:3000/profiles?...
https://ci.foo.redhat.com:1337/profiles?...
```
Maybe the opening `/` slash should be omitted to get the right base?

```
Using rswag-api 2.3.3
Using rswag-specs 2.3.3
Using rswag-ui 2.3.3
```

spec/integration/systems_spec.rb
```diff
@@ -45,7 +45,7 @@
     end
   end
 
-  path "#{Settings.path_prefix}/#{Settings.app_name}/systems/{id}" do
+  path '/systems/{id}' do
     get 'Retrieve a system' do
       tags 'host'
       description 'Lists all hosts requested'
```
LightOfHeaven1994, line 51: Can you change also description and operation ID for this systems/{id}. Seems it's copied from systems endpoint

swagger/v1/openapi.json
```diff
@@ -5,8 +5,9 @@
     "version": "v1",
     "description": "This is the API for Cloud Services for RHEL Compliance. You can find out more about Red Hat Cloud Services for RHEL at [https://cloud.redhat.com/](https://cloud.redhat.com/)"
   },
+  "basePath": "/api/compliance",
```
LightOfHeaven1994, line 8: Why do we need this one here? I checked spec with this raw and validator is complaining
`Structural error at 
should NOT have additional properties
additionalProperty: basePath
Jump to line 0`

spec/integration/systems_spec.rb
LightOfHeaven1994, line 51: Can you please fix description and operation ID for systems/{id} endpoint? Seems it was copied from /systems endpoint

LightOfHeaven1994 (changes requested):

spec/api/v1/openapi.rb
akofink: I believe that's set in the [version field](https://github.com/RedHatInsights/compliance-backend/blob/master/spec/api/v1/openapi.rb#L28).

spec/integration/profiles_spec.rb
akofink: I'm not sure why it's not working for you. If you check other openapi.json from other apps, paths are specified with leading slashes. Remediations, for example:

```js
{
  "openapi": "3.0.0",
  "info": {…},
  "servers": […],
  "tags": […],
  "paths": {   
    "/remediations": {…},
    "/remediations/download": {…},
    "/remediations/{id}": {…},
    ...
    "/playbook": {…},
    "/version": {…}
  },
  "components": {…}
}
```

I think basePath may not be the correct way to specify this, though. I saw it in approval's openapi.json, but it looks like most other services specify the base path in the `servers`. I will update this PR with that to see if it fixes your issue.

swagger/v1/openapi.json
akofink: Thanks for checking! I found it in approval's openapi.json, but I think the better way to specify this is via `servers`. I'll update it.

spec/integration/profiles_spec.rb
akofink: See http://spec.openapis.org/oas/v3.0.3#paths-object

akofink:
Ok, so this is actually another rswag bug :( I don't think we'll be able to fix it. It's one of the many cases where rswag expects v2 syntax to support tests but then generates an invalid v3 spec. See https://github.com/rswag/rswag/issues/318 for more information.

PR #768, opened 2021-03-03
feat(Rules): RHICOMPL-1403 implement associate rules mutation

skateman:
Implemented the `AssociateRules` mutation inspired by `AssociateSystems` analogously to the REST endpoint for updating rules under a profile.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


app/graphql/mutations/profile/associate_rules.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      argument :id, ID, required: true
+      argument :rule_ids, [ID], required: true
+      field :profile, Types::Profile, null: true
+
+      def resolve(args = {})
+        profile = find_profile(args[:id])
+        if profile
+          profile.update_rules(ids: args[:rule_ids])
+          audit_mutation(profile)
```
vkrizan, line 17: `update_rules` returns number of added and removed rules. Could it be used as a condition for the audit logging? The numbers might be included within the message.

app/graphql/mutations/profile/associate_rules.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+        { profile: profile }
+      end
+
+      include ProfileHelper
```
vkrizan, line 22: Nit: Could be moved to the top.

app/graphql/types/mutation.rb
```diff
@@ -13,6 +13,7 @@ class Mutation < Types::BaseObject
     field :deleteTestResults, mutation: Mutations::TestResult::Delete
     field :associateSystems, mutation: Mutations::Profile::AssociateSystems
     field :associateProfiles, mutation: Mutations::System::AssociateProfiles
+    field :associateRules, mutation: Mutations::Profile::AssociateRules
```
vkrizan, line 16: New types and mutations need the schema file to be updated. See test failure and/or run `bundle exec rake dump_schema`.

vkrizan (commented):
Looks good. Last thing is to update the GraphQL schema (see inline).

app/graphql/mutations/profile/associate_rules.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Mutations
+  module Profile
+    # Mutation to associate systems with a profile
+    class AssociateRules < BaseMutation
+      graphql_name 'associateRules'
+
+      argument :id, ID, required: true
+      argument :rule_ids, [ID], required: true
```
vkrizan, line 10: @bastilian Would frontend send UUIDs or does it use ref_ids instead?

app/graphql/mutations/profile/associate_rules.rb
skateman: Just making sure that I understand :thinking: you want me to emit an audit message like this:
```
Updated N rule associations of profile #{profile.id}
```
?

vkrizan: Yes. Check out profiles controller for inspiration. It logs smth similar.

app/graphql/mutations/profile/associate_rules.rb
bastilian: @vkrizan @skateman The Frontend does send ref_ids at the moment. I'm not opposed to changing that at all btw., but maybe later on, or if we encounter issues with using the ref_id.

vkrizan: @skateman @bastilian What do you guys suggest then? Can we use UUID ids to keep the APIs consistent and update the frontend to do the use them?

akofink: I recommend using UUIDs like the REST API if using `rule_ids`. If we want to support ref_ids in addition, we should add another argument `rule_ref_ids`.

akofink (approved):
Works as expected. Thanks, @skateman!

NB: Sending an empty rules array assigns the canonical set of rules, the same way the REST API works. This is how `Profile#update_rules` works. I'm not sure any user would want to set 0 rules on a profile, but it could come up again. In any case, this PR doesn't alter that behavior; if we want to change it, we should submit another jira and fix it separately.

test/graphql/mutations/associate_rules_test.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class AssociateRulesMutationTest < ActiveSupport::TestCase
+  setup do
+    profiles(:one).update account: accounts(:one),
+                          policy: policies(:one)
+    users(:test).update account: accounts(:one)
```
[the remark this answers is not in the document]
skateman:
```
No fixture named 'one' found for fixture set 'users'
```
:shrug:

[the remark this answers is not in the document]
akofink: Oops, sorry about that. I was thinking of the accounts fixture. :+1:

app/graphql/mutations/profile/associate_rules.rb
```diff
@@ -0,0 +1,36 @@
...
+      def resolve(args = {})
+        profile = find_profile(args[:id])
+        if profile
+          rules_added, rules_removed = profile.update_rules(
+            ids: args[:rule_ids]
+          )
+          audit_mutation(profile, rules_added, rules_removed)
+        end
+        { profile: profile }
+      end
```
akofink, line 24: Recommend wrapping this in a transaction

akofink: It can be done in a followup PR.

skateman: @akofink does this mean that we should have a transaction around the [associate hosts](https://github.com/RedHatInsights/compliance-backend/blob/master/app/graphql/mutations/profile/associate_systems.rb#L16) as well?

akofink: It can't hurt! Thanks for noticing that.

PR #762, opened 2021-03-01
devel(compose): consume the .env in the main rails container

skateman:
This allows the`SETTINGS__PLATFORM_BASIC_AUTH_USERNAME` and `SETTINGS__PLATFORM_BASIC_AUTH_PASSWORD` environment variables mentioned in the README to be consumed from this file.

PR #769, opened 2021-03-04
fix(test): RHICOMPL-1482 make disable_rbac: false when testing

skateman:
As now we [consume the .env file in the rails container](https://github.com/RedHatInsights/compliance-backend/pull/762), the `SETTINGS__DISABLE_RBAC` env var can override the default `nil`
value of `Settings.disable_rbac` which causes issues when running the tests.

In order to prevent this from happening, I'm changing its value to `true` when testing.

### DIFF
--- a/config/settings/test.yml
+++ b/config/settings/test.yml
@@ -2,3 +2,4 @@ host_inventory_url: http://test-host-inventory:8000
 rbac_url: http://test-rbac-url:9002
 # Disable Prometheus
 prometheus_exporter_host:
+disable_rbac: false


skateman:
This will never work as the `config` gem will always favor env variables over config files :cry: so I guess we have to live with it.

vkrizan:
The env vars from dotenv fies have different sections (filenames) to divide envs like dev and test.

skateman:
I think we can still tackle this, by making the tests completely ignore this setting. It really doesn't make sense for tests as it influences their results :confused: wdyt?

vkrizan:
I'd keep settings as close to prod as possible. Tests that need to disable can do so in setup.

PR #770, opened 2021-03-04
devel(ruby): RHICOMPL-1482 update .ruby-version to 2.6.6

skateman:
We were using 2.5.1 in the .ruby-version file, while our Dockerfile states 2.6, so if someone with rvm/rbenv tries to run the developer setup locally, they will get into a conflict.


.ruby-version
```diff
@@ -1 +1 @@
-2.5.1
+2.6.6
```
vkrizan, line 1: prod container uses `ruby-2.6.3-106.module+el8.1.0+3653+beb38eb0.x86_64`, would this change affect it?

skateman: Not really, just local deployment. The container is always influenced by what is in the dockerfile.

vkrizan (approved):
thx

.ruby-version
akofink: We should probably add this to .gitignore. I'm not sure anyone on our team even uses rbenv.

PR #771, opened 2021-03-04
RHICOMPL-1482: make the test output more readable

skateman:
The test output contained a lot of warnings and puts/logging output. I went through all of these and cleaned them up, see the commit messages for more details. There are still two deprecation warnings coming from GraphQL but they will probably need modifications outside the test folder, so I will try to tackle that in a separate PR.

### DIFF
--- a/test/controllers/v1/profiles_controller_test.rb
+++ b/test/controllers/v1/profiles_controller_test.rb
@@ -196,6 +196,7 @@ module V1
       test 'policy_profile_id is exposed' do
         profiles(:one).update!(external: false,
                                parent_profile: profiles(:two),
+                               policy: policies(:one),
                                account: accounts(:one))
         get v1_profiles_url
         assert_response :success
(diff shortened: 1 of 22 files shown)


test/controllers/v1/profiles_controller_test.rb
```diff
@@ -196,6 +196,7 @@ class IndexTest < ProfilesControllerTest
       test 'policy_profile_id is exposed' do
         profiles(:one).update!(external: false,
                                parent_profile: profiles(:two),
+                               policy_object: policies(:one),
```
vkrizan, line 199: `s/policy_object/policy/`

vkrizan (commented):
Awesome improvement! Can you please rebase on master after the refactoring has been made with the policy object association?

Thank you.

vkrizan (approved):
:+1: :1st_place_medal:

PR #774, opened 2021-03-04
devel(rake): RHICOMPL-1482 set the default rake task to test:validate

skateman:
By default minitest sets the `test` task as the default, so when running `bin/rake` it only runs the minitest tests, but not the others. This change makes the `test:validate` to be default, i.e. making the `bin/rake` and `bin/rake test:validate` commands identical.

There might be a reason why we didn't have this before, feel free to close this if that's the case :see_no_evil:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,4 +9,6 @@ task :log => :environment do
   ActiveRecord::Base.logger = Logger.new(STDOUT)
 end
 
+task :default => [:'test:validate']
+
 Rails.application.load_tasks


vkrizan:
I'm using tests only run most of the time. Would this impact running `rails test` also? Can we keep it as it is please?

skateman:
> I'm using tests only run most of the time. Would this impact running `rails test` also? Can we keep it as it is please?

It doesn't affett `rails test` or `rake test` only `rake`

vkrizan:
Oh, in that case if it just sets the default task then it's okay. Sorry for the noise.

akofink (commented):
@skateman this is a good change. :tada: We should merge it.

Victoremepunto:
@skateman @akofink  @vkrizan  planning to reopening this then or ... ?

akofink:
I think we should reopen and merge, but I'll leave it up to @skateman.

skateman:
Somehow this default task messes up the `rails test` command :confused: so it's maybe not a good idea to get it in.

akofink:
Ah, okay - I didn't realize that, though I'm not sure why that should matter. There's always `rake test` if you only want the tests and no linting.

PR #777, opened 2021-03-09
Create policy profiles for each assigned OS minor version

skateman:
Introduced a new method on the Policy model for updating OS versions and integrated it to the `update_hosts`.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -3,6 +3,8 @@
 # Host representation in insights compliance backend. Most of the times
 # these hosts will also show up in the insights-platform host inventory.
 class Host < ApplicationRecord
+  OS_MINOR_VERSION = Arel.sql("system_profile->'operating_system'->'minor'")
+
   self.table_name = 'inventory.hosts'
   self.primary_key = 'id'
 
@@ -17,6 +19,10 @@ class Host < ApplicationRecord
   has_many :test_result_profiles, through: :test_results, source: :profile
   has_many :policies, through: :policy_hosts
   has_many :assigned_profiles, through: :policies, source: :profiles
+
+  def self.os_minor_versions(hosts)
+    distinct.where(id: hosts).pluck(OS_MINOR_VERSION)
+  end
 
   def readonly?
     true
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -50,9 +50,8 @@ class Policy < ApplicationRecord
     return unless new_host_ids
 
     removed = policy_hosts.where.not(host_id: new_host_ids).destroy_all
-    imported = PolicyHost.import((new_host_ids - host_ids).map do |host_id|
-      { host_id: host_id, policy_id: id }
-    end)
+    imported = PolicyHost.import_from_policy(id, new_host_ids - host_ids)
+
     [imported.ids.count, removed.count]
   end
 
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -51,8 +51,21 @@ class Policy < ApplicationRecord
 
     removed = policy_hosts.where.not(host_id: new_host_ids).destroy_all
     imported = PolicyHost.import_from_policy(id, new_host_ids - host_ids)
+    update_os_minor_versions
 
     [imported.ids.count, removed.count]
+  end
+
+  def update_os_minor_versions
+    unassigned_minor_versions.each do |os_minor_version|
+      Profile.canonical_for_os(
+        initial_profile.os_major_version, os_minor_version
+      ).find_by(ref_id: initial_profile.ref_id)&.clone_to(
+        account: account,
+        os_minor_version: os_minor_version,
+        policy: self
+      )
+    end
   end
 
   def compliant?(host)
@@ -79,6 +92,14 @@ class Policy < ApplicationRecord
   end
 
   private
+
+  def unassigned_minor_versions
+    stored_versions = profiles.pluck(:os_minor_version)
+    Host.os_minor_versions(hosts).reject do |version|
+      # Ignore already stored minor versions
+      stored_versions.include?(version.to_s)
+    end
+  end
 
   def audit_bo_autoremove(removed_bos)
     return if removed_bos.empty?
--- a/app/models/policy_host.rb
+++ b/app/models/policy_host.rb
@@ -7,4 +7,10 @@ class PolicyHost < ApplicationRecord
 
   validates :policy, presence: true
   validates :host, presence: true, uniqueness: { scope: :policy }
+
+  def self.import_from_policy(policy_id, host_ids)
+    import(host_ids.map do |host_id|
+      { host_id: host_id, policy_id: policy_id }
+    end)
+  end
 end
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -14,12 +14,9 @@ class Profile < ApplicationRecord
   belongs_to :parent_profile, class_name: 'Profile', optional: true
 
   validates :ref_id, uniqueness: {
-    scope: %i[account_id benchmark_id external policy_id]
+    scope: %i[account_id benchmark_id os_minor_version policy_id],
+    message: 'must be unique in a policy OS version'
   }, presence: true
-  validates :ref_id, uniqueness: {
-    scope: %i[account_id benchmark_id policy_id],
-    message: 'must be unique in a policy'
-  }, if: :policy_id
   validates :ref_id, uniqueness: {
     scope: %i[account_id benchmark_id],
     conditions: -> { where(external: false) },
@@ -29,10 +26,19 @@ class Profile < ApplicationRecord
     scope: %i[ref_id account_id benchmark_id],
     conditions: -> { where(policy_id: nil) }
   }, unless: :policy_id
+  validates :os_minor_version, uniqueness: {
+    scope: %i[account_id policy_id]
+  }, if: -> { os_minor_version.present? }
   validates :name, presence: true
   validates :benchmark_id, presence: true
   validates :account, presence: true, if: -> { hosts.any? }
   validates :policy, presence: true, if: -> { policy_id }
+
+  scope :canonical_for_os, lambda { |os_major_version, os_minor_version|
+    canonical.ssg_versions(
+      SupportedSsg.latest_ssg_version_for_os(os_major_version, os_minor_version)
+    ).os_major_version(os_major_version)
+  }
 
   delegate :account_number, to: :account, allow_nil: true
 
@@ -74,7 +80,7 @@ class Profile < ApplicationRecord
     parent_profile_id.blank?
   end
 
-  def clone_to(account:, policy:)
+  def clone_to(account:, policy:, os_minor_version: nil)
     new_profile = in_account(account, policy)
     if new_profile.nil?
       (new_profile = dup).update!(account: account,
@@ -83,6 +89,9 @@ class Profile < ApplicationRecord
                                   policy: policy)
       new_profile.update_rules(ref_ids: rules.pluck(:ref_id))
     end
+
+    # Update the os minor version if not already set
+    new_profile.update_os_minor_version(os_minor_version)
 
     new_profile
   end
@@ -100,6 +109,10 @@ class Profile < ApplicationRecord
     ref_id.downcase.split(
       'xccdf_org.ssgproject.content_profile_'
     )[1] || ref_id
+  end
+
+  def update_os_minor_version(version)
+    update!(os_minor_version: version) if version && os_minor_version.empty?
   end
 
   private
--- a/test/models/policy_test.rb
+++ b/test/models/policy_test.rb
@@ -90,6 +90,10 @@ class PolicyTest < ActiveSupport::TestCase
   context 'update_hosts' do
     should 'add new hosts to an empty host set' do
       policies(:one).update!(hosts: [])
+      profiles(:one).update!(account: accounts(:test))
+      policies(:one).update!(profiles: [profiles(:one)])
+      profiles(:one).update!(parent_profile: profiles(:two))
+
       assert_empty(policies(:one).hosts)
       assert_difference('policies(:one).hosts.count', hosts.count) do
         changes = policies(:one).update_hosts(hosts.pluck(:id))
@@ -99,6 +103,9 @@ class PolicyTest < ActiveSupport::TestCase
 
     should 'add new hosts to an existing host set' do
       policies(:one).update!(hosts: hosts[0...-1])
+      profiles(:one).update!(account: accounts(:test))
+      policies(:one).update!(profiles: [profiles(:one)])
+
       assert_not_empty(policies(:one).hosts)
       assert_difference('policies(:one).hosts.count', 1) do
         changes = policies(:one).update_hosts(hosts.pluck(:id))
@@ -117,6 +124,9 @@ class PolicyTest < ActiveSupport::TestCase
 
     should 'add new and remove old hosts from an existing host set' do
       policies(:one).update!(host_ids: hosts.pluck(:id)[0...-1])
+      profiles(:one).update!(account: accounts(:test))
+      policies(:one).update!(profiles: [profiles(:one)])
+
       assert_not_empty(policies(:one).hosts)
       assert_difference('policies(:one).hosts.count', 0) do
         changes = policies(:one).update_hosts(hosts.pluck(:id)[1..-1])
@@ -178,6 +188,82 @@ class PolicyTest < ActiveSupport::TestCase
                    policies(:one).score(host: hosts(:one))
       assert_equal test_results(:two).score,
                    policies(:one).score(host: hosts(:two))
+    end
+  end
+
+  context 'update_os_minor_versions' do
+    setup do
+      @profile = Profile.new(parent_profile: profiles(:two),
+                             account: accounts(:one),
+                             policy: policies(:one)).fill_from_parent
+      policies(:one).update!(account: accounts(:one))
+      @profile.save!
+
+      @policy_host = PolicyHost.create!(
+        policy: @profile.policy,
+        host: hosts(:one)
+      )
+    end
+
+    should 'update initial_profile when host has the latest minor version' do
+      @profile.benchmark.update!(version: '0.1.33')
+
+      @profile.policy.update_os_minor_versions
+
+      assert_equal @profile.reload.os_minor_version, '4'
+    end
+
+    context 'child profiles' do
+      setup do
+        @new_benchmark = benchmarks(:one).dup
+        @new_benchmark.assign_attributes(version: '0.1.33')
+        @new_benchmark.save!
+
+        @new_profile = profiles(:two).dup
+        @new_profile.assign_attributes(benchmark: @new_benchmark)
+        @new_profile.save!
+      end
+
+      should 'create a new profile when it does not exist' do
+        @profile.policy.update_os_minor_versions
+
+        assert_equal @profile.reload.os_minor_version, ''
+        assert_equal @profile.policy.profiles.external(true)
+                             .first.os_minor_version, '4'
+      end
+
+      should 'update existing profile according the minor version' do
+        child_profile = @new_profile.clone_to(
+          account: accounts(:one),
+          policy: @profile.policy
+        )
+
+        @profile.policy.update_os_minor_versions
+
+        assert_equal @profile.reload.os_minor_version, ''
+        assert_equal child_profile.reload.os_minor_version, '4'
+      end
+
+      context 'multiple supported benchmarks with older assigned version' do
+        should 'not touch the newer profile' do
+          @profile.benchmark.update!(version: '0.1.49')
+          @profile.update!(os_minor_version: '9')
+
+          @new_benchmark.update!(version: '0.1.52')
+
+          child_profile = @new_profile.clone_to(
+            account: accounts(:one),
+            policy: @profile.policy
+          )
+
+          Host.expects(:os_minor_versions).returns([9])
+
+          @profile.policy.update_os_minor_versions
+
+          assert_equal @profile.reload.os_minor_version, '9'
+          assert_equal child_profile.reload.os_minor_version, ''
+        end
+      end
     end
   end
 end
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -31,6 +31,11 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
       end.map(&:version)
     end
 
+    def latest_ssg_version_for_os(os_major_version, os_minor_version)
+      ssg_versions_for_os(os_major_version, os_minor_version)
+        .max_by { |ssg_v| Gem::Version.new(ssg_v) }
+    end
+
     def versions
       all.map(&:version).uniq
     end
--- a/test/models/host_test.rb
+++ b/test/models/host_test.rb
@@ -247,4 +247,8 @@ class HostTest < ActiveSupport::TestCase
       assert_includes Host.search_for(search), hosts(:one)
     end
   end
+
+  test '#os_minor_versions' do
+    assert_equal Host.os_minor_versions([hosts(:one)]), [4]
+  end
 end
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -11,7 +11,8 @@ class ProfileTest < ActiveSupport::TestCase
   should have_many(:rule_results).through(:test_results)
   should belong_to(:policy).optional
   should validate_uniqueness_of(:ref_id)
-    .scoped_to(%i[account_id benchmark_id external policy_id])
+    .scoped_to(%i[account_id benchmark_id os_minor_version policy_id])
+    .with_message('must be unique in a policy OS version')
   should validate_presence_of :ref_id
   should validate_presence_of :name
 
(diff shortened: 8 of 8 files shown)


app/models/host.rb
```diff
@@ -18,6 +18,10 @@ class Host < ApplicationRecord
   has_many :policies, through: :policy_hosts
   has_many :assigned_profiles, through: :policies, source: :profiles
 
+  scope :os_versions, lambda { |hosts|
+    select(:system_profile).where(id: hosts).map(&:os_version).uniq
+  }
```
akofink, line 23: This isn't how scopes are typically used. They should return another ActiveRecord::Relation. You may want a class method instead. In either case, you can pluck the OS versions directly using SQL:

```rb
pluck(%w(system_profile->'operating_system'->'major' system_profile->'operating_system'->'minor'))
```

app/models/policy.rb
```diff
@@ -49,10 +49,15 @@ def fill_from(profile:)
   def update_hosts(new_host_ids)
     return unless new_host_ids
 
-    removed = policy_hosts.where.not(host_id: new_host_ids).destroy_all
-    imported = PolicyHost.import((new_host_ids - host_ids).map do |host_id|
-      { host_id: host_id, policy_id: id }
-    end)
+    removed = imported = nil
+
+    PolicyHost.transaction do
```
akofink, line 54: This was designed for the caller to wrap in a transaction see https://github.com/RedHatInsights/compliance-backend/blob/master/app/controllers/v1/profiles_controller.rb#L45. The GraphQL mutation could do the same, or we should remove the transation within our REST controller, but it should be consistent.

app/models/policy_host.rb
```diff
@@ -7,4 +7,10 @@ class PolicyHost < ApplicationRecord
 
   validates :policy, presence: true
   validates :host, presence: true, uniqueness: { scope: :policy }
+
+  def self.import_from_policy(policy_id, host_ids)
+    PolicyHost.import(host_ids.map do |host_id|
```
akofink, line 12: `import` is in the local scope:

```suggestion
    import(host_ids.map do |host_id|
```

app/services/os_version_updater.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# A service class to update OS version for each policy with assigned hosts
+class OsVersionUpdater
+  class << self
+    def run!
+      Policy.all.each do |policy|
```
akofink, line 7: find_each auto-batches by 1000 records at a time. Calling `all` tries to create an array in memory of all policies in the DB:

```suggestion
      Policy.find_each do |policy|
```

app/services/os_version_updater.rb
```diff
@@ -0,0 +1,12 @@
...
+        policy.send(:update_os_versions, policy.hosts.map(&:id))
```
akofink, line 8: IMO `update_os_versions` should just use the currently assigned hosts instead of a passed in list of IDs. That would make it fool-proof to use and always guarantee the correct result, and then it could be public without any issue so we don't have to `send` it.

akofink (commented):
Looking good! Some more comments inline. I'll want to test it more thoroughly as it matures. When assigning an OS to a profile, we must consult the SupportedSsgs matrix to ensure we're adding the supported OS to the profile with the matching SSG version.

app/models/host.rb
```diff
@@ -18,6 +18,15 @@ class Host < ApplicationRecord
   has_many :policies, through: :policy_hosts
   has_many :assigned_profiles, through: :policies, source: :profiles
 
+  def self.os_versions(hosts)
+    select(:system_profile).where(id: hosts).pluck(
+      %w[
+        system_profile->'operating_system'->'major'
+        system_profile->'operating_system'->'minor'
+      ]
+    ).map { |version| version.join('.') }.uniq
```
akofink, line 27: It'd be great to use `distinct` instead of `uniq`. You may need to store the array in `pluck` in a constant, since we'd need to use it twice:

```rb
Host.select(%w(system_profile->'operating_system'->'major' system_profile->'operating_system'->'minor')).distinct.pluck(%w(system_profile->'operating_system'->'major' system_profile->'operating_system'->'minor'))
```

resulting query:

```sql
SELECT DISTINCT system_profile->'operating_system'->'major', system_profile->'operating_system'->'minor' FROM "inventory"."hosts"
```

skateman: I just noticed that the `system_profile` is json-based :scream: last time (2013?) I tried to use this in rails, it was crashing :laughing:

app/models/policy.rb
skateman: So I can just rely on the fact that there's a `transaction` around?

app/services/os_version_updater.rb
skateman: I don't know why I left it as `private` :thinking: sorry :see_no_evil: 

Regarding the list of IDs, I just wanted to avoid that +1 query for the `hosts.pluck(:id)` but I assume you are okay with it.

db/schema.rb
```diff
@@ -94,12 +94,13 @@
...
     t.boolean "external", default: false, null: false
     t.uuid "policy_id"
+    t.string "os_version"
     t.index ["account_id"], name: "index_profiles_on_account_id"
     t.index ["external"], name: "index_profiles_on_external"
     t.index ["name"], name: "index_profiles_on_name"
     t.index ["parent_profile_id"], name: "index_profiles_on_parent_profile_id"
     t.index ["policy_id"], name: "index_profiles_on_policy_id"
-    t.index ["ref_id", "account_id", "benchmark_id", "external", "policy_id"], name: "uniqueness", unique: true
+    t.index ["ref_id", "account_id", "benchmark_id", "external", "os_version", "policy_id"], name: "uniqueness", unique: true
```
vkrizan, line 103: I guess we can drop `external` from the uniqueness, as the support for external policies has been dropped. `external` is can be only considered as a flag for initial policy profile.

app/models/policy.rb
```diff
@@ -50,12 +50,25 @@ def update_hosts(new_host_ids)
...
+
     [imported.ids.count, removed.count]
   end
 
+  def update_os_versions
+    os_versions = Host.os_versions(hosts.pluck(:id))
+    initial_profile.update(os_version: os_versions.pop) if os_versions.any?
+
+    os_versions.each do |os_version|
+      initial_profile.parent_profile.clone_to(
```
vkrizan, line 64: I'm not sure if this is correct to clone by the initial profile's parent. Each other profile would have a different parent since it is from a different benchmark. You can check the report parsing, where the clone to is used from/with a profile that is generated from the report.

db/migrate/20210308092402_add_os_version_to_profiles.rb
```diff
@@ -0,0 +1,18 @@
+class AddOsVersionToProfiles < ActiveRecord::Migration[5.2]
+  def up
+    remove_index :profiles, name: 'uniqueness'
+    add_column :profiles, :os_version, :string, default: nil
+    add_index :profiles,
+              %i[ref_id account_id benchmark_id external os_version policy_id],
+              unique: true, name: 'uniqueness'
+    OsVersionUpdater.run!
```
vkrizan, line 8: We can try to separate schema migration and data migration, if that is even possible.

app/models/policy.rb
```diff
@@ -50,12 +50,25 @@ def update_hosts(new_host_ids)
...
-    end)
+    imported = PolicyHost.import_from_policy(id, new_host_ids - host_ids)
+    update_os_versions
+
     [imported.ids.count, removed.count]
   end
 
+  def update_os_versions
+    os_versions = Host.os_versions(hosts.pluck(:id))
+    initial_profile.update(os_version: os_versions.pop) if os_versions.any?
```
vkrizan, line 61: Setting of the OS info on the initial profile seems non-deterministic. Initial profile should get the OS info set only if it is from the benchmark that an OS has a latest support. Once set, the OS info must not be changed.

app/models/profile.rb
```diff
@@ -100,9 +104,10 @@ def short_ref_id
 
   private
 
-  def in_account(account, policy)
+  def in_account(account, policy, os_version)
     Profile.find_by(account: account, ref_id: ref_id,
                     policy: policy,
+                    os_version: os_version,
```
akofink, line 110: No need to find accounts with OS info. We want to find it even if os_version is `nil`.

app/models/policy.rb
akofink: Yes, we can rely on the calling code to wrap this in a transaction. We should revisit the graphql mutation: https://github.com/RedHatInsights/compliance-backend/pull/768/files#r590519204

db/migrate/20210308092402_add_os_version_to_profiles.rb
akofink: Agreed. Let's add the os_version field with a default value of `''` and `null: false` and update existing profiles in another migration.

db/migrate/20210308092402_add_os_version_to_profiles.rb
```diff
@@ -0,0 +1,18 @@
+class AddOsVersionToProfiles < ActiveRecord::Migration[5.2]
+  def up
+    remove_index :profiles, name: 'uniqueness'
+    add_column :profiles, :os_version, :string, default: nil
```
akofink, line 4: We may want to add a normal index for searching here:

```rb
    add_index :profiles, :os_version
```

app/services/os_version_updater.rb
akofink: We should be able to avoid N+1 by including hosts like `Policy.includes(:hosts).find_each`...

app/services/os_version_updater.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# A service class to update OS version for each policy with assigned hosts
+class OsVersionUpdater
+  class << self
+    def run!
+      Policy.find_each do |policy|
+        policy.send(:update_os_versions, policy.hosts.map(&:id))
```
akofink, line 8: This currently fails when running the migration:

```
wrong number of arguments (given 1, expected 0)                                                                                                                                                                    
/app/app/models/policy.rb:59:in `update_os_versions'                                                                                                                                                               
/app/app/services/os_version_updater.rb:8:in `block in run!'                                                                                                                                                       
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:70:in `block (2 levels) in find_each'                                                                                            
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:70:in `each'                                                                                                                     
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:70:in `block in find_each'                                                                                                       
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:136:in `block in find_in_batches'                                                                                                
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:238:in `block in in_batches'                                                                                                     
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:222:in `loop'                                                                                                                    
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:222:in `in_batches'                                                                                                              
/usr/local/bundle/gems/activerecord-5.2.4.5/lib/active_record/relation/batches.rb:135:in `find_in_batches'
```

db/migrate/20210308092402_add_os_version_to_profiles.rb
akofink, line 4: I've started thinking about how the API is going to look. We currently already have `os_major_version` which comes from the benchmark. If we add this field as `os_version` here, it's partially redundant. Perhaps we should only store `os_minor_version` instead. @vkrizan @skateman what do you think? There's a issue with data integrity if we allow setting any `os_version` as a string without validating that it also matches the `os_major_version` from the benchmark.

app/models/policy.rb
akofink: Agreed. @skateman and I have already discussed this.

db/schema.rb
akofink: I'm fine with including this change with this PR.

db/migrate/20210308092402_add_os_version_to_profiles.rb
vkrizan: Hmm, part of me is saying no to the suggestion and part is interested in the change. I would be more for a split into os_major and os_minor (two fields) to match models... If you think about it, our Policy models are not technically tied to an os_major. All of that is inferred from the one of the profile's benchmark and even that is inferred from the benchmarks ref_id. There are lots of parts that can go out of sync, so I would not worry about them much at this point. If we go that way, I'd suggest a refactoring to the policy model to have an os_major_version instead of the the inferred one. However, any of the solutions would work nominally. I'm inclined to consistency.

skateman: I added the minor version field in a separate PR https://github.com/RedHatInsights/compliance-backend/pull/778

vkrizan: I've realized that later... The minor on the profile and inferred major within benchmark make sense.

As the os_major_version is inferred from benchmark, combination of `os_minor_version` and `benchmark_id` is good enough.

skateman:
Ready for another round of review, with a bunch of inline comments

akofink (commented):
Thanks, @skateman! It seems like you've got the general concept of what needs to be done. There were quite a few issues to be resolved to get it working but nothing major.

I recommend running the tests locally in the future to identify silly mistakes like method names not matching. Any errors (rather than failures) are good indicators of issues introduced and should help to point you in the right direction.

vkrizan (commented):
Looks very promising!

app/services/os_version_updater.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# A service class to update OS version for each policy with assigned hosts
+class OsVersionUpdater
+  class << self
+    def run!
+      Policy.find_each(&:update_os_versions)
```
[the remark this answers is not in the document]
skateman: Yeah, is there a more optimal way to do this?

app/models/policy.rb
```diff
@@ -50,12 +50,53 @@ def update_hosts(new_host_ids)
...
+        os_minor_version: version
+      )
+        initial_profile.update(os_minor_version: version)
+      else
+        # This is probably super ineffective and will optimize it after it
+        # has been validated as the right solution for this
+        Profile.canonical.find_by(
+          ref_id: initial_profile.ref_id,
+          benchmark: Xccdf::Benchmark.find_by(
+            version: SupportedSsg.ssg_versions_for_os(
```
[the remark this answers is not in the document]
skateman: @akofink are you sure about the `where` here? Shouldn't be rather `find_by`? I might be missing something but if there is more than one matching canonical profile, there's something wrong. But if that's possible, how should I pick the right one from the array returned by `where`?

app/models/policy.rb
```diff
@@ -50,12 +50,53 @@ def update_hosts(new_host_ids)
...
+              initial_profile.os_major_version,
+              version
+            ).max_by { |ssg_v| Gem::Version.new(ssg_v) }
+            # I'm not sure if I should pick the latest version here
+            # or go with all supported versions to find the right
+            # benchmark and rule out the old version a level above
+            # on the Profile somehow
```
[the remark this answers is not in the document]
skateman: > we should do nothing

as far as I understand, this is already sorted out by this:
```ruby
    stored_versions = profiles.pluck(:os_minor_version)
    versions = Host.os_minor_versions(hosts.pluck(:id)).reject do |version|
      stored_versions.include?(version)
    end
```

app/services/os_version_updater.rb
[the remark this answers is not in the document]
vkrizan: We can simplify this by setting it up based on the supported matrix.

Here is the idea:
1. Pick one of the supported OS versions (e.g. 7.9) and its supported SSG record
2. On every non-canonical (user) profile set OS minor matching the SSG version from the step one
3. (Optionally) Check (filter by) whether the policy has hosts with that minor version
4. Repeat

For the case of dual SSG support within one minor version, first set the OS minor on latest and then do a next pass. In the next pass find profiles that are matching and set the OS, but only if the OS minor isn't already set on other profile.

Depending on the amount of data flow and performacne we might skip option 3 and leave some policies with OS minor set even if there is no such host. That might be acceptable, but we might need to discuss it. If hosts has to be checked, then they should be do so right in the profile search from step 2 (using subselects, if that's possible).

Or, we might not do any migration and should let the customers resave/update existing policies to properly assign the OS versions to profiles.

I'd suggest to split this data migration concern, as it might need a bit more discussions and testing.

app/models/policy.rb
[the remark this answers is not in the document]
akofink: Any suggestion I give like this is just a suggestion, not a complete solution. This one is meant to demonstrate the available scopes only. `find_by` makes sense :+1: Sorry for the confusion.

app/models/profile.rb
```diff
@@ -74,12 +76,13 @@ def canonical?
...
+  def clone_to(account:, policy:, os_minor_version: '')
     new_profile = in_account(account, policy)
     if new_profile.nil?
       (new_profile = dup).update!(account: account,
                                   parent_profile: self,
                                   external: true,
+                                  os_minor_version: os_minor_version,
                                   policy: policy)
       new_profile.update_rules(ref_ids: rules.pluck(:ref_id))
     end
```
akofink, line 88: This is one way we could make `clone_to` smarter about when to add an `os_minor_version`. Now the caller doesn't have to know about any of this logic, and it's encapsulated here in the profile model.

```suggestion
                                  policy: policy)
      new_profile.update_rules(ref_ids: rules.pluck(:ref_id))
    end
    new_profile.update!(os_minor_version: os_minor_version) if os_minor_version && new_profile.os_minor_version.empty?
```

akofink, line 88: This is one way we could make `clone_to` smarter about when to add an `os_minor_version`. Now the caller doesn't have to know about any of this logic, and it's encapsulated here in the profile model.

```suggestion
  def clone_to(account:, policy:, os_minor_version: nil)
    new_profile = in_account(account, policy)
    if new_profile.nil?
      (new_profile = dup).update!(account: account,
                                  parent_profile: self,
                                  external: true,
                                  os_minor_version: os_minor_version,
                                  policy: policy)
      new_profile.update_rules(ref_ids: rules.pluck(:ref_id))
    end
    new_profile.update!(os_minor_version: os_minor_version) if os_minor_version && new_profile.os_minor_version.empty?
```

app/models/policy.rb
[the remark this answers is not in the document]
akofink: Yep, this should handle that case, though I think it'd be prudent (and more elegant, not that it matters) to also handle this in `clone_to` if a profile is passed that already exists on the policy.

app/services/os_version_updater.rb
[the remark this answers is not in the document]
skateman: I removed the migration from this PR

test/models/policy_test.rb
```diff
@@ -180,4 +190,56 @@ class PolicyTest < ActiveSupport::TestCase
                    policies(:one).score(host: hosts(:two))
     end
   end
+
+  context 'update_os_minor_versions' do
```
akofink, line 194: We should test this scenario:

1. Policy with 2 profiles of the same OS minor version (i.e. 0.1.49 and 0.1.52, both RHEL 7.9)
2. Profile 0.1.49 has os_minor_version set to '9'
3. Profile 0.1.52 has no os_minor_version set
4. Call update_os_minor_versions on the policy

Actual:
Both profiles on the policy have `'9'` set for their `os_minor_version`.

Expected:
Only the 0.1.49 profile should have `'9'` set - the 0.1.52 profile should not be altered and have `''` set for its `os_minor_version`.

Additionally, we need to add a validation and maybe a DB index to enforce that a policy can only ever have one profile with each os_minor_version.

app/models/policy.rb
```diff
@@ -81,6 +93,13 @@ def destroy_orphaned_business_objective
 
   private
 
+  def unassigned_minor_versions
+    stored_versions = profiles.pluck(:os_minor_version)
+    Host.os_minor_versions(hosts.pluck(:id)).reject do |version|
+      stored_versions.include?(version) # Ignore already stored minor versions
+    end
+  end
```
akofink, line 101: This doesn't quite work as expected (string vs integer comparison is the culprit):

```rb
[304] pry(main)> Policy.first.send(:unassigned_minor_versions)
  Policy Load (1.3ms)  SELECT  "policies".* FROM "policies" ORDER BY "policies"."id" ASC LIMIT $1  [["LIMIT", 1]]
   (1.0ms)  SELECT "profiles"."os_minor_version" FROM "profiles" WHERE "profiles"."policy_id" = $1  [["policy_id", "e1e6e24e-46e2-4f9f-8986-50432875a6f5"]]
   (19.3ms)  SELECT "inventory"."hosts"."id" FROM "inventory"."hosts" INNER JOIN "policy_hosts" ON "inventory"."hosts"."id" = "policy_hosts"."host_id" WHERE "policy_hosts"."policy_id" = $1  [["policy_id", "e1e6e24e-46e2-4f9f-8986-50432875a6f5"]]
   (18.2ms)  SELECT DISTINCT system_profile->'operating_system'->'minor' FROM "inventory"."hosts" WHERE "inventory"."hosts"."id" = $1  [["id", "78a1ce1a-6173-484a-bcaa-9b10a46e651b"]]
=> [9]
[305] pry(main)> Policy.first.profiles.map { |p| [p.id, p.os_minor_version, p.ssg_version] }
  Policy Load (1.0ms)  SELECT  "policies".* FROM "policies" ORDER BY "policies"."id" ASC LIMIT $1  [["LIMIT", 1]]
  Profile Load (0.7ms)  SELECT "profiles".* FROM "profiles" WHERE "profiles"."policy_id" = $1  [["policy_id", "e1e6e24e-46e2-4f9f-8986-50432875a6f5"]]
  Xccdf::Benchmark Load (0.6ms)  SELECT  "benchmarks".* FROM "benchmarks" WHERE "benchmarks"."id" = $1 LIMIT $2  [["id", "52a35239-0b11-46c0-a372-c32f7d42dc8f"], ["LIMIT", 1]]
  Xccdf::Benchmark Load (0.6ms)  SELECT  "benchmarks".* FROM "benchmarks" WHERE "benchmarks"."id" = $1 LIMIT $2  [["id", "2e08ee4c-7114-4671-9227-fcc636df8653"], ["LIMIT", 1]]
=> [["303dce8f-053d-4d38-974d-e464c27ed0e0", "9", "0.1.52"], ["4948c203-345b-4501-9c1d-48752dd97611", "", "0.1.54"]]
```

app/models/policy.rb
```diff
@@ -81,6 +93,13 @@ def destroy_orphaned_business_objective
 
   private
 
+  def unassigned_minor_versions
+    stored_versions = profiles.pluck(:os_minor_version)
+    Host.os_minor_versions(hosts.pluck(:id)).reject do |version|
```
akofink, line 98: No need to `pluck` the ids here. Doing it as suggested uses a nested subquery vs. 2 independent queries.

```suggestion
    Host.os_minor_versions(hosts).reject do |version|
```

```rb
[307] pry(main)> Host.os_minor_versions(Host.limit(10))
   (64.3ms)  SELECT DISTINCT system_profile->'operating_system'->'minor' FROM "inventory"."hosts" WHERE "inventory"."hosts"."id" IN (SELECT  "inventory"."hosts"."id" FROM "inventory"."hosts" LIMIT $1)  [["LIMIT", 10]]
=> [7, 9]
```

app/models/policy.rb
```diff
@@ -81,6 +93,13 @@ def destroy_orphaned_business_objective
...
+      stored_versions.include?(version) # Ignore already stored minor versions
```
akofink, line 99: `stored_versions` is a list of strings (i.e. `["9", ""]`), but `version` is an integer:

```suggestion
      stored_versions.include?(version.to_s) # Ignore already stored minor versions
```

akofink (commented):
Works mostly as expected, just one minor bug I found. Nice work!

test/models/policy_test.rb
skateman: Wouldn't this change sort this out?
```diff
diff --git a/app/models/profile.rb b/app/models/profile.rb
index f29083f..0c9a22f 100644
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -27,7 +27,7 @@ class Profile < ApplicationRecord
     conditions: -> { where(policy_id: nil) }
   }, unless: :policy_id
   validates :os_minor_version, uniqueness: {
-    scope: %i[account_id benchmark_id policy_id]
+    scope: %i[account_id policy_id]
   }, if: -> { os_minor_version.present? }
   validates :name, presence: true
   validates :benchmark_id, presence: true

```

akofink: Yes, we should make this validation change. :+1:

akofink (approved):
All my concerns are addressed, and it works as expected! There's one more update for the profile uniqueness validation, but once added, it's good to merge. Nice work, @skateman!

akofink:
@vkrizan any other comments/concerns? If not, this should be good to merge. Thanks, @skateman!

vkrizan (approved):
Looks great!

PR #780, opened 2021-03-09
devel(reload): explicit namespacing for AR fixes code reloads

skateman:
Due to the bad namespacing in activerecord concerns, code reloading was not really functional. By putting `::` before AR class calls inside concerns we are bridging this problem as the autoloader will not try to load a namespaced version of the AR classes.


akofink (approved):
Nice! Thanks!

PR #778, opened 2021-03-09
feat(Profile): RHICOMPL-1402 add the os_minor_version column

skateman:
As discussed [here](https://github.com/RedHatInsights/compliance-backend/pull/777#discussion_r590542227) and on meetings, I'm creating the `os_minor_version` field and adjusting the unique index on the profiles table.

## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Authentication and Password Management
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Error Handling and Logging
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management
- [x] General Coding Practices

PR #781, opened 2021-03-10
feat(APIs): RHICOMPL-1406 add profile os_minor_version

akofink:
Add os_minor_version and os_version to both GraphQL and REST APIs for
profiles.

~Requires https://github.com/RedHatInsights/compliance-backend/pull/778~ merged

Can be tested with https://github.com/RedHatInsights/insights-core/pull/2971

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/graphql/schema.graphql
+++ b/app/graphql/schema.graphql
@@ -103,6 +103,9 @@ type Profile implements Node & RulesPreload {
   ): String!
   majorOsVersion: String!
   name: String!
+  osMajorVersion: String!
+  osMinorVersion: String!
+  osVersion: String!
   policy: Profile
   policyType: String!
   profiles: [Profile!]
--- a/app/graphql/types/profile.rb
+++ b/app/graphql/types/profile.rb
@@ -78,6 +78,9 @@ module Types
     field :compliant_host_count, Int, null: false
 
     field :major_os_version, String, null: false
+    field :os_major_version, String, null: false
+    field :os_minor_version, String, null: false
+    field :os_version, String, null: false
     field :policy_type, String, null: false
 
     def last_scanned(args = {})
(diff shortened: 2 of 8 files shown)


skateman (approved):
The Seal of Approval

app/graphql/schema.graphql
```diff
@@ -102,6 +102,9 @@ type Profile implements Node & RulesPreload {
   ): String!
   majorOsVersion: String!
   name: String!
+  osMajorVersion: String!
+  osMinorVersion: String!
+  osVersion: String!
```
vkrizan, line 107: I'm wondering if we need a composite `os_version`. What's the benefit of the duplicity?

skateman: I guess parity, i.e. the same stuff defined on the model and in GraphQL. At least that's my perception as I see both `majorOsVersion` and `osMajorVersion` and they're aliased in the model. Or if I'm wrong, we're saving an operation on the frontend :laughing: :see_no_evil:

akofink: IMO this is important for API users to see the same thing (i.e. the tab name) as UI users. For the UI implementation, it's also helpful because there need not be any logic to join major and minor versions. In general, I prefer to do any data translations in the backend and expose them directly in the API this way.

The reason I've added osMajorVersion (in addition to majorOsVersion) is because our search terms here are `os_major_version` and `os_minor_version`. It makes sense for the attributes to match the search terms exactly where possible.

PR #783, opened 2021-03-10
devel(reload): explicit de-namespacing in all model concerns

skateman:
I still found some problems with reloading after [de-namespacing](https://github.com/RedHatInsights/compliance-backend/pull/780) everything in the `ClassMethods` modules under model concerns. I basically added the `::` for any class called in the concerns. It might be an overkill, but it was necessary under `ClassMethods` as well to use it on stuff coming from external gems and not our codebase.


akofink (approved):
Seems fine to me. It's pretty hard to track down all the places it will fail to reload. I did notice that this seems to be an issue more with the GraphQL API than REST.

PR #786, opened 2021-03-15
fix(migrations): RHICOMPL-1513 clean up duplicate canonical profiles

skateman:
Cleaning up possible duplicated canonical profiles that might mess up the integrity of the database after https://github.com/RedHatInsights/compliance-backend/pull/778

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


app/models/profile.rb
```diff
@@ -12,6 +12,11 @@ class Profile < ApplicationRecord
   belongs_to :account, optional: true
   belongs_to :benchmark, class_name: 'Xccdf::Benchmark'
   belongs_to :parent_profile, class_name: 'Profile', optional: true
+  has_many :child_profiles,
+           class_name: 'Profile',
+           foreign_key: :parent_profile_id,
+           inverse_of: :child_profiles,
```
akofink, line 18:
```suggestion
           inverse_of: :parent_profile,
```

app/services/duplicate_canonical_profile_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+# Class that deduplicates canonical profiles with identical benchmark/ref IDs
+class DuplicateCanonicalProfileResolver
+  class << self
+    def run!
+      duplicates.each do |duplicate|
+        profiles = Profile.canonical.includes(:child_profiles).where(
+          duplicate.slice(:ref_id, :benchmark_id)
+        ).order(created_at: :asc).offset(1)
+
+        next unless profiles.all? { |p| p.child_profiles.count.zero? }
```
akofink, line 12: Just to say it out loud: our strategy for now is to do nothing in the case of child profiles, assuming we have none currently in prod. If we do, however, find some duplicate canonical profiles that do have children, we should probably log something so we know we have more work to do.

app/services/duplicate_canonical_profile_resolver.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# Class that deduplicates canonical profiles with identical benchmark/ref IDs
+class DuplicateCanonicalProfileResolver
+  class << self
+    def run!
+      duplicates.each do |duplicate|
+        profiles = Profile.canonical.includes(:child_profiles).where(
+          duplicate.slice(:ref_id, :benchmark_id)
+        ).order(created_at: :asc).offset(1)
```
akofink, line 10: I'd prefer an explicit `where.not(id: duplicate)` instead of relying on the offset(1). I guess the idea here is that at least one of these profiles will survive deletion, even if it's not `duplicate`, and that's ok. Still, it seems less deterministic than explicitly excluding `duplicate`.

akofink (commented):
Looks great, thanks! Just a couple minor things. Also, this work isn't really complete (fixed and stay fixed) until we add the unique partial index and call this duplicate resolver from within a migration. Any reason to hold off adding the migration in this change?

app/services/duplicate_canonical_profile_resolver.rb
skateman: How should I do logging from a migration? :thinking:

app/services/duplicate_canonical_profile_resolver.rb
skateman: The question is, which one should I preserve? My idea was that the oldest one :thinking:

akofink: The oldest one works, but **it doesn't really matter which one we keep** if none have children. Of course, if one does have children, we'd prefer to keep that one.

app/services/duplicate_canonical_profile_resolver.rb
akofink: You should be able to use the rails logger like normal: `Rails.logger.error(...)` It should go to stdout in our migrations.

app/services/duplicate_canonical_profile_resolver.rb
skateman: Then I don't understand your `where.not(id: duplicate)`, that would make us do +1 query as we need to fetch the ID of that "duplicate" first :confused:

akofink: The `duplicate` record is already fetched and in memory in your `duplicates.each`. Accessing the ID or calling `duplicates.slice(...)` accesses the query result already stored in memory and does not introduce any additional queries.

app/services/duplicate_canonical_profile_resolver.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+# Class that deduplicates canonical profiles with identical benchmark/ref IDs
+class DuplicateCanonicalProfileResolver
+  class << self
+    def run!
+      duplicates.each do |duplicate|
```
akofink, line 7: Pretty much always use `find_each`. It batches by 1000 by default so we don't run out of memory building and storing 1000s of profile objects in memory.

```suggestion
      duplicates.find_each do |duplicate|
```

db/migrate/20210316092758_deduplicate_canonical_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class DeduplicateCanonicalProfiles < ActiveRecord::Migration[5.2]
+  def up
+    DuplicateCanonicalProfileResolver.run!
+  end
```
akofink, line 4: In the same migration (same transaction), we want to add a partial index that enforces canonical profile uniqueness. If it's in a different migration, there is some small chance that more duplicates are inserted before our index is created.

app/services/duplicate_canonical_profile_resolver.rb
skateman: Oh, but this way we're just pushing the non-deterministic thing elsewhere, because the ID that we get there will be also non-deterministic, right?

skateman: Also I can't ask for the ID in the select because that would mess up the group_by :confused:

akofink: Ok yeah, both good points :laughing:

db/migrate/20210316092758_deduplicate_canonical_profiles.rb
skateman: We can only do this if we fail the migration if there are child profiles under a duplicate one :thinking:

akofink: Yes, though the index will fail to be created if that happens, and the transaction will be rolled back, so I don't think we need to add any special code to handle that case.

skateman: Yeah, but would not that break the migration chain and not allow us to write any other migration afterwards? Or is that the goal? :see_no_evil:

akofink: Yes, it will break the migrations and prevent us from deploying. If this migration fails, we'll have to revisit `DuplicateCanonicalProfileResolver` to reassign children profiles. Optionally, we could go ahead and handle that case now, though it's _probably_ unnecessary (up to you). We could also run the SQL query against prod again before we deploy this to ensure no children profiles exist. @vkrizan any input about how you think we should proceed?

akofink (approved):
The resolver and new partial index work as expected. Thanks, @skateman!

akofink:
To get tests passing, you may need to remove the old `DuplicateProfileResolver` and associated tests. There's no reason we need to keep them around if they're causing issues (which they are with this new index). I'm still looking into the rubocop issue and why its result is different in CI than locally.

akofink:
Looks like profile.rb is too long. Make sure to pull the latest changes from master in order to see it locally.

db/migrate/20210316092758_deduplicate_canonical_profiles.rb
vkrizan: If the whole migration chain isn't rollbacked on any error within the chain, then having it within one migration step/file is our best option.

db/migrate/20210316092758_deduplicate_canonical_profiles.rb
```diff
@@ -0,0 +1,10 @@
+class DeduplicateCanonicalProfiles < ActiveRecord::Migration[5.2]
+  def up
+    DuplicateCanonicalProfileResolver.run!
```
vkrizan, line 3: Would it be possible to log how many were deleted?

vkrizan (approved):
Thank you

app/services/duplicate_canonical_profile_resolver.rb
```diff
@@ -0,0 +1,54 @@
...
+    def no_child_profiles?(profiles)
+      profiles.all? { |p| p.child_profiles.count.zero? }
+    end
+
+    def duplicate_success_msg(duplicate)
+      Rails.logger.info(%(
+        Deleted #{duplicate.count} duplicates of policy with
+        ref_id=#{duplicate.ref_id} and benchmark_id=#{duplicate.benchmark_id}.
+      ))
+    end
```
akofink, line 43: Shouldn't this be `s/duplicate/profiles/`? `duplicate.count` calls `count` on a profile vs. an ActiveRecord::Relation (i.e. `profiles`)

akofink: Oh! gotcha, I see now :)

app/services/duplicate_canonical_profile_resolver.rb
```diff
@@ -0,0 +1,54 @@
...
+      ).having('count(id) > 1')
+    end
+
+    def no_child_profiles?(profiles)
+      profiles.all? { |p| p.child_profiles.count.zero? }
+    end
+
+    def duplicate_success_msg(duplicate)
+      Rails.logger.info(%(
+        Deleted #{duplicate.count - 1} duplicates of policy with
```
akofink, line 40: `destroy_all` returns the records it destroyed. We usually rely on that to determine how many were deleted, but this would seem to match that count. It's only ever going to run once, so I'm not asking for a change here if @vkrizan is also OK with it.

vkrizan: Using `duplicate` seems reasonable, considering the error message printed prior to destroy.

@skateman You might want to check whether general logging is printed out on a migration.

skateman: Yeah, in the migration context there's `say` and `write` which isn't available from the context of the service. So I'm going with `warn` and `STDOUT.puts` instead as they migrations rely on these anyway and rubocop likes them too.

akofink (approved):
Let's merge it! :) Nice work, @skateman.

PR #787, opened 2021-03-15
feat(COMP-E-133): RHICOMPL-1520 add a feature flag

akofink:
Currently the flag only disables setting the os_minor_version on a
policy, and it also disables the os_minor_version attribute from being
serialized in our REST API.

Please test with https://github.com/RedHatInsights/insights-core/pull/2971

Later, we'll remove this flag, and it should be quite trivial to find the locations necessary by searching by the flag name, feature_comp_e_133 in our code and tests.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -21,3 +21,4 @@ redis_cache_ssl: false
 slack_webhook: 'this is set through a env var in Openshift and not shared for security reasons'
 disable_rbac: 'false'
 async: true
+feature_133_os_tailoring: false
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -51,7 +51,7 @@ class Policy < ApplicationRecord
 
     removed = policy_hosts.where.not(host_id: new_host_ids).destroy_all
     imported = PolicyHost.import_from_policy(id, new_host_ids - host_ids)
-    update_os_minor_versions
+    update_os_minor_versions if Settings.feature_133_os_tailoring
 
     [imported.ids.count, removed.count]
   end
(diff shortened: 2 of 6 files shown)


config/settings.yml
```diff
@@ -21,3 +21,4 @@ redis_cache_ssl: false
 slack_webhook: 'this is set through a env var in Openshift and not shared for security reasons'
 disable_rbac: 'false'
 async: true
+feature_comp_e_133: false
```
akofink, line 24: I opted this time for a top level feature flag vs.

```yaml
features:
  comp_e_133: false
```

so it's easier to toggle with an env variable, SETTINGS__FEATURE_COMP_E_133

skateman:
Do we really want to name it like this? :thinking:

akofink:
@skateman what's wrong with the name? Any other suggestions? I could go for the shorter `feature_133`.

skateman:
@akofink that it's not descriptive :see_no_evil: I'd expect something like `feature_os_tailoring`

akofink:
I'm willing to change it to anything, tbh. The flag is going to be here for only a couple weeks before we remove it. Though, I think that if we ever did have many feature flags at the same time, including the Aha feature number would make it way easier to see which features were enabled than having only english-like descriptors of the features. What about `feature_133_os_tailoring` or similar?

akofink:
Updated to `feature_133_os_tailoring`.

vkrizan (approved):
Looks good. Seems to work as expected. Thanks!

skateman (approved):
The Seal of Approval

PR #788, opened 2021-03-16
chore(tests): remove deprecation warning from profiles controller

skateman:
Another nil-to-nil assert comparison cleaned up :broom:


akofink (approved):
Thanks!

PR #792, opened 2021-03-18
feat(Benchmark): RHICOMPL-1527 add supported_os_minor_versions attribute on APIs

vkrizan:
Adds `supported_os_minor_versions` to Benchmark APIs.

Filtering is on the way...

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -15,25 +15,34 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
     ".content_benchmark_#{OS_NAME}-#{os_major_version}"
   end
 
+  def comparable_version
+    Gem::Version.new(version)
+  end
+
   class << self
     def supported?(ssg_version:, os_major_version:, os_minor_version:)
       ssg_versions_for_os(os_major_version, os_minor_version)
         .include?(ssg_version)
     end
 
-    def ssg_versions_for_os(os_major_version, os_minor_version)
+    def for_os(os_major_version, os_minor_version)
       os_major_version = os_major_version.to_s
       os_minor_version = os_minor_version.to_s
 
       all.select do |ssg|
         ssg.os_major_version == os_major_version &&
           ssg.os_minor_version == os_minor_version
-      end.map(&:version)
+      end
+    end
+
+    def ssg_versions_for_os(os_major_version, os_minor_version)
+      for_os(os_major_version, os_minor_version).map(&:version)
     end
 
     def latest_ssg_version_for_os(os_major_version, os_minor_version)
-      ssg_versions_for_os(os_major_version, os_minor_version)
-        .max_by { |ssg_v| Gem::Version.new(ssg_v) }
+      for_os(os_major_version, os_minor_version)
+        .max_by(&:comparable_version)
+        .version
     end
 
     def versions
@@ -67,7 +76,7 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
 
     def latest_per_os_major
       all.group_by(&:os_major_version).values.map do |ssgs|
-        ssgs.max_by { |ssg| Gem::Version.new(ssg.version) }
+        ssgs.max_by(&:comparable_version)
       end
     end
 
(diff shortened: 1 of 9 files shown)


app/models/supported_ssg.rb
```diff
@@ -15,25 +15,34 @@ def ref_id
     ".content_benchmark_#{OS_NAME}-#{os_major_version}"
   end
 
+  def comparable_version
```
skateman, line 18: If you're introducing this helper method, we probably want to use it across the repo where `Gem::Version˙ is being already used.

app/models/supported_ssg.rb
```diff
@@ -74,9 +74,28 @@ def available_upstream
...
+             major_ssgs
+               .group_by(&:os_minor_version)
+               .transform_values do |minor_ssgs|
+                 minor_ssgs.max_by(&:comparable_version)
+               end.freeze
+           end.freeze
+      end
+    end
+
+    def latest_map!
```
skateman, line 91: I'd rather have this as:
```ruby
def latest_map
  @latest_map ||= build_latest_map
end

def build_latest_map
  ...
end
```

skateman (changes requested):

app/models/supported_ssg.rb
vkrizan: This is mostly to use a shortened version for mapping and sorting.

skateman: Yeah, if you do a fulltext search, it's used on multiple places

app/models/supported_ssg.rb
vkrizan: Updated. I've kept the `latest_map!`.

skateman: :+1:

skateman: Wait, are you using the `latest_map!` only in tests? Because the purpose of `build_latest_map` would be to call that one in the tests :thinking:

app/models/supported_ssg.rb
vkrizan: Do you have a recommendation where to move it?

skateman: I was aiming towards `ApplicationRecord` but now I see that the `SupportedSsg` is a `Struct` :confused: having a separate module that we include in both sounds like an overkill. So I guess it can stay as it is ... sorry.

app/models/supported_ssg.rb
vkrizan: Yes, only used in tests atm. There are some integration tests that need to clear the cached value (instance variable). It can be `clear_latest_map!` (which I guess its the same) or the tests can mangle with the class instance variable.

skateman: Oh, you can just do `SupportedSsg.instance_variable_set(:@latest_map, nil)` which would clear the thing from the test and then you don't need any extra method.

vkrizan: Yep, I've used it before I've decided to do a public method that would also clear it. Ok, I'll roll it back...

skateman (approved):
The Seal of Approval

app/models/supported_ssg.rb
```diff
@@ -87,6 +101,16 @@ def raw_supported
...
     end
+
+    def build_latest_map
+      all.group_by(&:os_major_version).transform_values do |major_ssgs|
+        major_ssgs
+          .group_by(&:os_minor_version)
+          .transform_values do |minor_ssgs|
+            minor_ssgs.max_by(&:comparable_version)
+          end.freeze
+      end.freeze
```
akofink, line 112: Could you explain the intentions behind `freeze` here? I thought it essentially makes an object immutable, which would be great, but this method will return a new object on every call. Maybe we should memoize it with `||=`?

akofink (approved):
Awesome! Works as expected, and the code looks great. Just one minor question.

app/models/supported_ssg.rb
vkrizan: This is memoized and accessible within `latest_map` method. The freeze is exactly to prevent any modifications to cached/saved objects.

PR #793, opened 2021-03-19
feat(Benchmark): RHICOMPL-1538 searching by supported_os_minor_versions

vkrizan:
Adds scope `supported_os_minor_versions` for `Benchmark` model.

Adds scoped search fitlering `supported_os_minor_version`.
Examples:
```
supported_os_minor_version = 9
supported_os_minor_version ^ (9, 10)
os_major_version = 7 and supported_os_minor_version ^ (8, 9)
```

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/models/concerns/benchmark_searching.rb
+++ b/app/models/concerns/benchmark_searching.rb
@@ -19,6 +19,12 @@ module BenchmarkSearching
       where(os_major_version_query(major, equals))
     }
 
+    scope :supported_os_minor_versions, lambda { |minor_versions|
+      supported_os_minor_clauses(minor_versions).reduce(none) do |ors, clause|
+        ors.or(where(*clause))
+      end
+    }
+
     scope :latest_supported, lambda {
       SupportedSsg.latest_per_os_major.inject(none) do |supported, ssg|
         supported.or(
@@ -45,6 +51,22 @@ module BenchmarkSearching
       equals = operator == '=' ? ' ' : ' NOT '
       { conditions: "ref_id#{equals}like ?",
         parameter: [os_major_version_like_condition(value)] }
+    end
+
+    def supported_os_minor_clauses(minor_versions)
+      minor_versions = [minor_versions].flatten.map(&:to_s)
+
+      SupportedSsg.latest_map.map do |major, major_ssgs|
+        ssg_versions = minor_versions.map do |minor_version|
+          major_ssgs[minor_version]&.version
+        end.compact.uniq
+
+        next if ssg_versions.count.zero?
+
+        ['benchmarks.ref_id LIKE ? AND benchmarks.version IN (?)',
+         os_major_version_like_condition(major),
+         ssg_versions]
+      end.compact
     end
 
     def latest
--- a/app/models/concerns/benchmark_searching.rb
+++ b/app/models/concerns/benchmark_searching.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+# Holder of scopes and scoped_search
+module BenchmarkSearching
+  extend ActiveSupport::Concern
+
+  included do
+    scoped_search on: %i[id ref_id title version]
+    scoped_search relation: :profiles, on: :id, rename: :profile_ids,
+                  aliases: %i[profile_id]
+    scoped_search relation: :rules, on: :id, rename: :rule_ids,
+                  aliases: %i[rule_id]
+    scoped_search on: :os_major_version,
+                  ext_method: 'os_major_version_search',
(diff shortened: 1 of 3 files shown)


app/models/xccdf/benchmark_searching.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
```
skateman, line 1: I think the right location for this file should be:
```
app/models/concerns/xccdf/benchmark_searching.rb
```
Or maybe:
```
app/models/xccdf/concerns/benchmark_searching.rb
```
But I think the first one is the more rails-y

vkrizan: The latter did not worked. The former can do it maybe...

akofink: I'm ok to have this outside of the Xccdf module. One day, we'll probably remove benchmark.rb from Xccdf as well. It was intended to show our models that have a 1 to 1 counterpart in the SCAP model, but we never actually followed up with it for rules, profiles, test results, etc., and I don't really see why it's useful anyway. To use it within benchmark.rb, all we'd need is the root namespace, `include ::BenchmarkSearching`.

app/models/xccdf/benchmark_searching.rb
```diff
@@ -20,6 +20,12 @@ module BenchmarkSearching
         where(os_major_version_query(major, equals))
       }
 
+      scope :supported_os_minor_versions, lambda { |minor_versions|
+        supported_os_minor_clauses(minor_versions).reduce(none) do |ors, clause|
+          ors.or(where(*clause))
+        end
+      }
```
akofink, line 31: Seems to work as expected :+1: 

```rb
[8] pry(main)> Xccdf::Benchmark.os_major_version(7).supported_os_minor_versions([7, 8, 9]).count
   (0.5ms)  SELECT COUNT(*) FROM "benchmarks" WHERE "benchmarks"."ref_id" ILIKE '%RHEL-7' AND ((benchmarks.ref_id LIKE '%RHEL-6' AND benchmarks.version IN ('0.1.21','0.1.28','0.1.28')) OR (benchmarks.ref_id LIKE '%RHEL-7' AND benchmarks.version IN ('0.1.43','0.1.46','0.1.52')))
```

app/models/xccdf/benchmark_searching.rb
```diff
@@ -48,6 +54,22 @@ def os_major_version_search(_filter, operator, value)
           parameter: [os_major_version_like_condition(value)] }
       end
 
+      def supported_os_minor_clauses(minor_versions)
+        minor_versions = [minor_versions].flatten.map(&:to_s)
+
+        SupportedSsg.latest_map.map do |major, major_ssgs|
+          ssg_versions = minor_versions.map do |minor_version|
+            major_ssgs[minor_version]&.version
+          end.compact
```
akofink, line 67: Nit: we're sending non-unique minor versions in some cases. It doesn't affect the query result, but we might want to add a `.uniq` here.

```rb
[8] pry(main)> Xccdf::Benchmark.os_major_version(7).supported_os_minor_versions([7, 8, 9]).count
   (0.5ms)  SELECT COUNT(*) FROM "benchmarks" WHERE "benchmarks"."ref_id" ILIKE '%RHEL-7' AND ((benchmarks.ref_id LIKE '%RHEL-6' AND benchmarks.version IN ('0.1.21','0.1.28','0.1.28')) OR (benchmarks.ref_id LIKE '%RHEL-7' AND benchmarks.version IN ('0.1.43','0.1.46','0.1.52')))
```

The `benchmarks.version IN ('0.1.21','0.1.28','0.1.28')` part.

akofink (commented):
Seems to work great! Nice work getting this to just a single SQL query. The only issue is the included benchmark_searching concern, which we should be able to fix by just not having it in the Xccdf module. Thanks, @vkrizan!

app/models/xccdf/benchmark_searching.rb
vkrizan: :+1:

app/models/xccdf/benchmark_searching.rb
vkrizan: Moved to top model-level concern.

akofink (approved):
Thanks, @vkrizan!

PR #795, opened 2021-03-22
fix(Prometheus): RHICOMPL-1559 Scope metrics to compliance hosts

akofink:
Joins across the entire hosts view can get very sluggish. We always want
to only count compliance hosts in our metrics. The sluggishness is caused, in part, by the lack of indices, etc. which will be addressed in future PRs. This PR just aims to limit the scope of our queries to compliance hosts only.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/lib/prometheus/engineering_collector.rb
+++ b/lib/prometheus/engineering_collector.rb
@@ -28,13 +28,16 @@ class EngineeringCollector < PrometheusExporter::Server::TypeCollector
   end
 
   def dangling(model)
-    model.where.not(host_id: Host.select(:id))
+    model.where.not(host_id: Host.with_policies_or_test_results.select(:id))
+  end
+
+  def dangling_accounts
+    Account.where.not(account_number: Host.with_policies_or_test_results
+                                          .select(:account).distinct)
   end
 
   def collect
-    @dangling_accounts.observe(
-      Account.where.not(account_number: Host.select(:account)).count
-    )
+    @dangling_accounts.observe(dangling_accounts.count)
     @dangling_test_results.observe(dangling(TestResult).count)
     @dangling_rule_results.observe(dangling(RuleResult).count)
     @dangling_policy_hosts.observe(dangling(PolicyHost).count)
--- a/app/models/concerns/host_searching.rb
+++ b/app/models/concerns/host_searching.rb
@@ -32,6 +32,11 @@ module HostSearching
         where.not(id: ::PolicyHost.select(:host_id))
     }
 
+    scope :with_test_results, lambda { |with_test_results = true|
+      with_test_results && where(id: ::TestResult.select(:host_id)) ||
+        where.not(id: ::TestResult.select(:host_id))
+    }
+
     scope :os_major_version, lambda { |version, equal = true|
       condition = ['system_profile @> ?',
                    { operating_system: { major: version.to_i } }.to_json]
@@ -110,12 +115,11 @@ module HostSearching
     end
 
     def test_results?(_filter, _operator, value)
-      operator = ::ActiveModel::Type::Boolean.new.cast(value) ? '' : 'NOT'
-      host_ids = ::TestResult.select(:host_id).distinct.where.not(host_id: nil)
-      {
-        conditions: "hosts.id #{operator} "\
-                    "IN(#{host_ids.to_sql})"
-      }
+      hosts = ::Host.with_test_results(
+        ::ActiveModel::Type::Boolean.new.cast(value)
+      )
+
+      { conditions: hosts.arel.where_sql.gsub(/^where /i, '') }
     end
 
     private
--- a/app/models/concerns/host_searching.rb
+++ b/app/models/concerns/host_searching.rb
@@ -37,6 +37,10 @@ module HostSearching
         where.not(id: ::TestResult.select(:host_id))
     }
 
+    scope :with_policies_or_test_results, lambda {
+      with_policy.or(with_test_results)
+    }
+
     scope :os_major_version, lambda { |version, equal = true|
       condition = ['system_profile @> ?',
                    { operating_system: { major: version.to_i } }.to_json]
--- a/lib/prometheus/business_collector.rb
+++ b/lib/prometheus/business_collector.rb
@@ -51,7 +51,7 @@ class BusinessCollector < PrometheusExporter::Server::TypeCollector
     )
     @client_accounts.observe client_accounts.count
     @client_accounts_with_hosts.observe(
-      Host.where(
+      Host.with_policies_or_test_results.where(
         account: client_accounts.select(:account_number)
       ).select(:account).distinct.count
     )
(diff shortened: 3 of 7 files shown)


lib/prometheus/engineering_collector.rb
```diff
@@ -28,12 +28,14 @@ def type
...
   def dangling(model)
-    model.where.not(host_id: Host.select(:id))
+    model.where.not(host_id: Host.with_policy_or_test_results.select(:id))
   end
 
   def collect
     @dangling_accounts.observe(
-      Account.where.not(account_number: Host.select(:account)).count
+      Account.where.not(
+        account_number: Host.with_policy_or_test_results.select(:account)
```
vkrizan, line 37: Josef had a suggestion to add a `DISTINCT` here. Maybe it can also speed it up a bit.

akofink: distinct hosts or accounts?

vkrizan:
```
Host.with_policy_or_test_results.select(:account).distinct
```

akofink: Updated

lib/prometheus/engineering_collector.rb
```diff
@@ -28,13 +28,16 @@ def type
...
-    model.where.not(host_id: Host.select(:id))
+    model.where.not(host_id: Host.with_policy_or_test_results.select(:id))
+  end
+
+  def dangling_accounts
+    Account.where.not(account_number: Host.with_policy_or_test_results
+                                          .select(:account).distinct)
   end
 
   def collect
```
vkrizan, line 39: Would it be possible to reduce the interval of collection of these expensive queries, or an introduction of some caching?

vkrizan (commented):
There are some incorrect method calls (typo in the name), other than it looks good. We would likely see the improvement in the environment with lots of data.

Would you also mind adding tests to metric collectors?

lib/prometheus/engineering_collector.rb
akofink: The real silver bullet here would be to add a read only instance of the DB for queries like these that cannot affect performance of prod for our users. The /metrics endpoint is polled; I'm not sure if we can configure the polling interval. If you see some opportunity for caching, please point it out, though the data collected here are changing constantly (kind of the point of metrics).

lib/prometheus/engineering_collector.rb
```diff
@@ -28,13 +28,16 @@ def type
   end
 
   def dangling(model)
-    model.where.not(host_id: Host.select(:id))
+    model.where.not(host_id: Host.with_policy_or_test_results.select(:id))
+  end
+
+  def dangling_accounts
+    Account.where.not(account_number: Host.with_policy_or_test_results
+                                          .select(:account).distinct)
```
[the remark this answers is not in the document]
akofink: Happy to remove the `distinct`. It doesn't affect the query result.

It looks like the total number of rows (435) doesn't change in your test. Be sure to test it with duplicate accounts as well as many hosts with the same account.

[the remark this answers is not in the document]
vkrizan: The row count isn't real, as I have only 2 accounts and a few hosts. I guess it's some estimation (?).

I'll leave the distinct choice at your discretion.

lib/prometheus/engineering_collector.rb
vkrizan: I guess you've meant to have a read-only copy of the database synced with the main one. That's also an option. For our purposes it might be an overkill.

For caching we can leverage Redis to cache it for some time period. We don't need to have immediate value available at all times. We would not miss out any spikes that might happen during the cache period, would we?

vkrizan (approved):

akofink:
Wait to merge. I've figured out how to test the collectors! :man_dancing:

lib/prometheus/engineering_collector.rb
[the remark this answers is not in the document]
akofink: I'll leave it - it makes sense that it would serve to minimize the join with our accounts table. We expect some accounts to have 1000s of hosts.

lib/prometheus/engineering_collector.rb
akofink: >I guess you've meant to have a read-only copy of the database synced with the main one. That's also an option. For our purposes it might be an overkill.

Yes, exactly. I believe it is supported directly from app-interface's managed resources (i.e. an out of the box solution for RDS replication). I agree it's overkill at this point considering the amount of traffic we have. With the appropriate index in place on accounts, even joining with 600k hosts should not bring RDS to its knees.

I have been thinking about this 50 second timeout, though. It would seem to make sense to decrease it to something reasonable like 5 seconds. If a query result doesn't return in 5 seconds, there is something seriously wrong with the state of our DB or with the query itself. In any case, polling seems to happen (from prometheus /metrics logs) every 5 seconds, so a timeout greater than that makes little sense:

```
10.130.2.1 - - [24/Mar/2021:15:30:05 UTC] "GET /metrics HTTP/1.1" 200 32640
- -> /metrics
10.130.20.31 - - [24/Mar/2021:15:30:09 UTC] "GET /metrics HTTP/1.1" 200 4435
- -> /metrics
10.130.2.1 - - [24/Mar/2021:15:30:14 UTC] "GET /metrics HTTP/1.1" 200 32640
- -> /metrics
10.129.4.104 - - [24/Mar/2021:15:30:19 UTC] "GET /metrics HTTP/1.1" 200 4433
- -> /metrics
```

akofink: btw, I'm curious if you think a smaller timeout would have saved the DB from being overloaded. If the caller times out, I guess it would depend on if some message is sent to the DB to cancel the pending query or not. I'm not sure how the timeout works.

akofink:
Tests added in a new commit. Please take a look - the build is currently failing due to https://github.com/rails/rails/issues/41750

lib/prometheus/engineering_collector.rb
vkrizan: I'm not sure if the timeout on the exporter would help us much with what has happened. I've also believed that the queries are being terminated if the caller closes the connection, but from what we've seen the queries kept hanging in RDS. I could have been due to several factors like the exporter pods being terminated and/or the use of `dblink`.

Anyhow, a lower timout would definitely help with some cases. At least if there would be several long running queries, it could stop at the first one.

Does the exporter run on a single worker/thread, because if so, the longer timeout on collection could hold/timeout requests coming in the interim.

akofink: >Does the exporter run on a single worker/thread, because if so, the longer timeout on collection could hold/timeout requests coming in the interim.

Yes, it appears it's single threaded from my tests locally. I also don't see any option for setting more threads/workers.

>I've also believed that the queries are being terminated if the caller closes the connection, but from what we've seen the queries kept hanging in RDS.

There may be different behavior on timeout vs. if the pod is killed. I'd expect the timeout to clean up more, but the pod getting killed almost certainly cannot clean up any open connections. Also, the timeout should still use the same connection to the DB rather than creating a new pod with a new connection and leaving the old connection dangling. There is a configurable maximum number of DB connections available (50 in prod, I think?).

vkrizan:
@akofink looks great! Thank you.

PR #801, opened 2021-03-29
fix(reports): RHICOMPL-1173 compensate wrong benchmark versions

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -75,7 +75,13 @@ class XccdfReportParser
   end
 
   def save_missing_supported_benchmark
-    save_all_benchmark_info if SupportedSsg.versions.include?(benchmark.version)
+    save_all_benchmark_info if SupportedSsg.all.find do |ssg|
+      # There's a version mismatch in some SSGs that has to be compensated
+      if ssg.real_version.present? && ssg.real_version.to_s == benchmark.version
+        benchmark.version = ssg.version
+      end
+      ssg.version == benchmark.version
+    end
   end
 
   def check_for_missing_benchmark
--- a/config/supported_ssg.yaml
+++ b/config/supported_ssg.yaml
@@ -82,6 +82,7 @@ supported:
   RHEL-7.2:
     - package: scap-security-guide-0.1.25-3.el7
       version: 0.1.25
+      real_version: 0.9
       profiles:
         common:
         pci-dss:
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -4,7 +4,7 @@ require 'yaml'
 
 # rubocop:disable Metrics/BlockLength
 SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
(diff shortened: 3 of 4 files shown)


app/services/xccdf_report_parser.rb
```diff
@@ -75,7 +75,13 @@ def check_for_external_reports
   end
 
   def save_missing_supported_benchmark
-    save_all_benchmark_info if SupportedSsg.versions.include?(benchmark.version)
+    save_all_benchmark_info if SupportedSsg.all.find do |ssg|
+      # There's a version mismatch in some SSGs that has to be compensated
+      if ssg.real_version.present? && ssg.real_version.to_s == benchmark.version
+        benchmark.version = ssg.version
```
skateman, line 81: I'm not entirely sure if this is the right thing to do, but this seemed logical to me.

skateman: I don't really understand :confused: in the corner case the `benchmark.version` will be `0.9` and the 7.2 SSG doesn't have an `upstream_version` and its `version` is the correct `0.1.25` which would never match :thinking: or am I missing something?

akofink: You seem to understand the corner case: it's when `benchmark.version` doesn't match `upstream_version` (if available) or `version`. In that corner case, we want to overwrite the incorrect (0.9) `benchmark.version`, which is parsed from the report.

config/supported_ssg.yaml
```diff
@@ -82,6 +82,7 @@ supported:
   RHEL-7.2:
     - package: scap-security-guide-0.1.25-3.el7
       version: 0.1.25
+      real_version: 0.9
```
akofink, line 85: Ok, now revisiting this, I think this is a fine way to try and tackle this issue; you can disregard my earlier comments. One thing we should also change is to treat reports coming in against version 0.9 as 0.1.25 as well. I suspect we will only ever have this one version that's problematic, and luckily `0.9` doesn't conflict with any other SSG versions.

skateman: Wait, isn't my code already doing that? the `save_missing_supported_benchmark` is IN the report parser :thinking:

akofink: Hmm... maybe you're right! I guess that's the only place the SSG version is even referenced. :laughing: Ok, I'm going to test this so we can get it merged :+1:

akofink:
As discussed in slack, the 0.1.25 SSG is not the only culprit reporting as 0.9. :confounded: 

0.1.18 (RHEL 6.6)
0.1.19 (RHEL 7.1)
0.1.21 (RHEL 6.7)
0.1.25 (RHEL 7.2)

We'll have to think a bit more how we could work around this.

PR #815, opened 2021-04-22
WIP preparing entrypoint and Dockerfile candidates

Victoremepunto:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


new-entrypoint.sh
```diff
@@ -0,0 +1,58 @@
...
+	if [[ ! "$2" =~ ^[0-9]+$ ]]; then
+		echo "$1 needs to be a non-negative number"
+		exit 1
+	fi
+}
+
+if [ "$APPLICATION_TYPE" = "compliance-backend" ]; then
+	check_number PUMA_WORKERS     "${PUMA_WORKERS:-0}"
+	check_number PUMA_MIN_THREADS "${PUMA_MIN_THREADS:-0}"
+	check_number PUMA_MAX_THREADS "${PUMA_MAX_THREADS:-0}"
```
akofink, line 20: This syntax means the default is `0` right? That doesn't seem appropriate :thinking: Maybe 1 for all of them?

Victoremepunto: I grabbed this from [here](https://github.com/RedHatInsights/s2i-insights-compliance/blob/82d848378312103a3e1836cc788049acdf91820a/s2i/bin/run) , which at the same time looks "inspired" by [this guy](https://github.com/sclorg/s2i-ruby-container/blob/master/2.7/s2i/bin/run#L16-L18)

> This syntax means the default is `0` right?

yup. From `man bash`:
```
${parameter:-word}
    Use Default Values.  If parameter is unset or null, the expansion of word is substituted.  Otherwise, the value of parameter is substituted.
```
I don't know what effect would be to have 0.

Victoremepunto: IMO it's okay to accept a 0 , regardless if you find that unlikely or useless - it's a valid value after all - I'd keep it as it is for consistency with the existing reference scripts

akofink: I've confirmed that trying to run puma with `0` for these three values works fine and actually uses 5 threads (the default) and 1 worker.

new.Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM registry.access.redhat.com/ubi8/ruby-27 as build
+
+# Install dependencies and clean cache to make the image cleaner
+
+USER 0
+RUN rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
+    yum install -y hostname shared-mime-info && \
+    yum clean all -y
+
+COPY . /tmp/src
```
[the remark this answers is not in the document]
Victoremepunto: yes that is the expected behavior if we're going to use the tools that come from `S2i` image (the UBI8 one I mean). If we're not, we can just bundle install ourselves, define our own workdir, etc... ( see [how to use the Ubi8/Ruby27 image](https://github.com/sclorg/s2i-ruby-container/blob/master/2.7/README.md#3-prepare-an-application-inside-a-container) , 3.1 is what I'm doing to be as less disruptive as possible, 3.2 is what we could do if we don't wanna leverage anything from the image's baked in tools. ). 

I guess we could explore 3.2 as a second step IMO.

new.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN chown -R 1001:0 /tmp/src
+
+USER 1001
+#WORKDIR /tmp/src
+
+ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true
+
+# Install the dependencies
+RUN /usr/libexec/s2i/assemble
```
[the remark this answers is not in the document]
Victoremepunto: see https://github.com/RedHatInsights/compliance-backend/pull/815#discussion_r619088742

new-entrypoint.sh
```diff
@@ -0,0 +1,58 @@
...
+			if [ -f Gemfile ]; then
+				exec bundle exec "rackup -E ${RAILS_ENV:-$RACK_ENV} -P /tmp/rack.pid --host 0.0.0.0 --port 8080"
+			else
+				exec rackup -E "${RAILS_ENV:-$RACK_ENV}" -P /tmp/rack.pid --host 0.0.0.0 --port 8080
+			fi
+		else
+			echo "ERROR: Rubygem Rack is not installed in the present image."
+			echo "       Add rack to your Gemfile in order to start the web server."
+		fi
+	fi
```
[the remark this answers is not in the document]
Victoremepunto: this is because of https://github.com/RedHatInsights/compliance-backend/pull/815#discussion_r619088742 and because we have the "application name conditional". it's essentially the same `run`  script from the `S2i image` , with the "if -else" structure for the application name logic.

I'd say, let's keep it as less-disruptive as possible, then we can work on simplify the Dockerfile logic (entrypoint, etc)

build_deploy.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -exv
+
+IMAGE_NAME="quay.io/cloudservices/compliance-backend"
+IMAGE_TAG=$(git rev-parse --short=7 HEAD)
```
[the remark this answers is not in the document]
Victoremepunto: This comes from the Quay tagging for AppSRE

new-entrypoint.sh
Victoremepunto: Awesome, I guess then this actually makes sense, which is not a surprise.

Victoremepunto:
> Seems fine, thanks @Victoremepunto! I've made some comments about changing some of what you copied directly from the s2i repo. We can just ignore those for now so we don't introduce any issues; one change at a time!
> 
> I think we could probably organize this a bit better; I'm not fond of `new-entrypoint.sh` and `new.Dockerfile`. Maybe something showing that this is the deployed container? Put it in a directory, maybe, like `deploy/`? I'm not really sure laughing

Yes, actually I didn't know how this would result, IMO ideally this should replace the `entrypoint` and the `Dockerfile`. the `new` prefix thing was not going to make it in any way. I believe it makes sense that they co-exist for a while, so I'll move them under a subdir, it makes sense.

akofink:
Good to merge. This doesn't affect anything else; we can tweak it further if needed in future PRs.

build_deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+if [[ ! ( -z "$QUAY_USER" || -z "$QUAY_TOKEN" ) ]]; then
+    echo "QUAY_USER and QUAY_TOKEN must be set"
+    exit 1
+fi
+
+DOCKER_CONF="$PWD/.docker"
+mkdir -p "$DOCKER_CONF"
+
+#remove before merging
+podman build -f new.Dockerfile -t "${IMAGE_NAME}:${IMAGE_TAG}" .
```
akofink, line 19: Oh, this needs to be updated

deploy/Dockerfile
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
vkrizan, line 1: Is this supposed to be a `Dockerfile` or the `entry.sh`? (check filename)

build_deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Replace before merging
+#if [[ -z "$QUAY_USER" || -z "$QUAY_TOKEN" ]]; then
+if [[ ! ( -z "$QUAY_USER" || -z "$QUAY_TOKEN" ) ]]; then
+    echo "QUAY_USER and QUAY_TOKEN must be set"
+    exit 1
+fi
+
+DOCKER_CONF="$PWD/.docker"
+mkdir -p "$DOCKER_CONF"
```
vkrizan, line 16: This looks like a temporary directory. `mktemp -d` might be used here.

deploy/Dockerfile
Victoremepunto: Sorry, you're right, I've pushed the right files / names

build_deploy.sh
Victoremepunto: mmm , not really sure we're going to get an advantage from using `mktemp` as it will create a randomly named directory (see https://ss64.com/bash/mktemp.html) but won't remove it or anything. If anything we'll get a new directory every time we'll have to remove manually.

Besides, this line comes straightly from [this example](https://internal.cloud.redhat.com/docs/devprod/tutorial/03-building/#integrating-into-appsres-jenkins) - I used the script there as a template to create our scripts, not really sure if we can use a different name scheme at all, so I'd say it's okay to keep the `mkdir -p` approach.

build_deploy.sh
Victoremepunto: Yes I'll remove all of this before removing the "WIP" prefix

akofink (approved):
Looks good. Thanks, @Victoremepunto!

deploy/README.md
```diff
@@ -0,0 +1,9 @@
+# Deployment
+
+The Compliance application uses the [UBI8/ruby27](https://catalog.redhat.com/software/containers/ubi8/ruby-27/5f521b90acb94fdd904e98d7?gti-tabs=unauthenticated) base image. 
+
+You can create a new image using:
+
+```
+podman build -t compliance-backend:latest .
```
vkrizan, line 8:
```
podman build -t compliance-backend:latest -f deploy/Dockerfile .
```
since the docker file is not in the root. I guess for Clowder (for builds within app-interface) it would have to be in the root.

vkrizan (approved):
Looks good.

deploy/README.md
akofink: Let's make the production Dockerfile the root :) https://github.com/RedHatInsights/compliance-backend/pull/836

akofink: Sure, both is fine with me. :+1:

clowdapp.yaml
```diff
@@ -0,0 +1,260 @@
...
+        readinessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /playbooks
+            port: web
+            scheme: HTTP
+          initialDelaySeconds: 35
+          periodSeconds: 5
+          successThreshold: 1
+          timeoutSeconds: 120
```
akofink, line 107: Can we just use defaults? Like `failureThreshold` is 3 by default already. What's the magic `35` second initialDelay? I think that one is 1 second by default. It should take no time at all to spin up nginx, right?

deploy/README.md
Victoremepunto: Done, moved current `entrypoint.sh` to `devel.entrypoint.sh` and renamed it's usage in `devel.Dockerfile`, moved everything from `deploy/` to root dir.

clowdapp.yaml
Victoremepunto: sure, I agree. I commented out most of the stuff - the template however is very on a very WIP stage, we need to update Compliance to be able to read from Clowder config for this template to work at all.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN chown -R 1001:0 /tmp/src
+
+USER 1001
+
+ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true QMAKE=/usr/lib64/qt5/bin/qmake DISABLE_ASSET_COMPILATION=true
+
+# Install the dependencies
+RUN /usr/libexec/s2i/assemble
+
+CMD ["/opt/app-root/src/deploy/entrypoint.sh"]
```
akofink, line 20: This is no longer correct:

```
/usr/bin/container-entrypoint: line 2: /opt/app-root/src/deploy/entrypoint.sh: No such file or directory
```

Victoremepunto: :facepalm:

devel.Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
 
 COPY vendor/ ./vendor
 COPY Gemfile* ./
-COPY entrypoint.sh ./
+COPY devel.entrypoint.sh ./entrypoint.sh
 
 RUN bundle -j4
```
akofink, line 14: Please update the `ENTRYPOINT` on line 15.

akofink: Sorry, nevermind! :smile: I see it's renamed when copied.

entrypoint.sh
```diff
@@ -1,7 +1,58 @@
...
+elif [ "$APPLICATION_TYPE" = "compliance-prometheus-exporter" ]; then
+	exec bundle exec prometheus_exporter -b 0.0.0.0 --prefix compliance_ -t 50 --verbose -a lib/prometheus/graphql_collector.rb -a lib/prometheus/business_collector.rb
+elif [ "$APPLICATION_TYPE" = "compliance-import-remediations" ]; then
+	exec bundle exec rake import_remediations --trace
+elif [ "$APPLICATION_TYPE" = "compliance-import-ssg" ]; then
+	exec bundle exec rake ssg:import_rhel_supported --trace
+elif [ "$APPLICATION_TYPE" = "compliance-inventory-sync" ]; then
+	exec bundle exec rake sync_with_inventory --trace
+elif [ "$APPLICATION_TYPE" = "compliance-import-host-os-releases" ]; then
+	exec bundle exec rake import_host_os_releases --trace
```
akofink, line 57: These two are no longer used in any environment.

devel.Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get update && apt-get install -y qt5-default libqt5webkit5-dev \
 
 COPY vendor/ ./vendor
 COPY Gemfile* ./
-COPY entrypoint.sh ./
+COPY devel.entrypoint.sh ./entrypoint.sh
```
akofink, line 11: Ok, I think I figured out what the issue is! Please just copy this directly to `./` because we volume mount the compliance-backend repo into this devel container, and then it runs the prod entrypoint.sh. You'll also need to update line 15, which specifies the ENTRYPOINT.

akofink (approved):
Other than the devel fix mentioned related to the entrypoint, it works as expected. Thanks, @Victoremepunto! Feel free to merge once the entrypoint comment is addressed.

entrypoint.sh
Victoremepunto: Removed, as well as `compliance-consumer` . I just used what we were using currently on our images, but you're right :) 

Hopefully we can deprecate this "method" soon enough.

devel.Dockerfile
Victoremepunto: okay :D

devel.Dockerfile
Victoremepunto: I guess this is not needed as it serves the purpose of properly renaming the entrypoint.

akofink: Again, copying this to `./entrypoint.sh` and setting the Dockerfile to use `./entrypoint.sh` will not work in devel because we volume mount the entire repo.

devel.Dockerfile
akofink: This change is required. Sorry for the confusion.

akofink (approved):
Works great! Thanks, @Victoremepunto!

PR #824, opened 2021-05-12
Introduce sorting for our APIs

skateman:
Both the GQL and REST APIs now accept a `sort_by` and `sortBy` attribute, which has to be an array of fields optionally with an `:asc` or `:desc` suffix (defaults to `:asc`) to sort by the results. I implemented a `SORTABLE_BY` hash on the `Profile` and `Host` models that contain all the allowed sorting fields as keys and their actual DB-mapped column/arel values as values. This allows us to sort entities based on nested properties, like the OS minor/major version for hosts. This should make the enablement of new sortable fields trivial...

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


akofink:
>what should happen if a field name is passed that is not sortable (is the default AR exception okay?)

>how should we validate the passed sorting direction for the REST API (is the default AR exception okay?)

Ideally, both these cases would result in some 4XX error, indicating the request was malformed with a helpful message to the user as to what they did wrong. The inventory API does a pretty good job of this:

```
$ http https://ci.foo.redhat.com:1337/api/inventory/v1/hosts order_by==fqdn 
HTTP/1.1 400 BAD REQUEST
connection: close
content-type: application/problem+json
date: Wed, 12 May 2021 12:54:02 GMT
server: gunicorn/19.9.0
transfer-encoding: chunked

{
    "detail": "'fqdn' is not one of ['display_name', 'updated']\n\nFailed validating 'enum' in schema:\n    {'enum': ['display_name', 'updated'], 'type': 'string'}\n\nOn instance:\n    'fqdn'",
    "status": 400,
    "title": "Bad Request",
    "type": "about:blank"
}
```

akofink:
> how should we validate the passed sorting direction for the REST API (is the default AR exception okay?)

I think you should be able to leverage StrongerParamters for this, using enum or regexp type: https://github.com/zendesk/stronger_parameters#types

We already use this to ensure our create and update API calls come in with the correct format: https://github.com/RedHatInsights/compliance-backend/blob/master/app/controllers/concerns/parameters.rb#L9

skateman:
@akofink not sure how to test the GraphQL, I didn't find any examples :thinking:

akofink:
Like this? https://github.com/RedHatInsights/compliance-backend/blob/master/test/graphql/queries/profile_query_test.rb

In the past, we've struggled to write functional tests for global functionality like pagination separate from a specific resource (i.e. profiles), since testing pagination or sorting must use _some_ resource. Still, the available attributes to sort by should be limited in each resource to a subset of the total columns in the database. i.e. for profiles, `name` makes sense, but probably not `description`.

app/controllers/concerns/collection.rb
```diff
@@ -0,0 +1,37 @@
...
+    def search(data)
+      return data if params[:search].blank?
+
+      data.search_for(params[:search])
+    end
+
+    def sort(data)
+      return data if params[:sort_by].blank?
+
+      by = params[:sort_by].each_with_object({}) do |item, obj|
```
vkrizan, line 26: This does not handle single param cases gracefully:
```
GET /api/compliance/v1/systems?sort_by=id

undefined method `each_with_object' for "id":String
```

app/controllers/concerns/collection.rb
```diff
@@ -0,0 +1,37 @@
...
+        column, direction = item.split(':')
+
+        validate_sorting(data.column_names, column, direction)
+
+        obj[column] = direction || 'asc'
+      end
+
+      data.order(by)
```
vkrizan, line 34: For the first iteration of sorting this is a good start. For frontend, however, we would need to add sorting by computed columns like `ssg_version` from Profiles and `os_major_version` with `os_minor_version` from Hosts.

vkrizan (commented):
So far so good.

For tests you can use existing GraphQL types or create some dummy ones on test runtime.

app/controllers/concerns/collection.rb
skateman: Yeah, that's kinda against the specifications, the `sort_by` should be always an array.

app/controllers/concerns/collection.rb
skateman: Not sure how much can we deal with that on the database level :confused: at least not in a nice & simple way.

app/controllers/concerns/collection.rb
vkrizan: Can it be handled gracefully please? Instead of a `500 Internal Server Error` return 4XX, and not raise an alarm.

Personally, I'd still handle it even w/o using `[]`. I guess that there is not even a standard for handling an array from query string.

skateman: :thinking: well, it's handled as any other invalid param, I don't see the point why this should get a special treatment. And yes, arrays in HTTP are a pretty standard thing.

akofink: It's ok to extend the spec. Ruby idiom for forcing an array: `[maybe_an_array].flatten`

skateman: My bad, the spec actually has the non-array version :see_no_evil: fixing

skateman: Done

akofink (commented):
I'm sorry, I did not have time to thoroughly review or test this yet, but it looks great! Nice work to make this generic to be used with REST or GraphQL and extensible for the future. :1st_place_medal:

I will take another look Monday if it's still not merged.

app/graphql/resolvers/concerns/collection.rb
```diff
@@ -6,11 +6,14 @@ module Concerns
...
       extend ActiveSupport::Concern
 
+      include Sorting
+
       included do
         type ["Types::#{self::MODEL_CLASS}".safe_constantize], null: false
         argument :search, String, 'Search query', required: false
         argument :limit, Integer, 'Pagination limit', required: false
         argument :offset, Integer, 'Pagination offset', required: false
+        argument :sort_by, [String], 'Sort results', required: false
```
akofink, line 16: You can define a [union type](https://graphql-ruby.org/type_definitions/unions#defining-union-types). Please add a `BaseUnion` type as mentioned in that doc and then a `StringOrArrayOfStrings` union type, used like:

```suggestion
        argument :sort_by, StringOrArrayOfStrings, 'Sort results', required: false
```

akofink (commented):
The code looks great, and it works as expected! There are some issues with the type definitions in the GraphQL and OpenAPI schemas, and you also need to update the openapi.json (sorry, terribly hacky):

```
podman exec -it compliance-backend_rails_1 /usr/local/bin/ruby -I/usr/local/bundle/gems/rspec-core-3.9.2/lib:/usr/local/bundle/gems/rspec-support-3.9.3/lib /usr/local/bundle/gems/rspec-core-3.9.2/exe/rspec --pattern spec/integration/\*\*/\*_spec.rb --format Rswag::Specs::SwaggerFormatter --order defined
```

app/graphql/resolvers/concerns/collection.rb
skateman: Nope, union doesn't allow primitive types https://github.com/rmosolgo/graphql-ruby/issues/530

skateman: @akofink how bad it is if we only allow arrays of strings? :see_no_evil:

akofink: For GraphQL, I'm not too worried about it, since we're just targeting our own UI. For REST, IMO it's a UX concern, but it looks like there's no problem with that implementation :+1:

swagger/v1/openapi.json
```diff
@@ -76,46 +105,48 @@
               "application/vnd.api+json": {
                 "data": [
                   {
-                    "id": "aa67c98c-d81f-5a9c-b0bc-26caa0051aea",
+                    "id": "9d9cb70a-024a-4152-9a26-1df1f400f392",
```
akofink, line 108: `git add -i swagger` is your friend :) There's a lot of good example responses here from before FactoryBot that are now either empty or show less relationships. I'd like to go back through the specs and ensure we have similar data there as before. Until then, though, there's no need to overwrite parts of the openapi.json that aren't affected by this sorting change.

akofink (approved):
Thank you, @skateman! Works great for me, and the code looks great, too :) :shipit:

PR #825, opened 2021-05-12
feat(REST): RHICOMPL-1563 status API endpoint

akofink:
Add a new status endpoint that returns 200 if everything is OK and 500
otherwise. We can expand the statuses if needed in the future.

It also works with HEAD requests if you're not interested in the individual json statuses.

```
GET compliance/v1/status

HTTP/1.1 200 OK

{
    "data": {
        "api": true
    }
}
```

```
GET compliance/v1/status

HTTP/1.1 500 Internal Server Error

{
    "data": {
        "api": false
    }
}
```

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/controllers/v1/statuses_controller.rb
+++ b/app/controllers/v1/statuses_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module V1
+  # API for Compliance Status
+  class StatusesController < ApplicationController
+    skip_around_action :authenticate_user
+
+    ARB = ActiveRecord::Base
+
+    def show
+      render json: { data: all_statuses }, status: status_return_code
+    end
+
+    private
+
+    def all_statuses
+      @all_statuses ||= {
+        api: api_status
+      }
+    end
+
+    def status_return_code
+      all_statuses.values.all? ? :ok : :internal_server_error
+    end
+
+    def api_status
+      ensure_db_connection
+      ARB.connection.active?
+    rescue PG::Error
+      false
+    end
+
+    def ensure_db_connection
+      ARB.establish_connection unless ARB.connection.active?
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,7 @@ Rails.application.routes.draw do
     scope "#{prefix}/#{Settings.app_name}" do
       concern :rest_api_v1 do
         scope module: 'v1' do
+          resource 'status', only: :show
           resources :benchmarks, only: [:index, :show]
           resources :business_objectives, only: [:index, :show]
           resources :profiles do
--- a/spec/api/v1/schemas.rb
+++ b/spec/api/v1/schemas.rb
@@ -9,6 +9,7 @@ require './spec/api/v1/schemas/business_objectives'
 require './spec/api/v1/schemas/profiles'
 require './spec/api/v1/schemas/rule_results'
 require './spec/api/v1/schemas/rules'
+require './spec/api/v1/schemas/status'
 require './spec/api/v1/schemas/supported_ssgs'
 
 module Api
@@ -23,6 +24,7 @@ module Api
       include Profiles
       include RuleResults
       include Rules
+      include Status
       include SupportedSggs
 
       SCHEMAS = {
@@ -44,6 +46,7 @@ module Api
         rule_result_relationships: RULE_RESULT_RELATIONSHIPS,
         rule: RULE,
         rule_relationships: RULE_RELATIONSHIPS,
+        status: STATUS,
         supported_ssg: SUPPORTED_SSG
       }.freeze
     end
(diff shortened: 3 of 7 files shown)


app/controllers/v1/statuses_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module V1
+  # API for Compliance Status
+  class StatusesController < ApplicationController
+    skip_around_action :authenticate_user
```
akofink, line 6: Unauthenticated access

app/controllers/v1/statuses_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+        api: api_status
+      }
+    end
+
+    def status_return_code
+      all_statuses.values.all? ? :ok : :internal_server_error
+    end
+
+    def api_status
+      ActiveRecord::Base.establish_connection
```
vkrizan, line 25: Restablishing a connection does not feel right to do here. It seems that this call will create a new connection pool.

vkrizan (commented):
It is interesting that I cannot reproduce a db connection to be hanging after the db server comes back online. It does establish it somehow and (profiles) API works as expected. I'm not sure what has changed, or whether we use some other db calls that overcome the situation noted in RHICOMPL-751.

app/controllers/v1/statuses_controller.rb
skateman: Nope, if the connection is already established, the method just returns false.

skateman (approved):
The Seal of Approval

app/controllers/v1/statuses_controller.rb
vkrizan: Are you sure about this? It seems to call `remove_connection` within the pool https://github.com/rails/rails/blob/v5.2.4.5/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L973

akofink: It seems like calling `establish_connection` does `remove_connection`, so I have added a conditional to only call it if the connection is not already active. Please give it another test.

app/controllers/v1/statuses_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def show
+      render json: { data: all_statuses }, status: status_return_code
+    end
+
+    private
+
+    def all_statuses
+      @all_statuses ||= {
+        api: api_status
```
akofink, line 18: Do you think `api` is OK? Or, maybe we should just say `db` here?

app/controllers/v1/statuses_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      }
+    end
+
+    def status_return_code
+      all_statuses.values.all? ? :ok : :internal_server_error
+    end
+
+    def api_status
```
skateman, line 26: What about something like?

```suggestion
    def api_status
      ARB.connection.active? || !!ARB.establish_connection
    rescue PG::Error
      false
    end
```

I have a feeling that you don't even need the `!!`

vkrizan: I guess the `ARB.connection.active?` would be enough.

app/controllers/v1/statuses_controller.rb
vkrizan: I guess `api` is fine for now.

vkrizan (approved):

app/controllers/v1/statuses_controller.rb
akofink: I have never seen `ARB.establish_connection` ever return anything other than `false`. Are you seeing `true` when it reestablishes a connection?

`ARB.connection.active?` by itself doesn't seem to reestablish the connection once dropped.

vkrizan: Does it have to return anything, considering Ruby's default return?
It might be a completely unrelated return from another function/method call

```
From: /home/vkrizan/.gem/gems/activerecord-import-1.0.4/lib/activerecord-import/import.rb:245 ActiveRecord::Import::Connection#establish_connection:

    243: def establish_connection(args = nil)
    244:   super(args)
 => 245:   ActiveRecord::Import.load_from_connection_pool connection_pool
    246: end
```
lets see what it does..

vkrizan: This might be the one that the `false` comes from:
```
From: /home/vkrizan/.gem/gems/activerecord-import-1.0.4/lib/activerecord-import/base.rb:23 ActiveRecord::Import.require_adapter:

    22: def self.require_adapter(adapter)
 => 23:   require File.join(ADAPTER_PATH, "/#{base_adapter(adapter)}_adapter")
    24: rescue LoadError
    25:   # fallback
    26: end

[3] pry(ActiveRecord::Import)> require File.join(ADAPTER_PATH, "/#{base_adapter(adapter)}_adapter")
=> false
```

PR #830, opened 2021-05-13
devel: Adjust host factory for development

akofink:
I don't expect this to stay super stable, since it depends on the inventory's hosts table, which could change at any time, but for devel, this is still very useful to be able to create thousands of fake hosts, optionally in your own account.

```rb
User.current = User.new(account: Account.find_by(account_number: '123456')) # optional
FactoryBot.create(:host)
FactoryBot.create_list(:host, 1000)
```

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/test/factories/host.rb
+++ b/test/factories/host.rb
@@ -3,36 +3,78 @@
 FactoryBot.define do
   # This is necessary to convince spring that the class is not defined elsewhere
   Object.send(:remove_const, :WHost) if defined?(WHost)
+  Object.send(:remove_const, :SSGS) if defined?(SSGS)
+
+  SSGS = SupportedSsg.all
 
   # A writable version of the Host model for the factory only
   class WHost < Host
-    self.table_name = 'inventory.hosts_v1_1'
+    if Rails.env.test?
+      self.table_name = 'inventory.hosts_v1_1'
+    elsif Rails.env.development?
+      self.table_name = 'hosts'
+
+      establish_connection(
+        Rails.configuration.database_configuration[Rails.env].merge(
+          'database' => 'insights'
+        )
+      )
+    end
 
     def readonly?
       false
     end
   end
 
+  sequence(:account_number) { |n| format('%05<num>d', num: n) }
+
   factory :host, class: WHost do
     id { SecureRandom.uuid }
-    account { User.current&.account&.account_number }
-    display_name { Faker::Lorem.sentence }
+    account do
+      User.current&.account&.account_number || generate(:account_number)
+    end
+    display_name { Faker::Internet.domain_name }
     tags { {} }
-    created { Time.zone.now }
-    updated { Time.zone.now }
     stale_timestamp { 10.years.since(Time.zone.now) }
-    system_profile do
-      {
-        'operating_system' => {
-          'major' => os_major_version,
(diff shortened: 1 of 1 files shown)


test/factories/host.rb
```diff
@@ -4,35 +4,73 @@
...
 
     transient do
-      os_major_version { 7 }
-      os_minor_version { 9 }
+      os_version do
+        ssg = SSGS.sample
+        [ssg.os_major_version.to_i, ssg.os_minor_version.to_i]
+      end
+      os_major_version { os_version[0] }
+      os_minor_version { os_version[1] }
```
akofink, line 65: I also decided to go ahead and make the OS dynamic based on what we have defined in supported_ssgs.yaml, since we often need different OS versions to test with (i.e. RHEL 6)

test/factories/host.rb
```diff
@@ -4,35 +4,73 @@
...
+      system_profile do
+        system_profile_data
+      end
+    elsif Rails.env.development?
+      reporter { 'puptoo' }
+      modified_on { Time.zone.now }
+      facts { {} }
+      canonical_facts do
+        {
+          'fqdn' => Faker::Lorem.sentence
```
vkrizan, line 50: How about `Faker::Internet.domain_name`? I would maybe used it also for display_name (with the same value).

vkrizan (commented):
Nice! I was impatient to have this host generation. Thank you.

There might be more that would need to be filled out for Inventory, so it does not raises exceptions when a new regular host is being registered.

test/factories/host.rb
akofink: Thank you! I was looking for this :laughing:

vkrizan (approved):
Awesome!

vkrizan:
```
 test_: update_hosts should add new hosts to an existing host set. #PolicyTest (45.12s)
Minitest::UnexpectedError:         NoMethodError: undefined method `version' for nil:NilClass
            app/models/supported_ssg.rb:45:in `latest_ssg_version_for_os'
            app/models/profile.rb:43:in `block in <class:Profile>'
            app/models/policy.rb:61:in `block in update_os_minor_versions'
            app/models/policy.rb:60:in `each'
            app/models/policy.rb:60:in `update_os_minor_versions'
            app/models/policy.rb:54:in `update_hosts'
            test/models/policy_test.rb:139:in `block (3 levels) in <class:PolicyTest>'
            test/models/policy_test.rb:138:in `block (2 levels) in <class:PolicyTest>'
            test/models/policy_test.rb:146:in `instance_exec'
            test/models/policy_test.rb:146:in `block in create_test_from_should_hash'
```

akofink:
I'm still trying to figure out why the OS isn't being displayed in the inventory page:

image

akofink:
Ok, I'm quite sure the test issue was caused by 7d51ec1. I was debugging it locally - it's kind of difficult because the issue is transient, but as soon as I reset my test DB after David's change, I can no longer reproduce the failure.

PR #836, opened 2021-05-18
dev: Rename devel Dockerfile to make room for Clowder

akofink:
The Dockerfile at the root of the repository is meant to be our
production image. Ideally, it'd be the same as the devel image, but for
now, this is a good workaround.

See https://docs.docker.com/compose/compose-file/compose-file-v3/#dockerfile

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management


docker-compose.yml
```diff
@@ -183,7 +183,9 @@ services:
       - SETTINGS__PROMETHEUS_EXPORTER_HOST=prometheus
   rails:
     env_file: .env
-    build: .
+    build:
+      context: .
+      dockerfile: devel.Dockerfile
```
vkrizan, line 188: Why not with the suffix, `Dockerfile.devel`, which is more common?

Victoremepunto: I was about to say it's because of the filetype detection, but popular Vim plugins for Dockerfile syntax seem to support any version as long as they have "dockerfile" anywhere in the name

https://github.com/ekalinin/Dockerfile.vim/blob/master/ftdetect/Dockerfile.vim#L4-L8

personally I don't have strong opinions on any format :shrug:

akofink: There is no convention, and it seems to be by design: https://github.com/moby/moby/issues/15549

However, I've actually seen more of this format than what you suggest, Viliam, mostly due to file naming conventions where the extension indicates the file type (and editors/IDEs recognize extensions).

I also do not feel strongly about this, and I hope we can remove this specific devel Dockerfile in the near future.

vkrizan (approved):

skateman (approved):
The Seal of Approval

PR #839, opened 2021-05-19
feat(REST): RHICOMPL-1717 introduce pagination for benchmark->rules

skateman:
I haven't found any documentation on HTTP params regarding the pagination of included relations of the main resource, so I introduced my own `included_offset` and `included_limit`. ActiveRecord doesn't really support the pagination of nested entities out of the box, so I introduced it in the `BenchmarkSerializer` as the `FastJsonapi` allows the passing of custom params to the nested retrieval blocks. It's not the nices solution, but I haven't found any other good place to have this. 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/controllers/concerns/rendering.rb
+++ b/app/controllers/concerns/rendering.rb
@@ -27,6 +27,7 @@ module Rendering
     def serializer_opts
       opts = index? ? metadata : {}
       opts.merge!(include: params[:include].split(',')) if params[:include]
+      opts.merge!(params: pagination)
 
       opts
     end
@@ -50,5 +51,14 @@ module Rendering
     end
 
     def includes; end
+
+    def pagination
+      {
+        limit:
+          params[:include_limit].present? ? params[:include_limit].to_i : 10,
+        offset:
+          params[:include_offset].present? ? params[:include_offset].to_i : 1
+      }
+    end
   end
 end
--- a/app/serializers/benchmark_serializer.rb
+++ b/app/serializers/benchmark_serializer.rb
@@ -3,9 +3,12 @@
 # JSON API serialization for an OpenSCAP Benchmark
 class BenchmarkSerializer
   include FastJsonapi::ObjectSerializer
+
   attributes :ref_id, :title, :version, :description, :os_major_version,
              :latest_supported_os_minor_versions
-  has_many :rules
+  has_many :rules do |benchmark, params|
+    benchmark.rules.paginate(per_page: params[:limit], page: params[:offset])
+  end
   has_many :profiles do |benchmark|
     Pundit.policy_scope(User.current, Profile).where(benchmark: benchmark)
   end


app/serializers/benchmark_serializer.rb
```diff
@@ -3,9 +3,12 @@
 # JSON API serialization for an OpenSCAP Benchmark
 class BenchmarkSerializer
   include FastJsonapi::ObjectSerializer
+
   attributes :ref_id, :title, :version, :description, :os_major_version,
              :latest_supported_os_minor_versions
-  has_many :rules
+  has_many :rules do |benchmark, params|
+    benchmark.rules.paginate(per_page: params[:limit], page: params[:offset])
+  end
```
skateman, line 11: We can make this DRY by either creating a wrapper around `has_many`, e.g. `has_many_pages_of` :scream_cat: or by just extracting this into a method.

app/controllers/concerns/rendering.rb
```diff
@@ -50,5 +51,14 @@ def serializer
     end
 
     def includes; end
+
+    def pagination
+      {
+        limit:
+          params[:include_limit].present? ? params[:include_limit].to_i : 10,
```
bastilian, line 58: Just a note, if we set a default limit for rules, we'll need introduce server-side paginating in the frontend, because currently it requests/expects all rules.

skateman: We maybe also want an upper limit here :thinking: :question:

vkrizan:
Yeah, this seems pretty nonstandard to do a pagination of child entities in REST. It looks like we're trying to re-invent GraphQL (or similar) here... Also consider that the child pagination would happen on all children entity types at once.

A common standard in REST is to expose children within each parent entity. For example `GET /api/v1/benchmarks/<ID>/rules`. 
If there is a need to get children entities from more than one parent, then there can be other means in place that would support that. In this case of Benchmark's rules (considering that the benchmarks are on the same level as rules) the `/rules` endpoint can accept a list of parent IDs, `benchmark_id`. The parent ID have to be exposed on the Rules entity for a client to be able to pair them.

I'd recommend to do that parent children pairing, instead of paginating child entities. Otherwise we should stick to GraphQL :slightly_smiling_face: .

skateman:
@vkrizan that would not really solve the core issue, as an alternative we could disallow the included rules from the endpoint, but afaik this is a supported REST API feature (except the pagination ofc) :confused:

app/controllers/concerns/rendering.rb
akofink: I'd rather figure out a way to determine a query's complexity and return some 4XX if the DB complexity is beyond some limit. So, you can request all rules, perhaps, but if you also want to include all profiles, too, it'd be potentially too complex. I'm not sure if this is really feasible, though. Still, if you request all benchmarks and include rules, it's possible to have included something like 12k rules, which would require a very long response time and something we want to avoid servicing.

akofink:
> A common standard in REST is to expose children within each parent entity. For example GET /api/v1/benchmarks/<ID>/rules.
If there is a need to get children entities from more than one parent, then there can be other means in place that would support that. In this case of Benchmark's rules (considering that the benchmarks are on the same level as rules) the /rules endpoint can accept a list of parent IDs, benchmark_id. The parent ID have to be exposed on the Rules entity for a client to be able to pair them.

I generally agree with this sentiment. REST should allow all the IDs required to request more information in another request; that's already exposed via `relationships`, but like Viliam says, we currently do not support getting many rules in a single request by many rule IDs.

PR #840, opened 2021-05-19
feat(sorting): RHICOMPL-1869 allow sorting of rules

skateman:
The sorting by severity was a bit tricky, I'm assigning numerical values to the severities to sort them properly. When sorting ascending, the first one is `unknown` and the last is `high`.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


app/models/rule.rb
```diff
@@ -3,6 +3,22 @@
 # Stores information about rules, such as which profiles can it be
 # found in, what hosts are associated with it, etceter
 class Rule < ApplicationRecord
+  SORTED_SEVERITIES = Arel.sql(
+    <<-SQL
+    CASE WHEN severity = 'high' THEN 3
+         WHEN severity = 'medium' THEN 2
+         WHEN severity = 'low' THEN 1
+         ELSE 0
```
vkrizan, line 11: Have you tried what would happen if `severity IS NULL`?

vkrizan (approved):
Nice.

app/models/rule.rb
skateman: :thinking: what should happen when `severity IS NULL`?

akofink: Same as unknown IMO

vkrizan: Confirmed. Sorry, I'm being cautious when I see a NULL in postgresql.

skateman: You had a valid point, no worries :+1:

PR #844, opened 2021-05-21
fix(SSG): RHICOMPL-1863 import remediatons after SSG import

vkrizan:
Update available remediations on rules after all SSGs have been imported
to avoid a PG deadlock.

The remediations where being imported on every single SSG that has been
imported.  It has been updating all existing rules regardless of the
current SSG being imported.

That could have been one of the reasons for the deadlock.  Another fact
to consider here that it waited on the API response from the remediation
service within a transaction.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/services/datastream_importer.rb
+++ b/app/services/datastream_importer.rb
@@ -16,16 +16,6 @@ class DatastreamImporter
   def import!
     Xccdf::Benchmark.transaction do
       save_all_benchmark_info
-      import_remediations
     end
-  end
-
-  private
-
-  def import_remediations
-    RemediationsAPI.new(
-      Account.find_by(account_number: ENV['JOBS_ACCOUNT_NUMBER']) ||
-        Account.new
-    ).import_remediations
   end
 end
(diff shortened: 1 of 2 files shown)


akofink (approved):
Thanks, @vkrizan!

app/services/datastream_importer.rb
```diff
@@ -16,16 +16,6 @@ def initialize(datastream_filename)
...
-      import_remediations
     end
   end
-
-  private
-
-  def import_remediations
-    RemediationsAPI.new(
-      Account.find_by(account_number: ENV['JOBS_ACCOUNT_NUMBER']) ||
-        Account.new
```
akofink, line 28: It looks like some of our deployments rely on this fallback. I've added it back here: https://github.com/RedHatInsights/compliance-backend/pull/848

vkrizan: Oh, sorry, I haven't noticed it. I saw that the env var was configured and continued with the changes.

akofink: I guess it needs to also be able to work when there are no accounts, like in a smoke test run.

PR #849, opened 2021-05-24
fix(sorting): RHICOMPL-1883 sort string-based columns in lowercase

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -7,7 +7,7 @@ class Host < ApplicationRecord
   OS_MAJOR_VERSION = Arel.sql("system_profile->'operating_system'->'major'")
 
   SORTABLE_BY = {
-    name: :display_name,
+    name: Arel.sql('LOWER(display_name)'),
     os_minor_version: OS_MINOR_VERSION,
     os_major_version: OS_MAJOR_VERSION
   }.freeze
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -14,7 +14,7 @@ class Rule < ApplicationRecord
   )
 
   SORTABLE_BY = {
-    title: :title,
+    title: Arel.sql('LOWER(rules.title)'),
     severity: SORTED_SEVERITIES,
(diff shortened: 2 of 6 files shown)


app/models/rule.rb
```diff
@@ -14,7 +14,7 @@ class Rule < ApplicationRecord
   )
 
   SORTABLE_BY = {
-    title: :title,
+    title: Arel.sql('LOWER(rules.title)'),
```
akofink, line 17: why does this need `rules.`?

app/models/host.rb
```diff
@@ -7,7 +7,7 @@ class Host < ApplicationRecord
   OS_MAJOR_VERSION = Arel.sql("system_profile->'operating_system'->'major'")
 
   SORTABLE_BY = {
-    name: :display_name,
+    name: Arel.sql('LOWER(display_name)'),
```
vkrizan, line 10:
```
Host.arel_table[:display_name].lower
```

app/models/rule.rb
skateman: Without this I'm getting:
```
ERROR:  column reference "title" is ambiguous
```

app/models/host.rb
skateman: That is not compatible with our hash-sorting format that is being built in the controller :thinking:

vkrizan:
@skateman what are we trying to solve? It seems that we might fix this on a database level, without the need to call `LOWER`. (Sorry Jira was down)

skateman:
@vkrizan how?

skateman:
Okay, so we discussed this with @vkrizan and we think that the current sorting is intended for humans and that's what we want...

PR #851, opened 2021-05-26
fix(build): RHICOMPL-1889 Refactor and bugfix of build_deploy.sh

Victoremepunto:
- Fixes a bug using the container engine command within a variable,
  refactored to use functions instead
- Add a local build option to run locally without pushing to remote
  registy

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,24 @@ docker-compose logs -f sidekiq
 podman logs -f compliance-backend_sidekiq_1
 ```
 
+### Building the image
+
+This project uses `build_deploy.sh` to build and push this image to a remote registry.
+You must set the required variables to provide authentication to the required registries.
+You must set at least the Red Hat Registry credentials to be able to pull the base image.
+
+Example:
+```
+LOCAL_BUILD=true RH_REGISTRY_USER=guybrush RH_REGISTRY_TOKEN=M0nk3y ./build_deploy.sh
+```
+
+Optionally, if you want to push to the remote registry you must set the remote registry credentials as well (LOCAL_BUILD defaults to `false`)
+
+Example:
+```
+LOCAL_BUILD=true RH_REGISTRY_USER=guybrush RH_REGISTRY_TOKEN=monkey QUAY_USER=lechuck QUAY_TOKEN=Ela1ne ./build_deploy.sh
+```
+
 ## Development notes
 
 ### Running Sonarqube
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -2,8 +2,20 @@
 
 set -exv
 
+local_build() {
+  [ "$LOCAL_BUILD" = true ]
+}
+
+backwards_compatibility_enabled() {
+  [ "$BACKWARDS_COMPATIBILITY" = true ]
+}
+
+get_7_chars_commit_hash() {
+    echo "$(git rev-parse --short=7 HEAD)-new"
+}
+
 _check_command_is_present() {
-    command -v $1
+    command -v "$1"
 }
 
 login_container_registry() {
@@ -12,14 +24,31 @@ login_container_registry() {
     local PASSWORD="$2"
     local REGISTRY="$3"
 
-    $CONTAINER_ENGINE_CMD login "-u=${USER}" "-p=${PASSWORD}" "$REGISTRY"
+    container_engine_cmd login "-u=${USER}" "-p=${PASSWORD}" "$REGISTRY"
 }
 
 check_required_registry_credentials() {
 
-    if [[ -z "$QUAY_USER" || -z "$QUAY_TOKEN" || -z "$RH_REGISTRY_USER" || -z "$RH_REGISTRY_TOKEN" ]]; then
-        echo "QUAY_USER, QUAY_TOKEN, RH_REGISTRY_USER and RH_REGISTRY_TOKEN must be set"
+
+    if [[ -z "$RH_REGISTRY_USER" || -z "$RH_REGISTRY_TOKEN" ]]; then
+        echo "RH_REGISTRY_USER and RH_REGISTRY_TOKEN must be set"
         exit 1
+    fi
+
+    if ! local_build; then
+        if [[ -z "$QUAY_USER" || -z "$QUAY_TOKEN" ]]; then
+            echo "QUAY_USER, QUAY_TOKEN must be set"
+            exit 1
+        fi
+    fi
+}
+
+container_engine_cmd() {
+
+    if [ "$CONTAINER_ENGINE_CMD" = "podman" ]; then
+        podman "$@"
+    else
+        docker "--config=${DOCKER_CONF}" "$@"
     fi
 }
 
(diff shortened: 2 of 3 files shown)


entrypoint.sh
```diff
@@ -49,4 +54,6 @@ elif [ "$APPLICATION_TYPE" = "compliance-import-remediations" ]; then
 	exec bundle exec rake import_remediations --trace
 elif [ "$APPLICATION_TYPE" = "compliance-import-ssg" ]; then
 	exec bundle exec rake ssg:import_rhel_supported --trace
+else
+    echo "Application type '$APPLICATION_TYPE' not supported"
```
akofink, line 58: Nice :laughing:

build_deploy.sh
```diff
@@ -12,15 +26,32 @@ login_container_registry() {
...
+        if [[ -z "$QUAY_USER" || -z "$QUAY_TOKEN" ]]; then
+            echo "QUAY_USER, QUAY_TOKEN must be set"
+            exit 1
+        fi
+    fi
+}
+
+container_engine_cmd() {
+
+    if [ "$CONTAINER_ENGINE_CMD" == 'podman' ]; then
```
vkrizan, line 50: Nit: use bash conditional `[[ "$CONTAINER_ENGINE_CMD" == "podman" ]]`. Not sure if it is even required for the double equals `==`.

README.md
```diff
@@ -235,6 +235,24 @@ docker-compose logs -f sidekiq
...
 
+### Building the image
+
+This project uses `build_deploy.sh` to build and push this image to a remote registry.
+You must set the required variables to provide authentication to the required registries.
+You must set at least the Red Hat Registry credentials to be able to pull the base image.
+
+Example:
+```
+LOCAL_BUILD=true RH_REGISTRY_USER=guybrush RH_REGISTRY_TOKEN=M0nk3y ./build_deploy
```
akofink, line 246:
```suggestion
LOCAL_BUILD=true RH_REGISTRY_USER=guybrush RH_REGISTRY_TOKEN=M0nk3y ./build_deploy.sh
```

vkrizan, line 246: `./build_deploy.sh`

build_deploy.sh
```diff
@@ -2,6 +2,20 @@
 
 set -exv
 
+BACKWARDS_COMPATIBILITY="${BACKWARDS_COMPATIBILITY:-enabled}"
+LOCAL_BUILD="${LOCAL_BUILD:-false}"
+IMAGE_REGISTRY="quay.io"
+IMAGE_REPOSITORY="cloudservices/compliance-backend"
```
vkrizan, line 8: Maybe a suggestion, if thes can be also set by an env var so we can push to other repos.

build_deploy.sh
```diff
@@ -12,15 +26,32 @@ login_container_registry() {
...
 
-    if [[ -z "$QUAY_USER" || -z "$QUAY_TOKEN" || -z "$RH_REGISTRY_USER" || -z "$RH_REGISTRY_TOKEN" ]]; then
-        echo "QUAY_USER, QUAY_TOKEN, RH_REGISTRY_USER and RH_REGISTRY_TOKEN must be set"
+
+    if [[ -z "$RH_REGISTRY_USER" || -z "$RH_REGISTRY_TOKEN" ]]; then
+        echo "RH_REGISTRY_USER and RH_REGISTRY_TOKEN must be set"
         exit 1
     fi
+
+    if ! "$LOCAL_BUILD"; then
```
vkrizan, line 40: Wouldn't this run a command? Also please check other usages.

```
if [[ ! "$LOCAL_BUILD" ]]; then
```

entrypoint.sh
```diff
@@ -14,6 +14,11 @@ function check_number() {
 	fi
 }
 
+if [ -z "$APPLICATION_TYPE" ]; then
```
vkrizan, line 17: Nit: use bash conditional `[[ -z "$APPLICATION_TYPE" ]]`

vkrizan (commented):
Nice. Thank you Victor.

build_deploy.sh
```diff
@@ -2,6 +2,20 @@
 
 set -exv
```
akofink, line 3: Not related to the PR, but what's the reasoning for `-xv` here? It really makes the output and error messages you've added pretty difficult to see.

From the man page:

```
-v      Print shell input lines as they are read.
-x      After expanding each simple command, for command, case command, select command, or arithmetic for command, display the expanded value of PS4, followed by the command and its  expanded
        arguments or associated word list.
```

Without them, the script is nice and easy to figure out:

```
$ ./build_deploy.sh
RH_REGISTRY_USER and RH_REGISTRY_TOKEN must be set
```

akofink (approved):
Looks good, and works as expected. Just a couple minor comments.

build_deploy.sh
vkrizan: It is good for logging (and audit), as you can see what is actually happening. Not sure if `-x` sends it to `stdout`, but if it does we can print error messages to `stderr` to be able to look for them.

README.md
Victoremepunto: thanks ! updated! (sorry I didn't see the suggestion mode :D )

Victoremepunto: well spotted :+1:

build_deploy.sh
Victoremepunto: I think it's because this is meant to be run on Jenkins and they want all the possible output. grabbed it from the existing templates for pr_check.sh  script. I'm also not a fan.

From Bash manpage
> -e      Exit immediately if a pipeline (which may consist of a single simple command), a list, or a compound command (see SHELL GRAMMAR above), exits with a non-zero status. [ ... ]
> -v      Print shell input lines as they are read.
> -x      After expanding each simple command, for command, case command, select command, or arithmetic >for command, display the expanded value of PS4, followed by the command and its expanded arguments or >associated word list.

build_deploy.sh
Victoremepunto: Makes sense, I'll add it, as well as for the `IMAGE_REGISTRY`

entrypoint.sh
Victoremepunto: it helped with having something when deploying this locally

build_deploy.sh
Victoremepunto: refactored

build_deploy.sh
Victoremepunto: refactored

build_deploy.sh
Victoremepunto: @akofink , @vkrizan  I've refactored this set flags to change it's behavior if it's a local run or not , hope it's okay

entrypoint.sh
Victoremepunto: I changed this , regardless, to `[[ -z "foo" ]]` form , but I think I'd like to get to some agreement if we're going to maintain different Bash scripts. I'm fine with anything as long as we agree on something :)

Victoremepunto: Sorry, I actually didn't change it in this particular file, but in the `buid_deploy.sh` script only, as the entrypoint was based on this https://github.com/sclorg/s2i-ruby-container/blob/master/2.7/s2i/bin/run , and is the original entrypoint used in prod.

build_deploy.sh
```diff
@@ -29,49 +58,74 @@ initialize_container_engine_cmd() {
...
+        fi
     done
 }
 
-IMAGE_NAME="quay.io/cloudservices/compliance-backend"
-IMAGE_TAG="$(git rev-parse --short=7 HEAD)-new"
-WORKDIR="$PWD"
+if ! local_build; then
+    set -xv
+fi
```
akofink, line 108: This doesn't work the same way now when local build is false. It needs to be at the top of the file if you want all the function definitions to be printed to STDOUT.

Victoremepunto: reverted it back and left the flags as they were

entrypoint.sh
vkrizan: Oh, I now see that this was taken from our prev. s2i. Since we're using `bash`, I'd stick with it's constructs. `[[ ]]` and `==` in this case.

PR #862, opened 2021-05-27
perf(REST): RHICOMPL-1717 allow preloading records for explicit includes

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/controllers/v1/benchmarks_controller.rb
+++ b/app/controllers/v1/benchmarks_controller.rb
@@ -32,5 +32,11 @@ module V1
     def serializer
       BenchmarkSerializer
     end
+
+    def includes
+      return unless params[:include]&.split(',')&.include?('rules')
+
+      { rules: %i[profiles rule_identifier rule_references] }
+    end
   end
 end
--- a/test/controllers/v1/benchmarks_controller_test.rb
+++ b/test/controllers/v1/benchmarks_controller_test.rb
@@ -18,5 +18,12 @@ module V1
       get v1_benchmarks_url(FactoryBot.create(:benchmark))
       assert_response :success
     end
+
+    test '#index includes rules' do
+      FactoryBot.create(:rule, benchmark: FactoryBot.create(:benchmark))
+      get v1_benchmarks_url, params: { include: 'rules' }
+      assert_response :success
+      assert_equal(response.parsed_body['included'].first['type'], 'rule')
+    end
   end
 end


app/controllers/concerns/rendering.rb
```diff
@@ -49,6 +49,16 @@ def serializer
...
+    def includes_from_params
+      params[:include] ? params[:include].split(',') : []
+    end
+
+    def includes
+      return nil unless defined?(self.class::INCLUDES)
+
+      [nil, includes_from_params].flatten.map do |field|
+        self.class::INCLUDES[field.try(:to_sym)]
+      end
```
akofink, line 61: :+1: It makes sense

app/controllers/v1/benchmarks_controller.rb
```diff
@@ -3,6 +3,11 @@
 module V1
   # API for Benchmarks
   class BenchmarksController < ApplicationController
+    INCLUDES = {
+      nil => [],
```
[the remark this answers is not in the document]
skateman: I'm all in for implementing this in the serializer, but find someone who cleans up that codebase first :sweat_smile:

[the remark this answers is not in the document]
akofink: As discussed in slack, a constant is not really useful to have inherited behavior with slightly different data. You need to use some instance-based storage, like a method or instance variable. Since you don't have access to the controller's initializer, you can't define an instance variable. The only thing I can think to use is a method, as was originally implemented.

skateman:
Maybe I took the simplification too far :see_no_evil:

app/controllers/concerns/rendering.rb
vkrizan: What would happen if `params[:include]` would get repeated includes, like `include=rules,rules,rules,...`?
What would happen if the param would get invalid or empty entry?

config/environments/development.rb
```diff
@@ -52,12 +52,12 @@
...
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
-  config.after_initialize do
-    Bullet.enable = true
-    Bullet.alert = true
-    Bullet.bullet_logger = true
-    Bullet.console = true
-    Bullet.rails_logger = true
-    Bullet.add_footer = true
-  end
+  # config.after_initialize do
```
vkrizan, line 55: Is this a remnant, or are we disabling this by default?

app/controllers/concerns/rendering.rb
skateman: Sorry :see_no_evil: somehow I got 2 commits instead of one, should be okay now.

config/environments/development.rb
skateman: Pushed the wrong stuff, fixed.

skateman:
In order to measure the performance in development, you will have to turn off the Bullet benchmarks. It doesn't run in production and it somehow makes things slow in dev.

app/controllers/v1/benchmarks_controller.rb
```diff
@@ -32,5 +32,11 @@ def resource
     def serializer
       BenchmarkSerializer
     end
+
+    def include
```
vkrizan, line 36: This method does not seem to be called at all, does it? (Also indicated by missing coverage.)

vkrizan (approved):
LGTM. It seems 1/2 faster.

Would you mind adding a test coverage for the `includes`?

PR #866, opened 2021-05-27
Add Initializer for Clowder config for settings

Victoremepunto:
UPDATE - I've modified the `docker-compose` file to read the configuration as if it was on a Clowder environment. In order to do so, I've added the `devel.json` file, which would be added to each pod's volume, and I've added the variable `ACG_CONFIG` to each of Compliance's pods. This should emulate the "clowder config injection" that happens on a Clowder deployed app. The Clowder_config::isClowderEnabled? function now returns true and uses the values from the JSON file to provide the mocked config values that are pretended to be provided by Clowder.

If you wanna test this, you should be able to do a podman-compose up and see the application working normally. Please test thoroughly and provide feedback.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -13,7 +13,7 @@
     db_service = ENV.fetch("DATABASE_SERVICE_NAME","").upcase.sub("-", "_")
     username = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? "postgres" : ENV["POSTGRESQL_USER"]
     password = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? ENV["POSTGRESQL_ADMIN_PASSWORD"] : ENV["POSTGRESQL_PASSWORD"]
-    host = ENV.fetch("#{db_service}_SERVICE_HOST")
+    host = ENV.fetch("POSTGRES_SERVICE_HOST")
     database = Rails.env.test? ? ENV['POSTGRESQL_TEST_DATABASE'] || ENV["POSTGRESQL_DATABASE"] : ENV["POSTGRESQL_DATABASE"]
     port = ENV.key?("POSTGRESQL_PORT") ? ENV["POSTGRESQL_PORT"] : 5432
   end
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,18 +3,19 @@
 
     config = ClowderCommonRuby::Config.load
 
+    db_service = config.database.hostname
     username = config.database.username
     password = config.database.password
     host = config.database.hostname
     database = config.database.name
+    port = config.database.port
   else
-
     db_service = ENV.fetch("DATABASE_SERVICE_NAME","").upcase.sub("-", "_")
-
     username = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? "postgres" : ENV["POSTGRESQL_USER"]
     password = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? ENV["POSTGRESQL_ADMIN_PASSWORD"] : ENV["POSTGRESQL_PASSWORD"]
     host = ENV.fetch("#{db_service}_SERVICE_HOST")
     database = Rails.env.test? ? ENV['POSTGRESQL_TEST_DATABASE'] || ENV["POSTGRESQL_DATABASE"] : ENV["POSTGRESQL_DATABASE"]
+    port = ENV.key?("POSTGRESQL_PORT") ? ENV["POSTGRESQL_PORT"] : 5432
   end
  %>
 
@@ -26,6 +27,7 @@ default: &default
   password: <%= password %>
   host: <%= host %>
   database: <%= database %>
+  port: <%= port %>
 
 test:
   <<: *default
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,7 +3,6 @@
 
     config = ClowderCommonRuby::Config.load
 
-    db_service = 'postgres'
     username = config.database.username
     password = config.database.password
     host = config.database.hostname
@@ -13,7 +12,7 @@
     db_service = ENV.fetch("DATABASE_SERVICE_NAME","").upcase.sub("-", "_")
     username = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? "postgres" : ENV["POSTGRESQL_USER"]
     password = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? ENV["POSTGRESQL_ADMIN_PASSWORD"] : ENV["POSTGRESQL_PASSWORD"]
-    host = ENV.fetch("POSTGRES_SERVICE_HOST")
+    host = ENV.fetch("#{db_service}_SERVICE_HOST")
     database = Rails.env.test? ? ENV['POSTGRESQL_TEST_DATABASE'] || ENV["POSTGRESQL_DATABASE"] : ENV["POSTGRESQL_DATABASE"]
     port = ENV.key?("POSTGRESQL_PORT") ? ENV["POSTGRESQL_PORT"] : 5432
   end
@@ -31,6 +30,7 @@ default: &default
 
 test:
   <<: *default
+  database: <%= ENV['POSTGRESQL_TEST_DATABASE'] || database %>
 
 production:
   <<: *default
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,7 +3,7 @@
 
     config = ClowderCommonRuby::Config.load
 
-    db_service = config.database.hostname
+    db_service = 'postgres'
     username = config.database.username
     password = config.database.password
     host = config.database.hostname
--- a/devel.json
+++ b/devel.json
@@ -1,0 +1,86 @@
+{
+    "webPort": 8000,
+    "metricsPort": 9000,
+    "metricsPath": "/metrics",
+    "logging": {
+        "type": "cloudwatch",
+        "cloudwatch": {
+            "accessKeyId": "ACCESS_KEY",
+            "secretAccessKey": "SECRET_ACCESS_KEY",
+            "region": "EU",
+            "logGroup": "base_app"
+            }
+        },
+    "kafka": {
+        "brokers": [
+            {
+                "hostname": "kafka",
+                "port": 29092
+            }
+        ],
+        "topics": [
+            {
+                "requestedName": "platform.inventory.events",
+                "name": "platform.inventory.events",
+                "consumerGroup": "someGroupName"
+            },
+            {
+                "requestedName": "platform.upload.compliance",
+                "name": "platform.upload.compliance",
+                "consumerGroup": "someGroupName"
+            },
+            {
+                "requestedName": "platform.payload-status",
+                "name": "platform.payload-status",
+                "consumerGroup": "someGroupName"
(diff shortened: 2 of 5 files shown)


config/initializers/99_clowder-config.rb
```diff
@@ -0,0 +1,37 @@
+
+if ClowderCommonRuby::Config.clowder_enabled?
+
+  config = ClowderCommonRuby::Config.load
+
+  Settings.add_source!({kafka_producer_topics.inventory_events: config.kafka_topics['platform.inventory.events']})
```
akofink, line 6: This is not a valid symbol. I believe you need to nest the hashes:

```suggestion
  Settings.add_source!({kafka_producer_topics: { inventory_events: config.kafka_topics['platform.inventory.events']}})
```

Really, it'd be better to add only one source for clowder:

```
Settings.add_source!(Rails.root.join('config', 'clowder.yml'))
```

Then, you can use erb within clowder.yml to set any values you'd like.

vkrizan: Or, (if the YAML wouldn't look okay) you can try to construct a single Hash.

akofink: Yes, to be clear, Victor, one single gigantic hash for all the clowder config; Only one `add_source!` call.

Victoremepunto: gotcha, adapted to the "big hash" approach

config/initializers/99_clowder-config.rb
```diff
@@ -0,0 +1,43 @@
...
+    kafka_producer_topics.payload_tracker: config.kafka_topics['platform.payload-status'],
+    kafka_producer_topics.remediation_updates: config.kafka_topics['platform.remediations.events'],
+    prometheus_exporter_host: prometheus_exporter_hostname,
+    prometheus_exporter_port: prometheus_exporter_port,
+    rbac_url: rbac_url,
+    remediations_url: remediations_url,
+    host_inventory_url: host_inventory_url
+  }
+
+  Settings.add_source!(clowder_config)
```
akofink, line 40: :+1: yes! Thank you

config/initializers/99_clowder-config.rb
```diff
@@ -0,0 +1,43 @@
...
+  rbac_hostname = config.private_dependency_endpoints.rbac.rbac.hostname
+  rbac_port = config.private_dependency_endpoints.rbac.rbac.port
+  rbac_url = "#{rbac_hostname}:#{rbac_port}"
+
+  prometheus_exporter_hostname = config.private_dependency_endpoints.compliance.prometheus-exporter.hostname
+  prometheus_exporter_port = config.private_dependency_endpoints.compliance.prometheus-exporter.port
+
+  # TO-DO add in-memory db here for compliance-redis
+
+  Settings.add_source!(Rails.root.join('config', 'clowder.yml'))
```
akofink, line 26: Are you going to add this file as part of this PR?

config/initializers/99_clowder-config.rb
```diff
@@ -0,0 +1,43 @@
+
+if ClowderCommonRuby::Config.clowder_enabled?
+
+  config = ClowderCommonRuby::Config.load
+
+  # TO-DO: Refactor with a helper function (get_service_url(app_name, component)
+  host_inventory_hostname = config.private_dependency_endpoints.host-inventory.service.hostname
+  host_inventory_port = config.private_dependency_endpoints.host-inventory.service.port
+  host_inventory_url = "#{host_inventory_hostname}:#{host_inventory_port}"
```
vkrizan, line 9: Nit: This might be a bit more readable:
```
host_inventory = config.private_dependency_endpoints.host-inventory.service
host_inventory_url = "#{host_inventory.hostname}:#{host_inventory.port}"
```

config/initializers/99_clowder-config.rb
Victoremepunto: no that was an overlook, I'm removing it. Clowder config is only meant to be read through this config library object, so there's little value on using a YAML file this templating

config/initializers/99_clowder-config.rb
Victoremepunto: Thank you, I appreciate the suggestion, and I've refactored those lines accordingly  :)

docker-compose.yml
```diff
@@ -156,13 +156,14 @@ services:
...
-      - POSTGRESQL_PASSWORD=insights
-      - SETTINGS__PROMETHEUS_EXPORTER_HOST=prometheus
+#      - DATABASE_SERVICE_NAME=postgres
+#      - POSTGRES_SERVICE_HOST=db
+#      - POSTGRESQL_DATABASE=compliance_dev
+#      - POSTGRESQL_TEST_DATABASE=compliance_test
+#      - POSTGRESQL_USER=insights
+#      - POSTGRESQL_PASSWORD=insights
+#      - SETTINGS__PROMETHEUS_EXPORTER_HOST=prometheus
+      - ACG_CONFIG=/app/test.json
```
akofink, line 166: :clap: this is way better than duplicating env variables. Can we name it based on the devel env, like devel.json? Also, what does ACG mean? Also, let's definitely remove stuff during the PR cleanup instead of just commenting it out.

akofink (approved):
Seems to work great! Thanks, @Victoremepunto!

docker-compose.yml
```diff
@@ -156,13 +156,14 @@ services:
...
-      - DATABASE_SERVICE_NAME=postgres
-      - POSTGRES_SERVICE_HOST=db
-      - POSTGRESQL_DATABASE=compliance_dev
-      - POSTGRESQL_TEST_DATABASE=compliance_test
-      - POSTGRESQL_USER=insights
-      - POSTGRESQL_PASSWORD=insights
-      - SETTINGS__PROMETHEUS_EXPORTER_HOST=prometheus
+#      - DATABASE_SERVICE_NAME=postgres
+#      - POSTGRES_SERVICE_HOST=db
+#      - POSTGRESQL_DATABASE=compliance_dev
```
akofink, line 161: Sorry, still using this compose config for testing other work, and I found an issue. This compliance_dev DB name is removed completely and not used within test.json. As you can see, some services need the insights DB (i.e inventory-web and inventory) while all of our compliance services need compliance_dev DB.

akofink: Similar comment for the tests which use POSTGRESQL_TEST_DATABASE which is different than the devel DB.

Victoremepunto: okay, I updated the Clowder settings mock file to provide the "compliance_dev" database name, left the `POSTGRESQL_TEST_DATABASE` as fixed environment variable - since this is only for local deployment I think it should be okay.

docker-compose.yml
Victoremepunto: The `ACG_CONFIG` is the variable environment where the Clowder config library expects to read the JSON config file that Clowder injects into a pod it manages. 

Maybe it's `Application ConfiG` ? or something :) - better ask @psav or @kylape about the historical naming reasons :)

Victoremepunto: I also removed the comments and renamed the file to `devel.json`

docker-compose.yml
```diff
@@ -185,13 +180,9 @@ services:
       - prometheus
       - compliance-ssg
     environment:
-      - DATABASE_SERVICE_NAME=postgres
-      - POSTGRES_SERVICE_HOST=db
-      - POSTGRESQL_DATABASE=compliance_dev
       - POSTGRESQL_TEST_DATABASE=compliance_test
-      - POSTGRESQL_USER=insights
-      - POSTGRESQL_PASSWORD=insights
       - SETTINGS__PROMETHEUS_EXPORTER_HOST=prometheus
```
akofink, line 184: This should come from devel.json

akofink (commented):
A couple more comments inline. I also noticed host syndication breaks. For some reason the view no longer works. I haven't figured out the reason yet:

```
compliance_dev=# select count(*) from inventory.hosts;
 count 
-------
     0
(1 row)

compliance_dev=# \c insights
psql (11.12 (Debian 11.12-0+deb10u1), server 13.3 (Debian 13.3-1.pgdg100+1))
You are now connected to database "insights" as user "insights".
insights=# select count(*) from hosts;
 count 
-------
     2
(1 row)
```

or, from the rails console:

```rb
[14] pry(main)> WHost.count
   (0.5ms)  SELECT COUNT(*) FROM "hosts"
=> 2
[15] pry(main)> Host.count
   (0.3ms)  SELECT COUNT(*) FROM "inventory"."hosts"
=> 0
```

docker-compose.yml
Victoremepunto: they come from it, I removed from the file

config/database.yml
```diff
@@ -3,18 +3,19 @@
...
     host = config.database.hostname
     database = config.database.name
+    port = config.database.port
   else
-
     db_service = ENV.fetch("DATABASE_SERVICE_NAME","").upcase.sub("-", "_")
-
     username = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? "postgres" : ENV["POSTGRESQL_USER"]
     password = ENV.key?("POSTGRESQL_ADMIN_PASSWORD") ? ENV["POSTGRESQL_ADMIN_PASSWORD"] : ENV["POSTGRESQL_PASSWORD"]
     host = ENV.fetch("#{db_service}_SERVICE_HOST")
```
akofink, line 16: Let's just remove db_service:

```suggestion
    host = ENV.fetch("POSTGRES_SERVICE_HOST")
```

devel.json
```diff
@@ -0,0 +1,117 @@
...
+            "app": "compliance",
+            "hostname": "prometheus",
+            "port": 9394
+        },
+        {
+            "name": "endpoint2",
+            "app": "app2",
+            "hostname": "endpoint2.svc",
+            "port": 10000
+        },
```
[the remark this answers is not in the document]
Victoremepunto: indeed  - removed along other unused sections (like the S3 buckets one)

config/database.yml
```diff
@@ -3,18 +3,19 @@
 
     config = ClowderCommonRuby::Config.load
 
+    db_service = config.database.hostname
```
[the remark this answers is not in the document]
Victoremepunto: hardcoded to `postgres` - we can remove it later  if we need to.

config/database.yml
```diff
@@ -3,18 +3,19 @@
 
     config = ClowderCommonRuby::Config.load
 
+    db_service = 'postgres'
```
akofink, line 6: Just remove it :laughing: It's not used anywhere in the clowder logic.

akofink:
Ok, @Victoremepunto - I figured out the Cyndi issue. The test DB was not being used with RAILS_ENV=test, so both setup_cyndi_test.sql and setup_cyndi_devel.sql were running on the same DB, which definitely did not work :laughing: I've pushed a fix here!

akofink (approved):

PR #870, opened 2021-06-02
Do not evaluate relationship on included records in REST

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/spec/integration/profiles_spec.rb
+++ b/spec/integration/profiles_spec.rb
@@ -255,7 +255,7 @@ describe 'Profiles API' do
                        type: { type: :string },
                        id: ref_schema('uuid'),
                        attributes: ref_schema('benchmark'),
-                       relationships: ref_schema('benchmark_relationships')
+                       relationships: { type: :object, properties: {} }
                      }
                    }
                  }
--- a/app/controllers/concerns/rendering.rb
+++ b/app/controllers/concerns/rendering.rb
@@ -26,6 +26,7 @@ module Rendering
 
     def serializer_opts
       opts = index? ? metadata : {}
+      opts.merge!(params: { root_resource: resource })
       opts.merge!(include: params[:include].split(',')) if params[:include]
 
       opts
--- a/app/controllers/v1/benchmarks_controller.rb
+++ b/app/controllers/v1/benchmarks_controller.rb
@@ -36,7 +36,7 @@ module V1
     def includes
       return unless params[:include]&.split(',')&.include?('rules')
 
-      { rules: %i[profiles rule_identifier rule_references] }
+      { rules: %i[profiles] }
     end
   end
 end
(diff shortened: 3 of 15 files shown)


app/serializers/compliance_serializer.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(resource, options = {})
+    options[:params] ||= {}
+    options[:params][:root_resource] = resource.class
+
+    super(resource, options)
+  end
+
+  class_methods do
+    def relationships_hash(record, _, _, _, params = {})
+      return nil if params[:root_resource] != record.class
```
skateman, line 17: I'm not entirely sure about the `nil` here, see the JIRA issue for more...

vkrizan: Empty Hash/JSON Object could be more appropriate.

This makes `relationships` empty in any case, even on top level items.

skateman: That has been adjusted :wink: gonna change the `nil` to `{}` then...

app/serializers/compliance_serializer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# JSON serializer parent class
+module ComplianceSerializer
```
akofink, line 4: Nit: It's not a class :laughing: But you could make it one if you'd like! `ApplicationSerializer` would be the rails pattern.

app/serializers/compliance_serializer.rb
```diff
@@ -0,0 +1,15 @@
...
+  extend ActiveSupport::Concern
+  include ::JSONAPI::Serializer
+
+  class_methods do
+    def relationships_hash(record, _, _, _, params = {})
+      return {} if params[:root_resource] != record.class
+
+      super
+    end
```
akofink, line 13: Maybe we should add a comment about this being hacky and dependent on jsonapi_serializer's implementation. It seems like it'd be pretty stable, though, as long as the method name doesn't change :see_no_evil:

spec/integration/profiles_spec.rb
```diff
@@ -255,7 +255,7 @@
                        type: { type: :string },
                        id: ref_schema('uuid'),
                        attributes: ref_schema('benchmark'),
-                       relationships: ref_schema('benchmark_relationships')
+                       relationships: {}
```
akofink, line 258: This is required to be a schema. An always-empty object would look like this:

```suggestion
                       relationships: { type: :object, properties: {} }
```

akofink (approved):
LGTM! Just a couple minor comments.

akofink:
I forgot to check the openapi.json. If there are any included resource relationships there (I haven't checked), we should also update them to empty objects `{}`.

PR #873, opened 2021-06-03
fix(REST): RHICOMPL-1899 sort profiles by score only when unsorted

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/controllers/concerns/metadata.rb
+++ b/app/controllers/concerns/metadata.rb
@@ -13,7 +13,7 @@ module Metadata
     end
 
     def metadata(opts = {})
-      opts[:total] ||= resolve_collection.count
+      opts[:total] ||= aggregated_count(resolve_collection.count)
       options = {}
       options[:meta] = { total: opts[:total], search: params[:search],
                          limit: pagination_limit, offset: pagination_offset }
@@ -61,6 +61,12 @@ module Metadata
     end
 
     private
+
+    def aggregated_count(count)
+      return count if count.is_a?(Integer)
+
+      count.values.sum
+    end
 
     def base_link_url
       "#{path_prefix}/#{Settings.app_name}/#{controller_name}"
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -12,7 +12,8 @@ class Profile < ApplicationRecord
 
   SORTABLE_BY = {
     name: :name,
-    os_minor_version: :os_minor_version
+    os_minor_version: :os_minor_version,
+    score: :score
   }.freeze
 
   belongs_to :account, optional: true
@@ -42,6 +43,17 @@ class Profile < ApplicationRecord
   validates :benchmark_id, presence: true
   validates :account, presence: true, if: -> { hosts.any? }
   validates :policy, presence: true, if: -> { policy_id }
+
+  scope :with_score, lambda {
+    group('profiles.id')
+      .select(
+        Profile.arel_table[Arel.star],
+        TestResult.arel_table[:score].average.as('score')
+      ).joins("
+        LEFT OUTER JOIN (#{TestResult.latest.to_sql}) AS test_results
+        ON test_results.profile_id = profiles.id
+      ")
+  }
 
   scope :canonical_for_os, lambda { |os_major_version, os_minor_version|
     benchmarks = Xccdf::Benchmark.latest_for_os(
--- a/test/controllers/v1/profiles_controller_test.rb
+++ b/test/controllers/v1/profiles_controller_test.rb
@@ -206,9 +206,31 @@ module V1
                      profiles.dig('data', 0, 'attributes', 'policy_profile_id')
       end
 
+      test 'profiles are sorted by score by default' do
+        p1 = FactoryBot.create(:profile, name: 'a')
+        p2 = FactoryBot.create(:profile, name: 'b')
+        FactoryBot.create(:test_result, profile: p1, score: 0.3)
+        FactoryBot.create(:test_result, profile: p2, score: 0.5)
+
+        get v1_profiles_url, params: {
+          search: 'canonical=false'
+        }
+
+        assert_response :success
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(%w[a b], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
       test 'profiles can be sorted by a single dimension' do
-        FactoryBot.create(:profile, name: 'a')
-        FactoryBot.create(:profile, name: 'b')
+        p1 = FactoryBot.create(:profile, name: 'a')
+        p2 = FactoryBot.create(:profile, name: 'b')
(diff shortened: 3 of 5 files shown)


app/controllers/concerns/collection.rb
```diff
@@ -23,5 +23,9 @@ def sort(data)
 
       data.order(build_order_by(data.klass, params[:sort_by]))
     end
+
+    def sorted?(data)
+      data.to_sql.include?('ORDER')
```
skateman, line 28: I haven't found a better way of detecting if the dataset is sorted or not, but if there's any, please let me know. Another alternative would be a default scope, but that doesn't really work well with `sort_by` and it's also not recommended for use for anything more complex than ASC/DESC.

vkrizan: I guess the `.order` call writes the state somewhere. Try if that can be read. Otherwise only arel comes to mi mind, traversing the tree checking for an ORDER node.

vkrizan: Wouldn't it be better to move the check to the controller level? Basically checking whether the `sort_by` was used or not. There might be a default sort set somewhere in the upper levels which could impact the default sort of the controller.

akofink: I'd implement default sort in the same way we implement [default search in some controllers](https://github.com/RedHatInsights/compliance-backend/blob/master/app/controllers/v1/profiles_controller.rb#L16).

skateman: @akofink we can't really do that here as the `score` is not a field, but a dynamic value generated from a method. There might be a way though to fetch that value via some querying magic, will look into that...

akofink: Hmm, good point. I'm now questioning why we sort by score in the first place :laughing: Since it cannot be sorted by from the user, it doesn't make sense to use it by default, since we'd have to show it in the API response meta section. I went digging, and it doesn't seem to have been added with any particular use case in mind: https://github.com/RedHatInsights/compliance-backend/pull/35/files#diff-7a47d335107fc35c2e79cc9c2b0a195c863594dd2343bcd48a4c770ac4af673cR9 I wonder if we should remove it?

skateman: I fully support this idea as it would save me from a lot of SQL thinking :exploding_head:

app/controllers/concerns/metadata.rb
```diff
@@ -62,6 +62,12 @@ def last_offset(total)
 
     private
 
+    def aggregated_count(count)
+      return count if count.is_a?(Integer)
+
+      count.values.sum
```
vkrizan, line 68: Time to upgrade to Rails 6 and use [`reselect`](https://apidock.com/rails/ActiveRecord/QueryMethods/reselect)? :smile:

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -206,9 +206,30 @@ class IndexTest < ProfilesControllerTest
...
+          profile['attributes']['name']
+        end)
+      end
+
       test 'profiles can be sorted by a single dimension' do
-        FactoryBot.create(:profile, name: 'a')
-        FactoryBot.create(:profile, name: 'b')
+        p1 = FactoryBot.create(:profile, name: 'a')
+        p2 = FactoryBot.create(:profile, name: 'b')
+        FactoryBot.create(:test_result, profile: p1, score: 0.3)
```
vkrizan, line 231: Can you please also add a test case when there would be more than one test result for a profile?

vkrizan (approved):
Nice!

Would this also apply to GraphQL?

skateman:
@vkrizan this doesn't apply to GQL yet, not sure if it should at all :thinking:

app/controllers/concerns/metadata.rb
akofink: It says:

> This is short-hand for unscope(:select).select(fields). Note that we’re unscoping the entire select statement.

So, @vkrizan, what exactly was your suggestion?

vkrizan: Oh, my bad I haven't even noticed the unscope.

The suggestion was to use a db count. Even with the `unscope` it does not give the expected results:
```
=> Profile.with_score.unscope(:select).count
{"bdb7b0a0-439d-4fb2-894d-ff31479fe78f"=>1,
 "30080442-3bd4-4950-b744-4313b2a0620f"=>1,
...
```

vkrizan:
@skateman I don't see a reason why the same sorting cannot be available in GraphQL (to have the parity). But, I'm not sure if it would be as easier to add it there as in general REST controllers.

skateman:
> @skateman I don't see a reason why the same sorting cannot be available in GraphQL (to have the parity). But, I'm not sure if it would be as easier to add it there as in general REST controllers.

Short answer: it won't :laughing: long answe: let's make an issue for that :wink:

test/controllers/v1/profiles_controller_test.rb
skateman: Sure!

app/models/profile.rb
```diff
@@ -43,6 +44,15 @@ class Profile < ApplicationRecord
   validates :account, presence: true, if: -> { hosts.any? }
   validates :policy, presence: true, if: -> { policy_id }
 
+  scope :with_score, lambda {
+    select('profiles.*', 'AVG(test_results.score) AS score')
```
akofink, line 48:
```suggestion
    select(Profile.arel_table[Arel.star], TestResult.arel_table[:score].average.as('score'))
```

akofink (commented):
I really think we ought to create a new `score` column on policy and host and update them with an `after_save` callback on TestResult to solve this the 'right way'. The fact that we can't call `Profile.with_score.count` or `Profile.with_score.pluck(:score)` makes this solution less than ideal. It's weird, too, that Rails allows `Profile.with_score.order(:score)` but doesn't allow `select` or `where` clauses on score.

There are other fields like this as well that are computed, like all the *_host_count attributes on profiles. A similar hacky solution for each field might be required unless we store the values on the policy directly. The existing APIs that serve profiles can forward attributes to the policy like we already do with name, description, compliance threshold, etc.

What do you think? Specifically for this `score` attribute, should we step back and reevaluate if we're making the best decision or move forward and merge this PR? I'd argue the new column and `after_save` hooks would still be easier to understand than this `with_score` scope. TBH, as a scope, `Profile.with_score` makes me think it's returning profiles that have some test results :laughing:

skateman:
@akofink makes sense to cache the fields, Rails even provides `counter_cache` on `belongs_to` that can be used even with custom associations with `where` clauses and I think even scopes.  Also I am dealing with the host count problem on a different front, so we can join these two...

PR #877, opened 2021-06-07
feat(reports): RHICOMPL-1236 check download URL

vkrizan:
Check schema of the report download URL to be https/secured in
production.  Fail otherwise.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/app/services/safe_downloader.rb
+++ b/app/services/safe_downloader.rb
@@ -52,7 +52,7 @@ class SafeDownloader
       Rails.logger.audit_success("Downloaded report from URL: #{url}")
       urlopen
     rescue *DOWNLOAD_ERRORS
-      Rails.logger.audit_fail("Failed to dowload report from URL: #{url}")
+      Rails.logger.audit_fail("Failed to download report from URL: #{url}")
       raise
     end
 
--- a/app/services/safe_downloader.rb
+++ b/app/services/safe_downloader.rb
@@ -58,11 +58,18 @@ class SafeDownloader
 
     def encode_url(url)
       url = URI(url)
-      raise DownloadError, 'url was invalid' unless url.respond_to?(:open)
-
+      check_url(url)
       url
     rescue ArgumentError
       raise DownloadError, 'url was invalid'
+    end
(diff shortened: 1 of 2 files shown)


app/services/safe_downloader.rb
```diff
@@ -55,11 +55,18 @@ def encode_url(url)
...
+      check_url(url)
       url
     rescue ArgumentError
       raise DownloadError, 'url was invalid'
     end
 
+    def check_url(url)
+      return unless Rails.env.production?
+
+      raise DownloadError, 'not secured (not https)' if url.scheme != 'https'
```
skateman, line 67:
```suggestion
      raise DownloadError, 'not secure (non-https)' if url.scheme != 'https'
```

app/services/safe_downloader.rb
```diff
@@ -55,11 +55,18 @@ def encode_url(url)
       url = URI(url)
       raise DownloadError, 'url was invalid' unless url.respond_to?(:open)
 
+      check_url(url)
```
skateman, line 58: I'm wondering if we could "flatten out" the exception flow here, you're introducing a new one a level deeper and it makes the code a bit less followable :thinking:

skateman (approved):
Small nits only

app/services/safe_downloader.rb
vkrizan: You mean the rerise of the same type with additional info? I've kinda used the flow that is used there already.

skateman: It's just weird that you're raising an error similarly to the previous one, just a lever deeper. But keep it so if rubocop forced you to extract this to a method :shrug:

vkrizan: Oh, I see what you mean.

PR #880, opened 2021-06-09
RHICOMPL-1899 Sort profiles by their cached score only when they are unsorted

skateman:
Replaces #873 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/jobs/delete_host.rb
+++ b/app/jobs/delete_host.rb
@@ -23,10 +23,18 @@ class DeleteHost
 
   def remove_related(host_id)
     num_removed = 0
+    profiles_to_rescore = []
     Sidekiq.logger.info("Deleting related records for host #{host_id}")
     MODELS.each do |model|
-      num_removed += model.where(host_id: host_id).delete_all
+      to_remove = model.where(host_id: host_id)
+      # Mark profile IDs as to be rescored if the model is TestResult
+      profiles_to_rescore = to_remove.pluck(:profile_id) if model == TestResult
+      num_removed += to_remove.delete_all
     end
+
+    # Rescore all marked profiles in batches
+    Profile.where(id: profiles_to_rescore.uniq).find_each(&:calculate_score!)
+
     num_removed
   end
 
--- a/app/models/concerns/profile_policy_scoring.rb
+++ b/app/models/concerns/profile_policy_scoring.rb
@@ -3,10 +3,12 @@
 # Methods that are related to computing the score of a profile or a policy
 module ProfilePolicyScoring
   def score(host: nil)
-    results = test_results.latest
-    results = results.where(host: host) if host
-    return 0 if results.blank?
+    return super().to_f unless host
 
-    ((scores = results.pluck(:score)).sum / scores.size).to_f
+    test_results.latest.where(host: host).average(:score).to_f
+  end
+
+  def calculate_score!(*_args)
+    update!(score: test_results.latest.average(:score).to_f)
   end
 end
--- a/app/models/test_result.rb
+++ b/app/models/test_result.rb
@@ -16,6 +16,9 @@ class TestResult < ApplicationRecord
   validates :end_time, presence: true,
                        uniqueness: { scope: %i[host_id profile_id] }
 
+  after_save :update_profile_score!
+  after_destroy :update_profile_score!
+
   scope :latest, lambda {
     joins("JOIN (#{latest_without_ids.to_sql}) as tr on "\
           'test_results.profile_id = tr.profile_id AND '\
@@ -30,5 +33,9 @@ class TestResult < ApplicationRecord
   def self.latest_without_ids
     group(:profile_id, :host_id)
       .select(:profile_id, :host_id, 'MAX(end_time) as end_time')
+  end
+
+  def update_profile_score!
+    profile.nil? || profile.calculate_score!
   end
 end
--- a/db/migrate/20210609081621_add_score_to_profiles.rb
+++ b/db/migrate/20210609081621_add_score_to_profiles.rb
@@ -1,0 +1,28 @@
+class AddScoreToProfiles < ActiveRecord::Migration[5.2]
+  def up
+    add_column :profiles, :score, :decimal
+
+    Profile.transaction do
+      scores = TestResult.latest.joins(:profile).group(:profile_id).average(:score)
+
+      # Profile.canonical(false).where(score: nil).joins(:test_results).distinct.find_each(&:calculate_score!)
+      # TODO: replace this with upsert after upgrading to Rails 6
+      values = scores.map do |k, v|
+        "(#{ActiveRecord::Base.connection.quote(k)}, #{ActiveRecord::Base.connection.quote(v)})"
+      end.join(', ')
+
+      query = %{
+        UPDATE profiles P
+        SET score = uv.new_score
+        FROM (VALUES #{values}) AS uv (id, new_score)
+        WHERE P.id = uv.id::uuid
+      }.gsub(/\s+/, " ").strip
+
+      ActiveRecord::Base.connection.execute(query)
+    end
+  end
+
+  def down
+    remove_column :profiles, :score
+  end
+end
--- a/app/controllers/v1/profiles_controller.rb
+++ b/app/controllers/v1/profiles_controller.rb
@@ -14,7 +14,8 @@ module V1
 
     def index
       params[:search] ||= 'external=false and canonical=false'
-      render_json resolve_collection.sort_by(&:score)
+      params[:sort_by] ||= 'score'
+      render_json resolve_collection
     end
 
     def show
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -12,7 +12,8 @@ class Profile < ApplicationRecord
 
   SORTABLE_BY = {
     name: :name,
-    os_minor_version: :os_minor_version
+    os_minor_version: :os_minor_version,
+    score: :score
   }.freeze
 
   belongs_to :account, optional: true
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_05_17_182143) do
+ActiveRecord::Schema.define(version: 2021_06_09_081621) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "dblink"
@@ -96,6 +96,7 @@ ActiveRecord::Schema.define(version: 2021_05_17_182143) do
     t.boolean "external", default: false, null: false
     t.uuid "policy_id"
     t.string "os_minor_version", default: "", null: false
+    t.decimal "score"
     t.index ["account_id"], name: "index_profiles_on_account_id"
     t.index ["external"], name: "index_profiles_on_external"
     t.index ["name"], name: "index_profiles_on_name"
--- a/test/controllers/v1/profiles_controller_test.rb
+++ b/test/controllers/v1/profiles_controller_test.rb
@@ -206,9 +206,31 @@ module V1
                      profiles.dig('data', 0, 'attributes', 'policy_profile_id')
       end
 
+      test 'profiles are sorted by score by default' do
+        p1 = FactoryBot.create(:profile, name: 'a')
+        p2 = FactoryBot.create(:profile, name: 'b')
+        FactoryBot.create(:test_result, profile: p1, score: 0.3)
+        FactoryBot.create(:test_result, profile: p2, score: 0.5)
+
+        get v1_profiles_url, params: {
+          search: 'canonical=false'
+        }
+
+        assert_response :success
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(%w[a b], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
       test 'profiles can be sorted by a single dimension' do
-        FactoryBot.create(:profile, name: 'a')
-        FactoryBot.create(:profile, name: 'b')
+        p1 = FactoryBot.create(:profile, name: 'a')
+        p2 = FactoryBot.create(:profile, name: 'b')
+        FactoryBot.create(:test_result, profile: p1, score: 0.4)
+        FactoryBot.create(:test_result, profile: p1, score: 0.1)
+        FactoryBot.create(:test_result, profile: p2, score: 0.5)
 
         get v1_profiles_url, params: {
           search: 'canonical=false',
--- a/test/jobs/delete_host_test.rb
+++ b/test/jobs/delete_host_test.rb
@@ -7,11 +7,11 @@ class DeleteHostTest < ActiveSupport::TestCase
   setup do
     user = FactoryBot.create(:user)
     @host = FactoryBot.create(:host, account: user.account.account_number)
-    profile = FactoryBot.create(:profile, :with_rules, account: user.account)
-    tr = FactoryBot.create(:test_result, profile: profile, host: @host)
+    @profile = FactoryBot.create(:profile, :with_rules, account: user.account)
+    tr = FactoryBot.create(:test_result, profile: @profile, host: @host)
     FactoryBot.create(
       :rule_result,
-      rule: profile.rules.sample,
+      rule: @profile.rules.sample,
       test_result: tr, host: @host
     )
 
@@ -28,11 +28,13 @@ class DeleteHostTest < ActiveSupport::TestCase
   end
 
   test 'deletes a host if the passed ID is found' do
+    assert_not_equal(@profile.reload.score, 0)
     DeleteHost.perform_async(@message)
     assert_equal 1, DeleteHost.jobs.size
     assert_difference('RuleResult.count' => -1, 'TestResult.count' => -1) do
       DeleteHost.drain
     end
+    assert_equal(@profile.reload.score, 0)
     assert_audited 'Deleteted related records for host'
     assert_audited @host.id
   end


skateman:
@vkrizan this also takes care of the GQL sorting

app/models/concerns/profile_policy_scoring.rb
```diff
@@ -3,10 +3,12 @@
 # Methods that are related to computing the score of a profile or a policy
 module ProfilePolicyScoring
   def score(host: nil)
-    results = test_results.latest
-    results = results.where(host: host) if host
-    return 0 if results.blank?
+    return super().to_f unless host
```
vkrizan, line 6: How does this work? The `super` is supposed to be the `score` field on `Profile` which this module method overrides?

```
> Profile.first.score
  Profile Load (1.6ms)  SELECT  "profiles".* FROM "profiles" ORDER BY "profiles"."id" ASC LIMIT $1  [["LIMIT", 1]]
NoMethodError: super: no superclass method `score' for #<Profile:0x00007f45c88582b0>
from /home/vkrizan/.gem/gems/activemodel-5.2.4.5/lib/active_model/attribute_methods.rb:427:in `method_missing'
```

app/models/test_result.rb
```diff
@@ -16,6 +16,9 @@ class TestResult < ApplicationRecord
   validates :end_time, presence: true,
                        uniqueness: { scope: %i[host_id profile_id] }
 
+  after_save :update_profile_score!
+  after_destroy :update_profile_score!
```
vkrizan, line 20: Please check if this would not have a high impact on (bulk) Host deletion triggering a TestResult cleanup.

db/migrate/20210609081621_add_score_to_profiles.rb
```diff
@@ -0,0 +1,10 @@
+class AddScoreToProfiles < ActiveRecord::Migration[5.2]
+  def up
+    add_column :profiles, :score, :decimal, null: false, default: 0
+    Profile.canonical(false).joins(:test_results).find_each(&:calculate_score!)
```
vkrizan, line 4: Are you sure the `joins` is a right choice? For a profile with N test results wouldn't it call calculate score N times?

app/models/concerns/profile_policy_scoring.rb
skateman: I'm trying to keep compatibility with the old `score(host)` method. If the `host` argument is not passed, we're falling back to the DB column, otherwise we're calculating the score for only the given host.

app/models/test_result.rb
skateman: Well, I'm a bit worried that it will cause an `N-1` but I have to think a bit more about how to avoid that :confused:

db/migrate/20210609081621_add_score_to_profiles.rb
skateman: My goal was to eliminate profiles without test results but I guess I'm not doing that the right way. I am wondering if a simple `uniq` would fix this :thinking:

skateman: :see_no_evil: `distinct` :see_no_evil:

app/models/test_result.rb
skateman: Okay, so the [`delete_all`](https://github.com/RedHatInsights/compliance-backend/blob/master/app/jobs/delete_host.rb#L28) method here is skipping all the callbacks [according to the docs](https://guides.rubyonrails.org/active_record_callbacks.html#skipping-callbacks) so this should be safe as it is.

app/models/concerns/profile_policy_scoring.rb
akofink: Nit: parenthesis are unnecessary in ruby :)

app/models/test_result.rb
vkrizan: Wouldn't that leave the old score cached / not recalculated?

app/models/concerns/profile_policy_scoring.rb
skateman: Nope, `super()` and `super` aren't the same things in Ruby

akofink: Could you explain? Does `super` forward params and `super()` not forward params or something?

skateman: Yes, as you said

app/models/test_result.rb
skateman: True, so what about building a set of affected profile IDs in the `remove_related` method and batch-updating their score?

app/models/concerns/profile_policy_scoring.rb
vkrizan: This is super :smile:

vkrizan: I was interested in the method resolution that is done here. `ProfilePolicyScoring` is a module that is being included in a class, that overrides the dynamically added method (added by Rails). This looks like a composition rather than an inheritance, hence my curiosity.

app/models/test_result.rb
vkrizan: > True, so what about building a set of affected profile IDs in the `remove_related` method and batch-updating their score?

That sound great!

skateman: Sounds good, doesn't work :laughing: at least it might introduce useless computation. Depends on how many hosts are we deleting at once? :thinking:

akofink: Keep in mind the DeleteHostsJob. It uses `delete` and not `destroy`, so the score would need to be updated if the host had any test results in Compliance. An alternative approach would be to just 'invalidate' the score by setting it to `null` and only re-calculating it when it's requested. I'm not suggesting we do this, just suggesting we think about if it would be a better solution :)

skateman: @akofink already solved that, but I'm concerned about the performance when deleting a lot of hosts :thinking:

app/models/concerns/profile_policy_scoring.rb
skateman: Ruby is magic :laughing:

skateman:
@akofink @vkrizan can I get a :green_circle: here?

db/migrate/20210609081621_add_score_to_profiles.rb
```diff
@@ -0,0 +1,10 @@
+class AddScoreToProfiles < ActiveRecord::Migration[5.2]
+  def up
+    add_column :profiles, :score, :decimal
+    Profile.canonical(false).where(score: nil).joins(:test_results).distinct.find_each(&:calculate_score!)
```
vkrizan, line 4: This will likely be very heavy on the database. I have 9 non-canonical profiles (only one host assigned) and it does 16 queries:
```
[2] pry(main)> Profile.canonical(false).joins(:test_results).distinct.find_each(&:calculate_score!)
  Profile Load (2.0ms)  SELECT  DISTINCT "profiles".* FROM "profiles" INNER JOIN "test_results" ON "test_results"."profile_id" = "profiles"."id" WHERE "profiles"."parent_profile_id" IS NOT NULL ORDER BY "profiles"."id" ASC LIMIT $1  [["LIMIT", 1000]]
   (0.8ms)  SELECT AVG("test_results"."score") FROM "test_results" JOIN (SELECT "test_results"."profile_id", "test_results"."host_id", MAX(end_time) as end_time FROM "test_results" WHERE "test_results"."profile_id" = 'df0d257f-169c-473e-a4cb-82b854c3f342' GROUP BY "test_results"."profile_id", "test_results"."host_id") as tr on test_results.profile_id = tr.profile_id AND test_results.host_id = tr.host_id AND test_results.end_time = tr.end_time WHERE "test_results"."profile_id" = $1  [["profile_id", "df0d257f-169c-473e-a4cb-82b854c3f342"]]
   (0.2ms)  BEGIN
  Xccdf::Benchmark Load (0.4ms)  SELECT  "benchmarks".* FROM "benchmarks" WHERE "benchmarks"."id" = $1 LIMIT $2  [["id", "e7f39c61-40aa-4fe7-b8db-70b6764ca3e6"], ["LIMIT", 1]]
  Profile Exists (0.6ms)  SELECT  1 AS one FROM "profiles" WHERE "profiles"."ref_id" = $1 AND "profiles"."id" != $2 AND "profiles"."account_id" = $3 AND "profiles"."benchmark_id" = $4 AND "profiles"."os_minor_version" = $5 AND "profiles"."policy_id" = $6 LIMIT $7  [["ref_id", "xccdf_org.ssgproject.content_profile_C2S"], ["id", "df0d257f-169c-473e-a4cb-82b854c3f342"], ["account_id", "7f544c34-8d4d-4edb-8b77-09bb6ce0f8f7"], ["benchmark_id", "e7f39c61-40aa-4fe7-b8db-70b6764ca3e6"], ["os_minor_version", ""], ["policy_id", "e29c54de-6e9e-48ec-8bca-0e391b5e3b10"], ["LIMIT", 1]]
  Host Exists (12.9ms)  SELECT  1 AS one FROM "inventory"."hosts" INNER JOIN "test_results" ON "inventory"."hosts"."id" = "test_results"."host_id" WHERE "test_results"."profile_id" = $1 LIMIT $2  [["profile_id", "df0d257f-169c-473e-a4cb-82b854c3f342"], ["LIMIT", 1]]
  Account Load (0.4ms)  SELECT  "accounts".* FROM "accounts" WHERE "accounts"."id" = $1 LIMIT $2  [["id", "7f544c34-8d4d-4edb-8b77-09bb6ce0f8f7"], ["LIMIT", 1]]
  Policy Load (0.3ms)  SELECT  "policies".* FROM "policies" WHERE "policies"."id" = $1 LIMIT $2  [["id", "e29c54de-6e9e-48ec-8bca-0e391b5e3b10"], ["LIMIT", 1]]
   (0.3ms)  COMMIT
   (0.6ms)  SELECT AVG("test_results"."score") FROM "test_results" JOIN (SELECT "test_results"."profile_id", "test_results"."host_id", MAX(end_time) as end_time FROM "test_results" WHERE "test_results"."profile_id" = 'ed19d644-d57e-47cc-9e6a-9f46729f2482' GROUP BY "test_results"."profile_id", "test_results"."host_id") as tr on test_results.profile_id = tr.profile_id AND test_results.host_id = tr.host_id AND test_results.end_time = tr.end_time WHERE "test_results"."profile_id" = $1  [["profile_id", "ed19d644-d57e-47cc-9e6a-9f46729f2482"]]
   (0.2ms)  BEGIN
  Xccdf::Benchmark Load (0.2ms)  SELECT  "benchmarks".* FROM "benchmarks" WHERE "benchmarks"."id" = $1 LIMIT $2  [["id", "e7f39c61-40aa-4fe7-b8db-70b6764ca3e6"], ["LIMIT", 1]]
  Profile Exists (0.4ms)  SELECT  1 AS one FROM "profiles" WHERE "profiles"."ref_id" = $1 AND "profiles"."id" != $2 AND "profiles"."account_id" = $3 AND "profiles"."benchmark_id" = $4 AND "profiles"."os_minor_version" = $5 AND "profiles"."policy_id" = $6 LIMIT $7  [["ref_id", "xccdf_org.ssgproject.content_profile_hipaa"], ["id", "ed19d644-d57e-47cc-9e6a-9f46729f2482"], ["account_id", "7f544c34-8d4d-4edb-8b77-09bb6ce0f8f7"], ["benchmark_id", "e7f39c61-40aa-4fe7-b8db-70b6764ca3e6"], ["os_minor_version", ""], ["policy_id", "ea50e400-a460-4cce-b505-ca1fa70b599c"], ["LIMIT", 1]]
  Host Exists (11.0ms)  SELECT  1 AS one FROM "inventory"."hosts" INNER JOIN "test_results" ON "inventory"."hosts"."id" = "test_results"."host_id" WHERE "test_results"."profile_id" = $1 LIMIT $2  [["profile_id", "ed19d644-d57e-47cc-9e6a-9f46729f2482"], ["LIMIT", 1]]
  Account Load (0.5ms)  SELECT  "accounts".* FROM "accounts" WHERE "accounts"."id" = $1 LIMIT $2  [["id", "7f544c34-8d4d-4edb-8b77-09bb6ce0f8f7"], ["LIMIT", 1]]
  Policy Load (0.3ms)  SELECT  "policies".* FROM "policies" WHERE "policies"."id" = $1 LIMIT $2  [["id", "ea50e400-a460-4cce-b505-ca1fa70b599c"], ["LIMIT", 1]]
   (0.2ms)  COMMIT
   (0.6ms)  SELECT AVG("test_results"."score") FROM "test_results" JOIN (SELECT "test_results"."profile_id", "test_results"."host_id", MAX(end_time) as end_time FROM "test_results" WHERE "test_results"."profile_id" = 'faf7e0b2-f02d-4f6f-b530-473bb1aef85f' GROUP BY "test_results"."profile_id", "test_results"."host_id") as tr on test_results.profile_id = tr.profile_id AND test_results.host_id = tr.host_id AND test_results.end_time = tr.end_time WHERE "test_results"."profile_id" = $1  [["profile_id", "faf7e0b2-f02d-4f6f-b530-473bb1aef85f"]]
   (0.2ms)  BEGIN
  Xccdf::Benchmark Load (0.3ms)  SELECT  "benchmarks".* FROM "benchmarks" WHERE "benchmarks"."id" = $1 LIMIT $2  [["id", "e7f39c61-40aa-4fe7-b8db-70b6764ca3e6"], ["LIMIT", 1]]
  Profile Exists (0.4ms)  SELECT  1 AS one FROM "profiles" WHERE "profiles"."ref_id" = $1 AND "profiles"."id" != $2 AND "profiles"."account_id" = $3 AND "profiles"."benchmark_id" = $4 AND "profiles"."os_minor_version" = $5 AND "profiles"."policy_id" = $6 LIMIT $7  [["ref_id", "xccdf_org.ssgproject.content_profile_cis"], ["id", "faf7e0b2-f02d-4f6f-b530-473bb1aef85f"], ["account_id", "7f544c34-8d4d-4edb-8b77-09bb6ce0f8f7"], ["benchmark_id", "e7f39c61-40aa-4fe7-b8db-70b6764ca3e6"], ["os_minor_version", ""], ["policy_id", "f43f2b47-5afb-4b3d-ab81-3a746863bb76"], ["LIMIT", 1]]
  Host Exists (11.2ms)  SELECT  1 AS one FROM "inventory"."hosts" INNER JOIN "test_results" ON "inventory"."hosts"."id" = "test_results"."host_id" WHERE "test_results"."profile_id" = $1 LIMIT $2  [["profile_id", "faf7e0b2-f02d-4f6f-b530-473bb1aef85f"], ["LIMIT", 1]]
  Account Load (0.5ms)  SELECT  "accounts".* FROM "accounts" WHERE "accounts"."id" = $1 LIMIT $2  [["id", "7f544c34-8d4d-4edb-8b77-09bb6ce0f8f7"], ["LIMIT", 1]]
  Policy Load (0.3ms)  SELECT  "policies".* FROM "policies" WHERE "policies"."id" = $1 LIMIT $2  [["id", "f43f2b47-5afb-4b3d-ab81-3a746863bb76"], ["LIMIT", 1]]
   (0.1ms)  COMMIT
```

vkrizan (commented):
Implementation-wise looks good to me.
As for the migration, we might run into some issues when running it on a bigger set.

db/migrate/20210609081621_add_score_to_profiles.rb
skateman: This one runs only once, but I'm open for any kind of optimizations here.

akofink (approved):
Looks great, thanks! I agree with Viliam that we should be careful to test this with a large data set before deploying to prod. The run-migrations initContainer will block the deployment for as long as it takes to run the migration.

app/jobs/delete_host.rb
```diff
@@ -23,10 +23,18 @@ def perform(message)
 
   def remove_related(host_id)
     num_removed = 0
+    to_rescore = Set.new # Keep a set of profiles that need to be rescored
     Sidekiq.logger.info("Deleting related records for host #{host_id}")
     MODELS.each do |model|
-      num_removed += model.where(host_id: host_id).delete_all
+      to_remove = model.where(host_id: host_id)
+      # Mark profile IDs as to be rescored if the model is TestResult
+      to_rescore.merge(to_remove.pluck(:profile_id)) if model == TestResult
```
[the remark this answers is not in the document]
skateman: Yeah, that didn't feel too DRY to me, but I'll change :+1:

db/migrate/20210609081621_add_score_to_profiles.rb
skateman: So the only truly efficient way to optimize this into two queries needs to use some raw sql :disappointed: which might change if we upgrade to Rails 6 as it supports `upsert`. It probably has to be wrapped around a transaction and needs to be prettified, but:
```ruby
scores = TestResult.latest.joins(:profile).group(:profile_id).average(:score)
values = scores.map { |k, v| "(#{ActiveRecord::Base.connection.quote(k)}, #{ActiveRecord::Base.connection.quote(v)})" }.join(', ')

query = "
    UPDATE profiles T
    SET score = uv.new_score
    FROM (VALUES #{values}) AS uv (id, new_score)
    WHERE T.id = uv.id::uuid"

ActiveRecord::Base.connection.execute(query)
```
@akofink what do you think?

akofink: I'm OK with using raw SQL like this for a one-off migration if it's our only option. Let me investigate it a bit more - I haven't looked especially closely at this work. I will get back to you :)

app/jobs/delete_host.rb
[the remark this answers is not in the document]
skateman: So here's the thing, the order in deletion is important if we don't want to have DB integrity problems, so the test results should be deleted AFTER the rule results and the policy hosts links should be the last (I guess). It would not fire an error as the validations/callbacks aren't running, but it's not healthy as it's no longer wrapped in a transaction.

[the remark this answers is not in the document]
akofink: Yes, that's a good point :laughing: I can see how this refactor might be difficult. I wouldn't block this PR on it - we can always go back and refactor this code; it's at least contained within this single file.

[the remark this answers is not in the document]
skateman: Just warning you @akofink , the solution will be a bit esoteric :laughing:

akofink:
I'm mostly testing this for performance, then we can merge it. :)

skateman:
I'll push soon the new version with the query + the esoteric refactor :laughing:

db/migrate/20210609081621_add_score_to_profiles.rb
akofink: I tried benchmarking this original implementation, and it doesn't seem to be that problematic:

```
$ time podman exec -it compliance-backend_prometheus_1 bundle exec rake db:migrate
2021-06-18T17:17:48.125Z 5300 TID-8go INFO: GitLab reliable fetch activated!
== 20210609081621 AddScoreToProfiles: migrating ===============================
-- add_column(:profiles, :score, :decimal)
   -> 0.0009s
== 20210609081621 AddScoreToProfiles: migrated (0.3996s) ======================

podman exec -it compliance-backend_prometheus_1 bundle exec rake db:migrate  0.11s user 0.14s system 10% cpu 2.365 total
```

Under 2.5 seconds for 6033 hosts with 5930 latest test results.

app/models/concerns/profile_policy_scoring.rb
```diff
@@ -3,10 +3,12 @@
...
 
-    ((scores = results.pluck(:score)).sum / scores.size).to_f
+    test_results.latest.where(host: host).average(:score).to_f
+  end
+
+  def calculate_score!(*_args)
+    update(score: test_results.latest.average(:score).to_f)
```
akofink, line 12: This will quietly fail. We tend to prefer `update!` so we can catch if something goes wrong (and if it does for some certain cases, we can write an exception handler).

akofink (approved):
Thanks, @skateman! Let's merge this and try to test at scale in stage or QA.

PR #883, opened 2021-06-14
feat(cache): RHICOMPL-1706 store host count for policies in fields

skateman:
This is just a super naive implementation that somehow works :shrug: :laughing: 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/profile_policy_scoring.rb
+++ b/app/models/concerns/profile_policy_scoring.rb
@@ -11,4 +11,31 @@ module ProfilePolicyScoring
   def calculate_score!(*_args)
     update!(score: test_results.latest.average(:score).to_f)
   end
+
+  def update_counters!
+    # rubocop:disable Rails/SkipsModelValidations
+    update_columns(
+      total_host_count: hosts.count,
+      test_result_host_count: latest_supported_policy_test_result_hosts.count,
+      compliant_host_count: latest_supported_policy_test_result_hosts
+        .count { |host| compliant?(host) },
+      unsupported_host_count: latest_unsupported_policy_test_result_hosts.count
+    )
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+
+  private
+
+  def latest_supported_policy_test_result_hosts
+    ::Host.where(id: latest_test_results.supported.select(:host_id))
+  end
+
+  def latest_unsupported_policy_test_result_hosts
+    ::Host.where(id: latest_test_results.supported(false).select(:host_id))
+  end
+
+  def latest_test_results
+    ::TestResult.where(id: test_results.pluck(:id))
+                .or(::TestResult.where(profile_id: profiles.pluck(:id))).latest
+  end
 end
--- a/app/models/concerns/profile_scoring.rb
+++ b/app/models/concerns/profile_scoring.rb
@@ -5,25 +5,19 @@ module ProfileScoring
   include ProfilePolicyScoring
 
   def total_host_count
-    if policy
-      ::Host.where(id: assigned_hosts).count
-    else
-      ::Host.where(id: hosts).count
-    end
+    policy&.total_host_count.to_i || ::Host.where(id: hosts).count
   end
 
   def test_result_host_count
-    latest_supported_policy_test_result_hosts.count
+    policy&.test_result_host_count.to_i
   end
 
   def compliant_host_count
-    latest_supported_policy_test_result_hosts.count { |host| compliant?(host) }
+    policy&.compliant_host_count.to_i
   end
 
   def unsupported_host_count
-    ::Host.where(
-      id: latest_policy_test_results.supported(false).select(:host_id)
-    ).count
+    policy&.unsupported_host_count.to_i
   end
 
   def compliance_score(host)
@@ -48,17 +42,5 @@ module ProfileScoring
         %w[pass notapplicable notselected].include? rule_result.result
       end
     end
-  end
-
-  private
-
-  def latest_supported_policy_test_result_hosts
-    ::Host.where(id: latest_policy_test_results.supported.select(:host_id))
-  end
-
-  def latest_policy_test_results
-    ::TestResult.where(id: test_results)
-                .or(::TestResult.where(id: policy_test_results))
-                .latest
   end
 end
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -27,6 +27,8 @@ class Policy < ApplicationRecord
   after_update :destroy_orphaned_business_objective
   after_rollback :destroy_orphaned_business_objective
 
+  after_save :update_counters!
+
   scope :with_hosts, lambda { |hosts|
     joins(:hosts).where(hosts: { id: hosts }).distinct
   }
@@ -52,6 +54,7 @@ class Policy < ApplicationRecord
     removed = policy_hosts.where.not(host_id: new_host_ids).destroy_all
     imported = PolicyHost.import_from_policy(id, new_host_ids - host_ids)
     update_os_minor_versions
+    update_counters!
 
     [imported.ids.count, removed.count]
   end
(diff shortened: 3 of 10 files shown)


app/models/concerns/profile_policy_scoring.rb
```diff
@@ -9,4 +9,34 @@ def score(host: nil)
...
+  def latest_unsupported_policy_test_result_hosts
+    ::Host.where(id: latest_policy_test_results
+      .supported(false).select(:host_id))
+  end
+
+  def latest_policy_test_results
+    ::TestResult
+      .where(id: test_results.pluck(:id))
+      .or(::TestResult.where(profile_id: profiles.pluck(:id)))
+      .latest
```
akofink, line 40: I'd argue we should add this as a relationship:

```suggestion
    has_many :latest_test_results, -> { latest }, through: :profiles, source: :test_results
```

app/models/concerns/profile_policy_scoring.rb
```diff
@@ -9,4 +9,34 @@ def score(host: nil)
...
+        .count { |host| compliant?(host) },
+      unsupported_host_count: latest_unsupported_policy_test_result_hosts.count
+    )
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+
+  private
+
+  def latest_supported_policy_test_result_hosts
+    ::Host.where(id: latest_policy_test_results.supported.select(:host_id))
```
akofink, line 28: Technically, we shouldn't need to query hosts for this count. Since we're within a single policy, it should be the same as the number of latest supported test results. If you want to be sure, you can use `count(:host_id)`.

akofink (commented):
Looking good! I have a few early comments, but the reasoning seems sound to me.

app/models/concerns/profile_scoring.rb
```diff
@@ -5,25 +5,19 @@ module ProfileScoring
...
   def compliant_host_count
-    latest_supported_policy_test_result_hosts.count { |host| compliant?(host) }
+    policy&.compliant_host_count.to_i
   end
 
   def unsupported_host_count
-    ::Host.where(
-      id: latest_policy_test_results.supported(false).select(:host_id)
-    ).count
+    policy&.unsupported_host_count.to_i
```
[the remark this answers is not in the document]
skateman: Yeah, no :laughing: it crashes on GQL when it's `NULL` and I wasn't sure if I can set `null: false` in the type :thinking:

app/models/concerns/profile_policy_scoring.rb
```diff
@@ -9,4 +9,34 @@ def score(host: nil)
...
+  def update_counters!
+    # rubocop:disable Rails/SkipsModelValidations
+    update_columns(
+      total_host_count: hosts.count,
+      test_result_host_count: latest_supported_policy_test_result_hosts.count,
+      compliant_host_count: latest_supported_policy_test_result_hosts
+        .count { |host| compliant?(host) },
+      unsupported_host_count: latest_unsupported_policy_test_result_hosts.count
+    )
+    # rubocop:enable Rails/SkipsModelValidations
```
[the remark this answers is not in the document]
skateman: Disabling the validations is just a side effects, we can't allow callbacks here as it would end up in an endless recursion.

vkrizan:
@skateman How easy/hard would be to keep the cached fields up to date? Are there other options how to cache these values to prevent them being out of sync?

app/models/concerns/profile_policy_scoring.rb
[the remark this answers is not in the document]
skateman: Also it updates the `updated_at` which we probably don't want :thinking:

app/models/concerns/profile_policy_scoring.rb
skateman: We no longer need this as I went with the `has_many :through`

skateman:
@vkrizan my only idea would be to use counter_cache, but I don't think it's more effective than this one and I haven't seen any example of using it for more complex `has_many_through_but_with_a_twist` associations :thinking: storing the counts in Redis doesn't really increase the performance of our queries as we need these numbers together with the policy/profile, so Redis would actually produce more overhead when querying.

app/models/concerns/profile_policy_scoring.rb
skateman: Yeah, this doesn't work as I thought it would work :sweat_smile:

app/models/concerns/profile_policy_scoring.rb
skateman: Yeah, gonna simplify a bit

skateman: Nope, it won't work :disappointed: we need the host because of the `compliant?` call

app/models/concerns/profile_scoring.rb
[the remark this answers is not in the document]
akofink: Makes sense!

db/migrate/20210609114653_add_cached_counts.rb
```diff
@@ -0,0 +1,17 @@
+class AddCachedCounts < ActiveRecord::Migration[5.2]
+  def up
+    add_column :policies, :total_host_count, :integer
+    add_column :policies, :test_result_host_count, :integer
+    add_column :policies, :compliant_host_count, :integer
+    add_column :policies, :unsupported_host_count, :integer
+
+    Policy.find_each(&:update_counters!)
+  end
```
akofink, line 9: Did I make some comment earlier about these being `null: false`? I can't remember if it was this PR or another one, but don't you think it makes more sense to have a default value of 0 for these counts and force them to be non-null? I think my previous comment was about score, which doesn't make as much sense as 0 when there's no data.

skateman: The other one, but I can definitely put the `where` here.

akofink: Well, I mean, don't you think this migration makes sense to add `null: false, default: 0` or similar? Host counts being 0 by default make more sense than score being 0 by default.

skateman: True

skateman: Still an issue as the `policy` can be `nil` and then these fields ale `nil` as well and GQL doesn't like it, even if I do `allow_nil` in the delegation. So yeah, I do need the `to_i` but I will keep the `default: 0` as I like this :+1:

app/models/concerns/profile_scoring.rb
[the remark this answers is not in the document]
akofink: Since we've added `null: false` in the database, can't we then remove these `to_i`s?

akofink (commented):
Thanks, @skateman! I only found one issue, and I think we're good to merge it :)

app/models/policy.rb
```diff
@@ -27,6 +27,8 @@ class Policy < ApplicationRecord
   after_update :destroy_orphaned_business_objective
   after_rollback :destroy_orphaned_business_objective
 
+  after_save :update_counters!
```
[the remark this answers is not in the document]
skateman: @akofink updated + adjusted the existing tests to cover this and fail accordingly

app/models/concerns/profile_scoring.rb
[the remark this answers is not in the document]
skateman: No as this would crash for canonical profiles :confused:

akofink (approved):
Thanks, @skateman!

PR #894, opened 2021-06-18
fix(devel): RHICOMPL-1947 run the test db prepare in the test env

skateman:
So apparently when `db:test:prepare` runs in any other environment, it messes up stuff :laughing: who would think that :see_no_evil: 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -143,7 +143,7 @@ services:
         bundle exec rake db:setup;
       fi;
       RAILS_ENV=test bundle exec rake db:create;
-      bundle exec rake db:test:prepare;
+      RAILS_ENV=test bundle exec rake db:test:prepare;
       bundle exec rails db < db/cyndi_setup_devel.sql;
       RAILS_ENV=test bundle exec rails db < db/cyndi_setup_test.sql;
       '


docker-compose.yml
```diff
@@ -143,7 +143,7 @@ services:
         bundle exec rake db:setup;
       fi;
       RAILS_ENV=test bundle exec rake db:create;
-      bundle exec rake db:test:prepare;
+      RAILS_ENV=test bundle exec rake db:test:prepare;
```
Victoremepunto, line 146: Can we better define a env variable `RAILS_ENV`set to `test` in this pod ? I see this variable set for multiple commands already

Victoremepunto (changes requested):
makes sense.

docker-compose.yml
skateman: I don't like that idea, the purpose of this container is to set up the development environment and on top of that also sets up the testing one. I would rather keep the latter as explicit

skateman: In that case we would have to set the `RAILS_ENV=development` on the rest of the commands, even in the `if` statement above and the word itself is longer so I think this is simpler :laughing:

Victoremepunto: fine by me :+1:  

but I think that'd be explicit enough , the template would look like similar to:

```YAML
  build-rails-db:
    environment:
        - RAILS_ENV=test
    command: sh -c '
      if bundle exec rake db:version 2> /dev/null; then
        bundle exec rake db:migrate;
      else
        bundle exec rake db:setup;
      fi;
      bundle exec rake db:create;
      bundle exec rake db:test:prepare;
      bundle exec rails db < db/cyndi_setup_devel.sql;
      bundle exec rails db < db/cyndi_setup_test.sql;
      '
```
Though it's true it would affect the three other rake commands `db:version`, `db:migrate` and `db:setup` which - I don't know if that's okay either.

Victoremepunto (approved):

docker-compose.yml
skateman: Yeah, those would need the development env...

PR #902, opened 2021-06-23
devel(seeds): RHICOMPL-1328 introducing a seeds script

skateman:
To test this you need a clean DB with a single account (yours) and to start the seeding process, run the following command:
```
bin/rake db:seed
```
The script will generate:
* some hosts, with OS versions based on the hash
* all the supported profiles for the given major versions pulled out from hosts
* tailored child profiles with a slightly less amount (random -10) of rules
* up to 3 test results for each profile/host combo with up to half of the hosts passing all rules
* rule results that are matching the score of the test results (pass/fail ratio)


akofink:
Rails seeds are usually meant for required DB setup (kind of like how we sync SSG datastreams). We don't really have any static 'setup' data like this, but I also don't want it to be confusing for future devs. Could we define a whole new task for this instead, maybe db:devel:seed or devel:seed or idk, whatever you think :laughing: ?

db/seeds.rb
```diff
@@ -3,3 +3,47 @@
 # with its default values.
 # The data can then be loaded with the rails db:seed command (or created
 # alongside the database with db:setup).
+
+account = Account.first!
```
akofink, line 7: Part of what I'd really like to see with this work is prod-like data, meaning many accounts with a representative set of policies, each with some largish number of hosts assigned (like 2 to 1000 hosts per policy) and with at least some significant portion of hosts reporting across different SSG versions with different levels of supported/unsupported and compliant/non-compliant.

skateman: Yeah, I was trying to approach this from the UI perspective, so generate stuff that the user can navigate through. But your perspective also makes sense, I guess we can do both...

db/seeds.dev.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+FactoryBot.create_list(:account, 3)
+
+Account.find_each do |account|
+  host_ids = SupportedSsg.all.map { |s| [s.os_major_version, s.os_minor_version] }.uniq.flat_map do |version|
+    FactoryBot.create_list(:host, rand(2..3), account: account.account_number, os_version_arr: version).map(&:id)
```
bastilian, line 7: Would it make sense to use `ENV` vars or cli arguments for the `rand`, like `MIN_HOSTS`/`MAX_HOSTS`?

db/seeds.dev.rb
```diff
@@ -0,0 +1,60 @@
...
+    policy = Policy.new(profile.slice(:name, :description, :compliance_threshold, :account))
+    policy.save!
+    profile.update!(policy: policy)
+
+    policy.update_hosts(hosts.pluck(:id))
+
+    passing_hosts = policy.hosts.sample(rand(1..policy.hosts.count / 2))
+
+    hosts.find_each do |host|
+      next if [true, true, false].sample
```
bastilian, line 26: ...curious, is this to randomly skip some hosts?

skateman: Yup, up to 30% so you have the opportunity to assign hosts on your own through the UI/API

db/seeds.dev.rb
skateman: Some of the stuff yes, but not sure if it's a good idea to have a bunch of ENV vars. It's way easier to just edit the file when you want something custom before running it.

db/seeds.dev.rb
```diff
@@ -0,0 +1,60 @@
...
+      )
+
+      test_results.each do |tr|
+        rule_result_columns = profile.rules.pluck(:id).map do |rule_id|
+          [rule_id, tr.id, host.id, passing_hosts.include?(host) ? 'pass' : %w[pass pass fail].sample]
+        end
+
+        RuleResult.import(%i[rule_id test_result_id host_id result], rule_result_columns)
+
+        tr.update!(score: tr.rule_results.passed.count / tr.rule_results.count.to_f * 100)
```
vkrizan, line 54: Isn't there a method to calculate this?

db/seeds.rb
```diff
@@ -3,3 +3,4 @@
 # with its default values.
 # The data can then be loaded with the rails db:seed command (or created
 # alongside the database with db:setup).
+
```
vkrizan, line 6: This added line is pretty lonely here. :smiley:

db/seeds.dev.rb
bastilian: It could be useful for QEs to target different amounts via the ENV vars in their test/staging environments, where it could be cumbersome to edit a file.

vkrizan (commented):
It seems to do something. Is it possible to have a slightly verbose option?

The seeds file might get some formatting :grin:.

db/seeds.dev.rb
skateman: Nope, this is coming directly from the XML file

skateman:
I added some ENV vars, adjusted the formatting a bit and also added some logging.

db/seeds.dev.rb
```diff
@@ -0,0 +1,112 @@
...
+  host_ids = SupportedSsg.all.map { |ssg| [ssg.os_major_version, ssg.os_minor_version] }.uniq.flat_map do |version|
+    FactoryBot.create_list(
+      :host,
+      host_count,
+      account: account.account_number,
+      os_version_arr: version
+    ).map(&:id)
+  end
+
+  logger.info("Generated #{host_count} hosts for acc #{account.account_number}")
```
vkrizan, line 80: Got hosts generated for the same account twice. I guess it is because I've run it second time.
```
I, [2021-06-29T14:23:47.581223 #174581]  INFO -- : Generated 3 hosts for acc 00002
I, [2021-06-29T14:24:42.808221 #174581]  INFO -- : Generated 15 policies for account 00002
I, [2021-06-29T14:24:43.062831 #174581]  INFO -- : Generated 3 hosts for acc 00002
```

vkrizan (approved):
I approve.

[insert image]

PR #904, opened 2021-06-25
feat(clowder) - RHICOMPL-1363 - provide working Clowder template

Victoremepunto:
To test on ephemeral:

login to the ephemeral environment first via `oc login ...` login command copied from the web UI

Deploy:

``` 
bonfire deploy --no-remove-resources all compliance
```

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -122,6 +122,13 @@ objects:
             set -e
             RAILS_ENV="${RAILS_ENV}" bundle exec rake --trace db:debug db:migrate
             RAILS_ENV="${RAILS_ENV}" bundle exec rake --trace ssg:import_rhel_supported
+          resources:
+            limits:
+              cpu: ${CPU_LIMIT_SIDE}
+              memory: ${MEMORY_LIMIT_SIDE}
+            requests:
+              cpu: ${CPU_REQUEST_SIDE}
+              memory: ${MEMORY_REQUEST_SIDE}
           env:
           - name: RAILS_ENV
             value: "${RAILS_ENV}"
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -45,21 +45,21 @@ objects:
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
-        - name: wait-on-migrations
-          image: " "
-          imagePullPolicy: Always
-          command:
-            - bash
-            - -c
-            - |
-              for i in {1..120}; do sleep 1;
-              if bundle exec rake --trace db:abort_if_pending_migrations;
-              then exit 0; fi; done; exit 1
+        - command:
+          - bash
+          - -c
+          - |
+            for i in {1..120}; do sleep 1;
+            if bundle exec rake --trace db:abort_if_pending_migrations;
+            then exit 0; fi; done; exit 1
           env:
           - name: RAILS_ENV
             value: "${RAILS_ENV}"
           - name: PATH_PREFIX
             value: "${PATH_PREFIX}"
+          # Should be a secret ?
+          - name: SECRET_KEY_BASE
+            value: "secret_key_base"
         env:
         - name: APPLICATION_TYPE
           value: compliance-backend
@@ -83,43 +83,41 @@ objects:
           value: "${PUMA_MAX_THREADS}"
         - name: OLD_PATH_PREFIX
           value: "${OLD_PATH_PREFIX}"
+        # Should be a secret ?
+        - name: SECRET_KEY_BASE
+          value: "secret_key_base"
 # If we can configure the metrics endpoint to listen to 9000 this definitions should go away
         livenessProbe:
           httpGet:
             path: /api/compliance/v1/openapi.json
-            port: 8080
-            scheme: HTTP
+            port: web
           initialDelaySeconds: 5
         readinessProbe:
           httpGet:
             path: /api/compliance/v1/status
-            port: 8080
-            scheme: HTTP
+            port: web
           initialDelaySeconds: 5
         resources:
           limits:
-            cpu: "${CPU_LIMIT_SERV}"
-            memory: "${MEMORY_LIMIT_SERV}"
+            cpu: ${CPU_LIMIT_SERV}
+            memory: ${MEMORY_LIMIT_SERV}
           requests:
-            cpu: "${CPU_REQUEST_SERV}"
-            memory: "${MEMORY_REQUEST_SERV}"
+            cpu: ${CPU_REQUEST_SERV}
+            memory: ${MEMORY_REQUEST_SERV}
     - name: prometheus-exporter
       minReplicas: ${{REPLICAS_PROMETHEUS}}
-#      webServices:
-#        public:
-#          enabled: true
+      webServices:
+        private:
+          enabled: true
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
-        - name: run-migrations
-          image: " "
-          imagePullPolicy: Always
-          command:
-            - bash
-            - -c
-            - |
-              set -e
-              RAILS_ENV="${RAILS_ENV}" bundle exec rake --trace db:debug db:migrate
+        - command:
+          - bash
+          - -c
+          - |
+            set -e
+            RAILS_ENV="${RAILS_ENV}" bundle exec rake --trace db:debug db:migrate
           env:
           - name: RAILS_ENV
             value: "${RAILS_ENV}"
@@ -127,6 +125,9 @@ objects:
             value: "${PATH_PREFIX}"
           - name: RAILS_LOG_TO_STDOUT
             value: "${RAILS_LOG_TO_STDOUT}"
+          # Should be a secret ?
+          - name: SECRET_KEY_BASE
+            value: "secret_key_base"
         env:
         - name: APPLICATION_TYPE
           value: compliance-prometheus-exporter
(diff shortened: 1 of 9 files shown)


config/initializers/0_clowder-config.rb
```diff
@@ -14,7 +17,11 @@
   rbac_url = "#{rbac_config.hostname}:#{rbac_config.port}"
 
   # Prometheus
-  prometheus_exporter_config = config.private_dependency_endpoints.dig('compliance', 'prometheus-exporter')
+  if ENV["APPLICATION_TYPE"] == "compliance-prometheus-exporter"
+    prometheus_exporter_config = { hostname: "localhost", port: config.privatePort }
```
akofink, line 21: This should make the two branches of this conditional return an object that behaves similarly:

```suggestion
    prometheus_exporter_config = OpenStruct.new(hostname: "localhost", port: config.privatePort)
```

config/initializers/0_clowder-config.rb
```diff
@@ -14,7 +17,11 @@
   rbac_url = "#{rbac_config.hostname}:#{rbac_config.port}"
 
   # Prometheus
-  prometheus_exporter_config = config.private_dependency_endpoints.dig('compliance', 'prometheus-exporter')
+  if ENV["APPLICATION_TYPE"] == "compliance-prometheus-exporter"
+    prometheus_exporter_config = OpenStruct.new(hostname: "localhost", port: config.privatePort)
```
akofink, line 21:
```suggestion
    prometheus_exporter_config = ClowderCommonRuby::PrivateDependencyEndpoint.new(hostname: "localhost", port: config.privatePort)
```

config/initializers/0_clowder-config.rb
```diff
@@ -24,9 +23,9 @@
   redis_url = "#{config.inMemoryDb.hostname}:#{config.inMemoryDb.port}"
```
akofink, line 23: Needs the schema. See https://github.com/RedHatInsights/compliance-backend/pull/905/files#diff-274fbb104eb7e6e64d378ecb4f260a28440db9fc32916c19337dae8aa46cf85aR21

akofink: Just kidding! Do not make this change please :laughing:

deploy/get-clowder-common-bash.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -e
+
+WORKDIR=$(dirname "$0")
+VERSION='992e8a167dcdb96895c68cfab5f416e76caa57e6'
+FILENAME='clowder-config-main'
+FILEPATH="${WORKDIR}/${FILENAME}"
+INTEGRITY_FILE="${FILENAME}.sha256"
+URL="https://raw.githubusercontent.com/Victoremepunto/clowder-common-bash/${VERSION}/${FILENAME}"
```
vkrizan, line 10: Can you moved this to the RedHatInsights organization? It can be even in this repo.

entrypoint.sh
```diff
@@ -26,18 +28,25 @@ if [ "$APPLICATION_TYPE" = "compliance-backend" ]; then
...
+        PORT=`ClowderConfigWebPort`
+    else
+        PORT="8080"
+    fi
+
 	if is_puma_installed; then
 		export_vars=$(cgroup-limits) ; export $export_vars
 
-		exec bundle exec "puma --config ../etc/puma.cfg"
+		exec bundle exec "puma --config ../etc/puma.cfg -b tcp://0.0.0.0:${PORT}"
```
vkrizan, line 40: Observation: unquoted `PORT`. It is possible to drop the quotes within the exec and quote just the bind param.

entrypoint.sh
```diff
@@ -26,18 +28,25 @@ if [ "$APPLICATION_TYPE" = "compliance-backend" ]; then
...
 	else
+
 		echo "You might consider adding 'puma' into your Gemfile."
 
 		if bundle exec rackup -h &>/dev/null; then
 			if [ -f Gemfile ]; then
-				exec bundle exec "rackup -E ${RAILS_ENV:-$RACK_ENV} -P /tmp/rack.pid --host 0.0.0.0 --port 8080"
+				exec bundle exec "rackup -E ${RAILS_ENV:-$RACK_ENV} -P /tmp/rack.pid --host 0.0.0.0 --port ${PORT}"
```
vkrizan, line 47: Observation: unquoted `PORT`.

entrypoint.sh
```diff
@@ -49,7 +58,16 @@ elif [ "$APPLICATION_TYPE" = "compliance-inventory" ]; then
...
 elif [ "$APPLICATION_TYPE" = "compliance-prometheus-exporter" ]; then
-	exec bundle exec prometheus_exporter -b 0.0.0.0 --prefix compliance_ -t 50 --verbose -a lib/prometheus/graphql_collector.rb -a lib/prometheus/business_collector.rb
+
+    if isClowderEnabled; then
+        PORT=`ClowderConfigPrivatePort`
+    else
+        PORT="9394"
+    fi
+
+    echo "PORT:$PORT"
```
vkrizan, line 68: Observation: indentation.

docker-compose.yml
```diff
@@ -244,8 +244,8 @@ services:
     volumes:
       - .:/app:z
     ports:
-      - '9394:9394'
-    command: bundle exec prometheus_exporter -b 0.0.0.0 -t 50 --verbose -a lib/prometheus/graphql_collector.rb -a lib/prometheus/business_collector.rb -a lib/prometheus/engineering_collector.rb
+      - '10000:10000'
```
vkrizan, line 247: Is it possible to to keep the standard port of 9394?

clowdapp.yaml
```diff
@@ -343,7 +349,7 @@ parameters:
   value: 50m
 - name: REDIS_SSL
   description: 'Whether to use secured connection to Redis. Use string values of true or false'
-  value: "true"
+  value: "false"
```
vkrizan, line 352: This should be changed by the environment. If possible, please keep the secured variant as default (for hardening).

config/initializers/0_clowder-config.rb
akofink: Sorry, this change was for `rbac_url` not `redis_url`. Redis uses its own `redis://` schema. Anyway, it's fixed in https://github.com/RedHatInsights/compliance-backend/pull/905

docker-compose.yml
akofink: I don't see why not :) It just has to match what we put in devel.json.

deploy/get-clowder-common-bash.sh
akofink: Done: https://github.com/RedHatInsights/clowder-common-bash

entrypoint.sh
akofink: I think I'm going to have to add a whole commit to fix the indentation in this file. It also has mixed tabs and spaces :vomiting_face:

entrypoint.sh
akofink: Shellcheck does not complain because it is already quoted. The reason for that rule, SC2046, is to prevent word splitting, which this type of quoting prevents.

entrypoint.sh
akofink: It is quoted

clowdapp.yaml
akofink: Reverted this change as requested

entrypoint.sh
akofink: Fixed in https://github.com/RedHatInsights/compliance-backend/pull/904/commits/536952cb67ec5c8281529edeef9cfc4cb85d9445. Note that if you `git diff -w` there are no changes (it's just whitespace changes).

akofink:
I think the PR is in good shape now (all comments addressed). I'm going to try retesting this myself tomorrow with minikube to make sure it all works as expected still. Please feel free to also test yourself :)

entrypoint.sh
vkrizan: shellcheck likely does not complain because it is all quoted. I guess the `exec bundle exec` splits it itself, hence the problem stays.
```
$ myvar="echo test"
$ bundle exec "$myvar"
test
```

akofink: Your example doesn't actually demonstrate the code in question:

```sh
$ MSG="foo bar baz"
$ bundle exec "echo $MSG"
foo bar baz
$ bundle exec "echo ${MSG}"
foo bar baz
```

I think we're good here. It's also worth noting that the port will never contain spaces anyway, so even if the words were split, there would be no issue.

vkrizan: I guess all it needs another pair of quotes:
```
exec bundle exec "puma --config ../etc/puma.cfg -b tcp://0.0.0.0:'${PORT}'"
```
The double ones are for bundle exec, and the single ones for the subshell.

Here is a better example to get the idea of its power:
```
$ testvar="\$(touch /tmp/testfile)"
$ echo $testvar
$(touch /tmp/testfile)
$ be "$testvar"
$ ls /tmp/testfile 
/tmp/testfile
$ be "'$testvar'"
sh: line 1: $(touch /tmp/testfile): No such file or directory
```
I'm fine keeping as is, some hardening would not hurt :slightly_smiling_face:.

clowdapp.yaml
```diff
@@ -83,50 +145,61 @@ objects:
...
-            memory: "${MEMORY_REQUEST_SERV}"
+            cpu: ${CPU_REQUEST_SERV}
+            memory: ${MEMORY_REQUEST_SERV}
     - name: prometheus-exporter
       minReplicas: ${{REPLICAS_PROMETHEUS}}
-#      webServices:
-#        public:
-#          enabled: true
+      webServices:
+        private:
```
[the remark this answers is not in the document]
akofink: Yeah, that makes sense. I'll update it.

Victor mentioned that we ideally would expose metrics from each pod (like compliance-backend would get 8000 for rails and 10000 for metrics), but I don't really see how that would be feasible for non-web pods.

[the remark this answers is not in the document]
akofink: Hmm... I'm not sure how to do this with the metrics provider right now. All our services send metrics to the prometheus_exporter, which serves the metrics when requested. I'll ask around to see what we can do.

akofink (approved):
This is kind of my PR at this point, but I'll ACK it anyway. Thanks, Victor! :laughing: We should be able to merge this without affecting anything and fix any remaining issues we find later. This will unblock a lot of other work related to clowder, such as our pr_check.sh.

PR #906, opened 2021-06-25
feat: RHICOMPL-1933 get datastream config from compliance-ssg

akofink:
In addition to fetching the supported datastream config from compliance-ssg, this adds a ConfigDownloader which leverages the SafeDownloader to download config files. During failure to download, the backend still requires some kind of valid config, so the ConfigDownloader provides a fallback file fixture. This is also leveraged in our test environment, where no compliance-ssg service exists.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,8 @@ config/settings.local.yml
 config/settings/*.local.yml
 config/environments/*.local.yml
 
+config/supported_ssg.yaml
+
 brakeman_output.markdown
 
 Gemfile.local
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -6,7 +6,6 @@ require 'yaml'
 SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
                           :os_major_version, :os_minor_version,
                           keyword_init: true) do
-  self::SUPPORTED_FILE = Rails.root.join('config/supported_ssg.yaml')
   OS_NAME = 'RHEL'
   self::OS_NAME = OS_NAME
 
@@ -89,9 +88,21 @@ SupportedSsg = Struct.new(:id, :package, :version, :upstream_version, :profiles,
       rhel.scan(/(\d+)\.(\d+)$/)[0]
     end
 
+    def clear
+      @raw_supported = nil
+      @checksum = SsgConfigDownloader.ssg_ds_checksum
+    end
+
+    def ssg_ds_changed?
+      SsgConfigDownloader.update_ssg_ds
+
+      @checksum != SsgConfigDownloader.ssg_ds_checksum
+    end
+
     def raw_supported
-      # cached for the whole runtime
-      @raw_supported ||= YAML.load_file(self::SUPPORTED_FILE)
+      clear if ssg_ds_changed?
+
+      @raw_supported ||= YAML.safe_load(SsgConfigDownloader.ssg_ds)
     end
 
     def build_latest_map
--- a/app/services/ssg_config_downloader.rb
+++ b/app/services/ssg_config_downloader.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+# Service for dowloading SSG configuration files
+class SsgConfigDownloader
+  class << self
+    SSG_DS_FILE = File.new('config/supported_ssg.yaml')
+
+    def ssg_ds
+      File.read(SSG_DS_FILE)
+    end
+
+    def ssg_ds_checksum
+      Digest::MD5.file(SSG_DS_FILE).hexdigest
+    end
+
+    def update_ssg_ds
+      ssg_content = download(ssg_datastream_config_url)&.read
+      ssg_content && File.open(SSG_DS_FILE, 'w') do |f|
+        f.write(ssg_content)
+      end
+    end
+
+    def ssg_datastream_config_url
+      [
+        Settings.compliance_ssg_url,
+        Settings.supported_ssg_ds_config
+      ].join('/')
+    end
+
+    private
+
+    def download(url)
+      file = SafeDownloader.download(url)
+      Rails.logger.audit_success("Downloaded config from #{url}")
+      file
+    rescue StandardError => e
+      Rails.logger.audit_fail("Failed to download config from #{url}: #{e}")
+      nil
+    end
+  end
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -29,3 +29,4 @@ redis_cache_ssl: false
 slack_webhook: 'this is set through a env var in Openshift and not shared for security reasons'
 disable_rbac: 'false'
 async: true
+supported_ssg_ds_config: 'ssg-datastreams.yaml'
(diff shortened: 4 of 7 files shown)


vkrizan (approved):
Works great!
Fallbacking to a fixture might get some consideration.

app/services/config_downloader.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# Service for dowloading SSG configuration files
+class ConfigDownloader
+  class << self
+    delegate :ssg_ds_config, to: :new
+  end
+
+  SSG_DS_FALLBACK = File.new(
+    "test/fixtures/files/#{Settings.supported_ssg_ds_config}"
```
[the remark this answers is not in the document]
akofink: What is your proposal that we do if the config is not able to be fetched?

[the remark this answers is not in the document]
vkrizan: The same principle, but different (original) path: `config/ssg-datastreams-fallback.yaml`

The `Settings.supported_ssg_ds_config` constant shouldn't be included in this path as it is intended for the URL and not FS path (or call `basename` on it :wink:).

app/models/supported_ssg.rb
```diff
@@ -89,9 +88,13 @@ def os_version(rhel)
...
     end
 
+    def fetch_config
+      ConfigDownloader.ssg_ds_config
+    end
+
     def raw_supported
       # cached for the whole runtime
-      @raw_supported ||= YAML.load_file(self::SUPPORTED_FILE)
+      @raw_supported ||= YAML.safe_load(fetch_config)
```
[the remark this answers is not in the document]
akofink: I don't see why not. I also don't see what benefit it'd give us. In an initializer, if something goes wrong, it would seem much more difficult to correct (i.e. restart the pod?) than if some API request fails here and simply retries the next time. Note that this `@raw_supported` is cached after the first usage of SupportedSsg (per thread), so it behaves similarly as being in an initializer anyway, but it's a bit lazier.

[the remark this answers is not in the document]
vkrizan: This way the initial load is mapped to a first request that comes in (that uses this), hence the suggestion to initialize it as part of the application. It should not crash in some random place where it would be used for the first time.

Considering the caching, there is no error handling or retry, so a single failure can lead to incorrect supported matrix for the whole duration of pod. We'd likely need to keep the fallback file in sync with what's served by compliance-ssg. Or maybe the fallback can be completely disabled in production, which would require to have compliance-ssg serving it. Still there are sync problems if we'd deploy a new compliance-ssg along with backend.

app/services/config_downloader.rb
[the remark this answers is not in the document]
akofink: Done!

akofink:
Thanks for the feedback, @vkrizan! I've kind of redesigned how this works so the cache is the config file itself rather than being just in memory. I'm leveraging checksums to detect if the config changes so we can clear the SupportedSsg's `@raw_supported` variable (and now `@checksum`). I'd like the compliance-backend and compliance-ssg deployments to stay independent if possible, so the config checksum is compared every time SupportedSsg is used, but the yaml is not parsed every time. I don't think this should pose any performance issue, and it will allow us to deploy new SSG content and have the backend pick it up immediately.

Please have another look!

app/models/supported_ssg.rb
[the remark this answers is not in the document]
akofink: Still not using an initializer, but I did fix the caching issue. Thank you for pointing it out!

.gitignore
```diff
@@ -55,6 +55,8 @@ config/settings.local.yml
 config/settings/*.local.yml
 config/environments/*.local.yml
 
+config/supported_ssg.yaml
```
akofink, line 58: This doesn't seem to work as expected. Perhaps because it's already checked in to git?

app/services/config_downloader.rb
```diff
@@ -0,0 +1,41 @@
...
+class ConfigDownloader
+  class << self
+    SSG_DS_FILE = File.new('config/supported_ssg.yaml')
+
+    def ssg_ds
+      File.read(SSG_DS_FILE)
+    end
+
+    def ssg_ds_checksum
+      Digest::MD5.file(SSG_DS_FILE).hexdigest
```
vkrizan, line 13: Observation: MD5 can be flagged by some scanners. Suggestion is to use SHA256+.

app/services/config_downloader.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# Service for dowloading SSG configuration files
+class ConfigDownloader
```
vkrizan, line 4: Observation: This is SSG specific: maybe `SSGConfigDownloader`?

app/models/supported_ssg.rb
```diff
@@ -89,9 +88,21 @@ def os_version(rhel)
...
+    def ssg_ds_changed?
+      ConfigDownloader.update_ssg_ds
+
+      @checksum != ConfigDownloader.ssg_ds_checksum
+    end
+
     def raw_supported
-      # cached for the whole runtime
-      @raw_supported ||= YAML.load_file(self::SUPPORTED_FILE)
+      clear if ssg_ds_changed?
```
vkrizan, line 103: It looks like that every time any general `SupportedSsg` method would be called it would re-download the file, is that right? Can it be cached at least for the duration of the request or time cached?

vkrizan (commented):
With the current changes it would probably redownload the file on every supported ssg call. (see below)

app/services/config_downloader.rb
akofink: I only used MD5 for performance reasons. Do you really think this is some kind of security issue?

app/models/supported_ssg.rb
akofink: It definitely downloads on every method call, as advertised:

> the config checksum is compared every time SupportedSsg is used, but the yaml is not parsed every time

I don't want to prematurely optimize this, especially since we can go back and add it easily later. Note that remediations also downloads content on every request from compliance-ssg - it really shouldn't be a problem. I am also wary of caching and wouldn't add it unless it is necessary.

That said, I think the best way to improve this current design is to serve the checksums from compliance-ssg so very little work (or network) is necessary to determine if the config has changed. I don't see why we can't merge this PR and improve it in the meantime.

app/services/config_downloader.rb
akofink: Yeah, I guess so :laughing: I initially thought it'd be useful for other configs, but I can't imagine what others we'd need other than SSG information.

app/services/config_downloader.rb
vkrizan: For our purpose MD5 here is okay. The nuisance will be when some scanner starts to report this.

app/services/config_downloader.rb
akofink: Updated to `SsgConfigDownloader`.

PR #909, opened 2021-06-30
feat(metrics): RHICOMPL-1763 track OS major/minor version counts

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/lib/prometheus/business_collector.rb
+++ b/lib/prometheus/business_collector.rb
@@ -38,6 +38,12 @@ class BusinessCollector < PrometheusExporter::Server::TypeCollector
     @client_systems = PrometheusExporter::Metric::Gauge.new(
       'client_systems', 'Systems from clients (excludes Red Hat)'
     )
+    @total_systems_by_os = PrometheusExporter::Metric::Counter.new(
+      'total_systems_by_os', 'Systems by OS version'
+    )
+    @client_systems_by_os = PrometheusExporter::Metric::Counter.new(
+      'client_systems_by_os', 'Systems by OS version (excluded Red Hat)'
+    )
   end
 
   def type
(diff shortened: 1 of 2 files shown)


lib/prometheus/business_collector.rb
```diff
@@ -64,6 +70,28 @@ def collect
...
+      account: client_accounts.select(:account_number)
+    ).select(
+      "COUNT(id), concat(
+        #{Host::OS_MAJOR_VERSION},
+        '.',
+        #{Host::OS_MINOR_VERSION}
+      ) as version"
+    ).group('version').map(&:attributes).each do |item|
+      @client_systems_by_os.observe(item['count'], version: item['version'])
+    end
```
akofink, line 94: Be sure to add tests to business_collector_test.rb :)

skateman: Fixed

lib/prometheus/business_collector.rb
```diff
@@ -64,6 +70,28 @@ def collect
...
+
+    Host.with_policies_or_test_results.where(
+      account: client_accounts.select(:account_number)
+    ).select(
+      "COUNT(id), concat(
+        #{Host::OS_MAJOR_VERSION},
+        '.',
+        #{Host::OS_MINOR_VERSION}
+      ) as version"
+    ).group('version').map(&:attributes).each do |item|
```
vkrizan, line 92: I wonder how this version concat from JSON blob and group by it would perform in production.

vkrizan (approved):
LGTM

lib/prometheus/business_collector.rb
skateman: Do you have performance concerns?

akofink: We've introduced metrics in the past that affect performance because of the DB pressure they add. I'm not too concerned about Ruby performance, since it at least should not affect other services like compliance-backend.

akofink (approved):

sthirugn:
will this show up in Grafana once pushed to prod? or should we update some grafana config?

vkrizan:
@sthirugn No, it needs to have a graph configured at https://github.com/RedHatInsights/compliance-backend/blob/master/dashboards/grafana-dashboard-insights-compliance-general.configmap.yaml (and then deployed via app-interface). However, the data should start to be collected, and an ad-hoc graph would be also a temporary option.

sthirugn:
Thank you, viliam

lib/prometheus/business_collector.rb
```diff
@@ -38,6 +38,12 @@ def initialize
     @client_systems = PrometheusExporter::Metric::Gauge.new(
       'client_systems', 'Systems from clients (excludes Red Hat)'
     )
+    @total_systems_by_os = PrometheusExporter::Metric::Counter.new(
+      'total_systems_by_os', 'Systems by OS version'
+    )
+    @client_systems_by_os = PrometheusExporter::Metric::Counter.new(
+      'client_systems_by_os', 'Systems by OS version (excluded Red Hat)'
+    )
```
akofink, line 46: @skateman I'm quite sure we should update this to be a Gauge. From [the docs](https://prometheus.io/docs/concepts/metric_types/#counter):

> A counter is a cumulative metric that represents a single monotonically increasing counter whose value can only increase or be reset to zero on restart. For example, you can use a counter to represent the number of requests served, tasks completed, or errors.
> 
> Do not use a counter to expose a value that can decrease. For example, do not use a counter for the number of currently running processes; instead use a gauge.

I've updated it in https://github.com/RedHatInsights/compliance-backend/pull/932

PR #910, opened 2021-06-30
perf(GraphQL): RHICOMPL-1961 calculate passed rules of hosts in SQL

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/system_like.rb
+++ b/app/models/concerns/system_like.rb
@@ -48,7 +48,15 @@ module SystemLike
   end
 
   def compliance_score
-    score = (100 * (rules_passed.to_f / (rules_passed + rules_failed)))
+    counts = test_results.latest.joins(:rule_results)
+                         .group('rule_results.result').count
+    score = passed_count(counts).to_f / counts.values.map(&:to_i).sum * 100
     score.nan? ? 0.0 : score
+  end
+
+  private
+
+  def passed_count(result_hash)
+    result_hash.values_at(*RuleResult::PASSED).map(&:to_i).sum
   end
 end


app/models/concerns/system_like.rb
```diff
@@ -48,7 +48,16 @@ def rules_failed(profile = nil)
...
+    counts = test_results.latest.joins(:rule_results)
+                         .group('rule_results.result').count
+    score = passed_count(counts).to_f / counts.values.map(&:to_i).sum * 100
     score.nan? ? 0.0 : score
   end
+
+  private
+
+  def passed_count(result_hash)
+    result_hash.values_at('pass', 'notapplicable', 'notselected')
```
akofink, line 60:
```suggestion
    result_hash.values_at(*RuleResult::PASSED)
```

skateman: Thanks, fixed

app/models/concerns/system_like.rb
```diff
@@ -48,7 +48,15 @@ def rules_failed(profile = nil)
   end
 
   def compliance_score
-    score = (100 * (rules_passed.to_f / (rules_passed + rules_failed)))
+    counts = test_results.latest.joins(:rule_results)
+                         .group('rule_results.result').count
+    score = passed_count(counts).to_f / counts.values.map(&:to_i).sum * 100
     score.nan? ? 0.0 : score
   end
```
akofink, line 55: As far as I can tell, this is only used for filtering systems by compliance score. It seems to have the same behavior that is on master, but I found an issue on master with the way this search works. This `compliance_score` is for all profiles being reported against, so on the report details page, the 'Compliance Score' filter does not work as expected.

image

image

And here is the unfiltered graphql response. Note one of these systems has scanned against multiple profiles:

```js
{
    "data": {
        "systems": {
            "nodes": [
                {
                    "hasPolicy": true,
                    "id": "624e78e4-b6fc-4402-8a58-307c1ed0ea39",
                    "lastScanned": "2021-07-01T15:09:55Z",
                    "name": "rhel7-insights-client.virbr0.akofink-laptop",
                    "rulesFailed": 6,
                    "rulesPassed": 9,
                    "testResultProfiles": [
                        {
                            "compliant": true,
                            "external": true,
                            "id": "1e1e3f6d-908f-4282-b8cf-57dd50543aa4",
                            "lastScanned": "2021-07-01T15:09:55Z",
                            "osMajorVersion": "7",
                            "osMinorVersion": "7",
                            "refId": "xccdf_org.ssgproject.content_profile_rht-ccp",
                            "rulesFailed": 6,
                            "rulesPassed": 9,
                            "score": 63.541664,
                            "ssgVersion": "0.1.43",
                            "supported": true
                        }
                    ]
                },
                {
                    "hasPolicy": true,
                    "id": "6417c06d-e481-4e0b-9af2-cd34ef97e7d7",
                    "lastScanned": "2021-07-01T15:11:12Z",
                    "name": "rhel79-insights-client.virbr0.akofink-laptop",
                    "rulesFailed": 243,
                    "rulesPassed": 198,
                    "testResultProfiles": [
                        {
                            "compliant": true,
                            "external": true,
                            "id": "0def7dce-05f1-4a22-953a-0176806a0a7d",
                            "lastScanned": "2021-07-01T15:10:01Z",
                            "osMajorVersion": "7",
                            "osMinorVersion": "",
                            "refId": "xccdf_org.ssgproject.content_profile_rht-ccp",
                            "rulesFailed": 28,
                            "rulesPassed": 40,
                            "score": 71.25,
                            "ssgVersion": "0.1.52",
                            "supported": true
                        },
                        {
                            "compliant": false,
                            "external": true,
                            "id": "fc133cbe-9b79-4048-92e6-11b40ee976b4",
                            "lastScanned": "2021-07-01T15:09:54Z",
                            "osMajorVersion": "7",
                            "osMinorVersion": "",
                            "refId": "xccdf_org.ssgproject.content_profile_cui",
                            "rulesFailed": 78,
                            "rulesPassed": 26,
                            "score": 44.648766,
                            "ssgVersion": "0.1.52",
                            "supported": true
                        },
                        {
                            "compliant": true,
                            "external": true,
                            "id": "c8f71bd8-b65f-4861-93e6-20011d1a1269",
                            "lastScanned": "2021-07-01T15:11:12Z",
                            "osMajorVersion": "7",
                            "osMinorVersion": "",
                            "refId": "xccdf_org.ssgproject.content_profile_cis",
                            "rulesFailed": 137,
                            "rulesPassed": 132,
                            "score": 71.928329,
                            "ssgVersion": "0.1.52",
                            "supported": true
                        }
                    ]
                }
            ],
            "totalCount": 2
        }
    }
}
```

So, we may need to step back a second and try to figure out a fix for the bug that also will perform adequately with many hosts.

skateman: Okay, so there are multiple issues here :facepalm: like really bad ones :rofl: 
* the score displayed on the screenshots (A) is the value of the latest `TestResult#score` for the given host that is coming from the uploaded report
* the score for the filtering (B) is calculated from the number of ALL passed rules on a host, nost just for a single policy
* even if there is just one profile assigned to each hosts, the (A) and (B) values differ :cry: :thinking: 

My guess is that the uploaded OpenSCAP reports have a different way to calculate the score percentage than our `passed / total * 100` calculation and I'm not sure which one should we use. Right now we are relying on the values coming from the report (A) and the only place I know of where we use our calculated value (B) is here in the filtering.

If we trust the value coming from the reports (A) the solution is to eliminate the calculation and do the filtering based on the latest `TestResult#score` instead of the `Host#compliance_score` which should speed up the things even more significantly. But I am not entirely sure that we don't use our calculated (B) value elsewhere :thinking: 

cc @akofink

skateman: A wild guess about the different scores is that OpenSCAP might not respect the profile tailoring :thinking:

akofink: Correct, score (B) is not valid, and we should not use it anywhere. The SCAP spec defines how exactly the scoring works, if you're interested: [NIST 7275 r.4, section 7.3.2 Scoring Models, The Default Model](https://csrc.nist.gov/publications/detail/nistir/7275/rev-4/final)

image

app/models/concerns/system_like.rb
```diff
@@ -48,7 +48,15 @@ def rules_failed(profile = nil)
...
 
   def compliance_score
-    score = (100 * (rules_passed.to_f / (rules_passed + rules_failed)))
+    counts = test_results.latest.joins(:rule_results)
```
vkrizan, line 51: This bypasses `last_scanned` method and the context (`profile_id`) that reduces the scope of results for score count. I wonder if this would have a consequence, maybe not.

`rules_passed` and `rules_failed` are memoized on its model and are contextful. Now it depends if that context is before or after the models are constructed.

It would be good to verify this with a test.

skateman: This PR will be obsolete if #915 gets merged

PR #912, opened 2021-07-08
feat(REST): RHICOMPL-1284 implemented filtering by tags for systems

skateman:
I traced back how Rails and Rack handle the parsing of the query string and found that the `Rack::Request::Env#query_parser` method holds the reference to their parser. So I am monkeypatching this single method in an initializer to hold our parser, which is just a delegation to the original parser with the addition of overriding the already parsed tags with our correct values.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/controllers/concerns/collection.rb
+++ b/app/controllers/concerns/collection.rb
@@ -5,11 +5,19 @@ module Collection
   extend ActiveSupport::Concern
 
   include Sorting
+  include TagFiltering
 
   included do
     def resolve_collection
-      result = sort(search(policy_scope(resource)))
+      result = filter_by_tags(sort(search(policy_scope(resource))))
       result.paginate(page: pagination_offset, per_page: pagination_limit)
+    end
+
+    def filter_by_tags(data)
+      return data unless tags_supported? && params[:tags]&.any?
+
+      tags = parse_tags(params[:tags])
+      data.where('tags @> ?', tags.to_json)
     end
 
     def search(data)
@@ -22,6 +30,12 @@ module Collection
       return data if params[:sort_by].blank?
 
       data.order(build_order_by(data.klass, params[:sort_by]))
+    end
+
+    private
+
+    def tags_supported?
+      resource.column_names.include?('tags')
     end
   end
 end
--- a/app/controllers/concerns/metadata.rb
+++ b/app/controllers/concerns/metadata.rb
@@ -12,14 +12,25 @@ module Metadata
       response.headers['Content-Type'] = 'application/vnd.api+json'
     end
 
+    # This is part of a JSON schema, no need for strict metrics
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength
     def metadata(opts = {})
       opts[:total] ||= resolve_collection.count
-      options = {}
-      options[:meta] = { total: opts[:total], search: params[:search],
-                         limit: pagination_limit, offset: pagination_offset }
-      options[:links] = links(last_offset(opts[:total]))
-      options
+
+      {
+        meta: {
+          total: opts[:total],
+          search: params[:search],
+          tags: tags_supported? ? params.fetch(:tags, []) : nil,
+          limit: pagination_limit,
+          offset: pagination_offset
+        }.compact,
+        links: links(last_offset(opts[:total]))
+      }
     end
+    # rubocop:enable Metrics/AbcSize
+    # rubocop:enable Metrics/MethodLength
 
     def links(last_offset)
       {
--- a/lib/adjust_tags/middleware.rb
+++ b/lib/adjust_tags/middleware.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+# Common space for Insights API stuff
+module Insights
+  module API
+    module Common
+      module AdjustTags
+        # Rack middleware to adjust params[tags] to be an array
+        class Middleware
+          def initialize(app)
+            @app = app
+          end
+
+          def call(env)
+            qs = env['QUERY_STRING'].sub(/^tags=/, 'tags[]=')
+                                    .gsub(/&tags=/, '&tags[]=')
+
+            # Match the QUERY_STRING at the end of the line only
+            re = /#{Regexp.quote(env['QUERY_STRING'])}$/
+
+            env['REQUEST_URI'] = env['REQUEST_URI'].sub(re, qs)
+            env['QUERY_STRING'] = qs
+
+            @app.call(env)
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/integration/systems_spec.rb
+++ b/spec/integration/systems_spec.rb
@@ -18,6 +18,7 @@ describe 'Systems API' do
       auth_header
       pagination_params
       search_params
+      tags_params
       sort_params
 
       include_param
--- a/swagger/v1/openapi.json
+++ b/swagger/v1/openapi.json
@@ -2795,6 +2795,20 @@
             }
           },
           {
+            "name": "tags",
+            "in": "query",
+            "required": false,
+            "type": "array",
+            "description": "An array of tags to narrow down the results against.",
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "default": ""
+            }
+          },
+          {
             "name": "include",
             "in": "query",
             "type": "string",
@@ -3091,6 +3105,16 @@
           "filter": {
             "type": "string",
             "example": "name='Standard System Security Profile for Fedora'"
+          }
+        }
+      },
+      "tags": {
+        "type": "array",
+        "items": {
+          "type": "string",
+          "properties": {
+            "type": "string",
+            "example": "insights/environment=production"
           }
         }
       },
--- a/test/controllers/v1/systems_controller_test.rb
+++ b/test/controllers/v1/systems_controller_test.rb
@@ -75,6 +75,63 @@ module V1
         assert_response :success
         assert_empty JSON.parse(response.body).dig('meta', 'search')
       end
+
+      should 'allow filtering by tags' do
+        host1 = FactoryBot.create(
+          :host,
+          tags: [
+            {
+              key: 'env',
+              value: 'prod',
+              namespace: 'insights-client'
+            },
+            {
+              key: 'env',
+              value: 'stage',
+              namespace: 'insights-client'
+            }
+          ]
+        )
+
+        FactoryBot.create(
+          :host,
+          tags: [
+            {
+              key: 'env',
+              value: 'stage',
+              namespace: 'insights-client'
+            }
+          ]
+        )
+
+        FactoryBot.create(:policy, hosts: Host.all)
+
+        # The Insights API tags format cannot be constructed using a hash
+        get [
+          v1_systems_url,
+          'tags=insights-client/env=prod&tags=insights-client/env=stage'
+        ].join('?')
+
+        assert_response :success
+        results = JSON.parse(response.body)['data']
+        assert_equal results.count, 1
+        assert_equal results.first['id'], host1.id
+        assert_not_empty JSON.parse(response.body).dig('meta', 'tags')
+      end
+
+      %w[
+        tags=satellite/lifecycle_environment=Library
+        tags=satellite%2Flifecycle_environment=Library
+        tags=satellite/lifecycle_environment%3DLibrary
+        tags=satellite%2Flifecycle_environment%3DLibrary
+      ].each do |qstr|
+        should "properly parse #{qstr}" do
+          get [v1_systems_url, qstr].join('?')
+          tags = JSON.parse(response.body)['meta']['tags']
+
+          assert_equal tags, %w[satellite/lifecycle_environment=Library]
+        end
+      end
     end
 
     context 'show' do
--- a/README.md
+++ b/README.md
@@ -288,6 +288,10 @@ an identity header (i.e. rbac, remediations).
 
 To disable Prometheus (e.g. in develompent) clear `prometheus_exporter_host` setting (set to empty).
 
+### Tagging
+
+If there is a `tags` column defined in any model, it always should be a `jsonb` column and follow the structured representation of tags described in Insights, i.e. an array of hashes. If this convention is not kept, the controllers might break when a user tries to pass the `tags` attribute to a GET request.
+
 ## API documentation
 
 The API documentation can be found at `Settings.path_prefix/Settings.app_name`. To generate the docs, run `rake rswag:specs:swaggerize`. You may also get the OpenAPI definition at `Settings.path_prefix/Settings.app_name/v1/openapi.json`
--- a/app/controllers/concerns/tag_filtering.rb
+++ b/app/controllers/concerns/tag_filtering.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+# Concern to support filtering by tags
+module TagFiltering
+  extend ActiveSupport::Concern
+
+  TAG_RIGHT_RE = %r{^([^\/]+)\/([^\=]+)(\=(.+))?}.freeze
+
+  included do
+    def parse_tags(tags = [])
+      tags.map { |tag| parse_tag(tag) }.compact
+    end
+
+    private
+
+    def parse_tag(tag)
+      namespace, key, _, value = tag.scan(TAG_RIGHT_RE).first
+
+      return nil if namespace.nil? || key.nil?
+
+      {
+        namespace: Rack::Utils.unescape(namespace),
+        key: Rack::Utils.unescape(key),
+        value: value && Rack::Utils.unescape(value)
+      }
+    end
+
+    def decode_value(str)
+      str.gsub(/\+|%\h\h/, URI::TBLDECWWWCOMP_)
+         .force_encoding(Encoding::UTF_8)
+         .scrub
+    end
+  end
+end
--- a/config/application.rb
+++ b/config/application.rb
@@ -42,6 +42,9 @@ module ComplianceBackend
     # Attach audit logging for requests
     require 'audit_log/audit_log'
     config.middleware.use Insights::API::Common::AuditLog::Middleware
+    # Adjust params[tags] to be array
+    require 'adjust_tags/middleware'
+    config.middleware.use Insights::API::Common::AdjustTags::Middleware
 
     # GraphiQL
     if Rails.env.development?
--- a/spec/api/v1/schemas.rb
+++ b/spec/api/v1/schemas.rb
@@ -33,6 +33,7 @@ module Api
         relationship_collection: RELATIONSHIP_COLLECTION,
         error: ERROR,
         metadata: METADATA,
+        tags: TAGS,
         host: HOST,
         host_relationships: HOST_RELATIONSHIPS,
         links: LINKS,
--- a/spec/api/v1/schemas/metadata.rb
+++ b/spec/api/v1/schemas/metadata.rb
@@ -23,6 +23,17 @@ module Api
             }
           }
         }.freeze
+
+        TAGS = {
+          type: 'array',
+          items: {
+            type: 'string',
+            properties: {
+              type: 'string',
+              example: 'insights/environment=production'
+            }
+          }
+        }.freeze
       end
     end
   end
--- a/spec/swagger_helper.rb
+++ b/spec/swagger_helper.rb
@@ -88,6 +88,12 @@ def search_params
             schema: { type: :string, default: '' }
 end
 
+def tags_params
+  parameter name: :tags, in: :query, required: false, type: :array,
+            description: 'An array of tags to narrow down the results against.',
+            schema: { type: :array, items: { type: 'string' }, default: '' }
+end
+
 def sort_params
   parameter name: :sort_by, in: :query, required: false,
             type: { oneOf: [{ type: :string }, { type: :array }] },
--- a/test/controllers/concerns/tag_filtering_test.rb
+++ b/test/controllers/concerns/tag_filtering_test.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class TagFilteringTest < ActiveSupport::TestCase
+  class TagParserDummy
+    include TagFiltering
+  end
+
+  setup do
+    @parser = TagParserDummy.new
+  end
+
+  test '#parse_tags' do
+    {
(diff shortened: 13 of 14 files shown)


app/controllers/concerns/collection.rb
```diff
@@ -5,13 +5,21 @@ module Collection
...
 
   included do
     def resolve_collection
-      result = sort(search(policy_scope(resource)))
+      result = filter_by_tags(sort(search(policy_scope(resource))))
       result.paginate(page: pagination_offset, per_page: pagination_limit)
     end
 
+    def filter_by_tags(data)
+      tags = url_decode_tags(request.env['QUERY_STRING'])
```
akofink, line 17:
```suggestion
      tags = url_decode_tags(request.query_string)
```

app/controllers/concerns/metadata.rb
```diff
@@ -15,8 +15,10 @@ def set_headers
     def metadata(opts = {})
       opts[:total] ||= resolve_collection.count
       options = {}
-      options[:meta] = { total: opts[:total], search: params[:search],
-                         limit: pagination_limit, offset: pagination_offset }
+      options[:meta] = {
+        total: opts[:total], search: params[:search], tags: params[:tags],
```
skateman, line 19: @akofink I'm not entirely sure about including this in the response and if yes, I don't know if we should use the parsed or the original format (my bet is on the parsed one)...

skateman: It IS an array, but a but the tag itself is not just a simple string, but a hash with 3 elements (namespace, key, value) so the result right now is something like:
```
    "links": {
        "first": "/api/compliance/profiles?limit=10&offset=1&search=external%3Dfalse+and+canonical%3Dfalse",
        "last": "/api/compliance/profiles?limit=10&offset=2&search=external%3Dfalse+and+canonical%3Dfalse",
        "next": "/api/compliance/profiles?limit=10&offset=2&search=external%3Dfalse+and+canonical%3Dfalse"
    },
    "meta": {
        "limit": 10,
        "offset": 1,
        "tags": [
            {"namespace": "insights", "key": "environment", "value": "development"},
            {"namespace": "insights", "key": "environment", "value": "production"},
        ]
        "search": "external=false and canonical=false",
        "total": 15
    }
```

> I think it makes sense to include here, only for the systems endpoint, though

I think a simple `compact` can solve this problem and keep the complexity low (i.e. no tests for the hosts controller), we would show the tags only when they're passed...wdyt?

app/controllers/concerns/collection.rb
```diff
@@ -8,10 +8,17 @@ module Collection
 
   included do
     def resolve_collection
-      result = sort(search(policy_scope(resource)))
+      result = filter_by_tags(sort(search(policy_scope(resource))))
```
vkrizan, line 11: Observation: filtering before sort.

app/controllers/concerns/collection.rb
```diff
@@ -8,10 +8,17 @@ module Collection
...
       result.paginate(page: pagination_offset, per_page: pagination_limit)
     end
 
+    def filter_by_tags(data)
+      tags = params[:tags]
+      return data if !resource.column_names.include?('tags') || tags.empty?
+
+      data.where('tags @> ?', tags.to_json)
```
vkrizan, line 19: Observation: `tags` could start to be considered as a common global field on any model that would use them. Since this only checks for the field presence, it would have to be become a standard. Otherwise any table/model having `tags` _not_ as JSON blob would break on this lookup.

I'm wondering where we can document this...

config/initializers/1_rack_query_parser.rb
```diff
@@ -0,0 +1,8 @@
+require 'rack/request'
+require 'query_parser'
+
+Rack::Request::Env.class_eval do
+  def query_parser
```
vkrizan, line 5: Is there an other way than monkey-patching?

test/controllers/v1/systems_controller_test.rb
```diff
@@ -75,6 +75,49 @@ class SystemsControllerTest < ActionDispatch::IntegrationTest
...
+            }
+          ]
+        )
+
+        FactoryBot.create(:policy, hosts: Host.all)
+
+        # The Insights API tags format cannot be constructed using a hash
+        get [
+          v1_systems_url,
+          'tags=insights-client/env=prod&tags=insights-client/env=stage'
```
vkrizan, line 112: Nit: Can pass this as `params`.

lib/query_parser.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+# Common space for Insights API stuff
+module Insights
+  module API
+    module Common
+      # Query parser that can handle insights-flavored tags
+      class QueryParser < ::SimpleDelegator
```
vkrizan, line 8: Tests please? :slightly_smiling_face:

app/controllers/concerns/metadata.rb
akofink: I'm fine with the JSON object per tag. I'd include tags in the meta section even if no tags are specified, the same way we handle the other meta fields. That way, we communicate the default behavior (right now, default tags are just `[]`, but maybe in the future we will return hosts tagged with `compliance` or similar by default).

app/controllers/concerns/collection.rb
skateman: It is documented by the insights API, they have a very strict specification about how a tag should look like :thinking:

app/controllers/concerns/collection.rb
skateman: Ideally the ordering here doesn't really matter as ActiveRecord constructs the query at the end when the `result` is being used. However, somehow the `filter_by_tags` tends to give an array as a result and not a CollectionProxy :confused: which I don't really understand why, it might be connected with the fact that we're querying a jsonb field. So TL;DR if we swap the order here, the chaining will just not work :disappointed:

config/initializers/1_rack_query_parser.rb
vkrizan: Sounds like a job for a Middleware?

app/controllers/concerns/collection.rb
vkrizan: Interesting. I wonder where is the hiccup.

skateman: My guess is that the `jsonb` support in AR is relatively new and it might be just poorly implemented. Anyway, theoretically having these in different orders shouldn't alter the query, but I agree that it goes against human logic :see_no_evil:

lib/query_parser.rb
skateman: It's actually tested as a systems query, but sure I can write some unit tests :wink:

config/initializers/1_rack_query_parser.rb
skateman: Yes, that was my idea last week on Slack, we even discussed this with @akofink, but the problem is that the `params` method is defined in the context of `ActionController` and not the context of a middleware, so you can't have access to it from a middleware. We could possibly alter the `request.query_string` in a middleware e.g. by replacing any occurrence `tags=` with `tags[]=`, but we would also need to keep in mind that `=` and `/` characters can be escaped in the tags and the params parser declared in Rails would just unescape everything equally.

This would mean that without introducing a new language (to bypass the implicit unescaping) we could end up with strings like `compliance=hey/david=yay/foo=bar=bar` where we would not know which one are the "real" `=` and `/` delimiters and which were the escaped one. Then in the controllers we need to parse this new language explicitly and also alter the logging to output the tags in a parsed forms instead of using the default (which again, would be this new language). I guess I don't have to explain why is this a bad practice :thinking:

app/controllers/concerns/collection.rb
skateman: Okay, so what are you suggesting here? :thinking:

app/controllers/concerns/metadata.rb
skateman: Yeah, I've been trying point out that I would like to avoid a `tags if controller == 'Systems'` kind of condition in a global scope like the `Metadata` module :shrug: 

So do you have any preference on how to do this? I have too many ideas and I don't like any of them :sweat_smile:

skateman: Nevermind, figured it out :yawning_face: mornings are hard :sweat_smile:

app/controllers/concerns/collection.rb
vkrizan: Just to document it, that when a model with `tags` is exposed (has a controller) that the tags will follow the Insights tag structure.

config/initializers/1_rack_query_parser.rb
```diff
@@ -0,0 +1,8 @@
+require 'rack/request'
+require 'query_parser'
+
+Rack::Request::Env.class_eval do
```
vkrizan, line 4: Env module is being removed in rack (rack/rack#1751). We might monkey patch or change rack's `Utils.default_query_parser`. (would need to check whether that applies to our current version)

vkrizan: This one also did the job:
```
Rack::Utils.default_query_parser = Insights::API::Common::QueryParser.new(Rack::Utils.default_query_parser)
```

skateman: @vkrizan I'm a bit concerned about how would this behave in code reloads, but I'll do some experiments with it

vkrizan: Good point.

skateman: It works, changed it :medal_military: thanks

vkrizan (approved):
LGTM

I assume that you've tested it :slightly_smiling_face:

akofink:
Be sure to update the openapi.json and related specs to document this new parameter. Our users don't have access to the readme :)

swagger/v1/openapi.json
```diff
@@ -2794,6 +2794,22 @@
...
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "array",
+              "items": {
+                 "type": "string"
+              },
+              "default": ""
+            },
+            "description": "An array of tags to narrow down the results against.",
```
skateman, line 2809: @akofink please review :us: :pray:

swagger/v1/openapi.json
```diff
@@ -2794,6 +2794,22 @@
               "default": ""
             }
           },
+
+          {
+            "name": "tags",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "array",
```
skateman, line 2803: Not entirely sure about this, I stole it [from here](https://github.com/RedHatInsights/patchman-engine/blob/ffcac0203ac7fc1129daf6eab04f7083ac4f083e/docs/openapi.json#L128) but I guess this is the best match :shrug:

akofink:
@skateman the openapi.json is auto-generated. Everything is defined in ruby within the `spec/` directory. For example, you'll need to document the new tags attribute within the meta schema: https://github.com/RedHatInsights/compliance-backend/blob/master/spec/api/v1/schemas/metadata.rb

When you run rswag, the autogenerated examples should also contain the new field in all API responses. To demonstrate the usage, add it to an existing spec or define a new one within the [systems spec](https://github.com/RedHatInsights/compliance-backend/blob/master/spec/integration/systems_spec.rb).

You did similar steps when you implemented sorting. Pretty much any API change requires updating the specs like this, since the specs are used to generate the API docs.

skateman:
> @skateman the openapi.json is auto-generated. Everything is defined in ruby within the `spec/` directory. For example, you'll need to document the new tags attribute within the meta schema: https://github.com/RedHatInsights/compliance-backend/blob/master/spec/api/v1/schemas/metadata.rb
> 
> When you run rswag, the autogenerated examples should also contain the new field in all API responses. To demonstrate the usage, add it to an existing spec or define a new one within the [systems spec](https://github.com/RedHatInsights/compliance-backend/blob/master/spec/integration/systems_spec.rb).
> 
> You did similar steps when you implemented sorting. Pretty much any API change requires updating the specs like this, since the specs are used to generate the API docs.

:facepalm: sorry, I did it in the morning :facepalm: :scream: I thought there was something weird :laughing:

test/lib/query_parser_test.rb
```diff
@@ -0,0 +1,46 @@
...
+class QueryParserTest < ActiveSupport::TestCase
+  setup do
+    @parser = Insights::API::Common::QueryParser.new(
+      Rack::Utils.default_query_parser
+    )
+  end
+
+  test '#parse_nested_query' do
+    {
+      'foo=bar' => { foo: 'bar' },
```
akofink, line 15: How about one that encodes the slash `/`? `tags=insights-client%2Fenv%3Dprod`

akofink (changes requested):
It seems to work as expected for the most part, but the Inventory API accepts tags with encoded slashes, and this implementation does not.

```
# Encoded slash
GET /api/compliance/v1/systems?tags=insights-client%2Fenv=prod
...
    "meta": {
        "limit": 10,
        "offset": 1,
        "search": "has_test_results=true or has_policy=true",
        "tags": [],
        "total": 2
    }
}

# Not encoded slash
GET /api/compliance/v1/systems?tags=insights-client/env=prod
...
    "meta": {
        "limit": 10,
        "offset": 1,
        "search": "has_test_results=true or has_policy=true",
        "tags": [
            {
                "key": "env",
                "namespace": "insights-client",
                "value": "prod"
            }
        ],
        "total": 0
    }
}
```

skateman:
@akofink according to the documentation the Insights API accepts encoded `/` and `=` if they are part of a namespace, key or value, but not when delimiting these, so the `insights-client%2Fenv=prod` is an invalid tag as it doesn't have a delimiter between the namespace and the key.  A valid tag should always contain the unescaped delimiters, e.g. `insights-client/system%2Fenv=prod`

test/lib/query_parser_test.rb
skateman: Sure, I can cover that as well.

akofink: I took a look at the docs, which I assume is mostly under [this header](https://platform-docs.cloud.paas.psi.redhat.com/backend/inventory.html?highlight=cyndi#string-representation). It does not say that the delimiters must be escaped, it simply says that `=` or `/` in the namespace, key, or value must be escaped. I haven't found docs that say what you're claiming:

>  the Insights API accepts encoded / and = if they are part of a namespace, key or value, **but not when delimiting these**

The docs state:

> Note that in transit (e.g. when sent as an HTTP query parameter value) the tag will be URI-encoded again (as required by RFC3986). This happens automatically in modern HTTP browsers and http client libraries. This should not be confused with the aforementioned encoding of the tag segments described above which the client must always do.

Again, I found this discrepancy by comparing how the Inventory API works, and it accepts encoded delimiters.

skateman: Okay, so how should we handle if the delimiters and the `/` and `=` symbols in the fields are both escaped? There's no way to distinguish them, only if we allow double-escaping, but should we also expect double escaping if the delimiters are in the unescaped form? 

Btw I've been trying to follow [this one](https://clouddot.pages.redhat.com/docs/dev/services/inventory.html#_string_representation)

akofink: Here is an example request from the Inventory UI:

```
https://cloud.redhat.com/api/inventory/v1/hosts?per_page=50&page=1&order_by=updated&order_how=DESC&staleness=fresh&staleness=stale&tags=satellite%2Flifecycle_environment%3DLibrary&registered_with=insights&fields%5Bsystem_profile%5D%5B%5D=operating_system
```

I'm not sure I understand the issue with parsing the tags. The first `/` you get, encoded or not, will be the delimiter between the namespace and key. The first `=` you get after that, encoded or not, will be the delimiter between the key and the value. Any further `=` you get will be part of the value, and btw must be encoded.

skateman: > The first / you get, encoded or not, will be the delimiter between the namespace and key.

So there's no way for the namespace to contain a `/`? :thinking: 

 > The first = you get after that, encoded or not, will be the delimiter between the key and the value

So there's no way for the key to contain a `=`?

>  Any further = you get will be part of the value, and btw must be encoded.

So the key can contain a `/` symbol and the value can contain both `/` and `=`?

This is confusing :thinking: :rofl:

akofink: Please check with the Inventory API to verify your questions.

> So there's no way for the namespace to contain a `/`?

```
$ http https://cloud.redhat.com/api/inventory/v1/hosts tags==satellite/foo/lifecycle_environment=Library # NB: httpie escapes this query param
HTTP/1.1 400 Bad Request

{
    "detail": "'satellite/foo/lifecycle_environment=Library' does not match '^([^=/]+/)?[^=/]+(=[^=/]+)?$'\n\nFailed validating 'pattern' in schema['items']:\n    {'pattern': '^([^=/]+/)?[^=/]+(=[^=/]+)?$', 'type': 'string'}\n\nOn instance[0]:\n    'satellite/foo/lifecycle_environment=Library'",
    "status": 400,
    "title": "Bad Request",
    "type": "about:blank"
}
```

Nope, seems not. How nice of them to provide the regex :laughing: They should have included it in the docs!

> So there's no way for the key to contain a `=`

Nope, seems not. Nor can it contain a `/`

> So the key can contain a / symbol and the value can contain both `/` and `=`?

The key cannot contain a `/` according to the regex and testing with their API. The value can contain `=` but not `/` according to their API and the regex.

All that said, if you double escape anything, it's fair game.

i.e. `?tags=satellite%2Flifecycle_environment%3DLi%%3Dbrary`

skateman: @akofink from [this page](https://clouddot.pages.redhat.com/docs/dev/services/inventory.html#_string_representation):
> The fact that a namespace, key or value can contain / and = characters creates a potential ambiguity when parsing a tag in the string representation.

I am just more confused now :disappointed:

akofink: They have to be escaped. That's the very next sentence:

> In order to avoid this ambiguity it is required that / and = characters that occur in the namespace, key or value are escaped using RFC3986 (URI encoding).

This means the API requester (aka the client) must escape these values if they're in the namespace, key, or value. I demonstrated that above with the Inventory API. I definitely agree with you that it's confusing :laughing:

skateman: Okay, now I just understand it less than before :sweat:

vkrizan: It seems that we are confusing client side escaping with server side. Clients should escape it one their side, escaping the query string's param. It should be client's responsibility to do so.

If we are talking about using these two characters as a value within the tags, then the resulting query string sent by a client is double encoded. Server should get it in as if the URL wasn't decoded, but just for the value part.

```
// client sends
console.log(encodeURIComponent(encodeURIComponent('/=')));
> "%252F%253D"
// server gets 
console.log(encodeURIComponent('/='));
> "%2F%3D"
```

@akofink this case seems to be validate a few lines below.

akofink: Like I keep saying, please validate this works the same as inventory's tags param:

```
GET inventory/v1/hosts?tags=satellite/lifecycle_environment=Library # works
GET compliance/v1/systems?tags=satellite/lifecycle_environment=Library # works

GET inventory/v1/hosts?tags=satellite%2Flifecycle_environment%3DLibrary # works
GET compliance/v1/systems?tags=satellite%2Flifecycle_environment%3DLibrary # doesn't work
```

The fact that the tests pass indicate this case is not actually covered.

vkrizan: These four should be equivalent:
```
tags=satellite/lifecycle_environment=Library
tags=satellite%2Flifecycle_environment=Library
tags=satellite/lifecycle_environment%3DLibrary
tags=satellite%2Flifecycle_environment%3DLibrary
```

What we are missing is a decode of a raw query string.
```
> params
=> {"tags"=>"satellite/lifecycle_environment=Library"}
> query_string
=> "tags=satellite%2Flifecycle_environment%3DLibrary"
> Rack::Utils.unescape(query_string)
=> "tags=satellite/lifecycle_environment=Library
```

Speaking with @skateman, we kinda agree that replacing `tags=` to `tags[]=` (or allowing array params w/o `[]`)  could solve the first parsing problem. Then the values can be handed in the controller (or middleware).

lib/adjust_tags/middleware.rb
```diff
@@ -0,0 +1,27 @@
...
+        class Middleware
+          def initialize(app)
+            @app = app
+          end
+
+          def call(env)
+            qs = env['QUERY_STRING'].sub(/^tags=/, 'tags[]=')
+                                    .gsub(/&tags=/, '&tags[]=')
+
+            env['REQUEST_URI'] = env['REQUEST_URI'].sub(env['QUERY_STRING'], qs)
```
vkrizan, line 18: This could replace parts of URI that should not be replaced. Can it do the substitution from the end of the URI?

skateman:
I think everything should be fixed now...

swagger/v1/openapi.json
vkrizan: Yeah, it's kinda different from what the API actually returns within data.

vkrizan (approved):
Looks great!

swagger/v1/openapi.json
skateman: This is specified in the API docs

skateman: Well, there are some specs about how is it queried and how is it stored and they are different...

akofink: You did not add a usage example :disappointed: Within systems_spec.rb, we should show how to use the tags and an example response.

spec/integration/systems_spec.rb
```diff
@@ -18,6 +18,7 @@
       auth_header
       pagination_params
       search_params
+      tags_params
```
akofink, line 21: This should be added to all the specs because all our endpoints accept tags as implemented

skateman: @akofink only the systems endpoint will allow tags, in every other case it'll ignore the param. So not sure if it should be everywhere :thinking:

swagger/v1/openapi.json
skateman: :thinking: I'm not entirely sure what does this mean

PR #915, opened 2021-07-13
fix(score): RHICOMPL-2026 search hosts based on the correct scoring

skateman:
When searching for hosts with a certain profile score we were calculating this score based on the number of passing rules, however, this is incorrect and also slow. Instead we should rely the score that is already set in the latest test results.
So this change should not just correct the searching, but also speed up the process when having a lot of records.

Closes https://github.com/RedHatInsights/compliance-backend/pull/910

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/host_searching.rb
+++ b/app/models/concerns/host_searching.rb
@@ -4,6 +4,8 @@
 module HostSearching
   extend ActiveSupport::Concern
 
+  NUM_OPERATORS = ['=', '>', '<', '<=', '>=', '!='].freeze
+
   included do
     scoped_search on: %i[id display_name], only_explicit: true
     scoped_search on: :display_name, rename: :name
@@ -77,6 +79,8 @@ module HostSearching
                  .or(::Profile.where(policy_id: policy_or_profile_id))
                  .or(::Profile.where(policy_id: profiles.select(:policy_id)))
 
+      RequestStore.store['scoped_search_context_profiles'] = profiles
+
       { conditions: "hosts.id IN (#{
         ::TestResult.where(profile: profiles).select(:host_id).to_sql
       })" }
@@ -109,13 +113,20 @@ module HostSearching
     end
 
     def filter_by_compliance_score(_filter, operator, score)
-      ids = ::Host.includes(:test_result_profiles).select do |host|
-        host.compliance_score.public_send(operator, score.to_f)
+      unless NUM_OPERATORS.include?(operator)
+        raise ActiveRecord::StatementInvalid
       end
-      ids = ids.pluck(:id).map { |id| "'#{id}'" }
-      return { conditions: '1=0' } if ids.empty?
 
-      { conditions: "hosts.id IN(#{ids.join(',')})" }
+      profiles = RequestStore.store['scoped_search_context_profiles']
+
+      raise ScopedSearch::QueryNotSupported if profiles.nil?
+
+      hosts = ::TestResult.where("score #{operator} ?", score.to_f)
+                          .where(profile: profiles)
+                          .latest
+                          .select('test_results.host_id')
+
+      { conditions: "hosts.id IN(#{hosts.to_sql})" }
     end
 
     def test_results?(_filter, _operator, value)
--- a/Gemfile
+++ b/Gemfile
@@ -72,6 +72,9 @@ gem 'stronger_parameters'
 # Clowder config
 gem 'clowder-common-ruby'
 
+# Support for per request thread-local variables
+gem 'request_store'
+
 group :development, :test do
   gem 'brakeman'
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -264,6 +264,8 @@ GEM
     redis (4.2.1)
     reline (0.1.4)
       io-console (~> 0.5)
+    request_store (1.5.0)
+      rack (>= 1.4)
     rspec-core (3.9.2)
       rspec-support (~> 3.9.3)
     rspec-expectations (3.9.1)
@@ -396,6 +398,7 @@ DEPENDENCIES
   rack-cors
   rails (~> 5.2.4, >= 5.2.4.4)
   redis (~> 4.0)
+  request_store
   rspec-rails
   rswag-api
   rswag-specs
--- a/test/graphql/queries/system_query_test.rb
+++ b/test/graphql/queries/system_query_test.rb
@@ -780,6 +780,67 @@ class SystemQueryTest < ActiveSupport::TestCase
     assert graphql_host.assigned_profiles.pluck(:id).include?(@profile1.id)
   end
 
+  test 'search for systems with a specific score for a given profile' do
+    # setup_two_hosts
+    @host2 = FactoryBot.create(
+      :host,
+      account: @user.account.account_number,
+      os_minor_version: 7
+    )
+    @profile1.policy.update(hosts: [@host1, @host2])
+
+    FactoryBot.create(
+      :test_result,
+      host: @host1,
+      profile: @profile1,
+      score: 52
+    )
+    FactoryBot.create(
+      :test_result,
+      host: @host2,
+      profile: @profile1,
+      score: 40
+    )
+
+    query = <<-GRAPHQL
+      query getSystems($filter: String!, $policyId: ID) {
+        systems(search: $filter) {
+          totalCount
+          edges {
+            node {
+              id
+              name
+              testResultProfiles(policyId: $policyId) {
+                id
+                name
+                lastScanned
+                compliant
+                score
+              }
+            }
+          }
+        }
+      }
+    GRAPHQL
+
+    filter = <<-FILTER
+      (with_results_for_policy_id = #{@profile1.id}) and
+      (has_test_results = true and compliance_score >= 45 and compliance_score <= 55)
+    FILTER
+
+    result = Schema.execute(
+      query,
+      variables: { filter: filter },
+      context: { current_user: @user }
+    )
+
+    node = result['data']['systems']['edges'][0]['node']
+
+    assert_equal result['data']['systems']['totalCount'], 1
+    assert_equal node['id'], @host1.id
+    assert_equal node['testResultProfiles'][0]['score'], 52
+  end
+
   test 'query system rules when results contain wrong rule_ids' do
     query = <<-GRAPHQL
     query System($systemId: String!){
--- a/test/models/host_test.rb
+++ b/test/models/host_test.rb
@@ -154,15 +154,37 @@ class HostTest < ActiveSupport::TestCase
       assert_equal 50.0, @host1.compliance_score
     end
 
+    should 'fail to find based on compliance score without a policy' do
+      RequestStore.clear!
+      assert_raises(::ScopedSearch::QueryNotSupported) do
+        Host.search_for('compliance_score >= 40 and compliance_score <= 60')
+      end
+    end
+
     should 'be able to find based on compliance score' do
       assert_includes(
-        Host.search_for('compliance_score >= 40 and compliance_score <= 60'),
+        Host.search_for("
+          with_results_for_policy_id = #{@profile.policy_id} and (
+            compliance_score >= #{@host1.test_results.first.score - 10} and
+            compliance_score <= #{@host1.test_results.first.score + 10}
+          )
+        "),
         @host1
       )
+
+      RequestStore.clear!
+
       assert_not_includes(
-        Host.search_for('compliance_score >= 0 and compliance_score <= 39'),
+        Host.search_for("
+          with_results_for_policy_id = #{@profile.policy_id} and (
+            compliance_score >= 0 and
+            compliance_score <= #{@host1.test_results.first.score - 1}
+          )
+        "),
         @host1
       )
+
+      RequestStore.clear!
     end
 
     should 'be able to find based on compliance true/false' do
@@ -341,5 +363,11 @@ class HostTest < ActiveSupport::TestCase
 
   test '#os_minor_versions' do
     assert_equal Host.os_minor_versions([@host1]), [4]
+  end
+
+  should 'fail if a wrong operator is passed to filter_by_compliance_score' do
+    assert_raises(ActiveRecord::StatementInvalid) do
+      Host.filter_by_compliance_score(nil, '=\'\' or 1=1);--', 10)
+    end
   end
 end


app/models/concerns/host_searching.rb
```diff
@@ -109,9 +109,8 @@ def filter_by_compliance(_filter, operator, value)
     end
 
     def filter_by_compliance_score(_filter, operator, score)
-      ids = ::Host.includes(:test_result_profiles).select do |host|
-        host.compliance_score.public_send(operator, score.to_f)
-      end
+      ids = ::Host.joins(:test_result_profiles)
+                  .where("test_results.score #{operator} ?", score.to_f)
```
skateman, line 113: I'm not happy about the `#{operator} ?` here, but I am not sure if this can be done via Arel as the `operator` is coming from outside :confused: :shrug:

akofink: See the related filters in this module. We can't just use the raw `operator` directly because it's not necessarily SQL. For example, users can pass `!=` or `<>` for not equal. We can limit the available operators like other scoped searches do at the top of this file.

skateman: So what should be the allowed operators? `['>', '>=', '=', '<=', '<']`?

vkrizan: Try mapping operators to Arel. There might be a method to do that already...

Definitely do not open this for an SQL Injection. Consider at least asserting the values before formatting the string.

vkrizan (changes requested):

app/models/concerns/host_searching.rb
skateman: The `operator` method is already sanitized to the allowed operators, so that should be safe. I just want to know which ones should we allow above...

akofink: Using the `?` format should cause ActiveRecord to sanitize the SQL, so we shouldn't need to worry about injection, but we could certainly write a test that confirms it. Any/all operators that make sense for a float comparison should be included IMO. In fact, the [default seems ok](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/definition.rb#L255) if we can translate it to SQL:

```rb
['=', '>', '<', '<=', '>=', '!=', '^', '!^']  
```

vkrizan: @skateman where is it sanitized? It could be much safer to do some assertion in this method to be absolutely sure.

@akofink SQL placeholders would quote it as a value, which the SQL parser does not expect for an operator (IIRC).

akofink: Confirmed, this can be a problem in ActiveRecord:

```
[23] pry(main)> Profile.count
   (0.8ms)  SELECT COUNT(*) FROM "profiles"
=> 0
[24] pry(main)> op='=\'\' or 1=1);--'
=> "='' or 1=1);--"
[25] pry(main)> Profile.where("ref_id#{op}?", 'bar')
  Profile Load (0.6ms)  SELECT "profiles".* FROM "profiles" WHERE (ref_id='' or 1=1);--'bar')
=> [] # this would be all profiles if there were any in the DB, even those outside the user account
```

scoped_search's query language likely will not allow the operator to be injected in this way, since it explicitly validates the operator exists within the allowable operators, but it wouldn't hurt to harden (and test) this.

skateman: The `options` argument here is NOT an SQL operator, but a string-based parsed value coming from `scoped_search` that is luckily interpretable as an SQL operator. It is not vulnerable to an SQL injection described above by @akofink because it only allows fields that are declared using the `scoped_search` method in our models. Having an operator other than the ones listed above would [raise](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/definition.rb#L258) a `ScopedSearch::QueryNotSupported` exception and even the right side of the operator is sanitized using a [regular expression](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/definition.rb#L261). 

If we really want to do some kind of hardening, I would imagine it like this:
```ruby
    # These are dynamically defined, so can't just pull them out from the scoped_search library :(
    NUM_OPERATORS = ['=', '>', '<', '<=', '>=', '!=', '^', '!^'].freeze

    def filter_by_compliance_score(_filter, operator, score)
      unless NUM_OPERATORS.include?(operator)
        raise ActiveRecord::StatementInvalid
      end
```

But I kinda find it pointless and also ugly :see_no_evil: 

Regarding testing, I don't think it provides any coverage to just unit-test the `filter_by_compliance_score` method with arguments that would never come from `scoped_search` anyway. I am thinking hard about higher level tests that would try to reproduce the injection that @akofink described above (or similar), but I don't really know how to formulate it in a query without making it obviously too easy for the parser to catch it. So I'm kinda lost in here what should I really do :shrug: 

Just as an example, if you do `1=1` in the search, you will get the following error without even hitting the `filter_by_compliance_score` method:
```
Field '1' not recognized for searching!
```

vkrizan: As mentioned, the `operator` is scoped search one. It does not represent an SQL operator, only it resembles it. For example `^` and `!^` are not valid in numerical SQL context:
```
compliance_dev=# SELECT 1 FROM profiles where score ^ 1;
ERROR:  argument of WHERE must be type boolean, not type numeric
LINE 1: SELECT 1 FROM profiles where score ^ 1;
```

Considering that this is a public class method and the `operator` value is going directly to a raw SQL, it is our responsibility to harden and avoid any potential situation. Even if it might seem useless at the time. scoped_search might change its behavior or documentation, or one can use the method elsewhere in our app not considering the consequences. 

I highly suggest to take the `operator` value as "a switch from allowed values".

akofink: I agree with nearly everything you said, @skateman; the risk of injection is minimal or even nonexistent. Still, what @vkrizan says about the operator not being SQL is an issue. We need some mapping from the scoped search operator to SQL.

skateman: Yeah, it's already there

app/models/concerns/host_searching.rb
```diff
@@ -109,9 +111,12 @@ def filter_by_compliance(_filter, operator, value)
...
     def filter_by_compliance_score(_filter, operator, score)
-      ids = ::Host.includes(:test_result_profiles).select do |host|
-        host.compliance_score.public_send(operator, score.to_f)
+      unless NUM_OPERATORS.include?(operator)
+        raise ActiveRecord::StatementInvalid
       end
+
+      ids = ::Host.joins(:test_result_profiles)
+                  .where("test_results.score #{operator} ?", score.to_f)
       ids = ids.pluck(:id).map { |id| "'#{id}'" }
```
vkrizan, line 120: Can it use the previous query as a subquery into the condition?
We are putting trust on the ids coming from Inventory.

vkrizan (commented):
Seems to work okay. 
If there are NaN values they would be taken as an infinity, therefore `compliance_score > 9999` always yields results (if host and test results exist).

app/models/concerns/host_searching.rb
skateman: No, because we don't have access to the right context (profile/policy). Also if you read it further, we're building a "filter" based on those IDs and then it's being called on an already reduced subset of the Hosts in the context of the given profile/policy AND the user.

vkrizan: Can you please elaborate on the profile/policy context that it makes the set reduced? 
The lookup by score here does not seem to be affected by any context. It would take hosts by their score from any of its test results, regardless of a policy.

Regarding the ids formatted to raw SQL, there is a `parameter` that can be passed within the returning hash with those ids. It could work, unless some other condition/query would shift the order of placeholders.

skateman: There's a base query initiated by the browser and it already provides a reduced set of hosts (in our case under a policy). This helper query above gets a list of all IDs that meet the given conditions (in our case the score is below/above something) and the resulting IDs are being used to do a final sweep on the base query's result. The final result is an intersection of these two and this implementation was not much different before I introduced the changes.

Unfortunately there's no way to link these queries together as the second one is being called from inside `scoped_search` and the implementation doesn't allow us to pass any kind of context other than the three sanitized arguments.

vkrizan: (as discussed on chat) this is a more complex problem. 
Imagine a scenario with a Host has two policies. Policy A in which it has score 20, and Policy B in which it has score 100. Query of `policy = A and compliance_score > 50` would still yield that host.

I blame the introduction of virtual fields that are context based.

skateman: A possible solution could be to turn this around and instead of requesting hosts with (a limited set of) profiles underneath we could request the (limited set of) profiles and the hosts underneath. By turning the whole thing around, it should be possible to safely narrow down the results with my method in this PR.

skateman: Nevermind, this is not gonna work, for the same reason :disappointed: `scoped_search` still won't see the parent

skateman: So I guess we should go around `scoped_search` by introducing the `minScore` and `maxScore` optional arguments on the `testResultProfiles` and use that to narrow down our results. @akofink wdyt?

akofink: After looking into this, I now see what you mean about the scoped search context where each search term gets a filter method. I agree that we're hitting the limits of scoped search and maybe should opt to implement this directly by introducing one or two new parameters. We can validate parameters exist when others are present (i.e. `policy_id` and `policy_score`) or do something weird like `policy_score[id]=8b409091-567f-47a7-b79a-d368cd8372e1&policy_score[min]=70&policy_score[max]=80` or accept a JSON object like `policy_score={"id": "8b409091-567f-47a7-b79a-d368cd8372e1", "range": "70..80"}` (or min/max or whatever really).

btw, ActiveRecord supports ranges; Here's an example scope that, I think, does what we want:

```rb
scope :with_policy_score, lambda { |policy, range_or_value|
  where(
    id: TestResult.latest.where(
      profile_id: policy.profiles, score: range_or_value
    ).select(:host_id)
  )
}
```

And then:
```rb
Host.with_policy_score(Policy.first, 75..85) # between 75 and 85
Host.with_policy_score(Policy.first, 75..Float::INFINITY) # greater than 75
Host.with_policy_score(Policy.first, 0..75) # less than 75
Host.with_policy_score(Policy.first, 75) # or even exactly 75
```

As you've already identified, converting from REST params and GraphQL variables into a call to that scope is the trickier part. Sorry again for not quite following during our meeting about this.

skateman: For now I went with a minimal approach with the `current_scope` that @vkrizan suggested and it seems working. It also requires an additional param (to set the context), but without that param the query doesn't even make sense (hosts without test results)

app/models/concerns/host_searching.rb
```diff
@@ -109,13 +113,18 @@ def filter_by_compliance(_filter, operator, value)
...
+      unless NUM_OPERATORS.include?(operator)
+        raise ActiveRecord::StatementInvalid
       end
-      ids = ids.pluck(:id).map { |id| "'#{id}'" }
-      return { conditions: '1=0' } if ids.empty?
 
-      { conditions: "hosts.id IN(#{ids.join(',')})" }
+      profiles = Profile.current_scope.select(:id)
+
+      hosts = ::Host.joins(:latest_test_results)
```
vkrizan, line 122: Does this join give expected results?
Here is the SQL:
```
SELECT "inventory"."hosts".* 
FROM "inventory"."hosts" 
INNER JOIN "test_results" ON "test_results"."host_id" = "inventory"."hosts"."id" 
WHERE "inventory"."hosts"."account" = '00023'
```

skateman: There was a point in my tests when this was working as it should :see_no_evil: will investigate further...

skateman: Okay, so `has_many` doesn't like `join` calls in their lambda :disappointed: as a solution we can approach this from the other end via `::TestResult.latest` and narrow it down via `profile_id` and score from there to get the `host_id`s

skateman:
So I found another issue :facepalm: this is getting more and more exciting :skull_and_crossbones: :tada: because we support(ed) searching by score without specifying a profile. This means that when filtering for a score without specifying a single profile, we used to look for a ratio of all the passed rules for a given host in all assigned profiles. This calculation is obviously wrong as it doesn't use *The Right Way&trade;* of relying on the uploaded score coming from the report.

I see a few options as a solution:
* keep the old way of calculating the scores if no profile specified, just move it to SQL for a better performance
* require a profile for calculating the score and so get rid of the filtering via the overall score completely
* calculate the score as an average of the scores from the latest test results for a given host

app/models/concerns/host_searching.rb
skateman: Well, bad news is that the `current_scope` cannot be used as other things rely on it and it kinda breaks all our tests :confused: and using the underlying `Thread.current` can leak data to a next request, so that's pretty much unsafe and it might not be a good idea to clear its value in a different context (controller) as we can't really clean it in the `filter_by_compliance_score` because we might call this method twice (less than + greater than) :confused:

skateman: @akofink I guess the best way to get this through `scoped_search` is to pass a JSON-like object to the right side as it can hold both the ID and the interval...just need to look into this deeper if `scoped_search` likes it and how much can we force this :see_no_evil:

skateman: OR maybe https://github.com/steveklabnik/request_store :thinking:

app/models/concerns/host_searching.rb
```diff
@@ -77,6 +79,8 @@ def filter_with_results_for_policy(_filter, _operator, policy_or_profile_id)
                  .or(::Profile.where(policy_id: policy_or_profile_id))
                  .or(::Profile.where(policy_id: profiles.select(:policy_id)))
 
+      Thread.current['filtered_by_profiles'] = profiles
```
vkrizan, line 82: (as you mentioned) this might leak to another request.

How about making it more global like `Thread.current['scoped_search_context'] = {...}` and reset it after each call of `Model.search_for`?

```
class ApplicationRecord < ActiveRecord::Base
  def self.search_for
    super
  ensure
    # reset context
  end
```

skateman: Good idea, or maybe we could just use https://github.com/steveklabnik/request_store ?

skateman: Good idea, there are also some request-wide stores that use `Thread.current`, like [this](https://github.com/steveklabnik/request_store), that prevent leaking. However, I prefer your idea more :+1:

app/models/application_record.rb
```diff
@@ -17,6 +17,13 @@ def self.arel_find(collection, batch_size = 1000)
     end.inject([], &:+)
   end
 
+  def self.search_for(*args)
+    super
```
vkrizan, line 21:
```
NoMethodError (super: no superclass method `search_for'
```

skateman: Yeah, wrestling with that :see_no_evil:

vkrizan: here is what scoped_search does:
```
register_named_scope! unless klass.respond_to?(:search_for)
```
we need to do other magic...
probably adding a brand new class method.

skateman: I can redefine `respond_to` to answer correctly :rofl:

skateman: I am going to figure this out :laughing: no worries

skateman:
Okay, so I went with the  `request_store`, it is using thread-local variables with a middleware that cleans them among requests.

app/models/concerns/host_searching.rb
```diff
@@ -109,13 +113,20 @@ def filter_by_compliance(_filter, operator, value)
...
       end
-      ids = ids.pluck(:id).map { |id| "'#{id}'" }
-      return { conditions: '1=0' } if ids.empty?
 
-      { conditions: "hosts.id IN(#{ids.join(',')})" }
+      profiles = RequestStore.store['scoped_search_context']
+
+      raise ScopedSearch::QueryNotSupported if profiles.nil?
+
+      hosts = ::TestResult.where("score #{operator} ?", score.to_f)
```
vkrizan, line 124: Can it handle `NaN` values as not valid ones, meaning that they would not be selected in any search query?

vkrizan (approved):
Works very well!

app/models/concerns/host_searching.rb
skateman: Let me try...

skateman: So, I guess `NaN` is handled by PG as [infinity](https://www.youtube.com/watch?v=dL3AiuTsbOU) so if you're searching for a score in an interval, you will never actually get it as a result. However, if you are looking for all the results that are larger than 0, you will also have the `NaN` one :shrug:

vkrizan: Exactly. A test result score with a `NaN` (or inf.) isn't a correct one, therefore I think it should not be returned if one wants score 90 and up.

skateman: Well, that would require some messy hacks which we could do in a followup issue/PR

skateman:
@akofink waiting on your :+1:

akofink (approved):
It seems to work great. Nice work, @skateman!

If testing in the report details UI, note that there seem to be some minor issues with the current frontend implementation, addressed in https://github.com/RedHatInsights/compliance-frontend/pull/1319. This backend work is not affected.

PR #920, opened 2021-07-20
fix(Host): RHICOMPL-2063 deduplicate Host#test_result_profiles

skateman:
When requesting profiles through test results under hosts, it is possible that the JOIN between the two tables produces duplicate profiles between the results if there are outdated test results for a given profile. All the other fields, like score are taken from the latest test result, so these values will be correct. As a solution I passed a `distinct` into the `scope` that resolves this problem.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -23,7 +23,8 @@ class Host < ApplicationRecord
   has_many :test_results, dependent: :destroy
   include SystemLike
 
-  has_many :test_result_profiles, through: :test_results, source: :profile
+  has_many :test_result_profiles, -> { distinct },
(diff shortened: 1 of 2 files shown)


akofink (approved):
Makes sense. Thanks!

vkrizan (approved):

PR #927, opened 2021-07-23
feat(GraphQL): RHICOMPL-2037 allow filtering results by tags

skateman:
I (more or less) kept the same format as for the REST API, but here we can use proper arrays :wink: 
```
{
  "operationName": "getSystems",
  "query": "query getSystems($perPage: Int, $page: Int, $tags: [String!]) {
    systems(limit: $perPage, offset: $page, tags: $tags) {
      totalCount
      edges {
        node {
          id
          name
          osMajorVersion
          osMinorVersion
          __typename
        }
        __typename
      }
      __typename
    }
  }",
  "variables": {
    "perPage": 50,
    "page": 1, "tags": ["foo/bar=baz"]
  }
}
```

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/graphql/resolvers/concerns/collection.rb
+++ b/app/graphql/resolvers/concerns/collection.rb
@@ -7,6 +7,7 @@ module Resolvers
       extend ActiveSupport::Concern
 
       include Sorting
+      include TagFiltering
 
       included do
         type ["Types::#{self::MODEL_CLASS}".safe_constantize], null: false
@@ -14,6 +15,10 @@ module Resolvers
         argument :limit, Integer, 'Pagination limit', required: false
         argument :offset, Integer, 'Pagination offset', required: false
         argument :sort_by, [String], 'Sort results', required: false
+
+        if TagFiltering.tags_supported?(self::MODEL_CLASS)
+          argument :tags, [String], 'Filter by tags', required: false
+        end
       end
 
       def resolve(**kwargs)
@@ -24,11 +29,15 @@ module Resolvers
 
       private
 
-      def filter_list(search: nil, sort_by: nil, offset: nil, limit: nil)
+      def filter_list(
+        search: nil, sort_by: nil, tags: nil, offset: nil, limit: nil
+      )
         filters = []
         filters << search_filter(search: search) if search.present?
 
         filters << sort_filter(sort_by: sort_by) if sort_by.present?
+
+        filters << tags_filter(tags: tags) if permit_tags?(tags)
 
         if offset.present? || limit.present?
           filters << pagination_filter(offset: offset, limit: limit)
@@ -63,12 +72,23 @@ module Resolvers
         end
       end
 
+      def tags_filter(tags:)
+        lambda do |scope|
+          tags = parse_tags(tags)
+          scope.where('tags @> ?', tags.to_json)
+        end
+      end
+
       def model_class
         self.class::MODEL_CLASS
       end
 
       def authorized_scope
         Pundit.policy_scope(context[:current_user], model_class)
+      end
+
+      def permit_tags?(tags)
+        TagFiltering.tags_supported?(model_class) && tags.present?
       end
     end
   end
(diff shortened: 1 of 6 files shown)


vkrizan:
Rubocop:
```
test/graphql/queries/system_query_test.rb:550:61: C: Layout/MultilineMethodCallBraceLayout: Closing method call brace must be on the line after the last argument when opening brace is on a separate line from the first argument.
      tags: [{ namespace: 'foo', key: 'bar', value: 'baz' }])
                                                            ^
```

app/graphql/resolvers/concerns/collection.rb
```diff
@@ -7,13 +7,15 @@ module Collection
...
       include Sorting
+      include TagFiltering
 
       included do
         type ["Types::#{self::MODEL_CLASS}".safe_constantize], null: false
         argument :search, String, 'Search query', required: false
         argument :limit, Integer, 'Pagination limit', required: false
         argument :offset, Integer, 'Pagination offset', required: false
         argument :sort_by, [String], 'Sort results', required: false
+        argument :tags, [String], 'Filter by tags', required: false
```
vkrizan, line 18: Can `tags` be only available on types that supports it?

skateman: Okay, how can we do that? :sweat_smile:  :see_no_evil:

vkrizan: Can you try moving it to a concrete type? Would that work?

skateman: That's was the first I tried :confused:

akofink: `tags` is available in REST for all endpoints, even though only systems supports/does anything with it. This seems to be the same - why does it matter to remove it here but not in REST?

vkrizan: If `tags` param in REST is available on all resources, then the difference would that it is checked (I believe) there to use the tags scoping only for those that actually support it (have the field available).

skateman: It IS checked here as well, we're kinda using the same resolver. If the `tags` field is not available on the DB model, we're ignoring the passed tags ...

skateman: Nope, missed that :facepalm: so this opens a window for the question: "how it should be?" :laughing:

vkrizan: The quickest way how to achieve for the argument to be only available the type that supports it would be to add a condition here similar to the `tags_supported?` (by checking the field availability on the db model).

skateman: Fixed...

vkrizan: Is it possible to also add a similar condition to the `argument` call?
```
argument ... if self::MODEL_CLASS.column_names.include?('tags')
```

app/graphql/resolvers/concerns/collection.rb
```diff
@@ -7,13 +7,17 @@ module Collection
...
       include Sorting
+      include TagFiltering
 
       included do
         type ["Types::#{self::MODEL_CLASS}".safe_constantize], null: false
         argument :search, String, 'Search query', required: false
         argument :limit, Integer, 'Pagination limit', required: false
         argument :offset, Integer, 'Pagination offset', required: false
         argument :sort_by, [String], 'Sort results', required: false
+        if self::MODEL_CLASS.column_names.include?('tags')
```
vkrizan, line 18: If the `tags_supported?` method is made a static one (on `TagFiltering` module) the check can be kept on one place.
```
if TagFiltering.tags_supported?(self::MODEL_CLASS)
  ...
```

```
module TagFiltering
  def self.tags_supported?(resource)
    resource.column_names.include?('tags')
  end

...
```

vkrizan (approved):
Looks great!

PR #928, opened 2021-07-26
fix(openapi): RHICOMPL-2075 fix swagger/openapi for tagging

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/spec/integration/systems_spec.rb
+++ b/spec/integration/systems_spec.rb
@@ -5,7 +5,20 @@ require 'swagger_helper'
 describe 'Systems API' do
   before do
     @account = FactoryBot.create(:account)
-    @host = FactoryBot.create(:host, account: @account.account_number)
+    @host = FactoryBot.create(
+      :host,
+      account: @account.account_number,
+      tags: [{ namespace: 'foo', key: 'bar', value: 'baz' }]
+    )
+    policy = FactoryBot.create(:policy, account: @account, hosts: [@host])
+    profile = FactoryBot.create(
+      :profile,
+      :with_rules,
+      account: @account,
+      policy: policy,
+      external: true
+    )
+    FactoryBot.create(:test_result, profile: profile, host: @host)
   end
 
   path "#{Settings.path_prefix}/#{Settings.app_name}/systems" do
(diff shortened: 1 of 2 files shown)


skateman:
@akofink when I run swaggerize, the generated JSON doesn't contain any examples, so I'm not sure how to deal with this :confused:

vkrizan (approved):

akofink:
@skateman ever since the FactoryBot swap in our tests, this is painful. We still need to go back and create records from the factories for our specs which demonstrate how the API responses can look. For now, you'll have to generate the openapi.json and then I suggest using `git add -i` to add the examples hunk by hunk to the openapi.json (it should only be the one for systems).

skateman:
> For now, you'll have to generate the openapi.json and then I suggest using `git add -i` to add the examples hunk by hunk to the openapi.json (it should only be the one for systems).

@akofink the problem is that it doesn't generate anything else than what is in this PR and I suspect that's not enough, i.e due to the missing records we have no example related to the tags :shrug:

akofink:
> due to the missing records we have no example related to the tags

Ah! That makes sense :laughing: Can we modify a host to give it tags for the test, then? Seems like maybe we need to enhance the hosts factory.

skateman:
> > due to the missing records we have no example related to the tags
> 
> Ah! That makes sense  Can we modify a host to give it tags for the test, then? Seems like maybe we need to enhance the hosts factory or just add tags manually to the existing host: https://github.com/RedHatInsights/compliance-backend/blob/master/spec/integration/systems_spec.rb#L8

Yeah, this won't work either, at least not with `bin/rake rswag:specs:swaggerize` :confused: even if I have everything set up correctly in the `before` hook. It's suspicious that the hook doesn't even get called :disappointed: at least I can't access my custom breakpoint that I added there.

akofink:
The only way I've ever been able to regenerate the openapi.json is by modifying the command run with that rake task:

```
ruby -I/usr/local/bundle/gems/rspec-core-3.9.2/lib:/usr/local/bundle/gems/rspec-support-3.9.3/lib /usr/local/bundle/gems/rspec-core-3.9.2/exe/rspec --pattern spec/integration/\*\*/\*_spec.rb --format Rswag::Specs::SwaggerFormatter --order defined
```

Last I checked, the rake task was adding --dry-run. It's been a very long time since I looked into this or if it works today.

skateman:
@akofink we should document this somewhere, or create a rake task :shrug: because I didn't even know about it

akofink (approved):
Thanks, @skateman! Sorry about the rswag nonsense :disappointed:

PR #933, opened 2021-07-28
feat(metrics): RHICOMPL-2050 gauge for measuring huge policies

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/lib/prometheus/business_collector.rb
+++ b/lib/prometheus/business_collector.rb
@@ -42,7 +42,7 @@ class BusinessCollector < PrometheusExporter::Server::TypeCollector
       'total_50plus_policies', 'Policies having 50 or more hosts'
     )
     @client_50plus_policies = PrometheusExporter::Metric::Gauge.new(
-      'client_50plus_policies', 'Policies from clients having 50 or more hosts (excliedes Red Hat)'
+      'client_50plus_policies', 'Policies from clients having 50 or more hosts (excludes Red Hat)'
     )
     @total_systems = PrometheusExporter::Metric::Gauge.new(
       'total_systems', 'Systems'
--- a/lib/prometheus/business_collector.rb
+++ b/lib/prometheus/business_collector.rb
@@ -18,12 +18,12 @@ class BusinessCollector < PrometheusExporter::Server::TypeCollector
       'client_accounts_with_hosts',
       'Client accounts with 1 or more hosts (excludes Red Hat)'
     )
-    @total_accounts_with_huge_policies = PrometheusExporter::Metric::Gauge.new(
-      'total_accounts_with_huge_policies',
+    @total_accounts_with_50plus_hosts_per_policy = PrometheusExporter::Metric::Gauge.new(
+      'total_accounts_with_50plus_hosts_per_policy',
       'Total accounts with 1 policy or more having 50 or more hosts'
     )
-    @client_accounts_with_huge_policies = PrometheusExporter::Metric::Gauge.new(
-      'client_accounts_with_huge_policies',
+    @client_accounts_with_50plus_hosts_per_policy = PrometheusExporter::Metric::Gauge.new(
+      'client_accounts_with_50plus_hosts_per_policy',
       'Client accounts with 1 policy or more having 50 or more hosts (excludes Red Hat)'
     )
     @total_policies = PrometheusExporter::Metric::Gauge.new(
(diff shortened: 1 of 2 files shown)


lib/prometheus/business_collector.rb
```diff
@@ -38,6 +38,12 @@ def initialize
     @client_policies = PrometheusExporter::Metric::Gauge.new(
       'client_policies', 'Policies from clients (excludes Red Hat)'
     )
+    @total_huge_policies = PrometheusExporter::Metric::Gauge.new(
```
vkrizan, line 41: How about `@total_50plus_policies`?

lib/prometheus/business_collector.rb
```diff
@@ -111,6 +117,11 @@ def collect
     @client_accounts_with_huge_policies.observe(
       client_accounts.where(id: hosts_50_plus).count(:account_number)
     )
+
+    @total_huge_policies.observe(hosts_50_plus.count.size)
```
vkrizan, line 121: Observation: `hosts_50_plus` is a bit misleading. It might be smth like `policies_50plus_hosts`.

vkrizan (approved):
:+1:

lib/prometheus/business_collector.rb
```diff
@@ -18,12 +18,12 @@ def initialize
       'client_accounts_with_hosts',
       'Client accounts with 1 or more hosts (excludes Red Hat)'
     )
-    @total_accounts_with_huge_policies = PrometheusExporter::Metric::Gauge.new(
-      'total_accounts_with_huge_policies',
+    @total_accounts_with_50plus_policies = PrometheusExporter::Metric::Gauge.new(
+      'total_accounts_with_50plus_policies',
```
akofink, line 22: This makes it sound like the number of accounts that have 50 or more policies. Could we specify it as `total_accounts_with_50plus_hosts_per_policy` or similar instead?

skateman: I guess this naming is getting out of hand already :see_no_evil:

akofink: It doesn't matter how long the metric/variable name is - it should be as easy to understand as possible from within Grafana.

lib/prometheus/business_collector.rb
```diff
@@ -38,6 +38,12 @@ def initialize
     @client_policies = PrometheusExporter::Metric::Gauge.new(
       'client_policies', 'Policies from clients (excludes Red Hat)'
     )
+    @total_50plus_policies = PrometheusExporter::Metric::Gauge.new(
+      'total_50plus_policies', 'Policies having 50 or more hosts'
+    )
+    @client_50plus_policies = PrometheusExporter::Metric::Gauge.new(
+      'client_50plus_policies', 'Policies from clients having 50 or more hosts (excliedes Red Hat)'
```
akofink, line 45:
```suggestion
      'client_50plus_policies', 'Policies from clients having 50 or more hosts (excludes Red Hat)'
```

lib/prometheus/business_collector.rb
```diff
@@ -99,17 +105,22 @@ def collect
...
 
-    @total_accounts_with_huge_policies.observe(
-      Account.where(id: hosts_50_plus).count(:account_number)
+    @client_accounts_with_50plus_hosts_per_policy.observe(
+      client_accounts.where(id: policies_50plus_hosts).count(:account_number)
     )
 
-    @client_accounts_with_huge_policies.observe(
-      client_accounts.where(id: hosts_50_plus).count(:account_number)
+    @total_50plus_policies.observe(policies_50plus_hosts.count.size)
```
akofink, line 121: What's up with `count.size`?

https://ruby-doc.org/core-2.5.0/Integer.html#method-i-size

> Returns the number of bytes in the machine representation of int (machine dependent).

skateman: @akofink in some cases `count` behaves weird, it returns a hash :confused:

akofink: count will show the counts of each group when you use `group`. I believe you'll get what you want if you pass the single column you want to count, like `count(:policy_id)`.

skateman: No, it unfortunately doesn't work like that :confused: whatever I pass to the `count` it will return a hash :confused: because it's not being `SELECT`-ed, but we can't add anything to the `SELECT` above because then the query just won't work :shrug:

vkrizan: Indeed, the `group` changes the `count` method's behavior. I guess it is okay to leave the size there if we want to count number of groups.
https://guides.rubyonrails.org/active_record_querying.html#total-of-grouped-items

akofink: Gotcha! Sorry, I didn't realize we _wanted_ to count the groups :+1: I also could have tested it to find out (sorry about that)

PR #934, opened 2021-07-28
Allow an opt-out from using spring via an ENV variable

skateman:
When I'm running the containers for long, spring starts forking like crazy and overheats my CPU :disappointed:

### DIFF
--- a/bin/rails
+++ b/bin/rails
@@ -1,10 +1,12 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-begin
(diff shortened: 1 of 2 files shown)


skateman:
Nevermind, I'm :exploding_head:

akofink:
I mean, spring doesn't really make much sense in containers anyway. Why'd you close it?

skateman:
@akofink because it's already supported by spring OOTB :see_no_evil: @vkrizan warned me about it

PR #939, opened 2021-08-03
fix: RHICOMPL-1849 controller fix for limits <= 0 

xmicha82:
handling of limits equal to or below zero

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/controllers/concerns/pagination.rb
+++ b/app/controllers/concerns/pagination.rb
@@ -3,9 +3,12 @@
 # Default options and configuration for pagination
 module Pagination
   extend ActiveSupport::Concern
+
+  ParamType = ActionController::Parameters
+
   included do
     def pagination_limit
-      params[:limit].present? ? params[:limit].to_i : 10
+      params.permit(limit: ParamType.integer & ParamType.gt(0))[:limit] || 10
     end
 
     def pagination_offset
--- a/test/controllers/concerns/metadata_test.rb
+++ b/test/controllers/concerns/metadata_test.rb
@@ -193,5 +193,33 @@ class MetadataTest < ActionDispatch::IntegrationTest
       assert_match(/offset=#{Profile.canonical(false).count}/,
                    json_body['links']['last'])
     end
+
+    should 'return invalid parameter if limit=0' do
+      get profiles_url, params: { limit: 0 }
+      assert_response 422
+      assert_equal('Invalid parameter: limit must be greater than 0',
+                   json_body['errors'][0])
+    end
+
+    should 'return invalid parameter if limit<0' do
+      get profiles_url, params: { limit: -256 }
+      assert_response 422
+      assert_equal('Invalid parameter: limit must be greater than 0',
+                   json_body['errors'][0])
+    end
+
+    should 'return invalid parameter if limit is float' do
+      get profiles_url, params: { limit: 15.12 }
+      assert_response 422
+      assert_equal('Invalid parameter: limit must be an integer',
+                   json_body['errors'][0])
+    end
+
+    should 'return invalid parameter if limit is string' do
+      get profiles_url, params: { limit: '15.12' }
+      assert_response 422
+      assert_equal('Invalid parameter: limit must be an integer',
+                   json_body['errors'][0])
+    end
   end
 end


akofink (commented):
Thanks, @xmicha82! Your solution works great, but we have a gem installed that can do parameter type checking and validation, the stronger_parameters gem. It works similar to strong_parameters, if you're familiar with that. Basically anywhere we're using `params[:limit]`, we'd perform validation and use the validated data structure. In this case, `params[:limit]` is only used here: https://github.com/RedHatInsights/compliance-backend/blob/master/app/controllers/concerns/pagination.rb#L8

So you get the error handling for free with something like:

```rb
module Pagination
  extend ActiveSupport::Concern

  ParamType = ActionController::Parameters

  included do
    def pagination_limit
      params.permit(limit: ParamType.integer & ParamType.gt(0))[:limit] || 10
    end
...
```

Passing <= 0: 
```
{
    "errors": [
        "Invalid parameter: limit must be greater than 0"
    ]
}
```

Passing a float or string:

```
{
    "errors": [
        "Invalid parameter: limit must be an integer"
    ]
}
```

vkrizan:
@akofink Yes, indeed. Sorry we talked about it with Martin. I was with the impression that the strong params are for data passed in POST/PUT requests. I'm glad that it can be used for GET params. Thanks!

xmicha82:
@akofink Thank you for the suggestion, I used the stronger_parameters for validation and now it should work as expected.

app/controllers/concerns/pagination.rb
```diff
@@ -3,9 +3,13 @@
...
 module Pagination
   extend ActiveSupport::Concern
+
+  ParamType = ActionController::Parameters
+
   included do
     def pagination_limit
-      params[:limit].present? ? params[:limit].to_i : 10
+      params[:limit].present? ? 
+      params.permit(limit: ParamType.integer & ParamType.gt(0))[:limit].to_i : 10
```
vkrizan, line 12: You can move the `params.permit` to be a first statement. It would raise an exception before the method has time to return the value.

app/controllers/concerns/metadata.rb
```diff
@@ -66,6 +66,10 @@ def next_offset(last_offset)
     end
 
     def last_offset(total)
+      # if pagination_limit.to_f <= 0
```
vkrizan, line 69: Can you please remove the commented code?

app/controllers/concerns/pagination.rb
```diff
@@ -3,9 +3,13 @@
 # Default options and configuration for pagination
 module Pagination
   extend ActiveSupport::Concern
+
+  ParamType = ActionController::Parameters
```
akofink, line 7: Nit/optional: I wonder if we should go ahead and move this to the top of the application_controller.rb since we're doing it here and in `concerns/parameters.rb`.

app/controllers/concerns/pagination.rb
akofink: This can be done on a single line now. The `.present?` check is redundant now because we no longer need the `to_i`, which causes problems with implicit conversions (i.e. `nil.to_i -> 0`, `'foo'.to_i -> 0`). See my original snippet suggestion (I believe it works as desired).

test/controllers/concerns/metadata_test.rb
```diff
@@ -193,5 +193,19 @@ def authenticate
...
     end
+
+    should 'return invalid parameter if limit=0' do
+      get profiles_url, params: { limit: 0 }
+      assert_response 422
+      assert_equal('Invalid parameter: limit must be greater than 0',
+                   json_body['errors'][0])
+    end
+
+    should 'return invalid parameter if limit<0' do
```
akofink, line 204: It'd make sense to also validate/test string and float inputs.

akofink (commented):
Thanks, @xmicha82! Just a few more thoughts inline.

akofink (approved):
Works great! Thanks, @xmicha82!

akofink:
@vkrizan any other concerns here? Otherwise, good to merge.

vkrizan:
@xmicha82 Code looks good.

I've tested it out. Only thing I'm seeing is this output printed 5 times for each request:
```
Unpermitted parameter: :search
```

Let me check where it is coming from...

app/controllers/concerns/pagination.rb
```diff
@@ -3,9 +3,12 @@
...
+
   included do
     def pagination_limit
-      params[:limit].present? ? params[:limit].to_i : 10
+      params.permit(limit: ParamType.integer & ParamType.gt(0))[:limit] || 10
```
vkrizan, line 11: You can slice (select) only the limit to prevent `Unpermitted parameter` on extra parameters:
```
params.slice(:limit).permit(...
```

akofink: I do not recommend this. The unpermitted params log only happens in development and testing environments and is just a warning. Once we properly validate _all_ our params, the warnings will go away.

akofink: Scratch that:

> By default parameter keys that are not explicitly permitted will be logged in the development and test environment. In other environments these parameters will simply be filtered out and ignored.

We definitely don't want them filtered out and ignored :laughing:

vkrizan: I hope that they are filtered out only from the return value of `permit` and not from whole `params`. A slice would not hurt it here, as it creates a copy to leverage only one parameter validation.

akofink: I tried testing with RAILS_ENV=production, and it seems to be OK. The slice also seems fine.

vkrizan (approved):

PR #942, opened 2021-08-04
feat(metrics): RHICOMPL-2054 counter for reports per year/month

skateman:
I guess this is *The Right Way&trade;* to use a counter :thinking: 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/lib/prometheus/business_collector.rb
+++ b/lib/prometheus/business_collector.rb
@@ -56,6 +56,12 @@ class BusinessCollector < PrometheusExporter::Server::TypeCollector
     @client_systems_by_os = PrometheusExporter::Metric::Gauge.new(
       'client_systems_by_os', 'Systems by OS version (excluded Red Hat)'
     )
+    @total_reports_per_month = PrometheusExporter::Metric::Counter.new(
+      'total_reports_per_month', 'Reports uploaded per month'
+    )
+    @client_reports_per_month = PrometheusExporter::Metric::Counter.new(
+      'client_reports_per_month', 'Reports uploaded per month (excludes Red Hat)'
+    )
   end
 
   def type
@@ -122,6 +128,26 @@ class BusinessCollector < PrometheusExporter::Server::TypeCollector
     @client_50plus_policies.observe(
       policies_50plus_hosts.where('policies.account_id' => client_accounts.select(:id)).count.size
     )
+
+    @total_reports_per_month.observe(
+      TestResult.where(
+        'updated_at >= ? AND updated_at <= ?',
+        DateTime.now.beginning_of_month,
+        DateTime.now.end_of_month
+      ).count,
+      month: Time.zone.now.month,
+      year: Time.zone.now.year
+    )
+
+    @client_reports_per_month.observe(
+      client_accounts.joins(:test_results).distinct.where(
+        'test_results.updated_at >= ? AND test_results.updated_at <= ?',
+        DateTime.now.beginning_of_month,
+        DateTime.now.end_of_month
+      ).count,
+      month: Time.zone.now.month,
+      year: Time.zone.now.year
+    )
   end
 
   def metrics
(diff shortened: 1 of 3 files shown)


lib/prometheus/business_collector.rb
```diff
@@ -122,6 +128,26 @@ def collect
     @client_50plus_policies.observe(
       policies_50plus_hosts.where('policies.account_id' => client_accounts.select(:id)).count.size
     )
+
+    @total_reports_per_month.observe(
+      TestResult.where(
+        'updated_at >= ? AND updated_at <= ?',
+        DateTime.now.beginning_of_month,
+        DateTime.now.end_of_month
```
vkrizan, line 136: Nit: I'd rather use an open interval from right and use `DateTime.now.next_month.beginning_of_month`.

lib/prometheus/business_collector.rb
```diff
@@ -122,6 +128,26 @@ def collect
     @client_50plus_policies.observe(
       policies_50plus_hosts.where('policies.account_id' => client_accounts.select(:id)).count.size
     )
+
+    @total_reports_per_month.observe(
+      TestResult.where(
+        'updated_at >= ? AND updated_at <= ?',
```
vkrizan, line 134: What's the rationale behind using `updated_at` vs. `created_at`? Wouldn't the `updated_at` break the "Counter requirement" promise? :slightly_smiling_face:

lib/prometheus/business_collector.rb
```diff
@@ -56,6 +56,12 @@ def initialize
     @client_systems_by_os = PrometheusExporter::Metric::Gauge.new(
       'client_systems_by_os', 'Systems by OS version (excluded Red Hat)'
     )
+    @total_reports_per_month = PrometheusExporter::Metric::Counter.new(
```
vkrizan, line 59: TestResults (aka Reports) are deletable by users. A Gauge might be more appropriate (?).

lib/prometheus/business_collector.rb
```diff
@@ -56,6 +56,12 @@ def initialize
...
+      'total_reports_per_month', 'Reports uploaded per month'
```
vkrizan, line 60: Technically this returns counts from current month. To get the right number, this number would have to be collected and **stored** exactly at the end of the month.

@akofink ideas how we can report this correctly? Considering that reports can be deleted, there would have to be some sort of incremental counter.

Example situation:
1. we got 10 reports at this moment; observing 10
2. one report will come in; observing 11
3. two reports are removed by user; observing 9
4. one report will come in; observing 10
```

There were 12 reports, however the last number observing number (to be collected) is 10.

Are there some means we can capture this easily?
Would a moving 1 month window be a "better" choice? (Reports in last 30 days)

skateman: Well, the issue specifies the number of uploaded results by month, so a counter still might be a good solution, but I agree that the deletion messes this up :shrug:

akofink: Reports don't even need to be removed by the user - they are removed if a new result replaces it for the same host and policy. I'd say, don't stress too much about the per month/year aspect of this metric. Let's push the data into prometheus as reports per time and use grafana to see trends over larger amounts of time (we have a card to ask platform to store > 30 days worth of metrics in prometheus). We already have sidekiq job metrics which counts how many ParseReportJobs per time frame come into Compliance (`compliance_sidekiq_jobs_total{job_name="ParseReportJobs"}`). I don't see why we couldn't use that for this metric. It's a counter, as you intuited would be best for this metric. :+1:

skateman:
Closing this as sidekiq already counts it...

PR #945, opened 2021-08-10
feat: RHICOMPL-2076 Export metrics on each pod

akofink:
To test, spin up an ephemeral environment against this IMAGE_TAG and template ref:

```sh
IMAGE_TAG=$(git rev-parse --short=7 HEAD);
TEMPLATE_REF=$(git rev-parse HEAD)
bonfire deploy -sappsre --no-remove-resources compliance -i quay.io/cloudservices/compliance-backend=$IMAGE_TAG --single-replicas --set-template-ref compliance=$TEMPLATE_REF compliance
```

Check in each of the three pods (backend, sidekiq, consumer) that metrics are exposed:

```sh
oc rsh <pod>
$ curl localhost:9000/metrics | less
```

Sidekiq should report sidekiq metrics (only once at least one job is processed), the backend should be the only pod containing DB related metrics (i.e. `complianceclient_policies_by_os_major`), and all pods should have process related metrics (memory/cpu usage, etc).

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -5,14 +5,6 @@ metadata:
   name: "${APP_NAME}"
 objects:
 - apiVersion: cloud.redhat.com/v1alpha1
-  kind: ClowdJobInvocation
-  metadata:
-    name: compliance
-  spec:
-    appName: compliance
-    jobs:
-      - import-ssg
-- apiVersion: cloud.redhat.com/v1alpha1
   kind: ClowdApp
   metadata:
     name: "${APP_NAME}"
@@ -42,31 +34,17 @@ objects:
       enabled: true
       appName: "compliance"
       insightsOnly: true
-    jobs:
-    - name: compliance-import-ssg
+    deployments:
+    - name: db-migrations
       podSpec:
-        imagePullPolicy: Always
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
         - command:
             - bash
             - -c
             - |
-              for i in {1..120}; do sleep 1;
-              if bundle exec rake --trace db:abort_if_pending_migrations;
-              then exit 0; fi; done; exit 1
-          env:
-          - name: RAILS_ENV
-            value: "${RAILS_ENV}"
-          - name: PATH_PREFIX
-            value: "${PATH_PREFIX}"
-          - name: POSTGRESQL_MAX_CONNECTIONS
-            value: ${POSTGRESQL_MAX_CONNECTIONS}
-          - name: RAILS_LOG_TO_STDOUT
-            value: "${RAILS_LOG_TO_STDOUT}"
-          # Should be a secret ?
-          - name: SECRET_KEY_BASE
-            value: "secret_key_base"
+              bundle exec rake --trace db:debug db:migrate
+          inheritEnv: true
         resources:
           limits:
             cpu: ${CPU_LIMIT_SIDE}
@@ -76,7 +54,7 @@ objects:
             memory: ${MEMORY_REQUEST_SIDE}
         env:
         - name: APPLICATION_TYPE
-          value: compliance-import-ssg
+          value: sleep
         - name: RAILS_ENV
           value: "${RAILS_ENV}"
         - name: RAILS_LOG_TO_STDOUT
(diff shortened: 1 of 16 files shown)


clowdapp.yaml
```diff
@@ -166,74 +166,6 @@ objects:
...
-      podSpec:
-        imagePullPolicy: Always
-        image: ${IMAGE}:${IMAGE_TAG}
-        initContainers:
-        - command:
-          - bash
-          - -c
-          - |
-            set -e
-            RAILS_ENV="${RAILS_ENV}" bundle exec rake --trace db:debug db:migrate
```
vkrizan, line 183: Be aware that this was the only migration executor.

akofink:
Ok, I think this is working now. I'm going to test it in the ephemeral environment and mark as ready if all goes well there.

clowdapp.yaml
```diff
@@ -111,9 +111,8 @@ objects:
           - bash
           - -c
           - |
-            for i in {1..120}; do sleep 1;
-            if bundle exec rake --trace db:abort_if_pending_migrations;
-            then exit 0; fi; done; exit 1
+            set -e
+            RAILS_ENV="${RAILS_ENV}" bundle exec rake --trace db:debug db:migrate
```
vkrizan, line 115: We might need to extract this to a single pod just for migrations (like cyndi-init). I am not sure if the concurrency will work if it spins up more than one pod at the same time.

akofink: It seemed to work fine when deploying in the ephemeral environment yesterday. I'm confused by your statement, though:

> I am not sure if the concurrency will work if it spins up more than one pod at the same time.

What concurrency? How does this change the number of pods started simultaneously? It seems the wait_for_migrations check works great to keep other services waiting in their init containers.

vkrizan: The `wait_for_migrations` is run on non-backend pods (consumer, sidekiq...). Here the `db:migrate` is run on any backend pod. If there would be a deployment, there would be at least two pods started. So the question is more on what would happen if `db:migrate` is run simultaneously more than once. Database would guarantee concurrency to some level with transactions, but I'm not sure how it would behave within Rails.  Whether the second pod would reapply the same migration, after a lock is lifted, and fail or not.

config/initializers/prometheus.rb
```diff
@@ -1,18 +1,39 @@
...
+  def ensure_exporter_server
+    require 'socket'
+    TCPSocket.open("127.0.0.1", Settings.metrics_port) {}
+  rescue Errno::ECONNREFUSED
+    require 'prometheus_exporter/server'
+
+    server = PrometheusExporter::Server::WebServer.new(port: Settings.metrics_port)
+    server.start
+
+    if $0.include?('rails') # only one pod to report DB metrics
```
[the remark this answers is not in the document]
akofink: Yes - I was looking into this issue yesterday.

clowdapp.yaml
akofink: Ah! Yes, okay. I see now. Thank you for the explanation :) I guess we can add this add a job and include it in the CJI.

akofink: Ok, updated. Nice catch, btw! That probably wouldn't have worked very well :laughing: Let me know if you see issues with using a job like this.

vkrizan: One-time job with a finite number of retries will do I guess. :+1:

config/initializers/prometheus.rb
```diff
@@ -1,18 +1,39 @@
...
+  def ensure_exporter_server
+    require 'socket'
+    TCPSocket.open("127.0.0.1", Settings.metrics_port) {}
+  rescue Errno::ECONNREFUSED
+    require 'prometheus_exporter/server'
+
+    server = PrometheusExporter::Server::WebServer.new(port: Settings.metrics_port)
+    server.start
+
+    if $0[/rails|puma/] # only one pod to report DB metrics
```
akofink, line 17: This doesn't seem to be working still :thinking:

config/initializers/prometheus.rb
```diff
@@ -1,18 +1,39 @@
...
+  def ensure_exporter_server
+    require 'socket'
+    TCPSocket.open("127.0.0.1", Settings.metrics_port) {}
+  rescue Errno::ECONNREFUSED
+    require 'prometheus_exporter/server'
+
+    server = PrometheusExporter::Server::WebServer.new(port: Settings.metrics_port)
+    server.start
+
+    if $0.include?('puma') # only one pod to report DB metrics
```
akofink, line 17: I tried `rails`, `puma`, and `rack` here without success so far. I'm testing by deploying in the ephemeral environment, waiting for init containers to complete, then opening a terminal in each pod and running `curl 127.0.0.1:9000/metrics`. @vkrizan maybe you could investigate it as well? I tried adding some log statements to see what `$0` is for each pod, but the logs didn't show any output for some reason (I tried `puts` and `Rails.logger.info`). Otherwise, I'll continue digging tomorrow.

clowdapp.yaml
```diff
@@ -42,10 +34,42 @@ objects:
...
+        resources:
+          limits:
+            cpu: ${CPU_LIMIT_SIDE}
+            memory: ${MEMORY_LIMIT_SIDE}
+          requests:
+            cpu: ${CPU_REQUEST_SIDE}
+            memory: ${MEMORY_REQUEST_SIDE}
+        env:
+        - name: APPLICATION_TYPE
+          value: sleep
```
akofink, line 57: This used to run as part of the prometheus_exporter, so part of this PR is to rearchitect the migrations a bit. The clowder team is currently investigating how to run migrations without having $REPLICAS processes running at the same time. For now, I've updated the entrypoint.sh to include a sleep APPLICATION_TYPE that does not ever terminate the pod.

We'll be replacing the import-ssg logic with a CronJob like we discussed - that change is not part of this PR!

entrypoint.sh
```diff
@@ -58,21 +58,8 @@ elif [ "$APPLICATION_TYPE" = "compliance-inventory" ]; then
...
-    PORT="9394"
-  fi
-
-  echo "PORT:$PORT"
-  exec bundle exec prometheus_exporter -b 0.0.0.0 --port "$PORT" --prefix compliance_ -t 50 --verbose -a lib/prometheus/graphql_collector.rb -a lib/prometheus/business_collector.rb
-
-elif [ "$APPLICATION_TYPE" = "compliance-import-remediations" ]; then
-  exec bundle exec rake import_remediations --trace
-elif [ "$APPLICATION_TYPE" = "compliance-import-ssg" ]; then
-  exec bundle exec rake ssg:import_rhel_supported --trace
```
akofink, line 75: Add this back in

clowdapp.yaml
```diff
@@ -55,18 +79,8 @@ objects:
               for i in {1..120}; do sleep 1;
               if bundle exec rake --trace db:abort_if_pending_migrations;
               then exit 0; fi; done; exit 1
```
akofink, line 81: Check the exit logic

config/initializers/prometheus.rb
```diff
@@ -1,18 +1,39 @@
...
-  )
+
+  def ensure_exporter_server
+    require 'socket'
+    TCPSocket.open("127.0.0.1", Settings.metrics_port) {}
+  rescue Errno::ECONNREFUSED
+    require 'prometheus_exporter/server'
+
+    server = PrometheusExporter::Server::WebServer.new(port: Settings.metrics_port)
+    server.start
```
akofink, line 15: Consider adding another rescue in case two processes try to start the server at the same time.

config/initializers/prometheus.rb
```diff
@@ -1,18 +1,39 @@
...
+
+    if $0.include?('puma')
```
akofink, line 17: We'd still get this for each puma/backend pod

akofink:
Tested in stage with non-clowder and clowder deployments. This latest version seems to finally have all the issues worked out.

PR #946, opened 2021-08-10
feat(db): RHICOMPL-1561 make account numbers unique

skateman:
:thinking: not sure how to test this as the migration runs first and will not allow me to create non-unique accounts :thinking: we should reconsider to use rspec for migration tests that would support testing the up/down calls directly

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/db/migrate/20210809071351_deduplicate_accounts.rb
+++ b/db/migrate/20210809071351_deduplicate_accounts.rb
@@ -1,0 +1,12 @@
+class DeduplicateAccounts < ActiveRecord::Migration[5.2]
+  def up
+    DuplicateAccountResolver.run!
+    remove_index :accounts, :account_number
+    add_index :accounts, :account_number, unique: true
+  end
+
+  def down
+    remove_index :accounts, :account_number
+    add_index :accounts, :account_number
+  end
+end
--- a/app/services/duplicate_account_resolver.rb
+++ b/app/services/duplicate_account_resolver.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# A service class to resolve duplicated accounts
+class DuplicateAccountResolver
+  MAIN_ID_COLUMN = Arel.sql('((array_agg("id"))[1]) as main')
+  DUPL_ID_COLUMN = Arel.sql('unnest((array_agg("id"))[2:]) as dupl')
+
+  class << self
+    def run!
+      duplicates = Account.select(MAIN_ID_COLUMN, DUPL_ID_COLUMN)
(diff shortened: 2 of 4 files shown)


db/migrate/20210809071351_deduplicate_accounts.rb
```diff
@@ -0,0 +1,14 @@
+class DeduplicateAccounts < ActiveRecord::Migration[5.2]
+  def up
+    Account.transaction do
+      DuplicateAccountResolver.run!
+      remove_index :accounts, :account_number
+      add_index :accounts, :account_number, unique: true
+    end
```
akofink, line 7: No need for an explicit transaction. From https://edgeguides.rubyonrails.org/active_record_migrations.html#migration-overview:

> On databases that support transactions with statements that change the schema, migrations are wrapped in a transaction. If the database does not support this then when a migration fails the parts of it that succeeded will not be rolled back. You will have to rollback the changes that were made by hand.

akofink:
> thinking not sure how to test this as the migration runs first and will not allow me to create non-unique accounts thinking we should reconsider to use rspec for migration tests that would support testing the up/down calls directly

Wouldn't something like this work? https://github.com/RedHatInsights/compliance-backend/blob/master/test/services/duplicate_benchmark_resolver_test.rb#L12

vkrizan (approved):
LGTM. I recommend to do a thorough testing with some data in some pre-prod env.

vkrizan:
@skateman also please update `schema.rb`. Thanks.

db/migrate/20210809071351_deduplicate_accounts.rb
skateman: Oh, :facepalm: okay, it will :sweat_smile:

akofink (commented):
Thank you for the test! Still a redundant transaction in the migration, but other than that, it looks good. :+1: We should try to set up CI or QA at least with some duplicates before merging.

skateman:
@akofink can you be more specific?

db/migrate/20210809071351_deduplicate_accounts.rb
akofink: @skateman maybe you forgot to push this update?

akofink (approved):
Thanks! :+1:

PR #948, opened 2021-08-11
feat(metrics): RHICOMPL-2055 profile usage by hosts/accounts

skateman:


lib/prometheus/business_collector.rb
```diff
@@ -84,6 +101,50 @@ def collect
     @client_policies.observe(
       Policy.where(account_id: client_accounts.select(:id)).count
     )
+
+    sq = Policy.joins(:profiles).select(:account_id, "profiles.ref_id").distinct
+    Policy.select('sq.ref_id', 'COUNT(sq.account_id)')
+          .from("(#{sq.to_sql}) as sq")
+          .group('sq.ref_id').each do |item|
```
vkrizan, line 108: I guess the subquery in a from clause is to ensure distinct counting. Postgresql & Rails supports `COUNT(DISTINCT ...)`.
```
> Policy.joins(:profiles).select(:account_id, "profiles.ref_id").distinct.group(:ref_id).count(:account_id)
   (4.7ms)  SELECT COUNT(DISTINCT "policies"."account_id") AS count_account_id, "ref_id" AS ref_id FROM "policies" INNER JOIN "profiles" ON "profiles"."policy_id" = "policies"."id" GROUP BY "ref_id"
```

lib/prometheus/business_collector.rb
```diff
@@ -84,6 +101,50 @@ def collect
...
+
+    sq = Policy.where(account_id: client_accounts.pluck(:id))
+               .joins(:policy_hosts)
+               .joins(:profiles)
+               .select('policy_hosts.host_id', 'profiles.ref_id').distinct
+    Policy.select('sq.ref_id', 'COUNT(sq.host_id)')
+          .from("(#{sq.to_sql}) as sq")
+          .group('sq.ref_id').each do |item|
+      @client_policies_by_host.observe(
+        item.attributes['count'], ref_id: item.attributes['ref_id']
```
vkrizan, line 144: I'd also observe/include average (and st. deviation) of hosts per policy type (profile ref_id).

lib/prometheus/business_collector.rb
```diff
@@ -84,6 +101,50 @@ def collect
...
+    sq = Policy.joins(:profiles).select(:account_id, "profiles.ref_id").distinct
+    Policy.select('sq.ref_id', 'COUNT(sq.account_id)')
+          .from("(#{sq.to_sql}) as sq")
+          .group('sq.ref_id').each do |item|
+      @total_policies_by_account.observe(
+        item.attributes['count'], ref_id: item.attributes['ref_id']
+      )
+    end
+
+    sq = Policy.where(account_id: client_accounts.pluck(:id))
```
vkrizan, line 114: Observation: Variable override.
I guess we should start to spit these to separate methods :slightly_smiling_face:.

skateman: I created an issue for that yesterday :wink:

lib/prometheus/business_collector.rb
skateman: That we can do from grafana (I guess)

lib/prometheus/business_collector.rb
```diff
@@ -84,6 +101,37 @@ def collect
...
     )
+
+    Policy.joins(:profiles)
+          .select(:account_id, "profiles.ref_id").distinct
+          .group(:ref_id)
+          .count(:account_id).each do |ref_id, count|
+      @total_policies_by_account.observe(count, ref_id: ref_id)
+    end
+
+    Policy.where(account_id: client_accounts.pluck(:id))
```
vkrizan, line 112: Observation: can omit `pluck` on `client_accounts`.

lib/prometheus/business_collector.rb
```diff
@@ -84,6 +101,37 @@ def collect
...
+          .joins(:profiles)
+          .select(:account_id, "profiles.ref_id").distinct
+          .group(:ref_id)
+          .count(:account_id).each do |ref_id, count|
+      @client_policies_by_account.observe(count, ref_id: ref_id)
+    end
+
+    Policy.joins(:policy_hosts).joins(:profiles)
```
vkrizan, line 120: Just a note, that policy hosts and profiles do not have a direct relationship together. It would join hosts with any of the policy profiles. This works thanks to the distinct counting and the fact that a policy uses exactly one profile ref_id.

vkrizan (approved):
Looks good!

lib/prometheus/business_collector.rb
skateman: I am aware, that's why I was using the subquery-querying :wink: that looked more trivial, but you asked for this :see_no_evil: but I like this more now :laughing:

PR #952, opened 2021-08-16
cleanup(rules): RHICOMPL-2081 remove unused rules and reimport them

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/services/rules_cleanup_service.rb
+++ b/app/services/rules_cleanup_service.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# This class cleans up unused rules and reimports
+class RulesCleanupService
+  class << self
+    def run!
+      # Clean up rules that don't belong to any profile
+      Rule.where.not(id: ProfileRule.select(:rule_id)).delete_all
+
+      # Clean up rules that belong only to canonical profiles
+      delete_rules(select_rules)
+
+      # Reimport all rules if we're not in the testing environment
+      Rake::Task['ssg:import_rhel_supported'].execute unless Rails.env.test?
+    end
+
+    def select_rules
+      profiles = Profile.canonical(false).select(:parent_profile_id).distinct
+      used_rules = ProfileRule.select(:rule_id).where(profile_id: profiles)
+
+      Rule.where.not(id: RuleResult.select(:rule_id))
+          .where.not(id: used_rules).select(:id)
+    end
+
+    def delete_rules(rules)
+      RuleReferencesRule.where(rule_id: rules).delete_all
+      RuleReference.left_outer_joins(:rules).where('rules.id' => nil).delete_all
+      RuleIdentifier.where(rule_id: rules).delete_all
+      rules.delete_all
+    end
+  end
+end
--- a/test/services/rules_cleanup_service_test.rb
+++ b/test/services/rules_cleanup_service_test.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class RulesCleanupServiceTest < ActiveSupport::TestCase
+  setup do
+    @cp = FactoryBot.create(:canonical_profile, :with_rules)
+    acc = FactoryBot.create(:account)
+    @profile = FactoryBot.create(:profile, :with_rules, account: acc)
+
+    r1, r2 = @cp.rules.sample(2)
+    @unused_rr = FactoryBot.create(:rule_reference, rules: [r1, r2])
+    FactoryBot.create(:rule_reference, rules: [@profile.rules.sample, r1])
+    FactoryBot.create(:rule_identifier, rule: r1)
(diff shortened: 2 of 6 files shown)


app/services/rules_cleanup_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# This class cleans up unused rules and reimports
+class RulesCleanupService
+  class << self
+    def run!
+      # Clean up rules that don't belong to any profile
+      Rule.where.not(id: ProfileRule.pluck(:rule_id)).delete_all
```
vkrizan, line 8: Observation: Can use `all` instead of a `pluck` to do a subquery.

test/services/rules_cleanup_service_test.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class RulesCleanupServiceTest < ActiveSupport::TestCase
+  setup do
+    cp = FactoryBot.create(:canonical_profile, :with_rules)
+    acc = FactoryBot.create(:account)
+    FactoryBot.create(:profile, :with_rules, account: acc)
+    @unused_rules = cp.rules
+  end
+
+  test 'delete unused rules' do
```
vkrizan, line 13: Would it be possible to cover more cases with tests please? ... to ensure that it does not delete stuff that it should not. :slightly_smiling_face:

app/services/rules_cleanup_service.rb
```diff
@@ -0,0 +1,32 @@
...
+      profiles = Profile.canonical(false).select(:parent_profile_id).distinct
+      used_rules = ProfileRule.select(:rule_id).where(profile_id: profiles)
+
+      Rule.where.not(id: RuleResult.select(:rule_id))
+          .where.not(id: used_rules).select(:id)
+    end
+
+    def delete_rules(rules)
+      RuleReferencesRule.where(rule_id: rules).delete_all
+      RuleReference.joins(:rules).where('rules.id' => nil).delete_all
```
vkrizan, line 27: When `rules.id` will be `NULL` in the case of an `INNER JOIN`?

test/services/rules_cleanup_service_test.rb
skateman: Do you mean adding `RuleReference` and `RuleIdentifier` records? Or something around rules/profiles that the current one is missing?

app/services/rules_cleanup_service.rb
skateman: I thought Rails sorts this out OOTB, but better safe than sorry :see_no_evil: fixing

app/services/rules_cleanup_service.rb
```diff
@@ -0,0 +1,32 @@
...
+  class << self
+    def run!
+      # Clean up rules that don't belong to any profile
+      Rule.where.not(id: ProfileRule.select(:rule_id)).delete_all
+
+      # Clean up rules that belong only to canonical profiles
+      delete_rules(select_rules)
+
+      # Reimport all rules if we're not in the testing environment
+      Rake::Task['ssg:import_rhel_supported'].execute unless Rails.env.test?
```
skateman, line 14: I'm not so sure about this, we already have a job that imports the SSGs and on a fresh install this would make it obsolete :thinking:

vkrizan: I'd remove it from the migration. This should be a non-impactful cleanup.

test/services/rules_cleanup_service_test.rb
vkrizan: Yes references, identifiers as well as profile rules, canonical profiles, non-canonical profiles, used vs. unused rules. Each `delete_all` has at least two scenarios (one which it deletes and one it should not delete)...

I see that we've added more of these scenarios to the test. Checking counts only might be a limiting factor as it does not fully check instances which are being/should be kept.

skateman: Right, I'll add checks for that

app/services/rules_cleanup_service.rb
skateman: Okay, but somehow we should trigger a reimport in prod :thinking:

app/services/rules_cleanup_service.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def select_rules
+      profiles = Profile.canonical(false).select(:parent_profile_id).distinct
```
vkrizan, line 18: Oh, this is the line that selects canonical profiles that have a child. In other words "used profiles". This is also the reason that requires reimport.

PR #955, opened 2021-08-16
feat(Clowder): Add Remediations Clowder config

Victoremepunto:
Signed-off-by: Victor M <vmugicag@redhat.com>

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -23,8 +23,8 @@ objects:
     - host-inventory
     - ingress
     - rbac
-#  blocked https://issues.redhat.com/browse/RHCLOUD-10212
-#    - remediations
+    optionalDependencies:
+    - remediations
     database:
       name: compliance
       version: 12
@@ -38,8 +38,6 @@ objects:
       - topicName: platform.remediations.events
         partitions: 1
     inMemoryDb: true
-#    optionalDependencies:
-#    - remediations
     cyndi:
       enabled: true
       appName: "compliance"
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -3,10 +3,8 @@ if ClowderCommonRuby::Config.clowder_enabled?
 
   config = ClowderCommonRuby::Config.load
 
-# Blocked by https://issues.redhat.com/browse/RHCLOUD-10212
-#  remediations_hostname = config.private_dependency_endpoints.remediations.service.hostname
-#  remediations_port = config.private_dependency_endpoints.remediations.service.port
-#  remediations_url = "#{remediations_hostname}:#{remediations_port}"
+  remediations_config = config.dependency_endpoints.dig('remediations', 'api')
+  remediations_url = "http://#{remediations_config&.hostname}:#{remediations_config&.port}"
 
   # compliance-ssg
   compliance_ssg_config = config.private_dependency_endpoints.dig('compliance-ssg', 'service')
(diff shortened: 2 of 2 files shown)


akofink (approved):
Thanks, @Victoremepunto!

clowdapp.yaml
```diff
@@ -23,8 +23,7 @@ objects:
     - host-inventory
     - ingress
     - rbac
-#  blocked https://issues.redhat.com/browse/RHCLOUD-10212
-#    - remediations
+    - remediations
```
akofink, line 26: I wonder if we should add this as an optionalDependency. Our import SSG job gracefully fails when the remediations service is not available, notwithstanding the open SSL issue we found yesterday.

Victoremepunto: I agree, updated as an optional dependency :+1:

clowdapp.yaml
```diff
@@ -423,7 +421,7 @@ parameters:
   value: 50m
 - name: REDIS_SSL
   description: 'Whether to use secured connection to Redis. Use string values of true or false'
-  value: "true"
+  value: "false"
```
akofink, line 424: @vkrizan wanted this true by default. We should probably use the resourceTemplate or cicd-common definition to define it as false, like we did here: https://gitlab.cee.redhat.com/insights-platform/cicd-common/-/blob/master/bonfire_configs/ephemeral_apps.yaml#L35.

clowdapp.yaml
```diff
@@ -7,7 +7,7 @@ objects:
 - apiVersion: cloud.redhat.com/v1alpha1
   kind: ClowdJobInvocation
   metadata:
-    name: compliance
+    name: compliance-import-ssg-${RANDOM_SUFFIX}
```
akofink, line 10: This change isn't related to this PR. We're going to be moving this to a CronJob and removing this CJI in RHICOMPL-2145. If you want to add that work to this PR in another commit, I wouldn't mind :) Otherwise, we can merge just the remediations change and create a new PR for RHICOMPL-2145.

Victoremepunto: The CJI is not meant to be used as we're using it now in this template. The way is created, once deployed, you'll never be able to update it ever. A CJI is a one time job run. In order to be run every time, it requires to have a different name, hence the dynamic suffix. However it is true that this changes is maybe outside of this PR, and will be useless when the CJI is deprecated. I'll remove it from the final version.

clowdapp.yaml
Victoremepunto: The default *true* value is the reason why the client doesn't connect. we need it to be *false* as Clowder does not provide SSL routes for any service, that's handled by Istio . I would like to hear @vkrizan opinion here.

vkrizan: Yes please if possible use safe default (`"true"`). Environment where it needs to be turned off can override the env var. For Istio case there would be an exception for Redis' port number.
/cc @lphiri

clowdapp.yaml
vkrizan: Technically there would be an image change on a deployment which would likely trigger the another one-time job, wouldn't it?

Victoremepunto: nope, since the CJI has the same name, there would not be another deployment as that resource is already there. It must have different names.

clowdapp.yaml
Victoremepunto: I'm just saying, in all Clowder templates, we'll have a default `true` value, and we will have to force it to `false` in every place we need it. that sounds ... weird. All clowder-managed environments need this variable set to false is what I mean

vkrizan: That's exactly the point, to disable SSL/security knowingly. We can set the default on ci-cd common configuration.

vkrizan (approved):
Thanks Victor!

PR #968, opened 2021-08-24
cleanup(tests): RHICOMPL-2183 use response.parsed_body everywhere

skateman:
```ruby
JSON.parse(response)
```
vs
```ruby
response.parsed_body
```


akofink (approved):
Thanks!

PR #977, opened 2021-08-27
fix(api): RHICOMPL-2175 display the sortable fields in openapi.json

skateman:
:shrug: not sure about the validity of this spec if the sort_by isn't passed as an array, but I guess it should be fine :thinking:

### DIFF
--- a/swagger/v1/openapi.json
+++ b/swagger/v1/openapi.json
@@ -84,6 +84,16 @@
                   }
                 ]
               },
+              "items": {
+                "enum": [
+                  "title",
+                  "version",
+                  "title:asc",
+                  "title:desc",
+                  "version:asc",
+                  "version:desc"
+                ]
+              },
               "default": ""
             }
           },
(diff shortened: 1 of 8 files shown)


swagger/v1/openapi.json
```diff
@@ -72,6 +72,17 @@
...
+              "enum": [
+                "title",
+                "version",
+                "title:asc",
+                "title:desc",
+                "version:asc",
+                "version:desc"
+              ],
+              "type": "string"
+            },
```
Victoremepunto, line 85: Hi ! this section doesn't seem to belong to the `parameter` level but to the `schema` one, better something like 

```
        - name: sort_by
          in: query
          required: false
          description: >-
            An array of fields with an optional direction (:asc or :desc) to
            sort the results.
          schema:
            items:
              enum:
                - title
                - version
                - title:asc
                - title:desc
                - version:asc
                - version:desc
            type: string
```

Also check `type` should not be specified neither at the same level, but under schema. 

The same revision should be applied to the rest of the items in this PR.

Victoremepunto (commented):
Added a small review.

akofink (approved):
Works great! Nice job, @skateman - this is much more clear.

PR #982, opened 2021-08-30
feat(rules): RHICOMPL-2081 mark upstream rules with a new field

skateman:
This is the mark phase of the non-invasive cleanup of unused upstream rules from our database. I'm initially marking every rule as an upstream and during the SSG import, all the imported rules will set the `upstream` field to be `false`. This way we will know which rules should be kept in a future cleanup.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/db/migrate/20210830070917_add_upstream_for_rules.rb
+++ b/db/migrate/20210830070917_add_upstream_for_rules.rb
@@ -1,0 +1,6 @@
+class AddUpstreamForRules < ActiveRecord::Migration[5.2]
+  def change
+    add_column :rules, :upstream, :boolean, null: false, default: true
+    add_index :rules, :upstream
+  end
+end
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -77,7 +77,8 @@ class Rule < ApplicationRecord
     rule.assign_attributes(title: op_rule.title,
                            description: op_rule.description,
                            rationale: op_rule.rationale,
-                           severity: op_rule.severity)
+                           severity: op_rule.severity,
+                           upstream: false)
 
     rule
   end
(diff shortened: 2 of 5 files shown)


db/migrate/20210830070917_add_upstream_for_rules.rb
```diff
@@ -0,0 +1,5 @@
+class AddUpstreamForRules < ActiveRecord::Migration[5.2]
+  def change
+    add_column :rules, :upstream, :boolean, null: false, default: true
+  end
```
akofink, line 4: An index may be warranted:

```suggestion
    add_column :rules, :upstream, :boolean, null: false, default: true
    add_index :rules, :upstream
  end
```

akofink (approved):
Thanks, @skateman!

NB: None of this code may be run due to this check which compares the number of rules: https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/concerns/xccdf/benchmarks.rb#L26

It uses a strict equality check, which is flawed. If there are more rules in the benchmark saved in our DB (i.e. extra upstream rules) than the datastream XML file, it will also try re-saving any rules.

skateman:
@akofink so what do you propose, should we get rid of this check completely?

akofink:
> what about an ENV variable that we can set to force an import that skips this check?

Yep, my thoughts exactly :+1: No need to mess with env variables directly - just use Settings.* and env var SETTINGS__*.

akofink:
Yes, we'll likely do a deployment with the env var set, then during the next deployment unset it again. It should be pretty straightforward. I will be happy to help.

akofink (approved):
Thanks, @skateman!

PR #983, opened 2021-08-30
feat(profiles): RHICOMPL-2080 mark upstream profiles with a new field

skateman:
Analogous with https://github.com/RedHatInsights/compliance-backend/pull/982, but for profiles ...

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -56,15 +56,14 @@ class Profile < ApplicationRecord
 
   class << self
     def from_openscap_parser(op_profile, benchmark_id: nil, account_id: nil)
-      profile = find_or_initialize_by(
-        ref_id: op_profile.id,
-        benchmark_id: benchmark_id,
-        account_id: account_id
-      )
+      profile = find_or_initialize_by(ref_id: op_profile.id,
+                                      benchmark_id: benchmark_id,
+                                      account_id: account_id)
 
       profile.assign_attributes(
         name: op_profile.title,
-        description: op_profile.description
+        description: op_profile.description,
+        upstream: false
       )
 
       profile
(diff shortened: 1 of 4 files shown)


db/migrate/20210830080830_add_upstream_for_profiles.rb
```diff
@@ -0,0 +1,11 @@
+class AddUpstreamForProfiles < ActiveRecord::Migration[5.2]
+  def up
+    add_column :profiles, :upstream, :boolean
+    # Initially all canonical profiles are marked as upstream
+    Profile.canonical.update_all(upstream: true)
```
akofink, line 5: Why not use a default value? If we update this field on non-canonical profiles, what issues would that cause?

Also: we may want an index on this column.

akofink (approved):
Thanks, @skateman! I have similar concerns about this change as the rules change. We may need to force an import due to the way the code tries to exit early here: https://github.com/RedHatInsights/compliance-backend/blob/master/app/services/concerns/xccdf/benchmarks.rb#L26

db/migrate/20210830080830_add_upstream_for_profiles.rb
skateman: It would cause semantical issues :see_no_evil: my eyes would not like it :confused:

app/models/profile.rb
```diff
@@ -56,15 +56,14 @@ class Profile < ApplicationRecord
...
   class << self
     def from_openscap_parser(op_profile, benchmark_id: nil, account_id: nil)
-      profile = find_or_initialize_by(
-        ref_id: op_profile.id,
-        benchmark_id: benchmark_id,
-        account_id: account_id
-      )
+      profile = find_or_initialize_by(ref_id: op_profile.id,
+                                      benchmark_id: benchmark_id,
+                                      account_id: account_id)
```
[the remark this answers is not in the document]
skateman: Yup :see_no_evil:

PR #984, opened 2021-08-30
fix(Accounts): RHICOMPL-2114 set is_internal from the identity header

skateman:
I moved around the `account_from_header` method and generalized it into `Account.from_identity_header`. The method now tries to look up for both the account number and the `user.is_internal` boolean and if the latter one isn't set, an exception happens, the handler looks for the account by the number only, updates the `is_internal` field and retries. This way any user interaction will keep the information about the account being internal or not.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -13,6 +13,21 @@ class Account < ApplicationRecord
 
   validates :account_number, presence: true
 
+  class << self
+    def from_identity_header(identity_header)
+      account = Account.find_or_create_by(
+        account_number: identity_header.identity['account_number']
+      )
+
+      # Update the is_internal if set and differs from the stored one
+      unless identity_header.is_internal.nil?
+        account.update!(is_internal: identity_header.is_internal)
+      end
+
+      account
+    end
+  end
+
   # rubocop:disable Metrics/MethodLength
   def fake_identity_header
     {
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -36,7 +36,7 @@ class XccdfReportParser
     end
 
     @b64_identity = message['b64_identity']
-    @account = Account.find_or_create_by(account_number: message['account'])
+    @account = Account.from_identity_header(IdentityHeader.new(@b64_identity))
     @host = Host.find(message['id'])
     @test_result_file = OpenscapParser::TestResultFile.new(report_contents)
     set_openscap_parser_data
--- a/test/controllers/concerns/authentication_test.rb
+++ b/test/controllers/concerns/authentication_test.rb
@@ -125,7 +125,10 @@ class AuthenticationTest < ActionController::TestCase
         {
           'identity': {
             'account_number': '1234',
-            'user': { 'username': 'username' }
+            'user': {
+              'username': 'username',
+              'is_internal': true
+            }
           },
           'entitlements':
           {
@@ -143,6 +146,7 @@ class AuthenticationTest < ActionController::TestCase
         Account.find_by(account_number: '1234').account_number,
         '1234'
       )
+      assert Account.find_by(account_number: '1234').is_internal
       assert_not User.current, 'current user must be reset after request'
     end
 
@@ -170,6 +174,35 @@ class AuthenticationTest < ActionController::TestCase
         '1234'
       )
       assert_not User.current, 'current user must be reset after request'
+    end
+
+    should 'account found, is_internal unset, updates it' do
+      FactoryBot.create(:account, account_number: '1234')
+      assert_nil Account.find_by(account_number: '1234').is_internal
+
+      encoded_header = Base64.encode64(
+        {
+          'identity':
+          {
+            'account_number': '1234',
+            'user': {
+              'username': 'username',
+              'is_internal': false
+            }
+          },
+          'entitlements':
+          {
+            'insights': {
+              'is_entitled': true
+            }
+          }
+        }.to_json
+      )
+
+      process_test(headers: { 'X-RH-IDENTITY': encoded_header })
+      assert_response :success
+      assert_equal '1234', response.body
+      assert_equal false, Account.find_by(account_number: '1234').is_internal
     end
 
     should 'user not found, creates a new account, username missing' do
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -52,7 +52,7 @@ module Authentication
   end
 
   def user
-    User.new(account: account_from_header)
+    User.new(account: Account.from_identity_header(identity_header))
   end
 
   def rbac_allowed?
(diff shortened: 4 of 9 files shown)


app/models/account.rb
```diff
@@ -13,6 +13,26 @@ class Account < ApplicationRecord
...
+    end
+
+    private
+
+    def update_is_internal!(identity_header)
+      Account.find_or_create_by(
+        account_number: identity_header.identity['account_number']
+      ).update!(is_internal: identity_header.identity['user']['is_internal'])
+    end
+  end
```
akofink, line 34: I think the steps should be:

1. Find the account by account_id alone
2. Update is_internal to whatever is in the header, if needed (AR handles this condition for you)

```rb
account = Account.find_by(account_number: identity_header.identity['account_number'])
account.update!(is_internal: identity_header.identity['user']['is_internal'])
```

test/controllers/concerns/authentication_test.rb
```diff
@@ -172,6 +176,35 @@ def raising
       assert_not User.current, 'current user must be reset after request'
     end
 
+    should 'account found, is_internal unset, updates it' do
```
akofink, line 179: A slight variant, which I expect to hit in prod: 'account found, is_internal set, updates it'

akofink (commented):
Thanks, @skateman! I'm not a huge fan of the rescue/retry logic. It seems we can just rely on Active Record here instead.

test/controllers/concerns/authentication_test.rb
skateman: Do we really have that as a use-case in practice? If the `is_internal` is set, it means that we already know if the account is internal or not and that AFAIK can't change in the future :thinking:

app/models/account.rb
skateman: Yeah, I was considering the performance here, why should we do an insert on any user interaction? :thinking:

test/controllers/concerns/authentication_test.rb
akofink: Did you run a query on accounts in prod to see what the state of this existing column is? How are you sure they are set properly there?

app/models/account.rb
akofink: The snippet provided does not actually update unless the is_internal field is changed. That's what I mean by 'AR handles this condition for you'. You can try it out locally to confirm.

test/controllers/concerns/authentication_test.rb
skateman: Yes, they're all set to `nil`

app/models/account.rb
```diff
@@ -13,6 +13,18 @@ class Account < ApplicationRecord
 
   validates :account_number, presence: true
 
+  class << self
+    def from_identity_header(identity_header)
+      account = Account.find_or_create_by(
+        account_number: identity_header.identity['account_number']
```
akofink, line 19: Nit: We really ought to extend our identity header validation to include checking for account_number. It could be nil here.

app/models/account.rb
```diff
@@ -13,6 +13,18 @@ class Account < ApplicationRecord
...
+      )
+      account.update!(
+        is_internal: identity_header.identity['user']['is_internal']
+      )
```
akofink, line 23: Refactoring idea: define an `is_internal` method in the IdentityHeader class, and be sure to `dig` it to handle `nil` user values (i.e. cert auth). Your test changes adding in this field signaled to me that this was an issue, otherwise I wouldn't have realized it. :laughing:

test/controllers/concerns/authentication_test.rb
```diff
@@ -172,6 +176,35 @@ def raising
...
+              'is_entitled': true
+            }
+          }
+        }.to_json
+      )
+
+      process_test(headers: { 'X-RH-IDENTITY': encoded_header })
+      assert_response :success
+      assert_equal '1234', response.body
+      assert_equal false, Account.find_by(account_number: '1234').is_internal
```
akofink, line 205: I like to use `refute`, but rubocop doesn't like it:

```suggestion
      assert_not Account.find_by(account_number: '1234').is_internal
```

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -11,7 +11,10 @@ class ProfilesAuthenticationTest < ActionDispatch::IntegrationTest
           {
             'identity': {
               'account_number': '1234',
-              'auth_type': IdentityHeader::CERT_AUTH
+              'auth_type': IdentityHeader::CERT_AUTH,
+              'user': {
+                'is_internal': false
+              }
```
akofink, line 17: Cert auth doesn't have a 'user' key! These tests must pass without these changes.

akofink (changes requested):
Unfortunately, this breaks cert auth as written. It should be an easy fix to include safe navigation.

app/models/account.rb
skateman: I don't really understand :confused: do you mean [this case](https://github.com/RedHatInsights/compliance-backend/pull/984/files#diff-11ba21e2704420a5b829d0b858daa1c977614447eb9ef38f9d792e04e677e62dL39)?

test/controllers/concerns/authentication_test.rb
skateman: `assert_not` also accepts `nil` which we don't want

app/models/account.rb
akofink: I mean to validate account_number within this validation method: https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/identity_header.rb#L18

Not a part of this PR, just pointing out that there isn't any guarantee `account_number` is non-nil here.

test/controllers/v1/profiles_controller_test.rb
```diff
@@ -135,7 +135,10 @@ class ProfilesAuthenticationTest < ActionDispatch::IntegrationTest
           {
             'identity': {
               'account_number': account.account_number,
-              'auth_type': IdentityHeader::CERT_AUTH
+              'auth_type': IdentityHeader::CERT_AUTH,
+              'user': {
+                'is_internal': false
+              }
```
akofink, line 141: Cert auth headers do not contain a 'user' within their 'identity'. They instead contain a 'system'; it will not include an 'is_internal' field. I pasted a link to a google doc in this jira that further explains the identity.

akofink (commented):
I think we're good now! Just one test related to cert auth that should be reverted (comment inline).

akofink (approved):
Thanks, @skateman!

app/services/xccdf_report_parser.rb
```diff
@@ -36,7 +36,7 @@ def initialize(report_contents, message)
     end
 
     @b64_identity = message['b64_identity']
-    @account = Account.find_or_create_by(account_number: message['account'])
+    @account = Account.from_identity_header(IdentityHeader.new(@b64_identity))
```
akofink, line 39: Here, `@b64_identity` is always `nil`. It's never been used in this parser before this change, so we never detected it as an issue.

PR #986, opened 2021-08-31
chore(db): RHICOMPL-2194 remove dangling accounts

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/services/dangling_account_remover.rb
+++ b/app/services/dangling_account_remover.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# Service for removing dangling accounts
+class DanglingAccountRemover
+  class << self
+    def run!
+      Account.connection.execute(deleter_arel.to_sql)
+      User.delete_all
+    end
+
+    private
+
+    def deleter_arel
+      deleter = Arel::DeleteManager.new
+      deleter.from(Account.arel_table).where(
+        Account.arel_table[:id].not_in(
+          accounts_with_profiles_or_policies
+        ).and(
+          Account.arel_table[:account_number].not_in(accounts_with_hosts)
+        )
+      )
+      deleter
+    end
+
+    def accounts_with_profiles_or_policies
+      Arel::Nodes::Union.new(
+        Policy.arel_table.project(:account_id).distinct,
+        Profile.arel_table.project(:account_id).distinct.where(
+          Profile.arel_table[:parent_profile_id].not_eq(nil)
+        )
+      )
+    end
+
+    def accounts_with_hosts
+      Host.with_policies_or_test_results.select(:account).distinct.arel
+    end
+  end
+end
--- a/test/services/dangling_account_remover_test.rb
+++ b/test/services/dangling_account_remover_test.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class DanglingAccountRemoverTest < ActiveSupport::TestCase
+  test 'removes accounts that have no relationships' do
+    FactoryBot.create_list(:account, 30)
+
+    assert_difference('Account.count' => -30) do
+      DanglingAccountRemover.run!
+    end
+  end
+
+  test 'keeps accounts that have relationships' do
+    accounts = FactoryBot.create_list(:account, 5)
+    FactoryBot.create(:user, account: accounts[4])
+    host1 = FactoryBot.create(:host, account: accounts[0].account_number)
+    host2 = FactoryBot.create(:host, account: accounts[1].account_number)
+
+    profile1 = FactoryBot.create(:profile, account: accounts[0])
+    profile1.policy.update(hosts: [host1])
+
+    profile2 = FactoryBot.create(:profile, account: accounts[1])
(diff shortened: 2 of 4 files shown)


app/services/dangling_account_remover.rb
```diff
@@ -0,0 +1,13 @@
...
+
+# Service for removing dangling accounts
+class DanglingAccountRemover
+  class << self
+    def run!
+      Account.where.not(
+        account_number: Host.with_policies_or_test_results
+                            .select(:account).distinct
+      ).delete_all
+    end
```
akofink, line 11: I may have misled you to point you to the dangling accounts metric. No compliance hosts is only one thing we should check. We definitely also want to check if the account has any policies, even if the policies don't have hosts. There is also no foreign key relationship between profiles and accounts, so this could leave some profiles pointing to nonexistent accounts. Finally, there may be some users associated to this account - those should probably just be removed along with the account if all the above conditions hold.

test/services/dangling_account_remover_test.rb
```diff
@@ -0,0 +1,32 @@
...
+class DanglingAccountRemoverTest < ActiveSupport::TestCase
+  test 'removes accounts that have no relationships' do
+    FactoryBot.create_list(:account, 30)
+
+    assert_difference('Account.count' => -30) do
+      DanglingAccountRemover.run!
+    end
+  end
+
+  test 'keeps accounts that have hosts with test results or policies' do
```
akofink, line 14: Variants I expected in prod:

- keeps accounts that have policies but no hosts
- keeps accounts that have profiles but no policies or hosts
- removes users along with removed accounts

akofink (commented):
Thanks, @skateman!

test/services/dangling_account_remover_test.rb
```diff
@@ -0,0 +1,38 @@
...
+class DanglingAccountRemoverTest < ActiveSupport::TestCase
+  test 'removes accounts that have no relationships' do
+    FactoryBot.create_list(:account, 30)
+
+    assert_difference('Account.count' => -30) do
+      DanglingAccountRemover.run!
+    end
+  end
+
+  test 'keeps accounts that have relationships' do
```
vkrizan, line 14: The `DanglingAccountRemover` checks relationships that exist today. If it is going to be used more than once (e.g. in a cron job), then we should ensure that no new relationship is overlooked. Can there be maybe a test with auto-discovery of new Account relationships to prevent that?

app/services/dangling_account_remover.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+# Service for removing dangling accounts
+class DanglingAccountRemover
+  class << self
+    def run!
+      User.delete_all
```
vkrizan, line 7: This should delete only the relevant User records. If the User table is discontinued, it should be removed first.

vkrizan (approved):
SQL looks sane. I recommend another pair of eyes.

```sql
DELETE
FROM "accounts"
WHERE "accounts"."id" NOT IN ((
                                 (SELECT DISTINCT account_id
                                  FROM "policies")
                               UNION
                                 (SELECT DISTINCT account_id
                                  FROM "profiles"
                                  WHERE "profiles"."parent_profile_id" IS NOT NULL)))
  AND "accounts"."account_number" NOT IN
    (SELECT DISTINCT "inventory"."hosts"."account"
     FROM "inventory"."hosts"
     WHERE ("inventory"."hosts"."id" IN
              (SELECT "policy_hosts"."host_id"
               FROM "policy_hosts")
            OR "inventory"."hosts"."id" IN
              (SELECT "test_results"."host_id"
               FROM "test_results")))
```

app/services/dangling_account_remover.rb
skateman: We agreed with @akofink that we should delete all of these as we no longer use the records...

akofink: True, I don't see why we shouldn't just remove the table altogether.

akofink: On second thought, removing the table (and associated model) would cause issues with the way we store User.current.

skateman: The model can stay, we can make the fields accessible by `attr_accessor` and get rid of the table :thinking: but maybe in a separate PR/issue.

test/services/dangling_account_remover_test.rb
skateman: PG will just fail if we have the foreign keys :thinking: at least it did for me when I tried to do anything ugly

app/services/dangling_account_remover.rb
akofink: Some thoughts about users, not for this PR:

- Account.current in favor of `User.current = User.new(account: a)` seems to fit better with how we're using users/accounts
- Pundit, used for authorization, requires a user (not an account)
- We might need users again when we implement fine grained RBAC, but maybe not

PR #988, opened 2021-09-02
fix(sort): RHICOMPL-2179 overridable default sort of business objectives

skateman:
This is some legacy stuff that doesn't allow to override the default sort in business objectives. Replacing it with *The Right Way&trade;* :wink:


akofink (approved):
Thanks!

PR #989, opened 2021-09-02
fix(metrics): RHICOMPL-2114 adjust metrics to non-internal accounts

skateman:
If the account has the `is_internal` not set to true, i.e. `nil` or `false` it is treated as a client (non-employee) account.

### DIFF
--- a/lib/prometheus/business_collector.rb
+++ b/lib/prometheus/business_collector.rb
@@ -88,9 +88,7 @@ class BusinessCollector < PrometheusExporter::Server::TypeCollector
 
   def collect
     @total_accounts.observe Account.count
-    client_accounts = Account.distinct.where(
-      id: User.select(:account_id).where.not('email LIKE ?', '%redhat%')
-    )
+    client_accounts = Account.where(is_internal: [nil, false])
     @client_accounts.observe client_accounts.count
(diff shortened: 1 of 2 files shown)


lib/prometheus/business_collector.rb
```diff
@@ -88,9 +88,7 @@ def type
 
   def collect
     @total_accounts.observe Account.count
-    client_accounts = Account.distinct.where(
-      id: User.select(:account_id).where.not('email LIKE ?', '%redhat%')
-    )
+    client_accounts = Account.distinct.where.not(is_internal: true)
```
vkrizan, line 91: Accounts with NULL/nil are not returned. NULL is default and set on existing accounts.

```
> Account.distinct.where.not(is_internal: true)
  Account Load (1.6ms)  SELECT DISTINCT "accounts".* FROM "accounts" WHERE "accounts"."is_internal" != $1  [["is_internal", true]]
=> []

> Account.where(is_internal: nil).count
   (1.5ms)  SELECT COUNT(*) FROM "accounts" WHERE "accounts"."is_internal" IS NULL
=> 4
```

Distinct might be redundant since it is now unique by account_number and it does not matter when used in the subqueries.

PR #994, opened 2021-09-03
feat(Clowder) RHICOMPL-1985 Fix smoke tests

Victoremepunto:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -70,6 +70,8 @@ objects:
         # Should be a secret ?
         - name: SECRET_KEY_BASE
           value: "secret_key_base"
+        - name: SETTINGS__REPORT_DOWNLOAD_SSL_ONLY
+          value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
     - name: compliance-import-ssg
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
@@ -106,6 +108,8 @@ objects:
           value: "secret_key_base"
         - name: SETTINGS__FORCE_IMPORT_SSGS
           value: "${SETTINGS__FORCE_IMPORT_SSGS}"
+        - name: SETTINGS__REPORT_DOWNLOAD_SSL_ONLY
+          value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
     - name: compliance-backend
       # TODO: Check how to auto-scale with Clowder
       minReplicas: ${{REPLICAS_BACKEND}}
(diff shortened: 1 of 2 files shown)


akofink (approved):
Bravo! :) ACK pending a green build

Victoremepunto:
> Bravo! :) ACK pending a green build

Oh my ! almost a golden run ! 
> = 99 passed, 1 skipped, 193 deselected, 5 xfailed, 4 warnings, 1 error in 303.58s (0:05:03) =

akofink:
I'm a bit worried about blocking PRs on these tests if the deployment beforehand is not consistent. It seems like we should only fail if the smoke tests were able to run after a successful deployment, _and_ the results show failing tests. I wonder if we could check the status of other apps (i.e. inventory) as a test and only fail the build if that check passes but our compliance tests fail, that way we wouldn't block PRs on failures within other services. Thoughts?

vkrizan:
@Victoremepunto @akofink is it possible to just report the result of the smoke tests within a PR, if we want them to be non-blocking. Or any failed tests in general.

Victoremepunto:
> @Victoremepunto @akofink is it possible to just report the result of the smoke tests within a PR, if we want them to be non-blocking? Or any failed tests in general.

Not sure what is the question here - You mean, not to fail the build if the tests didn't pass ? 

We can do virtually _anything_ , what we should do is to agree on what to do. I'm not confident enough on the stability of the `pr_check` script with the smoke tests included to make it a condition to block a PR merge .

That said, running the unit tests on every PR should be a "must" and should definitely block the PR merge IMO - we can always force merge but these occasions shouldn't be the norm

vkrizan:
> Not sure what is the question here - You mean, not to fail the build if the tests didn't pass ?

Yes, not fail on smoke fail but report (e.g. as a comment). Like a warning. I'm not sure what the capabilities of `pr_check`.

Victoremepunto:
> > Not sure what is the question here - You mean, not to fail the build if the tests didn't pass ?
> 
> Yes, not fail on smoke fail but report (e.g. as a comment). Like a warning. I'm not sure what the capabilities of `pr_check`.

yeah - like I said, we can do virtually anything. Think of the pipeline status as a success / no-success boolean result . we can make force the pipeline to "success" as long as it reaches the unit tests stage successfully. 

The artifact collection would also happen regardless. 

Problem with that approach is , you can basically start ignoring all results as soon as they are "green" every time. 

We should *decide* what to do, which stages from the pr_check need to run always, and which should run "when we decide to".

akofink:
Looks like codecov failed to upload its report and caused the failure? The smoke tests didn't even try to run :disappointed:

akofink:
So far, we have 1 out of 3 builds passing :disappointed:

akofink:
Looks like the same IQE test failed (iqe_compliance.tests.test_basic.test_validate_compliance_report). We determined that IQE simply needs to wait until cyndi syndicates the host to fix this issue; right now, the host exists in inventory but does not yet exist in Compliance.

PR #996, opened 2021-09-03
chore(clowder): Bump up CPU to 800m max for sidekiq/ssg pods

akofink:
This affects sidekiq and import-ssg.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -359,9 +359,9 @@ parameters:
 - name: MEMORY_REQUEST_SIDE
   value: 500Mi
 - name: CPU_LIMIT_SIDE
-  value: 500m
+  value: 800m
 - name: CPU_REQUEST_SIDE
-  value: 50m
+  value: 80m
 - name: REDIS_SSL
   description: 'Whether to use secured connection to Redis. Use string values of true or false'
   value: "true"


Victoremepunto (changes requested):
please increase the CPU request up to 400m

Victoremepunto:
> please increase the CPU request up to 400m

So there's not such a big difference between request / limit

akofink:
> please increase the CPU request up to 400m

Why? The limit is supposed to be 10 times the request. For memory, limit should be 2 times the request.

Victoremepunto:
> > please increase the CPU request up to 400m
> 
> Why? The limit is supposed to be 10 times the request. For memory, limit should be 2 times the request.

No, I think there might be  a misunderstanding here :) - the maximum ratio is 10 , that doesn't mean it has to be 10... 

I'm thinking of this situation: the controller needs to bring up your pod, and since it just requires 80m cpu, it allocates it on a node that has 200m free. so it "fits". then if it needs to bump the CPU allocation,  it will have to evict the pod and spawn it on a different node. 

The bigger the distance between both request and limit is, the higher is the risk to fall into this situation. it's all going to be handled by Openshift, but there's no need to request under 100m if we're likely going to need CPU several orders of magnitude greater than what we just requested.

Victoremepunto (approved):
let's just leave it as it is

PR #997, opened 2021-09-07
perf(reports): RHICOMPL-2210 speed up filtering hosts by compliant

skateman:
Relying on the `RequestStore` once again for narrowing down the search

### DIFF
--- a/app/models/concerns/host_searching.rb
+++ b/app/models/concerns/host_searching.rb
@@ -98,18 +98,10 @@ module HostSearching
     end
 
     def filter_by_compliance(_filter, operator, value)
-      ids = ::Host.includes(test_results: :profile).select do |host|
-        host.compliant.values.all?(::ActiveModel::Type::Boolean.new.cast(value))
-      end
-      ids = ids.pluck(:id).map { |id| "'#{id}'" }
-
-      if ids.empty?
-        return { conditions: '1=0' } if operator == '='
-        return { conditions: '1=1' } if operator == '<>'
-      end
+      hosts = compliant_host_ids(value)
 
       operator = operator == '<>' ? 'NOT' : ''
-      { conditions: "hosts.id #{operator} IN(#{ids.join(',')})" }
+      { conditions: "hosts.id #{operator} IN(#{hosts.to_sql})" }
     end
 
     def filter_by_compliance_score(_filter, operator, score)
@@ -138,6 +130,25 @@ module HostSearching
     end
 
     private
+
+    def compliant_host_ids(value)
+      profiles = RequestStore.store['scoped_search_context_profiles']
+
+      TestResult.latest
+                .where(profile: profiles || User.current.account.profiles)
+                .joins(profile: :policy)
+                .group(:host_id, 'policies.id', 'policies.compliance_threshold')
+                .select(:host_id).having('AVG(test_results.score) <= 100')
+                .having("
+                  AVG(test_results.score)
+                  #{to_b(value) ? '>=' : '<'}
+                  policies.compliance_threshold
+                ")
+    end
+
+    def to_b(str)
+      ::ActiveModel::Type::Boolean.new.cast(str)
+    end
 
     def with_policy_lookup(policy_or_profile_id)
       policy_cond = { policies: { id: policy_or_profile_id } }
--- a/test/models/host_test.rb
+++ b/test/models/host_test.rb
@@ -162,6 +162,8 @@ class HostTest < ActiveSupport::TestCase
     end
 
     should 'be able to find based on compliance score' do
+      RequestStore.clear!
+
       assert_includes(
         Host.search_for("
           with_results_for_policy_id = #{@profile.policy_id} and (
@@ -188,6 +190,8 @@ class HostTest < ActiveSupport::TestCase
     end
 
     should 'be able to find based on compliance true/false' do
+      User.current = FactoryBot.create(:user, account: @account)
+      RequestStore.clear!
       assert_not_includes Host.search_for('compliant = true'), @host1
       assert_includes Host.search_for('compliant = false'), @host1
       assert_includes Host.search_for('compliant != true'), @host1


akofink (approved):
Thanks! Seems to continue to work well. I didn't benchmark this, but it's a fairly obvious performance win :+1:

app/models/concerns/host_searching.rb
```diff
@@ -139,6 +136,25 @@ def test_results?(_filter, _operator, value)
...
     private
 
+    def compliant_host_ids(value)
+      profiles = RequestStore.store['scoped_search_context_profiles']
+
+      TestResult.latest
+                .where(profile: profiles || Profile.all)
+                .joins(profile: :policy)
+                .group(:host_id, 'policies.id', 'policies.compliance_threshold')
+                .select(:host_id).having('AVG(test_results.score) <= 100')
```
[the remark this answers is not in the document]
skateman: `NaN`

app/models/concerns/host_searching.rb
```diff
@@ -98,10 +98,7 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
     end
 
     def filter_by_compliance(_filter, operator, value)
-      ids = ::Host.includes(test_results: :profile).select do |host|
-        host.compliant.values.all?(::ActiveModel::Type::Boolean.new.cast(value))
-      end
-      ids = ids.pluck(:id).map { |id| "'#{id}'" }
+      ids = compliant_host_ids(value).pluck(:host_id).map { |id| "'#{id}'" }
```
vkrizan, line 101: We can drop pluck and map and use a subselect to speed it up even further.

app/models/concerns/host_searching.rb
```diff
@@ -139,6 +136,25 @@ def test_results?(_filter, _operator, value)
 
     private
 
+    def compliant_host_ids(value)
+      profiles = RequestStore.store['scoped_search_context_profiles']
+
+      TestResult.latest
+                .where(profile: profiles || Profile.all)
```
vkrizan, line 143: How is this going to work without profiles context using all existing profiles, hence all hosts from all test results? The query would likely return a whole set that is in the database, wouldn't it? Considering that the ids are mapped into a string that is passed into the WHERE clause this might be a problem.

skateman: This is a bit more delicate, we're using these IDs to reduce an already reduced (pundit-scoped) set of hosts. If you look at the other custom search methods, I'm just following a pattern. These queries are usually context free (except of the ones I introduced for some special cases) and they're only used to append an `AND WHERE` to the already running query initiated by the incoming API/GQL request.

app/models/concerns/host_searching.rb
skateman: How can we do that? Considering what I wrote below :thinking:

app/models/concerns/host_searching.rb
vkrizan: This picks up TestResults out of requested or all profiles. Profiles should be scope limited here, not just hosts. Other custom search methods in this concern use limited set of profiles. `Profile.all` might be the problematic part that I'm concerned about.

Here is resulting query when using REST search:
```
SELECT "test_results"."host_id"
FROM   "test_results"
       INNER JOIN "profiles"
               ON "profiles"."id" = "test_results"."profile_id"
       INNER JOIN "policies"
               ON "policies"."id" = "profiles"."policy_id"
       JOIN (SELECT "test_results"."profile_id",
                    "test_results"."host_id",
                    Max(end_time) AS end_time
             FROM   "test_results"
             GROUP  BY "test_results"."profile_id",
                       "test_results"."host_id") AS tr
         ON test_results.profile_id = tr.profile_id
            AND test_results.host_id = tr.host_id
            AND test_results.end_time = tr.end_time
WHERE  "test_results"."profile_id" IN (SELECT "profiles"."id"
                                       FROM   "profiles")
GROUP  BY "test_results"."host_id",
          policies.id,
          policies.compliance_threshold
HAVING ( Avg(test_results.score) <= 100 )
       AND ( Avg(test_results.score) < policies.compliance_threshold )
```
I'm failing to see the part that limits the scope.

app/models/concerns/host_searching.rb
vkrizan:
```
operator = operator == '<>' ? 'NOT' : ''
{ conditions: "hosts.id #{operator} IN (#{compliant_host_ids(value).to_sql})" }
```
would that work?

skateman: Umm, we're not building SQL query here, the conditions are in the query language of the scoped search. I don't really want to stretch it too far :see_no_evil:

app/models/concerns/host_searching.rb
skateman: The query you described above is a helper query to return all the host IDs that are compliant. In the context of the `filter_by_compliance` method we have no contextual information (we actually dom since my hack with the `RequestStore` but that's not always there, so let's just ignore that for a moment). So the search query returns us all the host IDs that are compliant which then we use in the main query that the scoped search builds based on what this `filter_by_compliance` method returns:

```
SELECT *** FROM hosts WHERE account_id = ? AND host_id IN (scooped_search_method_kicks_in_here)
```

app/models/concerns/host_searching.rb
akofink: See other searches in this same file. The 'query language' used by scoped search is essentially ActiveRecord. Anything put in `conditions` here will be added to the `where` clause. Using a sub-query is definitely preferred than listing all the IDs because the `IN (...)` SQL syntax has a maximum length that we could run up against if the user has many compliance hosts.

app/models/concerns/host_searching.rb
vkrizan: > So the search query returns us all the host IDs that are compliant ...

The fact that it returns all ids and formats them as raw data to a WHERE clause is the problematic part. There can be thousands (and even more) systems. Passing it as as a subquery as mentioned [in the comment above](https://github.com/RedHatInsights/compliance-backend/pull/997#discussion_r705090927) could mitigate it. Considering that the hosts are coming from a separate database it also might become an issue when the number of compliant hosts would greatly increase.

skateman: @vkrizan I'm definitely going to make it a subquery, I thought there's some additional issue here :see_no_evil:

vkrizan (approved):
LGTM!

PR #1001, opened 2021-09-08
fix(Clowder): fix Clowder database config 

Victoremepunto:
- Sets correctly SSL mode and path to RDS CA cert
- Small refactor of Clowder template

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -295,25 +295,23 @@ objects:
           value: "secret_key_base"
 
 parameters:
-- name: MIN_REPLICAS
-  value: '1'
-- description: Image tag
-  name: IMAGE_TAG
+- name: IMAGE_TAG
+  description: Image tag
   required: true
   value: latest
-- description: Image name
-  name: IMAGE
+- name: IMAGE
+  description: Image name
   value: quay.io/cloudservices/compliance-backend
-- description: ClowdEnv Name
-  name: ENV_NAME
-- description: Replica count for backend service
-  name: REPLICAS_BACKEND
+- name: ENV_NAME
+  description: ClowdEnv Name
+- name: REPLICAS_BACKEND
+  description: Replica count for backend service
   value: "1"
-- description: Replica count for consumer
-  name: REPLICAS_CONSUMER
+- name: REPLICAS_CONSUMER
+  description: Replica count for consumer
   value: "1"
-- description: Replica count for sidekiq
-  name: REPLICAS_SIDEKIQ
+- name: REPLICAS_SIDEKIQ
+  description: Replica count for sidekiq
   value: "1"
 - name: RAILS_ENV
   required: true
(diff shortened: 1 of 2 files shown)


akofink (approved):
Seems reasonable. We'll have to test in stage, unfortunately.

config/database.yml
```diff
@@ -4,7 +4,10 @@
     if database_config.rdsCa
       rdsCaFile = File.new('tmp/rdsCa', 'wt')
       rdsCaFile.write(database_config.rdsCa)
-      database_config.rdsCaPath = File.expand_path(rdsCaFile.path)
+      database_config.sslrootcert = File.expand_path(rdsCaFile.path)
+      database_config.sslmode = 'verify-full'
+    else
+      database_config.sslmode = 'disable'
```
vkrizan, line 10: Please default to `prefer`. Even `require` could work on standard envs (dev, CI).

Victoremepunto: well I disabled it because on Clowder deployments on ephemeral environments it won't provide SSL at all. And `prefer` is not recommended anyways. 

> The default value for sslmode is prefer. As is shown in the table, this makes no sense from a security point of view, and it only promises performance overhead if possible. It is only provided as the default for backward compatibility, and is not recommended in secure deployments.

But ... yeah- maybe `prefer` is better than `disable` :)

config/database.yml
```diff
@@ -34,6 +38,8 @@ default: &default
   host: <%= host %>
   database: <%= database %>
   port: <%= port %>
+  sslmode: <%= sslmode || "disable" %>
```
vkrizan, line 41: Can we use the same default of `prefer` also here please? :slightly_smiling_face:

Victoremepunto: changed!

config/database.yml
```diff
@@ -13,8 +13,8 @@
     host = Settings.database.hostname
     database = Settings.database.name
     port = Settings.database.port
-    sslrootcert = Settings.database.rdsCaPath
-
+    sslrootcert = Settings.database.sslrootcert
+    sslmode = Settings.database.sslMode
```
Victoremepunto, line 17: Clowder provides the sslMode now so no need to set it up

akofink (approved):
Thanks! We might add some safe navigation in the future if it is problematic, like if the database_config is nil for some reason.

clowdapp.yaml
```diff
@@ -295,25 +295,23 @@ objects:
           value: "secret_key_base"
 
 parameters:
-- name: MIN_REPLICAS
```
vkrizan, line 298: `MIN_REPLICAS` is being removed. Is that correct?

Victoremepunto: yeah that's not used anywhere, it was probably a leftover, we're using per-service number-of-replicas variables, like `REPLICAS_BACKEND`

vkrizan (approved):
Thank you Victor!

PR #1007, opened 2021-09-14
feat(ssg_import): RHICOMPL-2148 correct profile->rule relationships

skateman:
If I am understanding the issue correctly, this PR corrects the relationships between profiles and rules by deleting the unnecessary links and not recreating the already existing ones upon SSG import. I would like to create some tests for this if the implementation is valid, so please take a look :pray: 

image


app/services/concerns/xccdf/profile_rules.rb
```diff
@@ -7,19 +7,31 @@ module ProfileRules
...
     included do
       def save_profile_rules
+        new_prs, old_prs = profile_rules.partition(&:new_record?)
+
+        # Get rid of all unnecessary connections under a given benchmark
+        ::ProfileRule.joins(:profile)
+                     .where(profile_id: profile_rules.map(&:profile_id))
+                     .where('profiles.parent_profile_id' => nil)
+                     .where.not(id: old_prs)
+                     .delete_all
```
vkrizan, line 17: First I'd suggest to import new/all and then remove those connections that are extra.

Please do not remove existing. Rather reconcile the extra ones. It has to be done for each profile separately.
Something like
```
profile_rules.group_by(&:profile_id).each do |profile_id, prs|
  ProfileRule.where(profile_id: profile_id)
    .where.not(rule_id: prs.map(&:rule_id))
    .delete_all
end
```
(that could be written with arel in one query, however I'm not sure if it would not hit a param limit)
This call is expected to be run on regular import, hence it does not need to check the canonical state of the profile.

Note that the `.where.not(id: old_prs)` (in the commited code) would not be enough as `old_prs` would hold those rules that are in the benchmark that is being imported. The deletion would lack removal of those extra rule connections.

app/services/concerns/xccdf/profile_rules.rb
```diff
@@ -7,19 +7,31 @@ module ProfileRules
...
+              profile_id: profile_id, rule_id: rule_id
+            )
           end
         end
-
-        ::ProfileRule.import!(@profile_rules.select(&:new_record?),
-                              ignore: true)
       end
 
-      private
```
vkrizan, line 21: Is the `private` removal deliberate?

skateman: Look above :wink:

app/services/concerns/xccdf/profile_rules.rb
skateman: @vkrizan the code above just in addition excludes the `new_prs` from the deletion :thinking: let's talk about this when you're at the office

vkrizan: The `old_prs` does not include all existing `ProfileRules` for that benchmark, specifically those that are not on the list of a benchmark that is being imported (those extra ones).

skateman: What about:
```ruby
     def save_profile_rules
        new_prs, old_prs = profile_rules.partition(&:new_record?)

        new_ids = ::ProfileRule.import!(new_prs, ignore: true)

        # Get rid of all unnecessary connections under a given benchmark
        ::ProfileRule.joins(:profile)
                     .where(profile_id: profile_rules.map(&:profile_id))
                     .where('profiles.parent_profile_id' => nil)
                     .where.not(id: old_prs)
                     .where.not(id: new_ids)
                     .delete_all
      end
```

skateman: Or is it still necessary to group by `profile_id`? :thinking:

app/services/concerns/xccdf/profile_rules.rb
```diff
@@ -7,18 +7,35 @@ module ProfileRules
 
     included do
       def save_profile_rules
```
vkrizan, line 9: Can you please wrap this in a transaction?

vkrizan, line 9: Can you please wrap this in a transaction?

vkrizan (approved):
LGTM. 

I recommend to do some kind of dry-run on production first.
Some logging of removed rules could also be helpful to know what is being removed.

app/services/concerns/xccdf/profile_rules.rb
skateman: Is it really necessary? Isn't the import already running in one?

vkrizan: Since this is a public method, and it might be run outside of an import, it might be good idea. DB also support sub-transactions. Anyhow, I leave that to your judgement.

vkrizan (approved):
@skateman yes, it looks good to me. Thanks.

PR #1009, opened 2021-09-15
fix(clowder): Clean up containers for unit tests

akofink:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/unit_test.sh
+++ b/unit_test.sh
@@ -5,13 +5,7 @@ export IMAGE="quay.io/cloudservices/compliance-backend"
 IMAGE_TAG=$(git rev-parse --short=7 HEAD)
 export IMAGE_TAG
 
-function teardown_docker {
-  docker rm -f "$DB_CONTAINER_ID" || true
-  docker rm -f "$TEST_CONTAINER_ID" || true
-  docker network rm "$NETWORK" || true
-}
-
-trap "teardown_docker" EXIT SIGINT SIGTERM
+DB_CONTAINER_NAME="compliance-db-${IMAGE_TAG}"
 
 NETWORK="compliance-test-${IMAGE_TAG}"
 POSTGRES_IMAGE="quay.io/cloudservices/postgresql-rds:cyndi-12-1"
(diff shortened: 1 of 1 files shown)


Victoremepunto (approved):
LGTM @akofink  thank you. Although, this is pretty similar to the solution @bsquizz prepared, but with the addition of the IMAGE_TAG prefix, which makes me think we can still leave behind come containers ...

akofink:
Yeah, agreed. I think we probably need to add `--rm` to guarantee the container is deleted. Still, this fixes the case when you have multiple builds of the same commit. When do we expect that to happen? When we merge a PR that is up to date with master, the build_master runs and tries to build the same SHA. That's really the only case I can think of.

Victoremepunto:
> Yeah, agreed. I think we probably need to add `--rm` to guarantee the container is deleted. Still, this fixes the case when you have multiple builds of the same commit. When do we expect that to happen? When we merge a PR that is up to date with master, the build_master runs and tries to build the same SHA. That's really the only case I can think of.

mmm, I'm worried the `--rm` flag actually breaks the deployment if the db container needs to restart for whatever reason (I think it does need to restart as part of the init process but I might be wrong). 

All I'm saying is that we still may leave behind some container behind. If there's a bug that can leave db containers leftovers then this PR doesn't fix it, though it reduces the chances of a collision (like you said). 

Let's see how it behaves

vkrizan:
@akofink You could have moved the `--rm` flag to container run. It would remove it after the container terminates, unless there is something unexpected with the container engine.

akofink:
Agreed. Victor thinks the container somehow restarts itself. I'm not super familiar with the 'init process' he mentions - I thought one `docker run` would result in a single container running without restarting itself :confused:

vkrizan:
@akofink @Victoremepunto I'm not sure what problematic was this fixing. :slightly_smiling_face: It moved the "teardown_docker" trap a few lines down with no other changes.

If there is another build for the same commit, the content and the built image should be same.

akofink:
Notice that `teardown_docker` uses env variables. Those variables must be defined _before_ executing the function. That's the gist of how this changes the behavior.

vkrizan:
@akofink Are they defined with the script execution, because between lines 1-24 they are not?

unit_test.sh
```diff
@@ -20,13 +14,21 @@ IMAGE_TAG=$(git rev-parse --short=7 HEAD)
 DATABASE_USER="compliance"
 DATABASE_PASSWORD="changeme"
 DATABASE_NAME="compliance-test"
-DATABASE_HOST="compliance-db"
+
+function teardown_docker {
+  docker rm -f "$DB_CONTAINER_ID" || true
+  docker rm -f "$TEST_CONTAINER_ID" || true
```
akofink, line 20: @vkrizan yep, you're right - I messed up the variable here. It should be `DB_CONTAINER_NAME` instead of the ID

PR #1011, opened 2021-09-16
feat(build): RHICOMPL-2280 enable multistage build to support tableau

skateman:
Introducing a new bundler group for tableau extraction and a next stage of the container build that creates a container with the dependencies for this bundler group. The exclusion/inclusion of these gems is enforced via bundler's `--without` and `--with` arguments, so please keep in mind that you need the proper depds when you try to `bundle install` without :exploding_head: the [dependencies for red-parquet](https://arrow.apache.org/install/).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ruby-27
-
+FROM registry.access.redhat.com/ubi8/ruby-27 AS compliance-backend
 
 USER 0
 
@@ -13,9 +12,28 @@ COPY --chown=1001:0 . /tmp/src
 
 USER 1001
 
-ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true QMAKE=/usr/lib64/qt5/bin/qmake DISABLE_ASSET_COMPILATION=true
+ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true QMAKE=/usr/lib64/qt5/bin/qmake DISABLE_ASSET_COMPILATION=true BUNDLE_WITHOUT=development:test:tableau
 
 # Install the dependencies
 RUN /usr/libexec/s2i/assemble
+
+CMD ["/opt/app-root/src/entrypoint.sh"]
+
+FROM compliance-backend as compliance-tableau
+
+USER 0
+
+RUN yum config-manager --enable "codeready-builder-for-rhel-8-$(arch)-rpms" && \
+    yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1).noarch.rpm && \
+    yum install -y https://apache.jfrog.io/artifactory/arrow/centos/$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1)/apache-arrow-release-latest.rpm && \
+    yum install -y parquet-glib-devel
+
+USER 1001
+
+ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true QMAKE=/usr/lib64/qt5/bin/qmake DISABLE_ASSET_COMPILATION=true BUNDLE_WITHOUT=development:test
+
+# Install the additional tableau dependencies
+RUN bundle install --deployment --retry 2 --path ./bundle --with tableau --without development:test && \
+    bundle clean -V
 
 CMD ["/opt/app-root/src/entrypoint.sh"]
--- a/deploy/build-deploy-common.sh
+++ b/deploy/build-deploy-common.sh
@@ -160,9 +160,9 @@ build_image() {
 
     if [ -n "$BUILD_ARGS" ]; then
         BUILD_ARGS_CMD=$(_get_build_args)
-        container_engine_cmd build --pull -f "$DOCKERFILE" $BUILD_ARGS_CMD -t "${IMAGE_NAME}:${IMAGE_TAG}" .
+        container_engine_cmd build --pull -f "$DOCKERFILE" $BUILD_ARGS_CMD -t "${IMAGE_NAME}:${IMAGE_TAG}" --target=${TARGET_BUILD_STAGE} .
     else
-        container_engine_cmd build --pull -f "$DOCKERFILE" -t "${IMAGE_NAME}:${IMAGE_TAG}" .
+        container_engine_cmd build --pull -f "$DOCKERFILE" -t "${IMAGE_NAME}:${IMAGE_TAG}" --target=${TARGET_BUILD_STAGE} .
     fi
 
 }
--- a/Gemfile
+++ b/Gemfile
@@ -107,6 +107,11 @@ group :development do
   gem 'spring-watcher-listen', '~> 2.0.0'
 end
 
+group :tableau do
+  gem 'aws-sdk-s3', require: false
+  gem 'red-parquet', require: false
+end
+
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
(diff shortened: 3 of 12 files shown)


akofink (approved):
Seems to work well! At least, it doesn't break anything that I've noticed. build_deploy.sh still works well, and so does the local development environment. Would you mind explaining a bit more in the Readme (~line 282) how to build this new tableau image via build_deploy (or even just docker/podman) and how to run it?

akofink:
Also, there seems to be some issue with enabling the sub-man repo in jenkins:

```
12:55:16 Step 11/17 : RUN subscription-manager repos --enable "codeready-builder-for-rhel-8-$(arch)-rpms" &&     yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1).noarch.rpm &&     yum install -y https://apache.jfrog.io/artifactory/arrow/centos/$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1)/apache-arrow-release-latest.rpm &&     yum install -y parquet-glib-devel
12:55:16  ---> Running in 02d51d7805ec
12:55:16 This system has no repositories available through subscriptions.
12:56:15 The command '/bin/sh -c subscription-manager repos --enable "codeready-builder-for-rhel-8-$(arch)-rpms" &&     yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1).noarch.rpm &&     yum install -y https://apache.jfrog.io/artifactory/arrow/centos/$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1)/apache-arrow-release-latest.rpm &&     yum install -y parquet-glib-devel' returned a non-zero code: 1
```

skateman:
> Seems to work well! At least, it doesn't break anything that I've noticed. build_deploy.sh still works well, and so does the local development environment. Would you mind explaining a bit more in the Readme (~line 282) how to build this new tableau image via build_deploy (or even just docker/podman) and how to run it?

Well :see_no_evil: for now the tableau container does nothing, my plan was to add some explanation into the readme *after* there are things that you can do with that container :thinking:

akofink:
I'm fine with that! :)

Dockerfile
```diff
@@ -13,7 +12,29 @@ COPY --chown=1001:0 . /tmp/src
...
+# Install the dependencies
+RUN /usr/libexec/s2i/assemble
+
+CMD ["/opt/app-root/src/entrypoint.sh"]
+
+FROM compliance-backend as compliance-tableau
+
+USER 0
+
+RUN subscription-manager attach && \
```
Victoremepunto, line 26: @skateman you need to remove this , the node where it runs should be already subscribed, this logic must not be part of the image build process.

Dockerfile
```diff
@@ -13,7 +12,29 @@ COPY --chown=1001:0 . /tmp/src
...
+    subscription-manager repos --enable "codeready-builder-for-rhel-8-$(arch)-rpms" && \
+    yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1).noarch.rpm && \
+    yum install -y https://apache.jfrog.io/artifactory/arrow/centos/$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1)/apache-arrow-release-latest.rpm && \
+    yum install -y parquet-glib-devel
+
+# Make an empty file to trick s2i to believe that it's moving the source files
+RUN install -o 1001 -g 0 /tmp/src/placeholder
```
Victoremepunto, line 33: Can we run a bundle install or similar to avoid doing these sort of hack please ? s2i after all does that (among other things) 
see https://github.com/sclorg/s2i-ruby-container/blob/master/3.0/s2i/bin/assemble for reference

skateman: Yes, I can look into that

Victoremepunto (changes requested):
Please review comments

deploy/build-deploy-common.sh
```diff
@@ -160,9 +160,9 @@ build_image() {
 
     if [ -n "$BUILD_ARGS" ]; then
         BUILD_ARGS_CMD=$(_get_build_args)
-        container_engine_cmd build --pull -f "$DOCKERFILE" $BUILD_ARGS_CMD -t "${IMAGE_NAME}:${IMAGE_TAG}" .
+        container_engine_cmd build --target=compliance-backend --pull -f "$DOCKERFILE" $BUILD_ARGS_CMD -t "${IMAGE_NAME}:${IMAGE_TAG}" .
```
Victoremepunto, line 163: please move this logic to the `BUILD_ARGS_CMD` variable. the `build-deploy-common.sh` is supposed to be a library of shared use and not meant to be updated regularly. 

the `BUILD_ARGS_CMD` variable is meant to be used here: https://github.com/RedHatInsights/compliance-backend/blob/442b82959741d82c9fdda58ab4b9d7e8bf70f296/build_deploy.sh#L10

skateman: :thinking: I can do that, but then we would need to run the command twice probably, for both targets :thinking:

Victoremepunto: Sorry, I didn't realize the `BUILD_ARGS_CMD` wouldn't actually work as it is not a Dockerfile build argument, but a `docker/podman` build argument. 

However your approach may not have the effect you want, see https://docs.docker.com/engine/reference/commandline/build/#specifying-target-build-stage---target

Just reorganize the commands order so the `--target` gets to be the last one used please. 

Also, while you're at it, try to fix the fact that you have to forcibly specify now a `target` for the build, when we actually didn't need to before. 

I'd suggest also to refactor the `if` clause that checks if there's build arguments , remove that check entirely as the result should be the same, as the `BUILD_ARGS_CMD` variable will be empty if not used, and use it to instead provide the `--target` option if the variable is provided. 

As a final nitpick - can you rename the variable to something more close to what it is, like `TARGET_BUILD_STAGE` or similar ? 

Thank you !

Victoremepunto: so with the change as it is now,  a `TARGET_BUILD_STAGE` variable is mandatory, which wasn't the case before. 

I think we could avoid this by adding an "if-else" clause to check if it's defined , and if it's not, don't pass the `--target` parameter at all. 

And to not to overly complicate the current logic, I think the if clause for build args is irrelevant, and as an empty `BUILD_ARGS`variable will not break the build command.

so I think setting the code like 

```
    local BUILD_ARGS_CMD=''

    if [ -n "$BUILD_ARGS" ]; then
        BUILD_ARGS_CMD=$(_get_build_args)
    fi

    if [ -n "$TARGET_BUILD_STAGE" ]; then
        container_engine_cmd build --pull -f "$DOCKERFILE" $BUILD_ARGS_CMD -t "${IMAGE_NAME}:${IMAGE_TAG}" --target=${TARGET_BUILD_STAGE} .

    else
        container_engine_cmd build --pull -f "$DOCKERFILE" $BUILD_ARGS_CMD -t "${IMAGE_NAME}:${IMAGE_TAG}" .
    fi
```

should work without creating a new dependency on `--target`.

vkrizan:
@skateman Using multistage builds to build two images that are pushed to a repository is non-standard. A result of a build is one image. However there would have to be two images (backend and tableau), each serving different purpose. Multistage builds are helpful for building/compiling artifacts that are copied to resulting image. 

Considering that it would likely be pushed to a separate name/repository it make sense to me to split it.
Instead of
```
FROM compliance-backend as compliance-tableau
```
it would use
```
# --build-arg BASE_TAG=commitshorthash
ARG BASE_TAG=latest
FROM quay.io/repository/cloudservices/compliance-backend:$BASE_TAG
```
This more common to build on top of previous layers. This should not be a problem when the backend image is build first and tagged prior to building the tableau image.

The tableau version can leverage multistage builds to build the binaries and then copy them to the resulting image. Copying just the necessities would likely reduce the bloatness and would harden the container.

skateman:
We're going with a different (and much simpler) solution...closing.

PR #1012, opened 2021-09-17
feat: RHICOMPL-2145 Use revision table to skip parsing datastreams

akofink:
Also contains https://github.com/RedHatInsights/compliance-backend/pull/1013

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -369,9 +369,9 @@ parameters:
 - name: MEMORY_REQUEST_SIDE
   value: 500Mi
 - name: CPU_LIMIT_SIDE
-  value: 800m
+  value: 1000m
 - name: CPU_REQUEST_SIDE
-  value: 80m
+  value: 100m
 - name: REDIS_SSL
   description: 'Whether to use secured connection to Redis. Use string values of true or false'
   value: "true"
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -36,8 +36,10 @@ objects:
       enabled: true
       appName: "compliance"
       insightsOnly: true
-    deployments:
-    - name: db-migrations
+    jobs:
+    - name: compliance-import-ssg
+      schedule: "*/5 * * * *" # every 5 minutes
+      concurrencyPolicy: Forbid
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
@@ -45,7 +47,9 @@ objects:
             - bash
             - -c
             - |
-              bundle exec rake --trace db:debug db:migrate
+              for i in {1..120}; do sleep 1;
+              if bundle exec rake --trace db:abort_if_pending_migrations;
+              then exit 0; fi; done; exit 1
           inheritEnv: true
         resources:
           limits:
@@ -56,7 +60,7 @@ objects:
             memory: ${MEMORY_REQUEST_SIDE}
         env:
         - name: APPLICATION_TYPE
-          value: sleep
+          value: compliance-import-ssg
         - name: RAILS_ENV
           value: "${RAILS_ENV}"
         - name: RAILS_LOG_TO_STDOUT
@@ -72,7 +76,8 @@ objects:
           value: "secret_key_base"
         - name: SETTINGS__REPORT_DOWNLOAD_SSL_ONLY
           value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
-    - name: compliance-import-ssg
+    deployments:
+    - name: db-migrations
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
@@ -80,8 +85,7 @@ objects:
             - bash
             - -c
             - |
-              while ! bundle exec rake --trace db:abort_if_pending_migrations; do sleep 1; done;
-              bundle exec rake ssg:import_rhel_supported --trace
+              bundle exec rake --trace db:debug db:migrate
           inheritEnv: true
         resources:
           limits:
@@ -124,7 +128,7 @@ objects:
           - -c
           - |
             for i in {1..120}; do sleep 1;
-            if bundle exec rake --trace db:abort_if_pending_migrations;
+            if bundle exec rake --trace db:abort_if_pending_migrations ssg:check_synced;
             then exit 0; fi; done; exit 1
           env:
           - name: RAILS_ENV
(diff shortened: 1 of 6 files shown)


clowdapp.yaml
```diff
@@ -36,16 +36,18 @@ objects:
...
+      concurrencyPolicy: Forbid
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
         - command:
             - bash
             - -c
             - |
-              bundle exec rake --trace db:debug db:migrate
+              while ! bundle exec rake --trace db:abort_if_pending_migrations; do sleep 1; done;
```
vkrizan, line 50: Can it wait up to some time out? The same as waits elsewhere with `for i in {1..120}; ...`?

lib/tasks/import_datastream.rake
```diff
@@ -4,6 +4,17 @@
...
 # rubocop:disable Metrics/BlockLength
 namespace :ssg do
+  desc 'Check if the latest SSG has been synced'
+  task check_synced: [:environment] do
+    message = "SSG datastreams not synced\nDatastream config revision: " \
+      "#{SupportedSsg.revision.inspect}\nDB revision: " \
+      "#{Revision.datastreams.inspect}"
+    if Revision.datastreams != SupportedSsg.revision
+      puts message
+      exit(1)
```
vkrizan, line 14: This would stop the whole rake process and would not allow to chain this task. Consider using `abort` instead.

vkrizan (commented):
Looks good. I might be able to test in a bit.

clowdapp.yaml
akofink: Sure, but why? What do you recommend doing on timeout? What's the problem with waiting infinitely? The timeout will only serve to exit(1) and cause openshift to restart the pod and retry the task. IMO that's more complex and more difficult to determine what's going on because we see multiple restarts just because it's still waiting.

vkrizan: The restart is deliberate here to claim the service as failed once the timeout is reached. The reason for this is to fulfill an expectation that a migration would not take longer than expected. If it does take longer, we do not want to have it hang indefinitely, but rather fail it and restart its lifecycle. I feel like keeping it wait without this "timeout" could lead to some hanging problems. If for example the migration will take too long or would not finished (would be killed over and over), then OpenShift/Kubernetes after a several failures might back off for a while and release cluster's resources.

These 120 cycles might be enough for debugging. Anyhow, this is a busy wait that depends on some other pod/service.

akofink: If the migrations fail, then the entire deployment will fail regardless of the state of this init container due to how clowder works. Your explanation does not convince me of any benefit; however, I'll make the change for consistency.

IMO we should either wait indefinitely (like this while loop) and not rely on openshift _or_ don't include any loop here and just let the pod crash until it's ready to start and rely completely on openshift. Doing some sort of hybrid complicates the deployment logic for no reason.

lib/tasks/import_datastream.rake
```diff
@@ -4,32 +4,24 @@
...
-    rescue StandardError => e
-      ExceptionNotifier.notify_exception(
-        e,
-        data: OpenshiftEnvironment.summary
-      )
-    end
+  desc 'Check if the latest SSG has been synced'
+  task check_synced: [:environment] do
+    message = "SSG datastreams not synced\nDatastream config revision: " \
+      "#{SupportedSsg.revision.inspect}\nDB revision: " \
```
vkrizan, line 10: Suggestion: can it also print/log when the revisions are matching, with the new/db revision included in the message?

clowdapp.yaml
```diff
@@ -70,16 +74,16 @@ objects:
...
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
         - command:
             - bash
             - -c
             - |
-              while ! bundle exec rake --trace db:abort_if_pending_migrations; do sleep 1; done;
-              bundle exec rake ssg:import_rhel_supported --trace
+              bundle exec rake --trace db:debug db:migrate
```
vkrizan, line 86: Out of scope for this effort: the migrations should probably run in a regular container not in initContainer, then the "sleep" workaround would not be needed.

lib/tasks/import_datastream.rake
```diff
@@ -4,32 +4,24 @@
...
-             ' seconds.'
-      end
-    rescue StandardError => e
-      ExceptionNotifier.notify_exception(
-        e,
-        data: OpenshiftEnvironment.summary
-      )
-    end
+  desc 'Check if the latest SSG has been synced'
+  task check_synced: [:environment] do
```
vkrizan, line 8: Do we have a coverage on this rake task?

vkrizan (approved):
Worked well in ephemeral. Thank you!

clowdapp.yaml
akofink: Unfortunately, clowder does not allow a `restartPolicy` of `OnFailure`. It's hardcoded to `Always` so the pod must not terminate.

lib/tasks/import_datastream.rake
akofink: Using `abort` prevents us from writing tests against this because it aborts the entire test run. I've resorted to using `raise` instead (updates incoming), which we use in other rake tasks as well.

clowdapp.yaml
vkrizan: Okay, I missed the fact that this is a deployment and not a job.

lib/tasks/import_datastream.rake
vkrizan: Isn't there a mock for that?

akofink: Aha! I figured it out: `assert_raises SystemExit do` (change incoming)

lib/tasks/import_datastream.rake
akofink: Done. I also added a similar message at the end of the import_ssg task.

lib/tasks/import_datastream.rake
akofink: Now we do! Thanks for pointing out that oversight.

akofink:
Merging without passing pr_check again. We've seen it green before the last rebase, and there's currently an SSO outage failing our builds.

PR #1019, opened 2021-09-27
perf(REST): RHICOMPL-2305 do not allow nested includes

skateman:
@akofink using just stronger params wasn't enough as the `includes` field according to the `openapi.json` should be comma separated and there's no matcher for that in the gem :disappointed: anyway, I tried to utilize the gem's errors so that will be consistent at lest


app/controllers/concerns/parameters.rb
```diff
@@ -36,6 +36,19 @@ def new_relationship_ids(relationship)
       end
     end
 
+    def include_params
+      splitted = params[:include].split(',')
+      valids = serializer.relationships_to_serialize.keys.map(&:to_s)
+
+      if splitted & valids != splitted
```
vkrizan, line 43: Can you please elaborate on what it does and how it compares the values?

app/controllers/concerns/parameters.rb
```diff
@@ -36,6 +36,19 @@ def new_relationship_ids(relationship)
       end
     end
 
+    def include_params
+      splitted = params[:include].split(',')
+      valids = serializer.relationships_to_serialize.keys.map(&:to_s)
```
akofink, line 41: I'd really appreciate defining available includes explicitly within each controller (i.e. an `allowed_includes` private method) rather than simply allowing includes on any relationship. Also, you're not really using StrongerParameters as it is intended here; let the library do its job and allow us to skip manual comparisons. You can leverage `ActionController::Parameters.regexp`.

skateman: @akofink will and should the explicit definition of these fields differ from the ones specified in the serializer? AFAIK it would be just duplication, that's why I went with the `relationships_to_serialize` method. I will look into the regexp :+1:

app/controllers/concerns/parameters.rb
skateman: Intersection with a set equals with the original set.

app/controllers/concerns/parameters.rb
akofink: Some of the relationships defined don't even exist as API resources, so you can't `include` them (i.e. test_results).

I guess, one simplification of this issue could be to define a simple regexp type that does not allow periods at all but continues to allow any include of any relationship, even if it does not exist. Still, the best input validation here would be matching against an enum of allowable values per API resource.

skateman: @akofink the problem with the enum is that I would have to reconstruct the params as it is defined as a comma separated string and stronger params can't deal with that.

akofink: We discussed this in person a bit. The simple regex that allows word characters, underscores, and commas can be used for initial validation with StrongerParams. If an include is specified that does not exist, the backend today returns a 500, an unhandled `JSONAPI::Serializer::UnsupportedIncludeError`. We could rescue that in our application_controller and return a 4XX instead, since it is a user input issue.

skateman:
@akofink here's what the error messages would look like:
```
* Invalid parameter: include must match ^[a-zA-Z0-9,_]+$"
* Invalid parameter: benchmarkk is not specified as a relationship
```

Are these okay?

akofink:
Looks good to me! If a user is able to use our JSON REST API, I don't think they would mind a little regex. I've seen similar regex 4XXs from the Inventory API as well when searching tags.

skateman:
@vkrizan can you please take a look at this?

PR #1021, opened 2021-09-30
feat(metrics): RHICOMPL-2316 LRU cache dangling items

vkrizan:
Adds LRU cache on engineering prometheus collector, that mostly
collects dangling items.

This should alleviate hits to database, and should fix database high
load.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/lib/prometheus/engineering_collector.rb
+++ b/lib/prometheus/engineering_collector.rb
@@ -6,7 +6,10 @@ end
 
 # when /metrics is called. This runs directly on prometheus_exporter.
 class EngineeringCollector < PrometheusExporter::Server::TypeCollector
+  LRU_DURATION = 1.hour
+
   def initialize
+    @cache = ActiveSupport::Cache::MemoryStore.new(expires_in: LRU_DURATION)
     @dangling_accounts = PrometheusExporter::Metric::Gauge.new(
       'dangling_accounts', 'Accounts without hosts'
     )
@@ -42,13 +45,15 @@ class EngineeringCollector < PrometheusExporter::Server::TypeCollector
   end
 
   def metrics
-    collect
+    @cache.fetch('metrics') do
+      collect
 
-    [
-      @dangling_accounts,
-      @dangling_test_results,
-      @dangling_rule_results,
-      @dangling_policy_hosts
-    ]
+      [
+        @dangling_accounts,
+        @dangling_test_results,
+        @dangling_rule_results,
+        @dangling_policy_hosts
+      ]
+    end
   end
 end


lib/prometheus/engineering_collector.rb
```diff
@@ -1,12 +1,17 @@
...
 
+require 'lru_redux'
+
 unless defined? Rails
   require File.expand_path('../../config/environment', __dir__)
 end
 
 # when /metrics is called. This runs directly on prometheus_exporter.
 class EngineeringCollector < PrometheusExporter::Server::TypeCollector
+  LRU_MINUTES = 15
```
skateman, line 11: I think for these specific metrics below the number can be much higher. I would go with 60 minutes, but having like 12 hours is also okay from the data integrity side. On the other hand, why should we keep something in-memory for days :shrug:

vkrizan: One hour it is.

skateman (approved):
The Seal of Approval
Looks good to me, works on my machine...but it probably needs the stress of the stage environment to see if it's working for real

lib/prometheus/engineering_collector.rb
```diff
@@ -1,12 +1,17 @@
...
 unless defined? Rails
   require File.expand_path('../../config/environment', __dir__)
 end
 
 # when /metrics is called. This runs directly on prometheus_exporter.
 class EngineeringCollector < PrometheusExporter::Server::TypeCollector
+  LRU_DURATION = 1.hour
+
   def initialize
+    @cache = LruRedux::TTL::ThreadSafeCache.new(1, LRU_DURATION.seconds.to_i)
```
akofink, line 14: Why introduce a new dependency when we could use redis with Rails.cache?

vkrizan: Can we specify timout with generic Rails caching?

skateman: If not, we can rely on redis' own methods, there you can set a TTL.

akofink: Yes, there is an `expires_in` option. See https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-fetch

akofink:
I hope we're viewing this as a band-aid and not a complete solution. Metrics per pod greatly increases the load on the DB for no benefit to us (backend and sidekiq both run with 6 pods each, for example). I'd like to figure out how we can re-architect our metrics rather than working around the poor existing architecture. Also, our metrics are broken right now in production due similarly to the poor architecture we've landed on.

vkrizan:
@akofink Indeed this is a kind of a band-aid. The LRU cache will likely stay, however the architecture would need some thought (as you've mentioned). Pods should export only pod-specific metrics. Extra metrics would likely need its own service/pod.

vkrizan:
@akofink Indeed this is a kind of a band-aid. The LRU cache will likely stay, however the architecture would need some thought (as you've mentioned). Pods should export only pod-specific metrics. Extra metrics would likely need its own service/pod.

lib/prometheus/engineering_collector.rb
akofink: I am opposed to adding a new unnecessary dependency. Looking at the gem, it doesn't even appear active (last changes were in 2017; last release was March 2015) or well documented (docs link broken: https://www.rubydoc.info/gems/lru_redux/1.1.0). We need to be careful when selecting new dependencies and ideally not introduce them at all where not necessary.

vkrizan: Yes, I'm on the same page. The lib was recommended by Racecar. I've checked activeness of that project and it seemed fine (code is stable and there are discussions happening).

vkrizan:
@akofink @skateman PTAL

skateman (approved):
The Seal of Approval

akofink:
Nice! Thanks, @vkrizan! Maybe we should also do the same for the business_collector.

PR #1022, opened 2021-09-30
feat(GQL): RHICOMPL-2209 expose upstream field for profiles/rules

skateman:


skateman:
Probably not needed, closing...

PR #1025, opened 2021-10-04
feat(Clowder): RHICOMPL-2317 default Kafka settings to enforce SALS_TLS

Victoremepunto:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -416,9 +416,6 @@ parameters:
   value: "true"
 - name: PUMA_MAX_THREADS
   value: "5"
-- name: POSTGRESQL_MAX_CONNECTIONS
-  description: Number of maximum database connections
-  value: "50"
 - name: RACECAR_OFFSET_COMMIT_INTERVAL
   required: true
   value: "5"
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -23,7 +23,7 @@ if ClowderCommonRuby::Config.clowder_enabled?
   first_kafka_server_config = config.kafka.brokers[0]
 
   kafka_server_config = {
-    brokers: "#{first_kafka_server_config.hostname}:#{first_kafka_server_config.port}",
+    brokers: "#{first_kafka_server_config.dig('hostname')}:#{first_kafka_server_config.dig('port')}",
     security_protocol: first_kafka_server_config.authtype
   }
 
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -29,17 +29,17 @@ if ClowderCommonRuby::Config.clowder_enabled?
 
   if kafka_security_protocol
     if kafka_security_protocol == 'sasl'
-      kafka_server_config['security_protocol'] = 'sasl_ssl'
+      kafka_server_config[:security_protocol] = 'sasl_ssl'
       kafkaCaFile = Tempfile.create(mode: 'wt')
       kafkaCaFile.write(first_kafka_server_config&.dig('cacert'))
-      kafka_server_config['ssl_ca_location'] = kafkaCaFile.path
-      kafka_server_config['sasl_username'] = first_kafka_server_config&.dig('sasl', 'username')
-      kafka_server_config['sasl_password'] = first_kafka_server_config&.dig('sasl', 'password')
+      kafka_server_config[:ssl_ca_location] = kafkaCaFile.path
+      kafka_server_config[:sasl_username] = first_kafka_server_config&.dig('sasl', 'username')
+      kafka_server_config[:sasl_password] = first_kafka_server_config&.dig('sasl', 'password')
     else
       raise "Unsupported Kafka security protocol '#{kafka_security_protocol}'"
     end
   else
-    kafka_server_config['security_protocol'] = 'plaintext'
+    kafka_server_config[:security_protocol] = 'plaintext'
   end
 
   clowder_config = {
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -21,7 +21,7 @@ if ClowderCommonRuby::Config.clowder_enabled?
 
   # Kafka
   first_kafka_server_config = config.kafka.brokers[0]
-  kafka_security_protocol = first_kafka_server_config.dig('authtype')
+  kafka_security_protocol = first_kafka_server_config&.dig('authtype')
 
   kafka_server_config = {
     brokers: config.dig('kafka', 'brokers')&.map { |b| "#{b&.dig('hostname')}:#{b&.dig('port')}" }.join(',')
@@ -31,10 +31,10 @@ if ClowderCommonRuby::Config.clowder_enabled?
     if kafka_security_protocol == 'sasl'
       kafka_server_config['security_protocol'] = 'sasl_ssl'
       kafkaCaFile = Tempfile.create(mode: 'wt')
-      kafkaCaFile.write(first_kafka_server_config.cacert)
+      kafkaCaFile.write(first_kafka_server_config&.dig('cacert'))
       kafka_server_config['ssl_ca_location'] = kafkaCaFile.path
-      kafka_server_config['sasl_username'] = first_kafka_server_config.dig('sasl', 'username')
-      kafka_server_config['sasl_password'] = first_kafka_server_config.dig('sasl', 'password')
+      kafka_server_config['sasl_username'] = first_kafka_server_config&.dig('sasl', 'username')
+      kafka_server_config['sasl_password'] = first_kafka_server_config&.dig('sasl', 'password')
     else
       raise "Unsupported Kafka security protocol '#{kafka_security_protocol}'"
     end
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -30,9 +30,9 @@ if ClowderCommonRuby::Config.clowder_enabled?
   if kafka_security_protocol
     if kafka_security_protocol == 'sasl'
       kafka_server_config['security_protocol'] = 'sasl_ssl'
-      kafkaCaFile = File.new('tmp/kafkaCa', 'wt')
+      kafkaCaFile = Tempfile.create(mode: 'wt')
       kafkaCaFile.write(first_kafka_server_config.cacert)
-      kafka_server_config['ssl_ca_location'] = File.expand_path(kafkaCaFile.path)
+      kafka_server_config['ssl_ca_location'] = kafkaCaFile.path
       kafka_server_config['sasl_username'] = first_kafka_server_config.dig('sasl', 'username')
       kafka_server_config['sasl_password'] = first_kafka_server_config.dig('sasl', 'password')
     else
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -24,7 +24,7 @@ if ClowderCommonRuby::Config.clowder_enabled?
 
   kafka_server_config = {
     brokers: "#{first_kafka_server_config.dig('hostname')}:#{first_kafka_server_config.dig('port')}",
-    security_protocol: first_kafka_server_config.authtype
+    security_protocol: first_kafka_server_config.dig('authtype')
   }
 
   if kafka_server_config['security_protocol'] == 'sasl_ssl'
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -19,13 +19,25 @@ if ClowderCommonRuby::Config.clowder_enabled?
   redis_url = "#{config.dig('inMemoryDb', 'hostname')}:#{config.dig('inMemoryDb', 'port')}"
   redis_password = config.dig('inMemoryDb', 'password')
 
+  # Kafka
+  first_kafka_server_config = config.kafka.brokers[0]
+
+  kafka_server_config = {
+    brokers: "#{first_kafka_server_config.hostname}:#{first_kafka_server_config.port}",
+    security_protocol: first_kafka_server_config.authtype
+  }
+
+  if kafka_server_config['security_protocol'] == 'sasl_ssl'
+    kafkaCaFile = File.new('tmp/kafkaCa', 'wt')
+    kafkaCaFile.write(first_kafka_server_config.cacert)
+    kafka_server_config['ssl_ca_location'] = File.expand_path(kafkaCaFile.path)
+    kafka_server_config['sasl_username'] = first_kafka_server_config.dig('sasl', 'username')
+    kafka_server_config['sasl_password'] = first_kafka_server_config.dig('sasl', 'password')
+  end
+
   clowder_config = {
     compliance_ssg_url: compliance_ssg_url,
-    kafka: {
-      brokers: config.dig('kafka', 'brokers')&.map { |b| "#{b&.dig('hostname')}:#{b&.dig('port')}" }.join(','),
-      # Not provided by clowder, not sure which of the following should be: [:plaintext, :ssl, :sasl_plaintext, :sasl_ssl]
-      security_protocol: 'plaintext'
-    },
+    kafka: kafka_server_config,
     kafka_consumer_topics: {
       inventory_events: config.kafka_topics&.dig('platform.inventory.events', 'name')
     },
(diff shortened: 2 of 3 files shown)


clowdapp.yaml
```diff
@@ -76,6 +76,8 @@ objects:
           value: "secret_key_base"
         - name: SETTINGS__REPORT_DOWNLOAD_SSL_ONLY
           value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
+        - name: SETTINGS__KAFKA__SECURITY_PROTOCOL
+          value: ${SETTINGS__KAFKA__SECURITY_PROTOCOL}
```
akofink, line 80: Not necessary here.

clowdapp.yaml
```diff
@@ -114,8 +116,9 @@ objects:
           value: "${SETTINGS__FORCE_IMPORT_SSGS}"
         - name: SETTINGS__REPORT_DOWNLOAD_SSL_ONLY
           value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
+        - name: SETTINGS__KAFKA__SECURITY_PROTOCOL
+          value: ${SETTINGS__KAFKA__SECURITY_PROTOCOL}
```
akofink, line 120: Not necessary

clowdapp.yaml
```diff
@@ -114,8 +116,9 @@ objects:
...
     - name: compliance-backend
-      # TODO: Check how to auto-scale with Clowder
```
akofink, line 118: Unrelated line is removed. This TODO is still valid.

Victoremepunto: you really cannot autoscale with Clowder, we saw that, there's no way to do it as of today, so we hardcoded the number of replicas, that's why I removed it. 

If you still think it's valid I can put it back.

akofink: I'd prefer to keep it so we remember to integrate with the feature once Clowder implements it, but I don't feel strongly.

vkrizan:
@Victoremepunto can you please elaborate on this change (setting `sasl_ssl` as default)? How would envs. set this up? And would it be possible to handle it within the clowder init (depending on the provided configuration)?
https://github.com/RedHatInsights/clowder/blob/master/docs/antora/modules/providers/pages/kafka.adoc#json-structure

Victoremepunto:
> @Victoremepunto can you please elaborate on this change (setting `sasl_ssl` as default)? How would envs. set this up? And would it be possible to handle it within the clowder init (depending on the provided configuration)? https://github.com/RedHatInsights/clowder/blob/master/docs/antora/modules/providers/pages/kafka.adoc#json-structure

The intent was to enforce this config on every environment, overriding the setting to "plaintext" on those environments that are not TLS auth. enabled. 

Seeing that this might not be the case for that many environments, it might be better to leave it as "plaintext" as it was before. What do you suggest @vkrizan ?

config/initializers/0_clowder-config.rb
```diff
@@ -25,9 +25,7 @@
   clowder_config = {
     compliance_ssg_url: compliance_ssg_url,
     kafka: {
-      brokers: config.dig('kafka', 'brokers')&.map { |b| "#{b&.dig('hostname')}:#{b&.dig('port')}" }.join(','),
-      # Not provided by clowder, not sure which of the following should be: [:plaintext, :ssl, :sasl_plaintext, :sasl_ssl]
-      security_protocol: 'plaintext'
+      brokers: config.dig('kafka', 'brokers')&.map { |b| "#{b&.dig('hostname')}:#{b&.dig('port')}" }.join(',')
```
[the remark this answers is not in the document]
Victoremepunto: added

Victoremepunto:
> @Victoremepunto getting stale. Can we move forward on this one?

Sure, sorry it took so long, other priorities .... wanna do the honors ?

Victoremepunto:
> @Victoremepunto I'd suggest to focus on the configuration init from Clowder-passed configuration. We expect envs to be Clowderized, maybe with the exception of local development.
> 
> There are two items that can be checked from Clowder config: SLL enabled with SSL cert provided (or use of system certs); and SASL authentication. If neither are present we can default back to "plaintext", and rely on Kafka rejecting incorrect configuration.

I removed the variables as discussed from the templates and relied on Clowder to configure Kafka auth based on what Clowder provides. If Auth is set to SASL then the corresponding variables will be set accordingly.

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,27 @@
...
 
+  # Kafka
+  first_kafka_server_config = config.dig('kafka', 'brokers')[0]
+
+  kafka_server_config = {
+    brokers: "#{first_kafka_server_config&.dig('hostname')}:#{first_kafka_server_config&.dig('port')}",
+    security_protocol: first_kafka_server_config.dig('authtype')
+  }
+
+  if kafka_server_config.security_protocol == 'sasl_ssl'
```
akofink, line 30: Use safe navigation

```suggestion
  if kafka_server_config&.security_protocol == 'sasl_ssl'
```

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,27 @@
   redis_url = "#{config.dig('inMemoryDb', 'hostname')}:#{config.dig('inMemoryDb', 'port')}"
   redis_password = config.dig('inMemoryDb', 'password')
 
+  # Kafka
+  first_kafka_server_config = config.dig('kafka', 'brokers')[0]
```
akofink, line 23: Are we guaranteed a single broker? Why not stick with the `map` logic below?

Victoremepunto: apps I've checked go by default with the first one (see [this](https://github.com/RedHatInsights/insights-ingress-go/blob/af9fff20de2cfd98a97b507fd5d088de5a907a9c/internal/config/config.go#L148))

Another point is , in Clowder, the authtype is tied to each broker, so each broker should have the same auth. mode in order for the config to work in every situation

Victoremepunto: See https://consoledot.pages.redhat.com/clowder/dev/providers/kafka.html

you can see the auth. info is on each `broker` section, so we'll have to do handle that somehow ...suggestions are welcome

clowdapp.yaml
```diff
@@ -416,9 +416,6 @@ parameters:
   value: "true"
 - name: PUMA_MAX_THREADS
   value: "5"
-- name: POSTGRESQL_MAX_CONNECTIONS
-  description: Number of maximum database connections
-  value: "50"
```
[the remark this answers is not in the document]
Victoremepunto: Yeah that's not used at all and thought it's worth removing it - is that an issue  ? ( I can separate that change in a different commit but otherwise, I don't see how fixing up stuff we see wrong while working on a PR can be an issue )

config/initializers/0_clowder-config.rb
Victoremepunto: ugh - I didn't see this suggestion and went with `kafka_server_config['security_protocol']` syntax :/ 

What is recommended in Rails better ?

akofink: Both should work and return `nil` if the key doesn't exist. The suggested syntax is a newer Ruby feature for safe navigation. In hashes, there is also `dig` which allows multiple levels, like `my_hash.dig('foo', 'bar', 'baz')` and will return `nil` if any key along the way doesn't exist.

clowdapp.yaml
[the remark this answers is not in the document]
akofink: Just checking :) I think it's fine to include.

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,25 @@
...
+
+  kafka_server_config = OpenStruct.new(
+    brokers: "#{first_kafka_server_config.hostname}:#{first_kafka_server_config.port}",
```
akofink, line 26:
```
undefined method `hostname' for {"hostname"=>"kafka", "port"=>29092}:Hash                                                                                                                                          
/app/config/initializers/0_clowder-config.rb:26:in `<main>
```

```suggestion
    brokers: "#{first_kafka_server_config.dig('hostname')}:#{first_kafka_server_config.dig('port')}",
```

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,25 @@
...
+    security_protocol: first_kafka_server_config.authtype
```
akofink, line 27:
```suggestion
    security_protocol: first_kafka_server_config.dig('authtype')
```

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,25 @@
   redis_url = "#{config.dig('inMemoryDb', 'hostname')}:#{config.dig('inMemoryDb', 'port')}"
   redis_password = config.dig('inMemoryDb', 'password')
 
+  # Kafka
+  first_kafka_server_config = config.dig('kafka', 'brokers')[0]
```
akofink, line 23: Or maybe, if you don't want to use `dig`:

```suggestion
  first_kafka_server_config = OpenStruct.new(config.dig('kafka', 'brokers')[0])
```

clowdapp.yaml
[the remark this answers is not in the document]
vkrizan: @Victoremepunto even if this isn't used, I'd rather not remove it here as it is a different concern. I'd suggest looking into why it is missing and why it was set... Anyhow, the default seems to be typically 100, as per documentation https://www.postgresql.org/docs/12/runtime-config-connection.html so removal should be fine (?).

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,25 @@
...
+  # Kafka
+  first_kafka_server_config = config.kafka.brokers[0]
+
+  kafka_server_config = {
+    brokers: "#{first_kafka_server_config.dig('hostname')}:#{first_kafka_server_config.dig('port')}",
+    security_protocol: first_kafka_server_config.dig('authtype')
+  }
+
+  if kafka_server_config['security_protocol'] == 'sasl_ssl'
+    kafkaCaFile = File.new('tmp/kafkaCa', 'wt')
```
vkrizan, line 31: Please use [`Tempfile`](https://ruby-doc.org/stdlib-3.0.0/libdoc/tempfile/rdoc/Tempfile.html). It would put the file in the right location with correct permissions.

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,25 @@
   redis_url = "#{config.dig('inMemoryDb', 'hostname')}:#{config.dig('inMemoryDb', 'port')}"
   redis_password = config.dig('inMemoryDb', 'password')
 
+  # Kafka
+  first_kafka_server_config = config.kafka.brokers[0]
+
+  kafka_server_config = {
+    brokers: "#{first_kafka_server_config.dig('hostname')}:#{first_kafka_server_config.dig('port')}",
```
vkrizan, line 26: (Continuation of https://github.com/RedHatInsights/compliance-backend/pull/1025#discussion_r743691416)

Can we put all brokers locations here? We can keep reading the base configuration (auth, cacert) from the first one.

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,25 @@
...
+    security_protocol: first_kafka_server_config.dig('authtype')
```
vkrizan, line 27: `authtype` would be `sasl` as per [Kafka provider example](https://redhatinsights.github.io/clowder/clowder/dev/providers/kafka.html#_json_structure). If there is a presence of cacert it should translate to `sasl_ssl`. This translation might happen always for `sasl` authtype, as Kafka would reject an access w/o SSL when SASL is used.

config/initializers/0_clowder-config.rb
Victoremepunto: Yeah that makes perfect sense ! I have a question though , I'm following the same approach as in [here](https://github.com/RedHatInsights/compliance-backend/blob/cea75b77678f2cc8699cf1a679337a0be94ff9d1/config/database.yml#L5) , also note I'm using the relative path to our project's tmp dir. 

Are you suggesting to switch it to the system default's TMPDIR ? (likely `/tmp`) or to use Tempdir as an interface and keep using the local tmp dir ?

config/initializers/0_clowder-config.rb
Victoremepunto: yeah I thought that too - however I don't know how that will work , if ... there's actual multiple brokers, each with different auth. modes

Victoremepunto: I've re-added all the broker locations as suggested and used the first broker's config for setting up the authentication config.

config/initializers/0_clowder-config.rb
Victoremepunto: I've refactored how I set the `security_protocol` . The three options from Clowder config are: 
- nothing (translates to `plaintext`)
- `sasl` (translate to `sasl_ssl`)
- `mtls` (raises an exception claiming that's unsupported)

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,32 @@
   redis_url = "#{config.dig('inMemoryDb', 'hostname')}:#{config.dig('inMemoryDb', 'port')}"
   redis_password = config.dig('inMemoryDb', 'password')
 
+  # Kafka
+  first_kafka_server_config = config.kafka.brokers[0]
+  kafka_security_protocol = first_kafka_server_config.dig('authtype')
```
vkrizan, line 24: Observation/Optional: This would crash if first broker config isn't set. Use safe navigation `&.dig(...`.

config/initializers/0_clowder-config.rb
```diff
@@ -19,13 +19,32 @@
...
+
+  kafka_server_config = {
+    brokers: config.dig('kafka', 'brokers')&.map { |b| "#{b&.dig('hostname')}:#{b&.dig('port')}" }.join(',')
+  }
+
+  if kafka_security_protocol
+    if kafka_security_protocol == 'sasl'
+      kafka_server_config['security_protocol'] = 'sasl_ssl'
```
vkrizan, line 32: Nit: `kafka_server_config` started with Symbols as key. This is not a problem, as the Settings load would map it correctly.

vkrizan (approved):
LGTM

config/initializers/0_clowder-config.rb
Victoremepunto: Sorry, as a Ruby noob I didn't even tell the difference at first, thanks ! I updated all the string keys to symbols

config/initializers/0_clowder-config.rb
Victoremepunto: Agree, I changed all the accesses to this object to use safe_navigation. However, it is true that if not even the first broker info is there , there will be bigger problems to solve :)

clowdapp.yaml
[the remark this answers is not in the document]
Victoremepunto: My guess is that this was added by error at some point - this `POSTGRESQL_MAX_CONNECTIONS` make sense on the PostgreSQL db container like can be seen [here](https://github.com/sclorg/postgresql-container/tree/generated/12#environment-variables-and-volumes) - we used this in [the CI, QA environments on the previous deployment templates](https://github.com/RedHatInsights/e2e-deploy/blob/ff96f1019828929453d33ad710ea688930d064d9/templates/compliance/compliance-db.yml#L106) , not anymore

config/initializers/0_clowder-config.rb
Victoremepunto: Yeah but they come from Clowder on a per-broker  basis - it is not expected to have different brokers with different auth methods though, so we decided to default with the first broker auth. config

config/initializers/0_clowder-config.rb
Victoremepunto: this was my suggestion initially but declined in order to keep it simple, as you also agreed on ;)

config/initializers/0_clowder-config.rb
Victoremepunto: I defaulted to whatever the TMPDIR is set and declined even to define a name (which is irrelevant), I only had to define the `mode` to ensure I have write permissions and that it was opened in text mode.

vkrizan: Yeah, keep the defaults of Tempfile.

PR #1026, opened 2021-10-04
fix: RHICOMPL-1113 Removes name from updatable policy attributes

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/graphql/mutations/profile/edit.rb
```diff
@@ -6,7 +6,7 @@ module Profile
     class Edit < BaseMutation
       graphql_name 'UpdateProfile'
 
-      POLICY_ATTRIBUTES = %i[name description
+      POLICY_ATTRIBUTES = %i[description
                              compliance_threshold business_objective_id].freeze
```
akofink, line 10: The graphql schema is autogenerated and must be re-generated pretty much whenever we change anything within `app/graphql`: `rake dump_schema`

akofink (commented):
Works well! It'd be great to limit the allowed parameters during update to only those allowed to be updated. Right now, sending name does not update the name, but it also does not error, so users may be confused why the profile name wasn't updated.

Any time we change the GraphQL or REST APIs, we must also update their generated documentation. `app/graphql/schema.graphql` for GraphQL or `swagger/openapi.json` for REST.

app/graphql/mutations/profile/edit.rb
marleystipich2: Working on these updates, thanks for the review!

PR #1028, opened 2021-10-05
fix(db): RHICOMPL-2328 get rid of the NULL account

skateman:
Also adding validation for mandatory value and `null: false`


akofink (approved):
:1st_place_medal: Thanks, @skateman!

PR #1054, opened 2021-11-04
chore(db): RHICOMPL-2363 verbose dry-run of the ProfileRule reconciliation

skateman:
This has to be deployed again with `SETTINGS__FORCE_IMPORT_SSGS`


db/migrate/20211104065354_reset_datastream_revisions.rb
```diff
@@ -0,0 +1,9 @@
+class ResetDatastreamRevisions < ActiveRecord::Migration[5.2]
+  def up
+    Revision.find_by(name: 'datastreams')&.delete
```
vkrizan, line 3: Is this enough for a proper force import?

vkrizan (approved):
LGTM

db/migrate/20211104065354_reset_datastream_revisions.rb
skateman: We also need the ENV var, but last time it worked :wink:

PR #1055, opened 2021-11-04
cleanup(db): RHICOMPL-2440 remove upstream profiles with test results

skateman:
Also removing any possible dangling account that has been created during the process.

### DIFF
--- a/app/services/upstream_profile_remover.rb
+++ b/app/services/upstream_profile_remover.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+# Service for removing upstream profiles
+class UpstreamProfileRemover
+  class << self
+    def run!
+      Profile.canonical(false)
+             .joins(:parent_profile)
+             .joins(:test_results)
+             .where(parent_profiles_profiles: { upstream: true })
+             .distinct.destroy_all
+    end
+  end
+end
--- a/db/migrate/20211104070445_delete_upstream_profiles.rb
+++ b/db/migrate/20211104070445_delete_upstream_profiles.rb
@@ -1,0 +1,10 @@
+class DeleteUpstreamProfiles < ActiveRecord::Migration[5.2]
+  def up
+    UpstreamProfileRemover.run!
+    DanglingAccountRemover.run!
+  end
+
+  def down
+    # nop
+  end
+end
--- a/test/services/upstream_profile_remover_test.rb
+++ b/test/services/upstream_profile_remover_test.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class UpstreamProfileRemoverTest < ActiveSupport::TestCase
+  setup do
+    @account = FactoryBot.create(:account)
+    @parent = FactoryBot.create(:canonical_profile, upstream: true)
+    @upstream = FactoryBot.create(
+      :profile,
+      account: @account,
+      parent_profile: @parent
+    )
+    @downstream = FactoryBot.create(:profile, account: @account)
+  end
+
+  test 'removes upstream non-canonical profiles' do
+    host = FactoryBot.create(:host, account: @account.account_number)
+    FactoryBot.create(:test_result, profile: @upstream, host: host)
+    FactoryBot.create(:test_result, profile: @downstream, host: host)
+
+    assert_difference(
+      '@downstream.test_results.count' => 0,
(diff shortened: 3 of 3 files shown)


app/services/upstream_profile_remover.rb
```diff
@@ -0,0 +1,16 @@
...
+class UpstreamProfileRemover
+  class << self
+    def run!
+      policies = Profile.canonical(false)
+                        .joins(:parent_profile)
+                        .joins(:test_results)
+                        .where(parent_profiles_profiles: { upstream: true })
+                        .select(:policy_id).distinct
+
+      Policy.where(id: policies).destroy_all
```
vkrizan, line 13: Why are we removing policies? AFAIK this should only remove profiles and keep other policy profiles intacted. Policy can be removed if no other profile exists, which should happen automatically (if I'm not mistaken).

app/services/upstream_profile_remover.rb
```diff
@@ -0,0 +1,16 @@
...
+
+# Service for removing upstream profiles
+class UpstreamProfileRemover
+  class << self
+    def run!
+      policies = Profile.canonical(false)
+                        .joins(:parent_profile)
+                        .joins(:test_results)
+                        .where(parent_profiles_profiles: { upstream: true })
+                        .select(:policy_id).distinct
```
vkrizan, line 11:
```
SELECT DISTINCT "profiles"."policy_id"
FROM   "profiles"
       INNER JOIN "profiles" "parent_profiles_profiles"
               ON "parent_profiles_profiles"."id" =
                  "profiles"."parent_profile_id"
       INNER JOIN "test_results"
               ON "test_results"."profile_id" = "profiles"."id"
WHERE  "profiles"."parent_profile_id" IS NOT NULL
       AND "parent_profiles_profiles"."upstream" = true  
```

app/services/upstream_profile_remover.rb
skateman: A policy is always based on a canonical profile, if the canonical profile is upstream, all the policies underneath are upstream.

skateman: @vkrizan where can I verify this automatic deletion?

vkrizan: > A policy is always based on a canonical profile, if the canonical profile is upstream, all the policies underneath are upstream.

Not necessarily. Policy is an umbrella/parent of profiles. Whether an initial profile is or isn't coming out of upstream does not necessarily make the whole policy as upstream. Users might have had created a policy at the time where the latest benchmark was the one that we now mark as upstream, however they might have had uploaded with a proper benchmark&profile that is not upstream.

Picture this scenario:
```
Policy A
  -> profile A1 (initial, upstream)
  -> profile A2 (downstream; has test results)
```

In this case it would remove everything, even-though the policy has downstream profiles with results.
If we are going to remove just the upstream ones from the scenario and the profile is initial one, then we need to find/elect a new profile or find some other means how to resolve this situation.

Policy is based off of initial profile. Canonical profile only plays role to determine "policy type", but only out of the initial profile.

> where can I verify this automatic deletion?

https://github.com/RedHatInsights/compliance-backend/blob/cea75b77678f2cc8699cf1a679337a0be94ff9d1/app/models/concerns/profile_policy_association.rb#L9

skateman:
@vkrizan adjusted it, can you :eyes: again please?

vkrizan:
@skateman Looks better. It doesn't solve when the profile that is being deleted is the one that is marked as initial (`external: false`). If this profile is removed then the policy does not have an underlying profile to connect to. Please review scenario described at https://github.com/RedHatInsights/compliance-backend/pull/1055#discussion_r743493709.

skateman:
@vkrizan the query I sent you proves that the profiles are lonely under each policy, can I get a :+1: on this?

app/services/upstream_profile_remover.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Service for removing upstream profiles
+class UpstreamProfileRemover
+  class << self
+    def run!
+      Profile.canonical(false)
+             .joins(:parent_profile)
+             .joins(:test_results)
```
vkrizan, line 9: What do we need test results for?

skateman: That was the requirement from PM, check the jira ticket

vkrizan (approved):

PR #1058, opened 2021-11-11
fix(Clowder): Use $ghprbPullId for IMAGE_TAG

bastilian:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/unit_test.sh
+++ b/unit_test.sh
@@ -16,6 +16,11 @@ DATABASE_USER="compliance"
 DATABASE_PASSWORD="changeme"
 DATABASE_NAME="compliance-test"
 
+# if this is a PR, use a different tag, since PR tags expire
+if [[ ! -z "$ghprbPullId" ]]; then
+  export IMAGE_TAG="pr-${ghprbPullId}-${IMAGE_TAG}"
+fi
+
 function teardown_docker {
   docker rm -f "$DB_CONTAINER_NAME" || true
   docker rm -f "$TEST_CONTAINER_NAME" || true


unit_test.sh
```diff
@@ -16,6 +16,11 @@ DATABASE_USER="compliance"
 DATABASE_PASSWORD="changeme"
 DATABASE_NAME="compliance-test"
 
+# if this is a PR, use a different tag, since PR tags expire
+if [ ! -z "$ghprbPullId" ]; then
```
Victoremepunto, line 20: nitpick:

```suggestion
if [ -n "$ghprbPullId" ]; then
```

Victoremepunto (approved):
LGTM - if you want to accept one suggestion, other than that seems this should work

unit_test.sh
```diff
@@ -16,6 +16,11 @@ DATABASE_USER="compliance"
 DATABASE_PASSWORD="changeme"
 DATABASE_NAME="compliance-test"
 
+# if this is a PR, use a different tag, since PR tags expire
```
vkrizan, line 19: Is this something to make it more convenient for us or what does it fix?

unit_test.sh
```diff
@@ -16,6 +16,11 @@ DATABASE_USER="compliance"
...
+if [ -n "$ghprbPullId" ]; then
```
vkrizan, line 20: Can this be moved closer where the `IMAGE_TAG` is being set? There are other variables that depend on this value.

Nit: use bash test `[[ ]]`.

unit_test.sh
bastilian: Just copied and pasted what I was told to copy and paste: https://github.com/RedHatInsights/bonfire/blob/master/cicd/bootstrap.sh#L23

vkrizan: Okay. What it is good for?

unit_test.sh
bastilian: There are already two places where a `IMAGE_TAG` is defined, line 5 and 14. In between other vars use `IMAGE_TAG`. To which one of these should it be closer to?

unit_test.sh
Victoremepunto: The expiry date on Images tags has been there since always, the only thing that has changed is the format (for whatever reason). 

So ... before we had this 7 characters long image tag from the commit's HASH, and if it was created on a PR, the tag had the `expiry-date` metadata added to it so the PR image tags didn't stay forever in Quay. 

For some reason, someone maintaining the shared "build.sh" script from `bonfire` project changed the format too, being this `pr-1058-52cf345` as an example. So this change is just affecting us because we use the shared `build.sh` script from Bonfire repo ([this one](https://github.com/RedHatInsights/bonfire/blob/master/cicd/build.sh))

vkrizan: Thanks @Victoremepunto for explanation. Those are perks of depending on a dynamic code from main branch.

unit_test.sh
vkrizan: I'd move it right under line 6, after the initial `IMAGE_TAG` definition.

bastilian: @vkrizan But then it would be defined again on line 14 and probably cause issues again.

PR #1061, opened 2021-11-15
feat(Profiles): RHICOMPL-2399 Add scope returning latest profiles

marleystipich2:
…chmark

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

Adding a method that will return all latest profiles on a benchmark. The current endpoint only gets us the profiles from the latest benchmark, but there might be missing profiles that are only available in older ones. This method returns the latest versions of all profiles for a certain benchmark, even if they are not on the latest version of the benchmark. 

Currently, there are too many lines in the ProfileSearching Module: `Metrics/ModuleLength: Module has too many lines. [111/100]`.

### DIFF
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -135,6 +135,18 @@ module ProfileSearching
                    ::Xccdf::Benchmark.where.not(version: value)
       { conditions: benchmark_id.in(benchmarks.pluck(:id)).to_sql }
     end
+
+    def profiles_for_os_major(major_os_version)
+      ::SupportedSsg.sorted_ssgs_for_os_major(major_os_version)
+                    .each_with_object({}) do |supported_ssg, latest_profiles|
+        supported_ssg.profiles.keys.each do |profile|
+          next if latest_profiles.values.flatten.include?(profile)
+
+          latest_profiles[supported_ssg.version] ||= []
+          latest_profiles[supported_ssg.version] << profile
+        end
+      end
+    end
   end
 
   class_methods do
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -53,6 +53,15 @@ class Profile < ApplicationRecord
 
   scope :joins_policy, -> { left_outer_joins(:policy) }
 
+  scope :supported_for_os_major, lambda { |major|
+    Profile.profiles_for_os_major(major)
+           .inject(none) do |union, (version, profs)|
+      ref_ids = profs.map { |p| "xccdf_org.ssgproject.content_profile_#{p}" }
+      union.or(where(ref_id: ref_ids).ssg_versions(version)
+                                     .os_major_version(major))
+    end
+  }
+
   delegate :account_number, to: :account, allow_nil: true
 
   class << self
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -69,6 +69,11 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
       all.group_by(&:os_major_version)
     end
 
+    def sorted_ssgs_for_os_major(os_major_version)
+      for_os_major(os_major_version).sort_by(&:comparable_version)
+                                    .reverse
+    end
+
     private
 
     def map_attributes(rhel, raw_attrs)
@@ -79,6 +84,11 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
 
     def os_version(rhel)
       rhel.scan(/(\d+)\.(\d+)$/)[0]
+    end
+
+    def for_os_major(os_major_version)
+      os_major_version = os_major_version.to_s
+      all.select { |ssg| ssg.os_major_version == os_major_version }
     end
 
     def clear
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -615,6 +615,42 @@ class ProfileTest < ActiveSupport::TestCase
     assert_equal profile.short_ref_id, 'xccdf_org.ssgproject.profile'
   end
 
+  test 'supported_for_os_major' do
+    FactoryBot.create(:benchmark, version: '0.1.54', ref_id: 'RHEL-7')
+    FactoryBot.create(:benchmark, version: '0.1.36', ref_id: 'RHEL-7')
+    FactoryBot.create(:benchmark, version: '0.1.30', ref_id: 'RHEL-7')
+    Xccdf::Benchmark.all.each do |benchmark|
+      profile2 = FactoryBot.create(
+        :profile,
+        account: @account,
+        ref_id: 'xccdf_org.ssgproject.content_profile_stig',
+        os_major_version: 7
+      )
+      profile2.update!(benchmark_id: benchmark.id)
+      next if benchmark.version == '0.1.54' || benchmark.version == '0.1.36'
+
+      profile1 = FactoryBot.create(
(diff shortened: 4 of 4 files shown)


app/models/concerns/profile_searching.rb
```diff
@@ -86,6 +86,26 @@ module ProfileSearching
             .or(search.where(profile_cond))
             .distinct
     }
+    scope :profile_benchmark, lambda { |benchmark_ref_id|
+      joins(:benchmark).where(benchmarks: { ref_id: benchmark_ref_id })
+    }
+    scope :latest_benchmark_profiles, lambda { |profiles|
+      latest_p = []
+      profiles.find_each do |new_p|
```
vkrizan, line 94: Preliminary suggestion: Try making the scope to produce a query (another scope in a sense). Scope should return a scope that can be further used (and scoped again).

marleystipich2: Oh right, okay thank you!

vkrizan: Note, that it does not have to be a scope on Benchmark (or scope in general). It can be a regular class method. Reduction of number of queries can apply.

marleystipich2:
@skateman Made the changes that you suggested for the scope! The `ssg_version => [profile1, profile2]` hash is being created correctly but I might be able to simplify the scope a bit. I also still need to build the query that returns the profiles.

app/models/profile.rb
```diff
@@ -53,6 +53,18 @@ class Profile < ApplicationRecord
 
   scope :joins_policy, -> { left_outer_joins(:policy) }
 
+  scope :supported_for_os_major, lambda { |major|
+    latest_profiles = Hash.new { |h, k| h[k] = [] }
+    ::SupportedSsg.sorted_ssgs_for_os_major(major).each do |supported_ssg|
```
skateman, line 58: You can combine the two lines above into an `each_with_object`

skateman: But you would need below:
```ruby
latest_profiles[supported_ssg.version] ||= []
```

app/models/supported_ssg.rb
```diff
@@ -69,6 +69,14 @@ def by_os_major
       all.group_by(&:os_major_version)
     end
 
+    def sorted_ssgs_for_os_major(os_major_version)
+      os_major_version = os_major_version.to_s
+      ::SupportedSsg.all
```
skateman, line 74:
```suggestion
      all
```
Also you can move this into the private block

marleystipich2: I moved part of this method into the private block but left the sorting part out of the private block because I needed to call it from the Profile model. Does that sound right?

app/models/profile.rb
```diff
@@ -53,6 +53,18 @@ class Profile < ApplicationRecord
 
   scope :joins_policy, -> { left_outer_joins(:policy) }
 
+  scope :supported_for_os_major, lambda { |major|
+    ::SupportedSsg.sorted_ssgs_for_os_major(major)
+                  .each_with_object({}) do |supported_ssg, latest_profiles|
+      latest_profiles[supported_ssg.version] ||= []
```
skateman, line 59: Move this above line 63, we only need this when there's something to add.

app/models/profile.rb
```diff
@@ -53,6 +53,16 @@ class Profile < ApplicationRecord
 
   scope :joins_policy, -> { left_outer_joins(:policy) }
 
+  scope :supported_for_os_major, lambda { |major|
+    Profile.profiles_for_os_major(major)
+           .inject(none) do |union, (version, profs)|
+      ref_ids = []
+      profs.each { |p| ref_ids << "xccdf_org.ssgproject.content_profile_#{p}" }
```
skateman, line 60: You can just use a `map` :wink:

app/models/profile.rb
```diff
@@ -53,6 +53,16 @@ class Profile < ApplicationRecord
...
+      union.or(where(ref_id: ref_ids).ssg_versions(version)
```
skateman, line 61: Nice DRY :+1: it took me some time to catch it

skateman (changes requested):
Looks great, looking forward for the tests. Also I would not call it search in the title as it is not, rather something like *added a scope listing all supported profiles under an OS major`

app/models/concerns/profile_searching.rb
```diff
@@ -135,6 +135,18 @@ def ssg_version_search(_filter, operator, value)
...
     end
+
+    def profiles_for_os_major(major_os_version)
+      ::SupportedSsg.sorted_ssgs_for_os_major(major_os_version)
+                    .each_with_object({}) do |supported_ssg, latest_profiles|
+        supported_ssg.profiles.keys.each do |profile|
+          next if latest_profiles.values.flatten.include?(profile)
+
+          latest_profiles[supported_ssg.version] ||= []
+          latest_profiles[supported_ssg.version] << profile
```
vkrizan, line 146: Up until now the profiles listed on SupportedSsgs were informational. This creates a dependency on the list of profile names with Profile objects. The naming convention is a bit different (names are short, the objects are exact). There might be some name mismatch, therefore if we are going with this implementation, we would need to revise the SupportedSsg list/yaml.

Other option, that I'd favor more (if it does not have a flaw), is do a union of all *supported canonical* profiles across selected major version. Those would be distinct by `ref_name` where the one with higher SSG would be returned. The sorting might be a bit trickier to do in SQL.

@skateman wdyt?

skateman: @vkrizan this is not true, if you check [how the current endpoint works](https://github.com/RedHatInsights/compliance-backend/blob/67630b7dac18c2db7be451ce2ee94e24078f7d60/app/models/concerns/benchmark_searching.rb#L32-L39) you'll see a lot of similarities. The list yaml is generated from the datastreams, so we have no direct way to influence the content of the YAML. But for sure, we can do a check if there's any mismatch.

vkrizan: > this is not true, if you check [how the current endpoint works](https://github.com/RedHatInsights/compliance-backend/blob/67630b7dac18c2db7be451ce2ee94e24078f7d60/app/models/concerns/benchmark_searching.rb#L32-L39) you'll see a lot of similarities.

@skateman indeed the version matching for supported benchmarks works with SupportedSSGs; the comment addresses the use of `SupportedSsg.profiles` ([example](https://github.com/RedHatInsights/compliance-backend/blob/4fcf793ecea42f2bf8293effff746f9bc696d098/config/supported_ssg.default.yaml#L254-L273)) which are just being used by this implementation.

> The list yaml is generated from the datastreams, so we have no direct way to influence the content of the YAML. 

Nope, the list of profiles is handcrafted manually. We get it from OpenSCAP team (by `state.yaml`).

Definitely this approach gives us more control on what is shown. The drawback, however, is the human factor.

skateman: From slack:
```
SELECT * FROM profiles WHERE benchmark.supported = TRUE GROUP BY benchmark_version DESC;
```
pseudo-SQL :wink: 

I am leaning towards this solution as we would not need to do 3 queries on our own, but GQL can self-optimize it.

PR #1062, opened 2021-11-18
feat(sorting): RHICOMPL-2461 enable sorting hosts by SSG version

skateman:


app/models/host.rb
```diff
@@ -11,6 +11,8 @@ class Host < ApplicationRecord
   sortable_by :name, :display_name
   sortable_by :os_major_version, OS_MAJOR_VERSION
   sortable_by :os_minor_version, OS_MINOR_VERSION
+  sortable_by :ssg_version, Arel.sql('benchmarks.version'),
```
vkrizan, line 14: Would this sort naturally?

skateman: Good catch, it doesn't inherit from the benchmarks model (would be nice) ... fixed :wrench:

app/models/host.rb
```diff
@@ -30,6 +35,19 @@ class Host < ApplicationRecord
   has_many :assigned_internal_profiles, -> { external(false) },
            through: :policies, source: :profiles
 
+  scope :with_benchmark, lambda { |profile = nil|
+    profile ||= RequestStore.store['scoped_search_context_profiles']
+
+    joins(:policies).joins(
+      <<-SQL
+      INNER JOIN "profiles" ON "profiles"."policy_id" = "policies"."id" AND
+      "profiles"."os_minor_version" = ("hosts".#{OS_MINOR_VERSION})::text
```
vkrizan, line 44: Can you please elaborate on the condition for OS minor version? 
It might not be necessary. When sorting by SSG version, it should sort by last test result's benchmark scoped by profile context.

app/models/host.rb
```diff
@@ -30,6 +35,14 @@ class Host < ApplicationRecord
   has_many :assigned_internal_profiles, -> { external(false) },
            through: :policies, source: :profiles
 
+  scope :with_benchmark, lambda { |profile = nil|
+    profile ||= RequestStore.store['scoped_search_context_profiles']
+
+    left_outer_joins(test_result_profiles: :benchmark).where(
```
skateman, line 41: @vkrizan something like this?

vkrizan: Yes. It just needs to join with exactly one test result, benchmark and profile: with latest test result for the given profile scope.

The `scoped_search_context_profiles` would likely be the initial profile or any profile for a given policy, wouldn't it?

skateman: It's all the child profiles under the policy

skateman:
@bastilian can you please verify if this works as it should before I squash?

bastilian (approved):
LGTM! Thank you @skateman!

app/models/host.rb
vkrizan: The "It just needs to join with exactly one test result, benchmark and profile: with latest test result for the given profile scope." is not resolved, yet.

skateman: @vkrizan after our discussion with @LightOfHeaven1994 are you okay with this?

vkrizan: Yes, since the system keeps one test result for each policy for a host, this implementation make sense.

PR #1066, opened 2021-11-23
fix(clowder): RHICOMPL-2473 open the kafkaCaFile for write only

skateman:
I guess the original author wanted text mode for writing but that's redundant in Ruby.

### DIFF
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -30,7 +30,7 @@ if ClowderCommonRuby::Config.clowder_enabled?
   if kafka_security_protocol
     if kafka_security_protocol == 'sasl'
       kafka_server_config[:security_protocol] = 'sasl_ssl'
-      kafkaCaFile = Tempfile.create(mode: 'wt')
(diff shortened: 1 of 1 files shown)


config/initializers/0_clowder-config.rb
```diff
@@ -30,7 +30,7 @@
   if kafka_security_protocol
     if kafka_security_protocol == 'sasl'
       kafka_server_config[:security_protocol] = 'sasl_ssl'
-      kafkaCaFile = Tempfile.create(mode: 'wt')
+      kafkaCaFile = Tempfile.create(mode: File::WRONLY)
```
vkrizan, line 33: It might need `CREAT` and `EXCL`.
```
File::WRONLY|File::CREAT|File::EXCL
```
But I guess it might be left as default (with read/write).

skateman: No https://github.com/ruby/ruby/blob/master/lib/tempfile.rb#L352

vkrizan (approved):

PR #1068, opened 2021-11-26
feat(Benchmarks): RHICOMPL-2399 all supported profiles for OS major

skateman:
Replaces https://github.com/RedHatInsights/compliance-backend/pull/1061

https://www.postgresql.org/docs/14/sql-select.html#SQL-DISTINCT

### DIFF
--- a/app/models/xccdf/benchmark.rb
+++ b/app/models/xccdf/benchmark.rb
@@ -30,6 +30,19 @@ module Xccdf
     end
     alias os_major_version inferred_os_major_version
 
+    # Returns all supported profiles for a given OS major version
+    # through a benchmark referencing this OS major
+    def supported_profiles
+      versions = SupportedSsg.by_os_major[os_major_version].map(&:version)
+
+      Profile.canonical.joins(:benchmark)
+             .where(benchmarks: { ref_id: ref_id, version: versions })
+             .order(:ref_id, Arel.sql('
+                string_to_array("benchmarks"."version", \'.\')::int[] DESC
+              '))
+             .select('DISTINCT ON ("profiles"."ref_id") "profiles".*')
+    end
+
     class << self
       def from_openscap_parser(op_benchmark)
         benchmark = find_or_initialize_by(
--- a/test/models/benchmark_test.rb
+++ b/test/models/benchmark_test.rb
@@ -451,5 +451,37 @@ module Xccdf
         assert_equal 1, returned.count
       end
     end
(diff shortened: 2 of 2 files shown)


app/models/concerns/benchmark_searching.rb
```diff
@@ -104,6 +104,17 @@ def supported_os_minor_search(_filter, _operator, value)
...
 
+    # Returns all supported profiles for a given OS major version
+    # through a benchmark referencing this OS major
+    def supported_profiles
+      Profile.canonical.joins(:benchmark)
+             .where(benchmarks: { ref_id: bm.ref_id })
+             .order(:ref_id, Arel.sql('
+                string_to_array("benchmarks"."version", \'.\')::int[] DESC
+              '))
+             .select('DISTINCT ON ("profiles"."ref_id") "profiles".*')
```
vkrizan, line 115: could work with `.distinct(:ref_id)`

skateman: It doesn't :disappointed:

skateman:
@marleystipich2 this is basically doing the same as your PR, but addresses the possible issues that @vkrizan pointed out and it's a bit more efficient. Are you okay with using this one instead, or would you favor your PR? In that case we might need to add additional safeguards that check if there are no broken links in the support matrix.

vkrizan (commented):
LGTM.

Lacks tests and use. =)

app/models/concerns/benchmark_searching.rb
```diff
@@ -104,6 +104,17 @@ def supported_os_minor_search(_filter, _operator, value)
         parameter: parameters }
     end
 
+    # Returns all supported profiles for a given OS major version
+    # through a benchmark referencing this OS major
+    def supported_profiles
+      Profile.canonical.joins(:benchmark)
+             .where(benchmarks: { ref_id: bm.ref_id })
```
skateman, line 111: Are all benchmarks we store supported? Or should it also have some kind of safeguard based on the versions in SupportedSsg for that given OS major.

skateman: @vkrizan :point_up:

vkrizan: The latter. Supported benchmarks are those listed in SupportedSsg.

marleystipich2 (approved):
Works great for me. Thanks @skateman!

PR #1070, opened 2021-11-30
fix(racecar): RHICOMPL-2496 explicit symbolization for config

skateman:
The `config.security_protocol` should and will be a symbol...

### DIFF
--- a/config/racecar.rb
+++ b/config/racecar.rb
@@ -3,9 +3,9 @@ Racecar.configure do |config|
   config.group_id_prefix = 'compliance'
 
   if Settings.kafka.security_protocol
-    config.security_protocol = Settings.kafka.security_protocol
+    config.security_protocol = Settings.kafka.security_protocol.try(:to_sym)
 
-    if config.security_protocol == 'sasl_ssl'
+    if config.security_protocol == :sasl_ssl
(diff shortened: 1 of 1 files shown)


config/racecar.rb
```diff
@@ -3,9 +3,9 @@
   config.group_id_prefix = 'compliance'
 
   if Settings.kafka.security_protocol
-    config.security_protocol = Settings.kafka.security_protocol
+    config.security_protocol = Settings.kafka.security_protocol.to_sym
 
-    if config.security_protocol == 'sasl_ssl'
+    if config.security_protocol == :sasl_ssl
```
vkrizan, line 8: Oneline Alternative: 
```
if Settings.kafka.security_protocol == 'sasl_ssl'
```
Clowder values are usually strings, unless our Settings would cast them (in future).

config/racecar.rb
```diff
@@ -3,9 +3,9 @@
   config.group_id_prefix = 'compliance'
 
   if Settings.kafka.security_protocol
-    config.security_protocol = Settings.kafka.security_protocol
+    config.security_protocol = Settings.kafka.security_protocol.to_sym
```
vkrizan, line 6: Is the `to_sym` safe for `nil` value?

vkrizan (approved):

PR #1071, opened 2021-11-30
feat(GraphQL): RHICOMPL-2485 expose Benchmark#supported_profiles

skateman:


app/graphql/schema.graphql
```diff
@@ -10,6 +10,7 @@ type Benchmark implements Node {
   profiles: [Profile!]
   refId: String!
   rules: [Rule!]
+  supportedProfiles: [Profile!]
```
vkrizan, line 13: Should this be an instance/object property? Shouldn't this be global (i.e. a separate query)?

vkrizan: Second issue: it returns canonical profiles only. It might be named accurately, e.g. `supportedCanonicalProfiles`.

skateman: Your proposal makes sense, I just wanted to have it simpler here, but I can change it.

Well, `profiles` also returns just canonical profiles.

app/graphql/types/benchmark.rb
```diff
@@ -15,10 +15,15 @@ class Benchmark < Types::BaseObject
...
     field :latest_supported_os_minor_versions, [String], null: false
     field :profiles, [::Types::Profile], null: true
+    field :supported_profiles, [::Types::Profile], null: true
     field :rules, [::Types::Rule], null: true
 
     def profiles
       object.profiles.canonical
     end
+
+    def supported_profiles
```
vkrizan, line 25: Can it not be on a Benchmark instance at all?

app/graphql/schema.graphql
vkrizan: Can it be moved to a standalone query, like the `latest_benchmarks` are?
https://github.com/RedHatInsights/compliance-backend/blob/f9fadd1faf51425156d693a16c358c24b825095c/app/graphql/types/query.rb#L63

vkrizan (commented):
Noice. :tophat:

app/policies/os_major_version_policy.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def show?
+    true
+  end
+
+  # All users should see all benchmarks currently
+  class Scope < ::ApplicationPolicy::Scope
+    def resolve
+      scope.all
```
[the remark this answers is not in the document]
skateman: It has the same scoping as the Benchmarks...

vkrizan (approved):
Seal of Approval.

PR #1073, opened 2021-12-03
feat(Benchmark): RHICOMPL-1896 allow reimports for backport versions

skateman:
I'm adding a new column for benchmarks that stores the SSG package name, if there's a difference between the imported vs current, the benchmark should be reimported on its own.

### DIFF
--- a/app/services/concerns/xccdf/benchmarks.rb
+++ b/app/services/concerns/xccdf/benchmarks.rb
@@ -7,12 +7,15 @@ module Xccdf
 
     included do
       def save_benchmark
-        ::Xccdf::Benchmark.import!([benchmark].select(&:new_record?),
-                                   ignore: true)
+        benchmark.package_name = package_name
+
+        return unless benchmark.new_record? || benchmark.package_name_changed?
+
+        benchmark.save!
       end
 
       def benchmark_saved?
-        benchmark.persisted?
+        benchmark.package_name == package_name && benchmark.persisted?
       end
 
       def benchmark_profiles_saved?
@@ -31,6 +34,14 @@ module Xccdf
 
       def benchmark
         @benchmark ||= ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
+      end
+
+      def package_name
+        @package_name ||= begin
+          SupportedSsg.by_os_major[benchmark.os_major_version].find do |item|
+            item.version == benchmark.version
+          end&.package
+        end
       end
     end
   end
--- a/db/migrate/20211203100649_add_package_name_to_benchmarks.rb
+++ b/db/migrate/20211203100649_add_package_name_to_benchmarks.rb
@@ -1,0 +1,9 @@
+class AddPackageNameToBenchmarks < ActiveRecord::Migration[5.2]
+  def up
+    add_column :benchmarks, :package_name, :string
+  end
+
+  def down
+    remove_column :benchmarks, :package_name
+  end
+end
--- a/db/migrate/20211211074855_clear_revisions.rb
+++ b/db/migrate/20211211074855_clear_revisions.rb
@@ -1,0 +1,9 @@
+class ClearRevisions < ActiveRecord::Migration[5.2]
+  def up
+    Revision.find_by(name: 'datastreams')&.delete
+  end
+
+  def down
+    #nop
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_12_131735) do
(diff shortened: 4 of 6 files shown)


app/services/concerns/xccdf/benchmarks.rb
```diff
@@ -7,12 +7,13 @@ module Benchmarks
...
     included do
       def save_benchmark
+        benchmark.package_name = package_name
         ::Xccdf::Benchmark.import!([benchmark].select(&:new_record?),
                                    ignore: true)
       end
 
       def benchmark_saved?
-        benchmark.persisted?
+        benchmark.package_name == package_name && benchmark.persisted?
```
skateman, line 16: Not sure if this change is actually needed :see_no_evil:

db/migrate/20211203100649_add_package_name_to_benchmarks.rb
```diff
@@ -0,0 +1,16 @@
+class AddPackageNameToBenchmarks < ActiveRecord::Migration[5.2]
+  def up
+    add_column :benchmarks, :package_name, :string
+
+    # Update already imported Benchmarks with their package names
+    SupportedSsg.all.each do |ssg|
+      ref_id = "#{Xccdf::Benchmark::REF_PREFIX}-#{ssg.os_major_version}"
+      Xccdf::Benchmark.where(ref_id: ref_id, version: ssg.version)
+                      .update(package_name: ssg.package)
```
vkrizan, line 9: @skateman how do you feel about trying it out with reimport. Empty packages would cause a reimport which should set the correct package on benchmark objects. This way we should be sure that what we have imported is actually what is coming out of package that is (going to be) set.
Although, there a potential risk of this deadlock, as it it would/should be have as a force import.

app/services/concerns/xccdf/benchmarks.rb
```diff
@@ -32,6 +33,14 @@ def benchmark_contents_equal_to_op?
       def benchmark
         @benchmark ||= ::Xccdf::Benchmark.from_openscap_parser(@op_benchmark)
       end
+
+      def package_name
+        @package_name ||= begin
+          SupportedSsg.by_os_major[benchmark.os_major_version].find do |item|
+            item.version == benchmark.version
+          end.package
```
vkrizan, line 41: Safe navigation?

vkrizan (commented):
Besides the one issue find it looks good code-wise.

marleystipich2 (approved):
This looks good to me. I went through code, including the test updates and everything looks good. Thanks, @skateman!

app/services/concerns/xccdf/benchmarks.rb
```diff
@@ -7,12 +7,15 @@ module Benchmarks
 
     included do
       def save_benchmark
-        ::Xccdf::Benchmark.import!([benchmark].select(&:new_record?),
-                                   ignore: true)
+        benchmark.package_name = package_name
+
+        return unless benchmark.new_record? || benchmark.package_name_changed?
+
+        benchmark.save!
```
skateman, line 14: @marleystipich2 I think the `import!` was an overkill for conditionally inserting one record and it was not really respecting the update of the package name. So I am changing it to a conditional save if the record is new OR the package name is being changed.

vkrizan: @skateman are we sure that the `benchmark.new_record?` condition would cover previous `ignore: true` to prevent unique constraint failure?

vkrizan (approved):
Code-wise LGTM. 

One comment down below.

app/services/concerns/xccdf/benchmarks.rb
skateman: I don't see why it would behave differently from its [documentation](https://apidock.com/rails/ActiveRecord/Base/new_record%3F) and the `from_openscap_parser` method uses `find_or_initialize_by`

vkrizan: If `find_or_initialize_by` is used then all is fine, unless it is created concurrently in between. `#new_record?` does not check object existence within db or does it? A new conflicting model instance can be initialized (having `new_record?` true) and then tried to be saved.

skateman: Well, I don't see how the import would help us with the package name :confused: maybe I could remove the `!` from the `save` that would ensure integrity, or catch the duplicate exception and ensure that the package name gets updated anyway.

vkrizan: Maybe an upsert? Anyhow, I guess it is good as it is. The concurrency might be ignored, for now.

skateman: Right, we have rails6 :see_no_evil:

skateman: Okay, keeping it as it is.

PR #1075, opened 2021-12-08
Rails 6 upgrade

skateman:
Recycling  #702 from @akofink  with a few adjustments. Note that batch enumerations on tables without a primary key no longer work, so I had to get rid of it from one migration.


config/environments/production.rb
```diff
@@ -82,6 +82,30 @@
     config.audit_logger = Logger.new(STDOUT)
   end
 
+  # Whitelist hosts
+  config.hosts << /((ci|qa)\.)?cloud\.(stage\.)?redhat\.com/
```
bastilian, line 86: `cloud` should probably be `console`. 

I wonder could this cause any issues in production behind a reverse-proxy/ingress?

skateman: Yes it can :disappointed:

config/environments/production.rb
```diff
@@ -82,6 +82,30 @@
     config.audit_logger = Logger.new(STDOUT)
   end
 
+  # Whitelist hosts
+  config.hosts << /((ci|qa)\.)?console\.(stage\.)?redhat\.com/
```
vkrizan, line 86: This needs more coverage for Ephemeral Envs. Currently the hostname that is used there is smth like `front-end-aggregator-ephemeral-NN.apps.c-rh-c-eph.HASH.p1.openshiftapps.com`. But that can change :disappointed: .

db/migrate/20210112221033_add_foreign_key_constraint_to_active_storage_attachments_for_blob_id.active_storage.rb
```diff
@@ -0,0 +1,10 @@
+# This migration comes from active_storage (originally 20180723000244)
+class AddForeignKeyConstraintToActiveStorageAttachmentsForBlobId < ActiveRecord::Migration[6.0]
+  def up
+    return if foreign_key_exists?(:active_storage_attachments, column: :blob_id)
+
+    if table_exists?(:active_storage_blobs)
```
vkrizan, line 6: I'm not aware of this table existing, so probably the migration isn't needed (?).

config/environments/production.rb
bastilian: How important is this to upgrade to rails 6? Could we disable it or just set a wildcard to match any hostname?

skateman: I don't think that's a wise decision, I would rather go with a more universal wildcard that would also allow something like `*ephemeral*.openshiftapps.com`

bastilian: 👍 It would also make sense to expose this setting to set an `ENV` var. ...Is there maybe a hostname ENV var already that could be used?

skateman: Yeah, thinking about that too.

vkrizan: How about: `.+.apps\.c-rh-c-eph\.(\w+)\.p1\.openshiftapps\.com$`? Instead of `(\w+)` we can use concrete cluster hash to make it more secure.

PR #1076, opened 2021-12-10
fix(DanglingAccountRemover): RHICOMPL-2537 do not depend on cyndi

skateman:
Sometimes cyndi is not available on initial setups and in this case not even needed (ephemeral, development, ci) and it fails on this migration. I'm adding a safeguard to return with an empty relation so the migration can continue. If there are no inventory hosts, this service should not do anything anyway.

### DIFF
--- a/app/services/dangling_account_remover.rb
+++ b/app/services/dangling_account_remover.rb
@@ -32,7 +32,12 @@ class DanglingAccountRemover
     end
 
     def accounts_with_hosts
-      Host.with_policies_or_test_results.select(:account).distinct.arel
+      if ApplicationRecord.connection.data_source_exists?(Host.table_name)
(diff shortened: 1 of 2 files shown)


app/services/dangling_account_remover.rb
```diff
@@ -32,7 +32,12 @@ def accounts_with_profiles_or_policies
     end
 
     def accounts_with_hosts
-      Host.with_policies_or_test_results.select(:account).distinct.arel
+      if ApplicationRecord.connection.data_source_exists?(Host.table_name)
```
vkrizan, line 35: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-data_source_exists-3F

vkrizan (approved):
LGTM

PR #1077, opened 2021-12-11
Rails 6 upgrade

skateman:
Recycling #702 from @akofink with a few adjustments. Note that batch enumerations on tables without a primary key no longer work, so I had to get rid of it from one migration.

As a last resort to fix the codecov issues, I'm replacing #1075 with this PR :disappointed:


LightOfHeaven1994 (approved):
if smoke tests pass, everything works as expected

PR #1080, opened 2021-12-15
fix(GraphQL): RHICOMPL-2257 narrow down querying for profiles/hosts

skateman:
Making the host lookup in mutations analogous to profile lookups and narrowing down profile querying in policy creation for speed.

### DIFF
--- a/app/graphql/mutations/profile/associate_systems.rb
+++ b/app/graphql/mutations/profile/associate_systems.rb
@@ -13,9 +13,10 @@ module Mutations
 
       def resolve(args = {})
         ::Profile.transaction do
+          hosts = find_hosts(args[:system_ids])
           profile = find_profile(args[:id])
           if profile&.policy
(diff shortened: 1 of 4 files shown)


app/graphql/mutations/profile/associate_systems.rb
```diff
@@ -13,9 +13,10 @@ class AssociateSystems < BaseMutation
 
       def resolve(args = {})
         ::Profile.transaction do
+          hosts = find_hosts(args[:system_ids])
           profile = find_profile(args[:id])
           if profile&.policy
-            profile.policy.update_hosts(args[:system_ids])
+            profile.policy.update_hosts(hosts.pluck(:id))
```
marleystipich2, line 19: Question about this change.. Is this more efficient than the way it was before? Just curious

marleystipich2 (approved):
These changes look good! This was a good idea. Thanks, @skateman!

PR #1081, opened 2021-12-15
feat(http): RHICOMPL-2389 add CSP/HSTS headers for the API

skateman:
API doesn't really like them, just to have the symmetry with the frontend


bastilian (approved):
LGTM!

PR #1082, opened 2021-12-15
cleanup(GraphQL): RHICOMPL-2256 drop the slow allProfiles query

skateman:
This query is no longer being used...


marleystipich2 (commented):
This looks good to me! It just looks like you need to update the tests.

marleystipich2 (approved):
Looks great! Thanks, @skateman!

PR #1089, opened 2021-12-22
fix(clowder): RHICOMPL-2553 livenessProbe timeout for sidekiq

skateman:
Racecar had it, I don't see why sidekiq shouldn't. Hopefully it fixes the livenessProbe triggered restarts under heavy loads.


vkrizan (approved):
lgtm

PR #1097, opened 2022-01-06
feat(clowder): RHICOMPL-2568 add tracing for readiness probes

skateman:
Rake task output looks better with `--trace` which was missing from two readiness probes (sidekiq, racecar).


LightOfHeaven1994 (approved):
LGTM :+1:

marleystipich2 (approved):
Also looks good to me!

PR #1098, opened 2022-01-07
fix(rails): RHICOMPL-2569 fixed db setup in development for rails61

skateman:
The `bin/rails db` command now required a `-p` argument if we want to avoid interactive password input (idk why) and I also added some tracing to the rake tasks in the compose file for better debugging.


LightOfHeaven1994 (approved):
LGTM

PR #1099, opened 2022-01-07
chore(rails6): RHICOMPL-2540 get rid of the remaining deprecations

skateman:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/


marleystipich2 (approved):
Looks good to me! Thanks, @skateman!

PR #1101, opened 2022-01-10
feat(notifications): RHICOMPL-2565 implement kafka producer

skateman:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/producers/notification.rb
+++ b/app/producers/notification.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# A Kafka producer client for notifications
+class Notification < ApplicationProducer
+  TOPIC = Settings.kafka_producer_topics.notifications
+  BUNDLE = 'rhel'
+  VERSION = 'v1.1.0'
+
+  # rubocop:disable Metrics/MethodLength
+  def self.deliver(event_type:, account:, host:, policy:, **extra)
+    payload = {
+      host_id: host.id,
+      host_name: host.display_name,
+      policy_id: policy.id,
+      policy_name: policy.name
+    }.merge(extra)
+
+    msg = {
+      version: VERSION,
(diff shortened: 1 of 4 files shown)


marleystipich2 (approved):
Looks good to me. Thanks, @skateman!

app/producers/notification.rb
```diff
@@ -0,0 +1,37 @@
...
+      bundle: BUNDLE,
+      application: SERVICE,
+      event_type: event_type,
+      timestamp: DateTime.now.iso8601,
+      account_id: account.account_number,
+      events: [{
+        metadata: {},
+        payload: payload
+      }],
+      context: payload
```
vkrizan, line 29: @skateman should the context be the same as payload? A context is usually a resource that is getting the notification, i.e. a host. Payload then has event data.

skateman: Context is not a mandatory field, payload is...but drift was consuming data from the context, probably because it's easier to read (no need to iterate it). I can change it in a followup, but I think it's safer to get it from the context than iterate in the template and have weird results if there are multiple payloads in the notification :confused:

vkrizan: Oh. It seems easier to iterate, but it is not the right place for payload data to be read from (context). To my understanding context determines what is getting an event, i.e. shared context between payloads. The array of events can be used also in our case, when a system is assigned to more than one policy and it's not compliant for at least two of them.

@pilhuhn might provide more details on the event production. I've tried to take it from the consumer stand-point.

PR #1102, opened 2022-01-10
fix(hosts): RHICOMPL-2570 Fix ssg version sorting

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/concerns/host_searching.rb
+++ b/app/models/concerns/host_searching.rb
@@ -74,10 +74,7 @@ module HostSearching
     end
 
     def filter_with_results_for_policy(_filter, _operator, policy_or_profile_id)
-      profiles = ::Profile.where(id: policy_or_profile_id)
-      profiles = profiles
-                 .or(::Profile.where(policy_id: policy_or_profile_id))
-                 .or(::Profile.where(policy_id: profiles.select(:policy_id)))
+      profiles = profiles_by_id(policy_or_profile_id)
 
       RequestStore.store['scoped_search_context_profiles'] = profiles
 
@@ -87,6 +84,10 @@ module HostSearching
     end
 
     def filter_by_policy(_filter, _operator, policy_or_profile_id)
+      profiles = profiles_by_id(policy_or_profile_id) + Profile.select('NULL AS id').distinct
+
+      RequestStore.store['scoped_search_context_profiles'] = profiles
+
       with_policy = with_policy_lookup(policy_or_profile_id).select(:id)
       with_profile = with_external_profile_lookup(policy_or_profile_id)
                      .select(:id)
@@ -167,6 +168,12 @@ module HostSearching
           }
         }
       )
+    end
+
+    def profiles_by_id(policy_or_profile_id)
+      profiles = ::Profile.where(id: policy_or_profile_id)
+      profiles.or(::Profile.where(policy_id: policy_or_profile_id))
+              .or(::Profile.where(policy_id: profiles.select(:policy_id)))
     end
   end
 


app/models/host.rb
```diff
@@ -35,12 +35,8 @@ class Host < ApplicationRecord
   has_many :assigned_internal_profiles, -> { external(false) },
            through: :policies, source: :profiles
 
-  scope :with_benchmark, lambda { |profile = nil|
-    profile ||= RequestStore.store['scoped_search_context_profiles']
-
-    left_outer_joins(test_result_profiles: :benchmark).where(
-      profiles: { id: profile.pluck(:id) }
-    )
```
marleystipich2, line 43: @skateman The issue with the sorting of the ssg version was caused by `profile` being nil. I tried to figure out how to set the profile correctly but then after looking at it for a little while, I was wondering if the last part of the query was necessary? It seems like `left_outer_joins(test_result_profiles: :benchmark)` might be sufficient? Or maybe I'm not understanding what `test_result_profiles` is exactly?
It's hard to test because the ssg versions are showing up as "Not available" on this table.. I asked Egor if he noticed the same thing when he was testing

skateman: @marleystipich2 your approach would only work if one host could be only assigned to a single profile and so have one test result and one benchmark. However, we need the profile(s) as a parameter to distinguish which policy we're underneath at the moment. If you delete the `profile` variable, it will also break the SSG sorting on a report's page.

So in order to have this working, you need to set the `profile` variable that's coming from the `RequestStore`. 

I'd suggest you to [look](https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/concerns/host_searching.rb#L82) where and how the `RequestStore.store['scoped_search_context_profiles']` is being set when opening a report's page.Then you should figure out why this is not being called when opening a policy/systems page.

marleystipich2: @skateman Okay, right I figured that that was the reason that checking the profile was necessary. This make sense! 
I'll look into why `RequestStore.store['scoped_search_context_profiles']` isn't being set correctly when you open the policy/systems page. I was looking at the method where it's being set and noticed that it wasn't being called when you're at this page. Thanks for the help!!

app/models/concerns/host_searching.rb
```diff
@@ -87,6 +90,10 @@ def filter_with_results_for_policy(_filter, _operator, policy_or_profile_id)
     end
 
     def filter_by_policy(_filter, _operator, policy_or_profile_id)
+      profiles = profiles_by_id(policy_or_profile_id).to_a.push(Profile.new)
```
marleystipich2, line 93: @skateman This did end up working but I had to do `to_a` before I pushed the new profile(apparently pushing directly onto an activerecord relation was possible before Rails 5 but now you have to convert it to an array). I also had to move the set up of `profiles` into a scope to fix a few rubocop issues. 
Does this look good to you? I left the other solution in `with_benchmark` in case we wanted to go back to that one but I'll remove that if we decide this new solution looks good.

skateman: I looked into the `ActiveRecord::Relation` and as it is an `Enumerable` it inherently supports the `+` operation, which probably also converts it to an array, but it seems more elegant and readable. I'm not sure if it will like a `+` with a `Profile` class, but you could have another relation to work with by tricking rails of creating a record with `nil` as its ID. The simplest way I think is `Profile.select('NULL AS id').distinct` that should return you a relation having exactly one profile with `nil` as its id. I am 100% sure you could do a `+` operation between two relations :wink: 

Regarding the scope, I'd rather have it as a method, ideally a private one, but maybe after my proposal, Rubocop will stop complaining.

marleystipich2: The `+` operation worked well with `Profile.select('NULL AS id').distinct` 👍  I also moved the set up of `profiles` into a private method instead of a scope.

PR #1113, opened 2022-01-17
dev(compose): improve dependency management in the compose file

skateman:
Renaming the `import-ssg` job to follow the clowder namings. Also adjusting some of the dependencies for smoother setup.


marleystipich2 (approved):
Looks good! Thanks, @skateman!

PR #1118, opened 2022-01-19
fix(Clowder): RHICOMPL-2579 Disable liveness/readiness check for jobs

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -76,6 +76,12 @@ objects:
           value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
         - name: SETTINGS__FORCE_IMPORT_SSGS
           value: "${SETTINGS__FORCE_IMPORT_SSGS}"
+        livenessProbe:
+          exec:
+            command:
+              - bash
+              - -c
+              - ps ax | grep [r]ake
     - name: floorist
       schedule: ${FLOORIST_SCHEDULE}
       suspend: ${{FLOORIST_SUSPEND}}
@@ -112,6 +118,12 @@ objects:
           value: "/tmp/floorplan/floorplan.yaml"
         - name: LOGLEVEL
           value: ${FLOORIST_LOGLEVEL}
+        livenessProbe:
+          exec:
+            command:
+              - bash
+              - -c
+              - ps ax | grep [r]un
         volumeMounts:
         - name: floorplan-volume
           mountPath: "/tmp/floorplan"


marleystipich2:
@skateman @vkrizan AppSRE said that it's a best practice to have a liveness probe for jobs but it's okay not to disable the readiness probe. 
Should I just do this command for the liveness check like David mentioned the other day? `ps ax | grep import`

skateman:
For ssg-import: `ps ax | grep [r]ake` - you need the `[]` so the ps process doesn't find itself, try it without and you'll understand :wink: 

For floorist: `ps ax | grep [r]un` - assuming from [here](https://github.com/RedHatInsights/floorist/blob/master/Dockerfile#L19), we might need to test this in ephemeral cc @vkrizan

vkrizan:
Rake is the main process of the container. I do not think that checking its existence does any benefit. It is bound to the container, so whenever it terminates the container will terminate with it. Liveness probe could rather check progress of the jobs, checking that they are not stuck waiting indefinitely.

clowdapp.yaml
```diff
@@ -38,6 +38,8 @@ objects:
     - name: import-ssg
       schedule: "*/5 * * * *" # every 5 minutes
       concurrencyPolicy: Forbid
+      annotations:
+        ignore-check.kube-linter.io/no-readiness-probe: "This job does not need a readiness probe because it does not handle incoming traffic."
```
vkrizan, line 42: Do you have a link to docs on the `ignore-check.kube-linter.io/no-readiness-probe` annotation at hand? Thanks.

skateman: > Rake is the main process of the container. I do not think that checking its existence does any benefit. It is bound to the container, so whenever it terminates the container will terminate with it. Liveness probe could rather check progress of the jobs, checking that they are not stuck waiting indefinitely.

@vkrizan we can't really do that easily :laughing: that's why I proposed what I proposed to go around this. We could cook up some DRB magic but I don't really see the point :disappointed:

marleystipich2: Yeah, here's the link that Luis sent me (https://www.civo.com/learn/yaml-best-practices-using-kubelinter#ignoring-a-check)

I was going to ask if you guys thought this was correct or not too. It's a little confusing to test

clowdapp.yaml
```diff
@@ -8,6 +8,8 @@ objects:
   kind: ClowdApp
   metadata:
     name: "${APP_NAME}"
+    annotations:
+      ignore-check.kube-linter.io/no-readiness-probe
```
marleystipich2, line 12: @vkrizan @skateman After looking at [this article](https://www.civo.com/learn/yaml-best-practices-using-kubelinter#ignoring-a-check) closer, I think that the `annotations` need to be under `metadata` like this. I'm not sure how to only ignore the checks for the jobs and not the deployments. Maybe I can add `metadata:` under `jobs:` and then put this annotation in there? 

The article that I attached here is the one that AppSRE sent me, by the way.

vkrizan: This is starting to be tricky with the ClowdApp object. This annotation would (likely) suppress it for whole ClowdApp. Unfortunately there are no metadata nor annotation parameters on [Clowder's Job](https://github.com/RedHatInsights/clowder/blob/41827e006117220fb7e10ed04bd57343ebf3fadf/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go#L76-L115) and neither on [Clowder's PodSpec](https://github.com/RedHatInsights/clowder/blob/41827e006117220fb7e10ed04bd57343ebf3fadf/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go#L206-L248). 

This leaves us four options:
1) leave it as you written it - which might not be acceptable,
2) write some readiness probe,
3) don't do any changes and ignore the issue,
4) ask platform to add metadata&annotations to `cloud.redhat.com/v1alpha1/PodSpec`.

I'd like to understand where this problem is coming from. Whether AppSRE has some guidance that a template has to comply with some linter. I'll ask them on the Jira.

marleystipich2: Right, I think it will correctly disable the check if we leave it how I have it but like you said, it will probably suppress it for everything. Maybe that could be okay, at least as a short term solution, since we know that we have the checks passing for everything else besides these few failing checks?
I also have a [separate ticket open](https://github.com/RedHatInsights/compliance-backend/pull/1120) for a similar issue where I need to suppress the min replica check for `compliance-ssg-service` and I don't know what a good workaround for that one would be besides disabling the alert like this

vkrizan: The linter warning suppression for everything might be okay. Try reaching out to AppSRE to get it confirmed. Maybe they have their own policy.

For the replicas, the workaround can be to have three replicas. It would likely take some of namespace resources, but it would resolve the warning.

skateman:
@marleystipich2 what about the readiness probes?

marleystipich2:
@skateman AppSRE referenced [this](https://developers.redhat.com/blog/2020/11/10/you-probably-need-liveness-and-readiness-probes) article saying that readiness probes aren't needed for jobs but liveness probes are(at least it's best practice)

skateman (approved):
The Seal of Approval

PR #1121, opened 2022-01-24
fix(ParseReportJob): RHICOMPL-1021 read reports from S3, not redis

skateman:
This should reduce the memory footprint of Redis as we would be no longer channel the reports through it. The racecar task pushes the URL to the job and Sidekiq consumes it directly from the S3. It's a bit less efficient as it does N+1 downloads (each report once + validation in racecar) though, but it's fully compatible with the old setup, i.e. 1 job for 1 report if N reports are in an upload. Also it's capable of draining old jobs from sidekiq, so we should clean that up later...

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/jobs/parse_report_job.rb
+++ b/app/jobs/parse_report_job.rb
@@ -6,19 +6,35 @@ require 'xccdf_report_parser'
 class ParseReportJob
   include Sidekiq::Worker
 
+  # rubocop:disable Metrics/MethodLength
   def perform(file, message)
     return if cancelled?
 
-    @file = file
     @msg_value = message
     Sidekiq.logger.info(
       "Parsing report for account #{@msg_value['account']}, "\
       "system #{@msg_value['id']}"
     )
+
+    # TODO: clean this up after the legacy jobs are drained in production
+    @file = if file.is_a?(Numeric)
+              begin
+                SafeDownloader.download_reports(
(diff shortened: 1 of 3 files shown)


app/jobs/parse_report_job.rb
```diff
@@ -6,11 +6,22 @@
...
 
-    @file = file
     @msg_value = message
+
+    # TODO: clean this up after the legacy jobs are drained in production
+    @file = if file.is_a?(Numeric)
+              SafeDownloader.download_reports(
+                @msg_value['url'],
+                ssl_only: Settings.report_download_ssl_only
+              )[file]
```
vkrizan, line 20: Observation: Lack of error reporting if the report on a given index does not exist.

vkrizan (approved):
LGTM

app/jobs/parse_report_job.rb
skateman: Right, this would need some additional error management...

PR #1132, opened 2022-01-28
fix(db:setup): RHICOMPL-2645 do not convert Arel to SQL on startup

skateman:
Apparently calling `.to_sql` on an Arel node in a model constant crashes if the database is not available, which also crashes a ` rake db:setup` call. Ideally we would not need the `to_sql` conversion, but [Rails can't deal with Arel nodes in an order hash](https://github.com/rails/rails/issues/44282) and until the [fix](https://github.com/rails/rails/pull/44284) gets in, we need to compensate somehow. So as a solution I moved the conversion from the constant to a bit deeper.

I know it's ugly, but we have to deal with it for some time :disappointed: 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/sortable.rb
+++ b/app/models/concerns/sortable.rb
@@ -30,7 +30,14 @@ module Sortable
         column, direction = field.underscore.split(':')
         assert_sortable_by!(column, direction)
 
-        rule = @sortable_by[column.to_sym]
+        # FIXME: Rails cannot deal with Arel nodes in order hashes.
+        # After the bug is resolved, the compensation will be no longer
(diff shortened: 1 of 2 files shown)


vkrizan (approved):
LGTM

Mind creating a unit test for the Arel Node case?

skateman:
@vkrizan it's already covered with our tests :wink:

PR #1146, opened 2022-02-03
fix(clowder): Make it possible to add a suffix to dependencies in clowdapp.yaml

jhutar:
This would allow to configure the deployment in Perf cluster so dependencies can have `-perf` suffix.

Pete S. said it might work when discussing over Slack, although it is possible we have missed how it is supposed to work.

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -14,9 +14,9 @@ objects:
       iqePlugin: compliance
     dependencies:
     - compliance-ssg
-    - host-inventory
-    - ingress
-    - rbac
+    - host-inventory${COMPLIANCE_DEPENDENCY_SUFFIX}
+    - ingress${COMPLIANCE_DEPENDENCY_SUFFIX}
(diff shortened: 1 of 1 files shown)


clowdapp.yaml
```diff
@@ -473,3 +473,6 @@ parameters:
 - name: FLOORIST_LOGLEVEL
   description: Floorist loglevel config
   value: 'INFO'
+- name: COMPLIANCE_DEPENDENCIES_SUFFIX
```
skateman, line 476: I'd make it a singular `COMPLIANCE_DEPENDENCY_SUFFIX`

skateman (commented):
@vkrizan what do you think about this? Performance cluster needs the suffixes in order to function correctly.

clowdapp.yaml
jhutar: Changed. Thank you

jhutar:
OK, in the end, looks like this is not needed.

Note: `dependencies` lists ClowdApp CRD resource name, not namespace where the service is deployed.

PR #1147, opened 2022-02-03
fix(ssgs): RHICOMPL-2649 only import the latest revision for OS major

skateman:
As discussed with the OpenSCAP team, we are safe with importing only the latest revision for each benchmark version per OS major. This should prevent the pingponging of rule changes between revisions that should not happen but they actually do.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -18,6 +18,12 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
     Gem::Version.new(version)
   end
 
+  def version_with_revision
+    Gem::Version.new(
+      package.sub(/^scap-security-guide-(.*)\.el.*$/, '\1')
+    )
+  end
+
   class << self
     def supported?(ssg_version:, os_major_version:, os_minor_version:)
       ssg_versions_for_os(os_major_version, os_minor_version)
--- a/app/services/datastream_downloader.rb
+++ b/app/services/datastream_downloader.rb
@@ -8,8 +8,9 @@ class DatastreamDownloader
 
   def default_supported_ssgs
     ::SupportedSsg.all
-                  .sort_by(&:comparable_version)
+                  .sort_by(&:version_with_revision)
                   .reverse
+                  .uniq { |ssg| [ssg.version, ssg.os_major_version] }
   end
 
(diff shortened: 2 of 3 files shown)


app/services/datastream_downloader.rb
```diff
@@ -10,6 +10,7 @@ def default_supported_ssgs
     ::SupportedSsg.all
                   .sort_by(&:comparable_version)
                   .reverse
+                  .uniq { |ssg| [ssg.version, ssg.os_major_version] }
```
vkrizan, line 13: How is it with the determinism of the uniqueness and sorting? It is sorted by a `comparable_version` based off of `version` which for some packages would the same.

```
scap-security-guide-0.1.57-3.el8_4: 0.1.57
scap-security-guide-0.1.57-5.el8: 0.1.57
```

skateman: @vkrizan adjusted

app/models/supported_ssg.rb
```diff
@@ -18,6 +18,12 @@ def comparable_version
     Gem::Version.new(version)
   end
 
+  def version_with_revision
+    Gem::Version.new(
+      package.sub(/^scap-security-guide-(.*)\.el.*$/, '\1')
```
vkrizan, line 23: Observation: for an unmatched package pattern this would raise `Malformed version number string ... (ArgumentError)`. That might or might not be deliberate.

vkrizan (approved):
LGTM

app/models/supported_ssg.rb
skateman: I think we want this to fail then, smoke tests would catch it.

vkrizan: Smoke tests would, however we get the data from external source (compliance-ssg).

skateman: Correction: we'd catch this in stage :laughing: and we rather want to

PR #1148, opened 2022-02-06
fix(HostSearching): RHICOMPL-2189 allow ^ operator when looking for OS

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/host_searching.rb
+++ b/app/models/concerns/host_searching.rb
@@ -10,9 +10,9 @@ module HostSearching
     scoped_search on: %i[id display_name], only_explicit: true
     scoped_search on: :display_name, rename: :name
     scoped_search on: :os_major_version, ext_method: 'filter_os_major_version',
-                  only_explicit: true, operators: ['=', '!=']
+                  only_explicit: true, operators: ['=', '!=', '^', '!^']
     scoped_search on: :os_minor_version, ext_method: 'filter_os_minor_version',
-                  only_explicit: true, operators: ['=', '!=']
+                  only_explicit: true, operators: ['=', '!=', '^', '!^']
     scoped_search on: :compliant, ext_method: 'filter_by_compliance',
                   only_explicit: true
     scoped_search on: :compliance_score,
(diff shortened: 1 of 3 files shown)


app/models/concerns/host_searching.rb
```diff
@@ -43,28 +43,26 @@ module HostSearching
...
+    scope :os_minor_version, lambda { |versions, equal = true|
+      where(::Host.os_version_query(:minor, versions, equal))
     }
   end
 
   # class methods for Host searching
   module ClassMethods
     def filter_os_major_version(_filter, operator, value)
-      hosts = ::Host.os_major_version(value, operator == '=')
+      values = value.split(',').map(&:to_i)
```
vkrizan, line 58: This would likely raise HTTP 500 if user provides invalid input. Does the lib provide some input validation?

app/models/host.rb
```diff
@@ -59,6 +59,17 @@ def readonly?
     true
   end
 
+  def self.os_version_query(path, values, include = true)
+    query = include ? :in : :not_in
+    path = %i[major minor].include?(path) && path
```
vkrizan, line 64: What would happen with `false` when `path` does not match allowed values? Can it raise exception early in the process to signal a code mistake?

app/models/host.rb
```diff
@@ -59,6 +59,17 @@ def readonly?
...
+
+    Arel::Nodes::NamedFunction.new('CAST', [Arel::Nodes::InfixOperation.new(
+      '->',
```
vkrizan, line 67: Interesting that `->` is changed to `@>` for the query.

app/models/host.rb
```diff
@@ -59,6 +59,17 @@ def readonly?
     true
   end
 
+  def self.os_version_query(path, values, include = true)
```
vkrizan, line 62: Observation: This fn might be in host_searching, if rubocop agrees.
Nit: `include` in function signature but `equal` as variable name outside of it.

vkrizan (approved):
Noice.

app/models/host.rb
skateman: `@>` is for partial matching

app/models/host.rb
skateman: It doesn't agree :disappointed:

app/models/concerns/host_searching.rb
skateman: The parser only lets through strings, any string can be splitted, worst case the result will be an empty string that converted to an integer returns a 0. Which could match 8.0 when looking for a minor one, but I guess that's okay.

app/models/host.rb
vkrizan: My bad, I forgot to check out the PR's code. :-(

app/models/host.rb
```diff
@@ -59,6 +59,17 @@ def readonly?
...
 
+  def self.os_version_query(path, values, equal = true)
+    query = equal ? :in : :not_in
+    raise ArgumentError unless %i[major minor].include?(path)
+
+    Arel::Nodes::NamedFunction.new('CAST', [Arel::Nodes::InfixOperation.new(
+      '->',
+      arel_table[:system_profile],
+      Arel::Nodes::SqlLiteral.new("'operating_system'->'#{path}'")
+    ).as('int')]).send(query, values)
```
vkrizan, line 70: What would happen with hosts that have empty string on major or minor OS version.

app/models/concerns/host_searching.rb
vkrizan: I had the `to_i` conversion in mind.

app/models/host.rb
skateman: I don't really understand the question :thinking:

vkrizan: The db. This queries JSON in `system_profile` field, populated by Inventory via Cyndi. Some hosts either might lack `operating_system.minor` (or major) values and can have it as an empty string. In such cases this would fall app art like house of cards. :grin:

skateman: The `CAST` in pg doesn't convert `NULL` to `0` so I don't see how it would happen:

```
SELECT CAST(NULL as int), CAST(0 as int);
 int4 | int4 
------+------
      |    0
(1 row)
```

vkrizan:
```
# SELECT CAST('' as int);
ERROR:  invalid input syntax for type integer: ""
LINE 1: SELECT CAST('' as int);
```

skateman: I resolved it with an additional layer of casting+nullif

PR #1153, opened 2022-02-09
feat(ProfileTailoring): RHICOMPL-1736 order rules by precedence

skateman:
Not sure if it's really needed but it was mentioned in the JIRA ticket. Also the order in a hash is only maintainable safely if we treat it as immutable, which we do. 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/profile_tailoring.rb
+++ b/app/models/concerns/profile_tailoring.rb
@@ -17,7 +17,7 @@ module ProfileTailoring
   end
 
   def added_rules
-    rules - parent_profile.rules
+    rules.order(:precedence) - parent_profile.rules
   end
 
   def removed_rules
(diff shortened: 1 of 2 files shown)


app/models/concerns/profile_tailoring.rb
```diff
@@ -17,7 +17,7 @@ def tailored?
   end
 
   def added_rules
-    rules - parent_profile.rules
+    rules.order(:precedence) - parent_profile.rules
```
vkrizan, line 20: We might add this as a default order within APIs.

vkrizan (approved):
LGTM

app/models/concerns/profile_tailoring.rb
skateman: In the followup issue that deals with remediations.

PR #1169, opened 2022-02-16
fix(notifications): RHICOMPL-2722 semantically correct messages

skateman:
@vkrizan this is to make you happy :wink: there's also a second half that should be merged first https://github.com/RedHatInsights/notifications-backend/pull/1002

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/producers/notification.rb
+++ b/app/producers/notification.rb
@@ -7,7 +7,8 @@ class Notification < ApplicationProducer
   VERSION = 'v1.1.0'
 
   # rubocop:disable Metrics/MethodLength
-  def self.deliver(event_type:, account:, host:, policy:, **extra)
+  # rubocop:disable Metrics/AbcSize
+  def self.deliver(event_type:, account_number:, host:, policy:, **extra)
     payload = {
       host_id: host.id,
       host_name: host.display_name,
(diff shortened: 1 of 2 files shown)


app/producers/notification.rb
```diff
@@ -26,7 +26,10 @@ def self.deliver(event_type:, account:, host:, policy:, **extra)
         metadata: {},
         payload: payload
       }],
-      context: payload
+      context: {
+        host: host.id,
```
vkrizan, line 30: Can we follow a convention similar to Advisor? Here's an example:
```
'context': {
  'display_name': 'rhel8desktop',
  'host_url': 'https://console.redhat.com/insights/inventory/10000000-1000-4000-8000-100000000000',
  'hostname': 'rhel8desktop',
  'inventory_id': '10000000-1000-4000-8000-100000000000',
  'rhel_version': '8.3',
  'tags': [{'key': 'Team', namespace': 'insights-client',  'PMs'}]
}
```

vkrizan (commented):
Thanks!

app/producers/notification.rb
vkrizan: It is not a convention yet, we're working on it... From what I've heard, the `context` can have anything that is common within the `events` array. Hence the `policy` id can be included as well.

I've added Advisor's example, as it includes more verbose host identification, including display name and RHEL version. There is no standard on the identification yet, unfortunately.

skateman: Sorry to say that, but this is a mess then :see_no_evil: :confused: the policy and host are equally part of the context and with this logic, however, if we aggregate, the policy will have a higher priority. We surely can have a single host assigned to multiple policies, but the opposite (multiple hosts under a policy) is the way people use compliance. So I am not entirely sure if a host even should be part of our context :thinking:

vkrizan: I should say that the concern is mainly on the identification of a system/host. If it would be an event with a single policy having multiple systems reporting, then the policy can be in context and individual systems in the events array.

skateman:
@vkrizan can I get your :+1: please?

vkrizan (approved):
LGTM

PR #1171, opened 2022-02-19
fix(remediations): RHICOMPL-2870 pass profile_ref_id for rules

skateman:
Fixes the remediation issue ID in GraphQL.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/rule_remediation.rb
+++ b/app/models/concerns/rule_remediation.rb
@@ -11,10 +11,13 @@ module RuleRemediation
   def remediation_issue_id
     return nil unless remediation_available
 
-    # FIXME: Nondetermenistic canonical selection
-    profile = profiles.select(&:canonical?)&.first
-    return unless profile
+    if attributes['profile_ref_id'] # use the cached profile_ref_id if available
+      "ssg:rhel#{benchmark.os_major_version}|#{ShortRefId.short_ref_id(attributes['profile_ref_id'])}|#{ref_id}"
+    else # select the first canonical profile if there's a cache miss
+      profile = profiles.select(&:canonical?)&.first
+      return unless profile
 
-    "ssg:rhel#{profile.os_major_version}|#{profile.short_ref_id}|#{ref_id}"
+      "ssg:rhel#{profile.os_major_version}|#{profile.short_ref_id}|#{ref_id}"
+    end
   end
 end
--- a/app/graphql/types/interfaces/rules_preload.rb
+++ b/app/graphql/types/interfaces/rules_preload.rb
@@ -35,7 +35,7 @@ module Types
       end
 
       def all_rules
-        ::CollectionLoader.for(::Profile, :rules).load(object).then do |rules|
+        ::CollectionLoader.for(::Profile, :rules_with_profile_ref_ids).load(object).then do |rules|
           rules
         end
       end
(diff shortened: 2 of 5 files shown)


app/models/concerns/rule_remediation.rb
```diff
@@ -9,10 +9,13 @@ module RuleRemediation
   #  `.includes(profiles: :benchmark)`
   #
   def remediation_issue_id
-    # FIXME: Nondetermenistic canonical selection
-    profile = profiles.select(&:canonical?)&.first
-    return unless profile
+    if attributes['profile_ref_id'] # use the cached profile_ref_id if available
+      "ssg:rhel#{benchmark.os_major_version}|#{ShortRefId.short_ref_id(attributes['profile_ref_id'])}|#{ref_id}"
+    else # select the first canonical profile if there's a cache miss
```
skateman, line 14: @vkrizan referring to what I mentioned in the jira task, we could theoretically just remove this part and return `nil` if the rule is not retrieved as a child of a profile. I am not exactly sure about how the remediations work here though...

vkrizan: So to summarize, we need to have this context/profile aware. Returning `nil` can be okay to know what would break :slightly_smiling_face:. Using some random profile's for an unknown context (current else block) is just random, but at least it provides something. I'll leave it up to you...

Honestly, the API could have been improved to make sense like having rules as a subresource of a profile: `/v2/profiles/profileid/rules`. But with regards to what return on general all `/rules` endpoint, it's questionable whether to return each rule duplicated for each profile it is used in, or return it general unique and profile-less rules.

skateman: This PR fixes the issue for `v1/profiles/<id>/rules`, the column will be there and the proper remediation will be returned. 

The problem is with the `/rules` endpoint, there it keeps the old and nondeterministic behavior, which I don't really like.

app/serializers/profile_serializer.rb
```diff
@@ -6,7 +6,7 @@ class ProfileSerializer < ApplicationSerializer
   belongs_to :account
   belongs_to :benchmark
   belongs_to :parent_profile, record_type: :profile
-  has_many :rules
+  has_many :rules, &:rules_with_profile_ref_ids
```
vkrizan, line 9: This solves the issue for a single profile. Within `/profiles?include=rules` this isn't solvable, as the structure is flat and objects are identified by uuid. Only the GraphQL has the profile-aware solved for a list of profiles.

```
profileA
  + rules
    + ruleAID
profileB
  + rules
    + ruleAID
included
  + rules
    + ruleA
```
`ruleA` cannot have `remediation_issue_id` for both profiles.

skateman: In this case I'm only fixing this for GQL where it makes sense :see_no_evil:

InsightsDroid:
0caacd6d-b2fc-4a71-8147-ca5d0c3b4c49
79da5df0-5ac4-4a90-a500-fc4dbe6f2233

InsightsDroid:
5dde6e37-d927-46a9-bc2c-98c8102a2eef
d3b8c322-6317-44be-9ae3-6fff86a46643

vkrizan (approved):
:+1:

PR #1207, opened 2022-03-11
perf(SSG): RHICOMPL-2793 determine playbook existence from autoindex

skateman:
Instead of downloading each playbook one-by-one, we can simply download the newly provided JSON autoindex from the right folder and match against the existing files. The performance increase is significant, the original import takes :sloth: ~70 seconds :sloth: on my local machine and the new one requires :drum: :drum:  only :zap: 2-3 seconds :zap: 

Note that for testing this you'd need to pull the latest SSG container from quay.
1. Clear the revisions from rails console
```ruby
Revision.find_by(name: 'remediations')&.delete
```
2. Import the remediations
```bash
time bin/rake import_remediations
```

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/services/playbook_downloader.rb
+++ b/app/services/playbook_downloader.rb
@@ -3,35 +3,35 @@
 # Service for dowloading SSG remediation playbooks
 class PlaybookDownloader
   class << self
-    def playbook_exists?(rule, profile = nil)
-      download(playbook_url(rule, profile&.short_ref_id)).present?
+    def playbooks_exist?(rules)
+      rules.inject({}) do |obj, rule|
+        obj.merge(rule.id => playbook_exists?(rule))
+      end
     end
 
-    def playbooks_exist?(rules)
-      rules.map do |rule|
-        [rule.id, playbook_exists?(rule)]
-      end.to_h
+    def playbook_exists?(rule)
+      playbook_list(rule.benchmark.os_major_version).include?(rule.short_ref_id)
     end
 
     private
 
-    def playbook_url(rule, profile_short_ref_id = nil)
+    def playbook_list(os_major_version)
+      @cache ||= {}
+      @cache[os_major_version] ||= begin
+        file = SafeDownloader.download(playbook_list_url(os_major_version))
+        JSON.parse(file.read).map do |row|
+          row['name'].sub(/\.yml$/, '')
+        end
+      end
+    end
+
+    def playbook_list_url(os_major_version)
       [
         Settings.compliance_ssg_url,
         'playbooks',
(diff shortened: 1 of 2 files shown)


app/services/playbook_downloader.rb
```diff
@@ -3,35 +3,35 @@
...
+
+    def playbook_list_url(os_major_version)
       [
         Settings.compliance_ssg_url,
         'playbooks',
-        "rhel#{rule.benchmark.os_major_version}",
-        profile_short_ref_id || 'all',
-        "#{rule.short_ref_id}.yml"
+        "rhel#{os_major_version}",
+        'all'
```
vkrizan, line 33: "all"? Where has the `profile_short_ref_id` gone?

skateman: Where the `profile_short_ref_id` has been used? :shrug: :wink:

vkrizan: I see. This is only used to update remediation_available boolean. 

Technically it could/should be profile specific. Could it happen that for some of the profiles the remediation would not be available?

skateman: If that's even possible, our current DB structure has no means to store this kind of info and we'd need a remodel.

skateman: @vkrizan I created a [task](https://issues.redhat.com/browse/RHICOMPL-2795) to verify it, but it should be independent from the current implementation as here I'm not changing behavior, just improving performance.

app/services/playbook_downloader.rb
```diff
@@ -3,35 +3,35 @@
 # Service for dowloading SSG remediation playbooks
 class PlaybookDownloader
   class << self
-    def playbook_exists?(rule, profile = nil)
-      download(playbook_url(rule, profile&.short_ref_id)).present?
+    def playbooks_exist?(rules)
+      rules.inject({}) do |obj, rule|
+        obj.merge(rule.id => playbook_exists?(rule))
```
vkrizan, line 8: Nit: `merge!`?

vkrizan (approved):
Nice!

app/services/playbook_downloader.rb
skateman: It's 20x slower :disappointed:

PR #1215, opened 2022-03-12
fix(racecar): RHICOMPL-2801 retrieve account from the right location

skateman:
The `@msg_value['account']` is being used in our jobs, but it is not consistent with what we have in the kafka consumer. The correct location for the account is under `@msg_value['platform_metadata']['account']` that can be deducted from the fact that the inventory consumer passes the [`metadata`](https://github.com/RedHatInsights/compliance-backend/blob/4061c4a56ffdf43160d76e74c6dc2e69a307565b/app/consumers/concerns/report_parsing.rb#L96-L102) object that lives in Sidekiq as `@msg_value` further (different from the same variable name in the consumer). 

This caused our audit logging to omit the account information in inventory-consumer messages and that the account and host information in notifications was incorrect. Our tests of course were also incorrectly setting the account number at the wrong location, so there everything was passing.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/consumers/concerns/report_parsing.rb
+++ b/app/consumers/concerns/report_parsing.rb
@@ -48,8 +48,8 @@ module ReportParsing
       logger.error error_message
       logger.audit_fail error_message
       notify_payload_tracker(:error, error_message)
-      ReportUploadFailed.deliver(host: Host.find_by(id: id, account: @msg_value['account']),
-                                 account_number: @msg_value['account'], error: msg_for_notification(exc))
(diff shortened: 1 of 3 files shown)


romanblanco (approved):
LGTM :+1:

PR #1218, opened 2022-03-13
chore(SSG): RHICOMPL-2804 do not query RuleRef(Rule)s when importing

skateman:
The `find_unique` used to detect existing `RuleReference` and `RuleReferencesRule` generates huge queries and the code complexity around it is huge and hard to follow. The activerecord-import gem is fully capable of just ignoring already existing records or overwrite certain fields, basically providing us an upsert-like feature. It also updates the IDs of the imported records if asked, therefore, our explicit logic for finding existing records is obsolete. It's also ~40 seconds faster :zap: :zap:  

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


skateman:
I gabied databases from the void and ran the 2 queries below:
```
SELECT md5(string_agg(concat(href, '|', label), ',' order by href, label)) FROM rule_references;
SELECT md5(string_agg(concat(rule_id, '|', rule_reference_id), ',' order by rule_id, rule_reference_id)) FROM rule_references_rules;
```
Then I cleared the revisions using `bin/rails r 'Revision.destroy_all'` and force-imported via `SETTINGS__FORCE_IMPORT_SSGS=1 bin/rake ssg:import_rhel_supported` to try to break stuff. After this I re-ran the queries and compared the hashes and they were identical.

romanblanco (approved):
LGTM :+1:

app/services/concerns/xccdf/rule_references.rb
```diff
@@ -7,30 +7,15 @@ module RuleReferences
...
-      def new_op_rule_references
-        @existing_href_labels = existing_rule_references.pluck(:href, :label)
-
-        @op_rule_references.reject do |op_reference|
-          @existing_href_labels.include? [op_reference.href, op_reference.label]
-        end
+        ::RuleReference.import!(@rule_references,
+                                on_duplicate_key_update: {
+                                  conflict_target: %i[label href],
+                                  columns: %i[label href]
```
marleystipich2, line 17: I'm not sure if `on_duplicate_key_update` is necessary here because if `label` and `href` are the target columns, that would mean that if there's another record with the same values for those 2 columns, it would update the `label` and `href` which would just make them the same thing, I think. So I think this is the same thing as just doing `ignore: true` ?

skateman: It is, even if it "doesn't do anything" :see_no_evil: I know it sounds weird, but the point here is to update the IDs of all objects in the `@rule_references` array. Notice that I removed the `find_or_initialize_by` to avoid a bunch of `find_by` calls to our database. If I'd do an import with `ignore: true`, the `@rule_references` array would only have updated IDs for the newly created entities.

PR #1220, opened 2022-03-14
[RHICOMPL-2815] Patch clowdapp

Victoremepunto:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -43,13 +43,8 @@ objects:
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
-        - command:
-            - bash
-            - -c
-            - |
-              for i in {1..120}; do sleep 1;
-              if bundle exec rake --trace db:abort_if_pending_migrations;
-              then exit 0; fi; done; exit 1
+        - command: ["/bin/sh"]
+          args: ["-c", "$HOME/scripts/abort_if_pending_migrations.sh"]
           inheritEnv: true
         resources:
           limits:
@@ -78,12 +73,11 @@ objects:
           value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
         - name: SETTINGS__FORCE_IMPORT_SSGS
           value: "${SETTINGS__FORCE_IMPORT_SSGS}"
+        - name: MAX_INIT_TIMEOUT_SECONDS
+          value: "${MAX_INIT_TIMEOUT_SECONDS}"
         livenessProbe:
           exec:
-            command:
-              - bash
-              - -c
-              - ps ax | grep [r]ake
+            command: ["pgrep" ,"-f", "rake"]
     - name: floorist
       schedule: ${FLOORIST_SCHEDULE}
       suspend: ${{FLOORIST_SUSPEND}}
@@ -122,10 +116,7 @@ objects:
           value: ${FLOORIST_LOGLEVEL}
         livenessProbe:
           exec:
-            command:
-              - bash
-              - -c
-              - ps ax | grep [r]un
+            command: ["pgrep", "-f", "run"]
         volumeMounts:
         - name: floorplan-volume
           mountPath: "/tmp/floorplan"
@@ -142,7 +133,6 @@ objects:
             memory: "${MEMORY_REQUEST_FLOO}"
     deployments:
     - name: service
-      # TODO: Check how to auto-scale with Clowder
       minReplicas: ${{REPLICAS_BACKEND}}
       webServices:
         public:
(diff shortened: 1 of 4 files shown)


clowdapp.yaml
```diff
@@ -46,10 +46,11 @@ objects:
         - command:
             - bash
             - -c
-            - |
-              for i in {1..120}; do sleep 1;
```
vkrizan, line 50: This 120 cycle limit was put in to leverage backoff and aid for failed job runs. Can we keep some kind of limit other than infinite loop?
Have a look at convo at https://github.com/RedHatInsights/compliance-backend/pull/1012#discussion_r712831489

Victoremepunto: @vkrizan  check: https://github.com/RedHatInsights/compliance-backend/pull/1220/commits/efad82845a795c998a99dc8ac0fe78c817980446

clowdapp.yaml
```diff
@@ -473,3 +525,6 @@ parameters:
 - name: IMPORT_SSG_SCHEDULE
   description: Cronjob schedule for import-ssg
   value: "*/5 * * * *" # every 5 minutes
+- name: MAX_INIT_TIMEOUT_SECONDS
+  description: Number of seconds for timeout init container operation
+  value: "120"
```
skateman, line 530: :thinking: I would maybe rename this more explicitly to something related to maximum SSG import wait(ing?) time(?out?). Also 240 seconds sound more reasonable as the import takes like ~5 minutes in ephemeral. 

That's another thing that locally for me it takes only ~3 minutes and it is not even parallelized yet. With Marley's rule grouping I measured a 100-120 seconds longer import time.

Victoremepunto: Well, we can discuss about names, but this is a timeout value for the init operation of whichever pod we're using - that is, the import-ssg and the run-migrations tasks - they're 2 different scripts, with very similar implementation. 

That said, the good thing is we can define this env variable per environment, so we can tune it based on which environment we're targeting. 

I can create 2 variables, for the 2 different init tasks timeout, and we can tune that per-environment on our deployment templates per namespace. LMK what you think

clowdapp.yaml
```diff
@@ -317,6 +337,38 @@ objects:
...
           value: ${SETTINGS__REPORT_DOWNLOAD_SSL_ONLY}
+        - name: MAX_INIT_TIMEOUT_SECONDS
+          value: "${MAX_INIT_TIMEOUT_SECONDS}"
+
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: init-scripts
+  data:
+    abort_if_pending_migrations.sh: |
```
vkrizan, line 348: Can we add these scripts to the image? That would avoid custom maps and mounts. Thanks.

clowdapp.yaml
vkrizan: Thanks Victor. 
I don't find the option to be essential.

clowdapp.yaml
Victoremepunto: Using the ConfigMaps has the point of avoiding having to rebuild the image and adding those there :) I mean that's what I was trying to avoid.

clowdapp.yaml
Victoremepunto: I find it a nice to have, and quite honestly, being able to fine tune it gives us the advantage of adjust it based on the environment we're deploying to - i.e. -> on a fresh deployment, it is reasonable to extend the timeout, as it is expected most dependencies won't be ready right away. 

On the contrary, on a long living environment, a deployment should take less time and dependent services are likely to be available so we can cut that timeout down and crash earlier if there's a real ongoing problem.

clowdapp.yaml
vkrizan: I know. The image is build nevertheless for each PR and merge.

Victoremepunto: Working on moving the scripts to the image - once it's tested I'll ping you for re-review

Victoremepunto: finally, it's green :)

LightOfHeaven1994:
@skateman @vkrizan I think this one should be ready to merge. All checks are green

vkrizan (approved):
LGTM. Thanks @Victoremepunto !

These are the mentioned changes:
```
- Moves init container commands into separate scripts
- Adds active wait loops to all initial containers, with configurable
  timeout before crashing
- Refactor of some liveness probes
- Reduced deployment timeout value to 10 minutes
```

@Victoremepunto, each change can have its own commit.

Victoremepunto:
> LGTM. Thanks @Victoremepunto !
> 
> These are the mentioned changes:
> 
> ```
> - Moves init container commands into separate scripts
> - Adds active wait loops to all initial containers, with configurable
>   timeout before crashing
> - Refactor of some liveness probes
> - Reduced deployment timeout value to 10 minutes
> ```
> 
> @Victoremepunto, each change can have its own commit.

Thanks for the review @vkrizan - I thought we squashed as a commit strategy - I can definitely split the changes again, wait for it :)

vkrizan:
@Victoremepunto no need to split them. I was meant to say it as a friendly reminder to split concerns. It's beneficial when going through history, understanding changes and why they were made.

PR is good to be merged.

PR #1225, opened 2022-03-16
feat(RBAC): RHICOMPL-2711 RBAC permission loading

romanblanco:
https://consoledot.pages.redhat.com/docs/dev/services/rbac.html#_integrating_with_rbac
https://insights-rbac.readthedocs.io/en/latest/management.html#permissions

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/models/concerns/rbac.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+# Role-based Access Control
+module Rbac
+  extend ActiveSupport::Concern
+
+  attr_accessor :loaded_permission
```
skateman, line 7: I suspect there are more than one, so:
```suggestion
  attr_accessor :loaded_permissions
```
Or if this is a boolean, semantically it's more correct to have:
```suggestion
  attr_accessor :permissions_loaded?
```

test/models/user_test.rb
```diff
@@ -32,4 +32,35 @@ class UserTest < ActiveSupport::TestCase
...
+    current_user.stubs(:request_permission_data).returns(data)
+
+    assert current_user.authorized_to?('app', 'resource0', '*')
+    assert current_user.authorized_to?('app', 'resource0', 'destroy')
+    assert current_user.authorized_to?('app', 'resource1', 'write')
+    assert_not current_user.authorized_to?('app', '*', 'link')
+    assert_not current_user.authorized_to?('app', '*', '*')
+    assert_not current_user.authorized_to?('app', 'resource1', 'read')
+    assert_not current_user.authorized_to?('app', 'resource1', '*')
+  end
```
skateman, line 65: I'd like to see a test that checks if the RBAC API is only called once across a request (caching).

app/services/rbac_api.rb
```diff
@@ -1,32 +1,39 @@
...
+      config.password = Settings.rbac_password
+    end
+    @api_instance = RBACApiClient::AccessApi.new
     @b64_identity = b64_identity
   end
 
-  def check_user
+  # rubocop:disable Metrics/MethodLength
+  def load_user_permissions
+    app = 'compliance,inventory'
```
skateman, line 19: These should go into a constant...

app/services/rbac_api.rb
```diff
@@ -1,32 +1,39 @@
...
+
 # This class is meant to handle calls to the RBAC light API
 class RbacApi
   def initialize(b64_identity)
-    @url = URI.parse("#{URI.parse(Settings.rbac_url)}"\
-                     "#{Settings.path_prefix}/rbac/v1/access/")
+    RBACApiClient.configure do |config|
+      config.host = Settings.rbac_url
+      config.username = Settings.rbac_username
+      config.password = Settings.rbac_password
```
skateman, line 11: I guess temporarily until you figure out the clowder stuff, you can just point at the constants in `Platform`

app/services/rbac_api.rb
```diff
@@ -1,32 +1,39 @@
...
+    end
+    @api_instance = RBACApiClient::AccessApi.new
```
skateman, line 13: Looks more conventional:
```suggestion
    @client = RBACApiClient::AccessApi.new
```

skateman (changes requested):

app/models/concerns/rbac.rb
romanblanco: Not boolean, will use `:loaded_permissions`. Thanks! :+1:

app/services/rbac_api.rb
```diff
@@ -1,32 +1,47 @@
 # frozen_string_literal: true
 
+require 'insights-rbac-api-client'
+
 # This class is meant to handle calls to the RBAC light API
 class RbacApi
+  APPS = 'compliance,inventory'
+  HEADER = { header_params: { 'X_RH_IDENTITY': @b64_identity } }.freeze
```
skateman, line 8: This will never work, the `HEADER` gets evaluated upon load time and then the `@b64_identity˙ is `nil`.

config/initializers/0_clowder-config.rb
```diff
@@ -10,6 +10,8 @@
   # RBAC
   rbac_config = config.dependency_endpoints.dig('rbac', 'service')
   rbac_url = "http://#{rbac_config&.hostname}:#{rbac_config&.port}"
+  rbac_username = config.dig('database', 'username')
+  rbac_password = config.dig('database', 'password')
```
skateman, line 14: Why database?

app/services/rbac_api.rb
```diff
@@ -1,32 +1,46 @@
...
-    rescue Faraday::ClientError => e
-      Rails.logger.info("#{e.message} #{e.response}")
+      result = @client.get_principal_access(
+        RbacApi::APPS,
+        opts.merge({ header_params: { 'X_RH_IDENTITY': @b64_identity } })
+      )
+      result.data.map(&:permission)
+    rescue RBACApiClient::ApiError => e
+      Rails.logger.info(e.message)
+      nil
```
skateman, line 33: Why the `nil`? Should not we fail with unauthorized?

spec/swagger_helper.rb
```diff
@@ -13,6 +13,7 @@
   config.before(:each) do
     stub_request(:get, /#{Settings.rbac_url}/)
       .to_return(status: 200,
+                 headers: { 'Content-Type' => 'application/vnd.api+json' },
```
skateman, line 16: How is this part of the scope of the PR?

app/models/concerns/rbac.rb
```diff
@@ -0,0 +1,60 @@
...
+module Rbac
+  extend ActiveSupport::Concern
+
+  attr_accessor :loaded_permissions
+
+  def permission
+    return loaded_permissions if loaded_permissions.present?
+
+    self.loaded_permissions = parse_rbac_permissions(request_permission_data)
+  end
```
skateman, line 13: Below you're using memoization, but here you're juggling with attr_accessor. Why?

app/models/concerns/rbac.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def parse_rbac_permissions(data)
+    return {} if data.blank?
+
+    data.map do |d|
+      app, resource, action = d.split(':')
+      { "#{app}": { "#{resource}": action } }
+    end.freeze
+  end
```
skateman, line 54: I think this should be part of the RBAC API class.

app/models/concerns/rbac.rb
skateman: Also why singular? Is it retrieving a single permission?

app/models/concerns/rbac.rb
```diff
@@ -0,0 +1,60 @@
...
+  private
+
+  def action_permitted?(resource_permissions, resource, action)
+    resource_permissions.any? do |permission|
+      [
+        { '*'.to_sym => '*' },
+        { resource.to_sym => '*' },
+        { resource.to_sym => action }
+      ].include?(permission)
+    end
```
skateman, line 34: The complexity here is heavy, what is this doing?

skateman (commented):
It's really hard to follow the code in the RBAC concern. In general I'd move as much stuff into the `RbacApi` class as possible, but also think about working with objects provided by the client gem instead of dealing with hashes if that's something we can do.

spec/swagger_helper.rb
romanblanco: Unit tests are failing when the header is not set. See https://github.com/RedHatInsights/insights-rbac-api-client-ruby/pull/11.

skateman: Ideally compensations should be marked with a comment and a FIXME to get it removed after the bug is fixed on the other side.

app/services/rbac_api.rb
```diff
@@ -1,32 +1,47 @@
...
+  APPS = 'compliance,inventory'
+
   def initialize(b64_identity)
-    @url = URI.parse("#{URI.parse(Settings.rbac_url)}"\
-                     "#{Settings.path_prefix}/rbac/v1/access/")
+    RBACApiClient.configure do |config|
+      config.host = Settings.rbac_url
+      config.username = Platform::BASIC_AUTH[0] # FIXME: Settings.rbac_username
+      config.password = Platform::BASIC_AUTH[1] # FIXME: Settings.rbac_password
+    end
```
skateman, line 15: This might be something we could extract into an initializer under `config/initializers`

app/services/rbac_api.rb
```diff
@@ -1,32 +1,63 @@
...
+    }
     begin
-      body = JSON.parse(access_check_response.body)
-      body['data'].find do |access_check|
-        return true if access_check['permission'] == 'compliance:*:*'
-      end
-    rescue Faraday::ClientError => e
-      Rails.logger.info("#{e.message} #{e.response}")
+      result = @client.get_principal_access(
+        RbacApi::APPS,
```
skateman, line 24:
```suggestion
        self::APPS,
```

app/services/rbac_api.rb
```diff
@@ -1,32 +1,63 @@
...
+# This class is meant to handle calls to the RBAC API
 class RbacApi
+  COMPLIANCE_FULL_ACCESS = 'compliance:*:*'
+  APPS = 'compliance,inventory'
+  ANY = '*'
+
   def initialize(b64_identity)
-    @url = URI.parse("#{URI.parse(Settings.rbac_url)}"\
-                     "#{Settings.path_prefix}/rbac/v1/access/")
+    @client = RBACApiClient::AccessApi.new
```
skateman, line 12: I'm waging if we should have this service a instantiable or to convert it to a singleton. As far as I understand, this `AccessApi` does not hold information that can leak into other contexts, i.e. we could theoretically share it among all requests:

```
class Rbac
  # ...
  API_CLIENT = RBACApiClient::AccessApi.new

  class << self
    def load_user_permissions(identity)
      # ...
    end

    private

    # ...
  end
end

```

skateman: I guess the only downside would be to pass the identity to the load method, but as it would be called only once and from a single location (where the header actually lives) I think it is fine, wdyt?

app/models/concerns/rbac.rb
```diff
@@ -0,0 +1,25 @@
...
+module Rbac
+  extend ActiveSupport::Concern
+
+  attr_accessor :rbac_permissions
+
+  def permissions
+    return rbac_permissions if rbac_permissions.present?
+
+    self.rbac_permissions = request_access_data
+  end
```
skateman, line 13: Memoization is much simpler than juggling with accessors and easier to read:
```suggestion
  def rbac_permissions
    @rbac_permissions ||= Rbac.load_users_permissions(account.b64_identity)
  end
```

skateman: Also, could this be a private method?

app/models/concerns/rbac.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def authorized_to?(access_request)
+    permissions.any? { |a| @rbac_api.verify(a.permission, access_request) }
```
skateman, line 16:
```suggestion
    rbac_permissions.any? { |a| Rbac.verify(a.permission, access_request) }
```

app/services/rbac_api.rb
```diff
@@ -1,32 +1,63 @@
...
     @b64_identity = b64_identity
   end
 
-  def check_user
+  # rubocop:disable Metrics/MethodLength
+  def load_users_permissions
```
skateman, line 17:
```suggestion
  def load_user_permissions
```
Also should it not be a constant below, if you extracted it?

app/services/rbac_api.rb
```diff
@@ -1,32 +1,63 @@
...
-    rescue Faraday::ClientError => e
-      Rails.logger.info("#{e.message} #{e.response}")
+      result = @client.get_principal_access(
+        RbacApi::APPS,
+        opts.merge({ header_params: { 'X_RH_IDENTITY': @b64_identity } })
+      )
+      result.data
+    rescue RBACApiClient::ApiError => e
+      Rails.logger.info(e.message)
+      # TODO: unauthorized
```
skateman, line 30: This HAS to be covered, we can't merge it without. Our API should return a 403 if the RBAC service is unavailable.

skateman: Maybe you could mock the result somehow for future consumption?

app/services/rbac_api.rb
```diff
@@ -1,32 +1,63 @@
...
-      @url, { application: 'compliance' },
-      X_RH_IDENTITY: @b64_identity
-    )
+  def structurize(access_entry)
+    app, resource, action = access_entry.split(':')
+    struct = OpenStruct.new
+    struct.app = app
+    struct.resource = resource
+    struct.action = action
+    struct
```
skateman, line 61: Can be simplified to:
```suggestion
    OpenStruct.new(app: app, resource: resource, action: action)
```

app/services/rbac_api.rb
```diff
@@ -1,32 +1,63 @@
...
+  def verify(permitted, requested)
+    permitted_access = structurize(permitted)
+    requested_access = structurize(requested)
+    return false if permitted_access.app != requested_access.app
+    return false if permitted_access.resource != ANY &&
+                    permitted_access.resource != requested_access.resource
+
+    permitted_access.action == ANY ||
+      permitted_access.action == requested_access.action
+  end
```
skateman, line 53: I thinki I understand what you're trying to do here, do you think it's possible to substitute it with [pattern matching](https://docs.ruby-lang.org/en/3.0/doc/syntax/pattern_matching_rdoc.html)? Maybe you'd have to use hashes instead of the structs though, but I guess it would require less jumps with guard cases.

skateman (changes requested):
More understandable than before :+1: there are still a few things that could be done better. I haven't checked the tests yet as I expect them to change as well...

app/controllers/concerns/authentication.rb
```diff
@@ -60,7 +60,7 @@ def rbac_allowed?
     return valid_cert_auth? if identity_header.cert_based?
 
     @rbac_api ||= ::RbacApi.new(raw_identity_header)
-    @rbac_api.check_user
+    @rbac_api.full_access?
   end
```
skateman, line 64: This should be replaced with `current_user.something?` to actually leverage what you implemented.

app/services/rbac_api.rb
romanblanco: Sounds reasonable to me, will try it :+1:

skateman: The point is to have the singleton class as a glue between our controller(s) and the API client.

app/controllers/concerns/authentication.rb
```diff
@@ -59,8 +57,8 @@ def rbac_allowed?
     return true if ActiveModel::Type::Boolean.new.cast(Settings.disable_rbac)
     return valid_cert_auth? if identity_header.cert_based?
 
-    @rbac_api ||= ::RbacApi.new(raw_identity_header)
-    @rbac_api.check_user
+    set_authenticated_user
```
skateman, line 60: Do we really need this method? Can't be hidden lazily in the next line? Also, where it is defined? :thinking:

romanblanco: Without using the method, the `User.current` does not get initiated and so `current_user` is `nil`.

Defined right below:

```ruby
 66   def set_authenticated_user
 67     User.current = user
 68     Insights::API::Common::AuditLog.audit_with_account(
 69       current_user.account_number
 70     )
 71   end
```

skateman: Okay, Ignore my comment here :see_no_evil:

skateman, line 60: Thinking this further, why this has to be called explicitly when it already has a callsite? Won't that mess up things?

skateman: Look at how `around_action` calls this method in other places and reorganize things please.

skateman: Anyway, setting the auth user here can lead to pseudo-authorizations as `current_user` might be tested in future locations even if the user is not authorized to do something...

app/controllers/concerns/authentication.rb
```diff
@@ -59,8 +57,7 @@ def rbac_allowed?
     return true if ActiveModel::Type::Boolean.new.cast(Settings.disable_rbac)
     return valid_cert_auth? if identity_header.cert_based?
 
-    @rbac_api ||= ::RbacApi.new(raw_identity_header)
-    @rbac_api.check_user
+    user.authorized_to?(Rbac::COMPLIANCE_FULL_ACCESS)
```
marleystipich2, line 60: @romanblanco Are we planning on using the `authorized_to?` method for other things in the future besides only checking for full Compliance access? Just curious! I see that there are a lot of tests covering this and they all look great

marleystipich2 (approved):
This all looks great to me! Very easy to follow and the tests look really good. Thanks, @romanblanco!

app/controllers/concerns/authentication.rb
romanblanco: @marleystipich2 Exactly. As the next step of enabling RBAC, we'll have defined resources and actions _(according to the document attached to the story)_ and test users permission using the `authorized_to?` method.

e.g.:
```
current_user.authorized_to?('compliance:custom_role:create')
```

skateman (approved):
The Seal of Approval

PR #1242, opened 2022-03-30
fix(RHICOMPL-2851): nullify remediation_issue_id if no remediation

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/models/concerns/rule_remediation.rb
+++ b/app/models/concerns/rule_remediation.rb
@@ -9,6 +9,8 @@ module RuleRemediation
   #  `.includes(profiles: :benchmark)`
   #
   def remediation_issue_id
+    return nil unless remediation_available
+
     # FIXME: Nondetermenistic canonical selection
     profile = profiles.select(&:canonical?)&.first
     return unless profile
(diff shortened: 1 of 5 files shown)


vkrizan:
@skateman unit test? :grin:

skateman:
> @skateman unit test? grin

@vkrizan if Jenkins is on PTO, I can be a bit slower with unit tests :wink:

Victoremepunto:
@skateman  I'm wondering if we should leave the former field intact and simply expose the `remediation_available` rule attribute. That way we would avoid change the API spec (making the remediation_issue_id nullable).

skateman:
I don't want to adjust the API if we can avoid it.

skateman:
> @skateman This looks good to me but I see Victor's comment above and I'm kind of confused what the purpose of adding `nullable: true ` on the `remediation_issue_id` is? Was this necessary?

Some rules have no remediation available, their issue ID should be `nil`

PR #1246, opened 2022-04-04
feat(PolicyHost): RHICOMPL-2830 Check if host is supported

marleystipich2:
Still working on this and testing so leaving it as a draft for now. 
The purpose of these changes is to check that the host version is included in the list of supported os versions for the profile associated with the policy that it's being assigned to. 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/policy_host.rb
+++ b/app/models/policy_host.rb
@@ -8,10 +8,26 @@ class PolicyHost < ApplicationRecord
   validates :policy, presence: true
   validates :host, presence: true, on: :create
   validates :host_id, presence: true, uniqueness: { scope: :policy }
+  validate :host_supported?, on: :create
+
+  scope :supported_os_versions, lambda { |policy_id, host_os_major_version|
+    profile = Profile.find_by(policy_id: policy_id)
+    if host_os_major_version.to_s == profile&.os_major_version.to_s
+      profile&.supported_minor_versions
+    end
+  }
 
   def self.import_from_policy(policy_id, host_ids)
     import(host_ids.map do |host_id|
       { host_id: host_id, policy_id: policy_id }
     end)
+  end
+
+  def host_supported?
+    if PolicyHost.supported_os_versions(policy&.id, host&.os_major_version).include?(host&.os_minor_version.to_s)
+      return
+    end
+
+    errors.add(:host, 'os version is unsupported for this policy')
   end
 end
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -80,6 +80,16 @@ class Profile < ApplicationRecord
     end.flatten.uniq.sort.reverse
   end
 
+  def supported_minor_versions
+    benchmark_versions = self.class.canonical.where(
+      ref_id: ref_id,
+      upstream: false
+    ).joins(:benchmark).pluck('benchmarks.version')
+    SupportedSsg.by_os_major[os_major_version]
+                .select { |ssg| benchmark_versions.include? ssg.version }
+                .map(&:os_minor_version)
+  end
+
   private
 
   def bm_versions
--- a/test/controllers/v1/profiles_controller_test.rb
+++ b/test/controllers/v1/profiles_controller_test.rb
@@ -148,6 +148,7 @@ module V1
 
   class ProfilesControllerTest < ActionDispatch::IntegrationTest
     setup do
+      PolicyHost.any_instance.stubs(:host_supported?).returns(true)
       ProfilesController.any_instance.stubs(:authenticate_user).yields
       User.current = FactoryBot.create(:user)
     end
--- a/test/controllers/v1/systems_controller_test.rb
+++ b/test/controllers/v1/systems_controller_test.rb
@@ -5,6 +5,7 @@ require 'test_helper'
 module V1
   class SystemsControllerTest < ActionDispatch::IntegrationTest
     setup do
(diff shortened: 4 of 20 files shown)


marleystipich2:
@skateman I'm still working on this but here's what I have so far. I haven't set up a validation yet but would something like this be good to include along with a validation?

app/models/policy_host.rb
```diff
@@ -9,8 +9,19 @@ class PolicyHost < ApplicationRecord
   validates :host, presence: true, on: :create
   validates :host_id, presence: true, uniqueness: { scope: :policy }
 
+  scope :supported_host_ids, lambda { |policy_id, host_ids|
```
skateman, line 12: You'd need a `supported_os_versions` instead, maybe `supported_os_minors` is enough...

app/models/policy_host.rb
```diff
@@ -9,8 +9,19 @@ class PolicyHost < ApplicationRecord
...
+      host = Host.find_by(id: host_id)
+      if profile.supported_os_versions.include?(Gem::Version.new("#{host&.os_major_version}.#{host&.os_minor_version}"))
+        host_id
+      end
+    end.compact
+  }
+
   def self.import_from_policy(policy_id, host_ids)
-    import(host_ids.map do |host_id|
+    supported_host_ids = supported_host_ids(policy_id, host_ids)
```
skateman, line 23: We should not silently ignore, but fail if something is not supported. Something like:
```
validates :host, -> { supported_host_ids.include?(os_minor_version) }
```

skateman (changes requested):

app/models/policy_host.rb
marleystipich2: Good idea, this is kind of what I had in mind too for the validation.

app/models/policy_host.rb
```diff
@@ -8,10 +8,24 @@ class PolicyHost < ApplicationRecord
   validates :policy, presence: true
   validates :host, presence: true, on: :create
   validates :host_id, presence: true, uniqueness: { scope: :policy }
+  validate :host_supported?
```
skateman, line 11: I would move this into the `host` validation with a custom lambda because:
* it semantically belongs there and should error on it
* it should be validated only when creating (we should not break existing policies by preventing to update them)
* less lines :wink:

app/models/policy_host.rb
```diff
@@ -8,10 +8,24 @@ class PolicyHost < ApplicationRecord
...
+
+  scope :supported_versions, lambda { |policy_id|
+    profile = Profile.find_by(policy_id: policy_id)
+    profile&.supported_os_versions&.map(&:version)
```
skateman, line 15: Be careful about `supported_os_versions` it's heavily inefficient if queried directly on profiles as it will produce a cache miss [here](https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/profile.rb#L87). The method was optimized for calling it from GraphQL [underneath](https://github.com/RedHatInsights/compliance-backend/blob/master/app/models/os_major_version.rb#L56) the `OsMajorVersion` pseudo-model.

You can either recreate somehow that query optimized in your context, or come up with something else to make it fast.

app/models/policy_host.rb
```diff
@@ -8,10 +8,24 @@ class PolicyHost < ApplicationRecord
...
+  }
 
   def self.import_from_policy(policy_id, host_ids)
     import(host_ids.map do |host_id|
       { host_id: host_id, policy_id: policy_id }
     end)
   end
+
+  def host_supported?
+    if PolicyHost.supported_versions(policy&.id).include?("#{host&.os_major_version}.#{host&.os_minor_version}")
```
skateman, line 25: Your context is already reduced by the major version, so when you construct the query above in a custom way, you can focus only on the minor version IMO.

skateman (changes requested):

app/models/policy_host.rb
marleystipich2: I thought about this but figured to be safe I'd check the major version too because when I was testing, it was letting me create a PolicyHost with a host that wasn't supported based on the `supported_os_versions` of the profile but when I only checked the minor version, it was thinking that it was supported. But maybe I don't need to test this scenario since it's already constrained by the major version anyway?

app/models/profile.rb
```diff
@@ -80,6 +80,16 @@ def supported_os_versions
...
 
+  def supported_os_versions_by_os_major
+    benchmarks = self.class.canonical.where(
+      ref_id: ref_id,
+      upstream: false
+    ).joins(:benchmark).pluck('benchmarks.version')
+    SupportedSsg.by_os_major[os_major_version].select { |ssg| benchmarks.include? ssg.version }.map do |ssg|
+      "#{ssg.os_major_version}.#{ssg.os_minor_version}"
+    end.uniq
+  end
```
marleystipich2, line 91: @skateman I spent a good amount of time looking into ways to make a query that's more efficient than calling the `bm_versions` method and this was the best I could come up with.. It is a little faster technically 🥲 and we're not caching anything by doing it this way.

PR #1251, opened 2022-04-06
feat(PolicyHost): RHICOMPL-2830 Check if host is supported

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -94,6 +94,15 @@ class Policy < ApplicationRecord
     audit_bo_autoremove(removed_bos)
   end
 
+  def supported_os_minor_versions
+    profile_os_major_version = initial_profile.os_major_version
+    Xccdf::Benchmark.including_profile(initial_profile).flat_map do |bm|
+      SupportedSsg.by_ssg_version[bm.version]
+                  .select { |ssg| ssg.os_major_version == profile_os_major_version }
+                  .map(&:os_minor_version)
+    end
+  end
+
   private
 
   def unassigned_minor_versions
--- a/app/models/policy_host.rb
+++ b/app/models/policy_host.rb
@@ -8,10 +8,29 @@ class PolicyHost < ApplicationRecord
   validates :policy, presence: true
   validates :host, presence: true, on: :create
   validates :host_id, presence: true, uniqueness: { scope: :policy }
+  validate :host_supported?, on: :create
 
   def self.import_from_policy(policy_id, host_ids)
     import(host_ids.map do |host_id|
       { host_id: host_id, policy_id: policy_id }
     end)
+  end
+
+  private
+
+  def host_supported?
+    if !os_major_supported?
+      errors.add(:host, 'Unsupported OS major version')
+    elsif !os_minor_supported?
+      errors.add(:host, 'Unsupported OS minor version')
+    end
+  end
+
+  def os_major_supported?
+    !host&.os_major_version.nil? && host&.os_major_version.to_s == policy&.initial_profile&.os_major_version.to_s
+  end
+
+  def os_minor_supported?
+    policy&.supported_os_minor_versions&.include?(host&.os_minor_version.to_s)
   end
 end
--- a/test/graphql/mutations/associate_profiles_test.rb
+++ b/test/graphql/mutations/associate_profiles_test.rb
@@ -15,6 +15,7 @@ class AssociateProfilesMutationTest < ActiveSupport::TestCase
   GRAPHQL
 
   setup do
+    PolicyHost.any_instance.stubs(:host_supported?).returns(true)
     @user = FactoryBot.create(:user)
     @profiles = FactoryBot.create_list(:profile, 2, account: @user.account)
     @host = FactoryBot.create(:host, account: @user.account.account_number)
--- a/test/models/policy_host_test.rb
+++ b/test/models/policy_host_test.rb
@@ -5,4 +5,71 @@ require 'test_helper'
 class PolicyHostTest < ActiveSupport::TestCase
   should belong_to(:policy)
   should validate_presence_of(:host)
+
+  setup do
+    account = FactoryBot.create(:user).account
+
+    supported_ssg1 = SupportedSsg.new(version: '0.1.50',
+                                      os_major_version: '7', os_minor_version: '1')
+    supported_ssg2 = SupportedSsg.new(version: '0.1.52',
+                                      os_major_version: '7', os_minor_version: '9')
+
+    SupportedSsg.stubs(:all).returns([supported_ssg1, supported_ssg2])
+
+    bm1 = FactoryBot.create(
+      :benchmark,
+      version: supported_ssg1.version,
+      os_major_version: '7'
+    )
+
+    bm2 = FactoryBot.create(
+      :benchmark,
+      version: supported_ssg2.version,
+      os_major_version: '7'
+    )
+
+    @policy1 = FactoryBot.create(:policy, account: account)
+    @policy2 = FactoryBot.create(:policy, account: account)
+
+    FactoryBot.create(:profile, policy: @policy1, account: account, benchmark: bm1)
+    FactoryBot.create(:profile, policy: @policy2, account: account, benchmark: bm2)
+
+    @policy1.stubs(:supported_os_minor_versions).returns(['1'])
+    @policy2.stubs(:supported_os_minor_versions).returns(['9'])
+
+    @host1 = Host.find(FactoryBot.create(
+      :host,
+      account: account.account_number,
+      os_major_version: 7,
+      os_minor_version: 1
+    ).id)
+
+    @host2 = Host.find(FactoryBot.create(
+      :host,
+      account: account.account_number,
+      os_major_version: 8,
+      os_minor_version: 3
+    ).id)
+  end
+
+  test 'an unsupported host based on minor version cannot be assigned to a policy' do
+    exception = assert_raises(Exception) do
+      @policy2.hosts << @host1
+    end
+    assert_equal(exception.message, 'Validation failed: Host Unsupported OS minor version')
+  end
+
+  test 'an unsupported host based on major version cannot be assigned to a policy' do
+    exception = assert_raises(Exception) do
+      @policy1.hosts << @host2
+    end
+    assert_equal(exception.message,
+                 'Validation failed: Host Unsupported OS major version')
+  end
+
+  test 'a supported host can be assigned to a policy' do
+    assert_nothing_raised do
+      @policy1.hosts << @host1
+    end
+  end
 end
--- a/app/models/xccdf/benchmark.rb
+++ b/app/models/xccdf/benchmark.rb
@@ -23,6 +23,12 @@ module Xccdf
     validates :ref_id, uniqueness: { scope: %i[version] }, presence: true
     validates :version, presence: true
 
+    scope :including_profile, lambda { |profile|
+      joins(:profiles).where(profiles: { parent_profile_id: nil,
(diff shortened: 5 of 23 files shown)


app/models/profile.rb
```diff
@@ -80,6 +80,16 @@ def supported_os_versions
     end.flatten.uniq.sort.reverse
   end
 
+  def supported_minor_versions
+    benchmark_versions = self.class.canonical.where(
+      ref_id: ref_id,
+      upstream: false
+    ).joins(:benchmark).pluck('benchmarks.version')
```
skateman, line 87: This could be extracted into a method called `supported_benchmark_versions`.

app/models/policy_host.rb
```diff
@@ -8,10 +8,26 @@ class PolicyHost < ApplicationRecord
   validates :policy, presence: true
   validates :host, presence: true, on: :create
   validates :host_id, presence: true, uniqueness: { scope: :policy }
+  validate :host_supported?, on: :create
+
+  scope :supported_os_versions, lambda { |policy_id, host_os_major_version|
+    profile = Profile.find_by(policy_id: policy_id)
+    if host_os_major_version.to_s == profile&.os_major_version.to_s
```
skateman, line 15: No, this should definitely not be part of the scope, ideally it should live in the `host_supported?` method.

app/models/policy_host.rb
```diff
@@ -8,10 +8,26 @@ class PolicyHost < ApplicationRecord
...
       { host_id: host_id, policy_id: policy_id }
     end)
   end
+
+  def host_supported?
+    if PolicyHost.supported_os_versions(policy&.id, host&.os_major_version).include?(host&.os_minor_version.to_s)
+      return
+    end
+
+    errors.add(:host, 'os version is unsupported for this policy')
```
skateman, line 31: Using a scope here is a bit of an antipattern, especially in the same exact class. My oversimplified idea would be something like:

```suggestion
    errors.add(:host, 'Unsupported OS major version') unless host.os_major_version == initial_profile.os_major_version
    
    errors.add(:host, 'Unsupported OS minor version') if SupportedSsg.ssg_versions_for_os(host.os_major_version, host.os_minor_version).include?(supported_benchmark_versions)
```

skateman (changes requested):

app/models/policy_host.rb
marleystipich2: Yep, I like this idea! Thanks

app/models/policy_host.rb
```diff
@@ -8,10 +8,25 @@ class PolicyHost < ApplicationRecord
...
+    errors.add(:host, 'Unsupported OS major version') unless host&.os_major_version.to_s ==
+                                                             profile&.os_major_version.to_s
+
+    errors.add(:host, 'Unsupported OS minor version') unless supported_os_minor_version(profile)
+  end
+
+  def supported_os_minor_version(profile)
+    ssg_supported_versions = SupportedSsg.ssg_versions_for_os(host&.os_major_version, host&.os_minor_version).uniq.sort
+    supported_versions_intersect = ((profile&.supported_benchmark_versions || []) & ssg_supported_versions)&.sort
+    supported_versions_intersect == ssg_supported_versions
```
skateman, line 30: This seems to me a bit suspicious, first of all you have a lot of `uniq` ans `sort` calls everywhere and the `|| []` should not be necessary. Maybe you can use the `ssg_supported_versions` when constructing the list of `supported_benchmarks` and just check if the result is empty. I need to look into this more, but my guess is that the complexity can be lowered here significantly.

test/graphql/mutations/associate_profiles_test.rb
```diff
@@ -15,6 +15,7 @@ class AssociateProfilesMutationTest < ActiveSupport::TestCase
   GRAPHQL
 
   setup do
+    PolicyHost.any_instance.stubs(:host_supported?).returns(true)
```
skateman, line 18: When you know the `@host` instance, you can just mock the `:host_supported?` method on it directly. Calling the `any_instance` method is costly as it has to maintain a list of the instances and constantly traverse the object space.

skateman (changes requested):

test/graphql/mutations/associate_profiles_test.rb
marleystipich2: Okay, thanks! I didn't know I could call it on the `@host` instance because `host_supported?` is a `PolicyHost` method but I'll do that instead

app/models/policy_host.rb
marleystipich2: Yeah, this was a little strange because sometimes `profile` is nil so `profile&.supported_benchmark_versions` would return nil and then this would mess up the intersection of `profile&.supported_benchmark_versions & ssg_supported_versions` because instersecting `nil & [..]` returns `false`
But I see what you mean.. I think I can make this less complex by doing something like what you suggested.

test/graphql/mutations/associate_profiles_test.rb
marleystipich2: It didn't work with the `@host` instance and I think it might be for the reason that I said in the comment above? I think the instance that the test would need to use is the `PolicyHost` instance where host is the`@host' and policy is whatever policy is being used in the tests

app/models/policy.rb
```diff
@@ -94,8 +94,23 @@ def destroy_orphaned_business_objective
     audit_bo_autoremove(removed_bos)
   end
 
+  def supported_os_minor_versions
+    profile_os_major_version = initial_profile.os_major_version
+    supported_benchmarks.map do |bm|
+      SupportedSsg.by_ssg_version[bm.version]
+                  .select { |ssg| ssg.os_major_version == profile_os_major_version }
+                  .map(&:os_minor_version)
+    end.flatten.uniq
```
skateman, line 103: No need for the `uniq`, doesn't matter how many times the value is there if it's there.

```suggestion
    end.flatten
```

app/models/policy.rb
```diff
@@ -94,8 +94,23 @@ def destroy_orphaned_business_objective
     audit_bo_autoremove(removed_bos)
   end
 
+  def supported_os_minor_versions
+    profile_os_major_version = initial_profile.os_major_version
+    supported_benchmarks.map do |bm|
```
skateman, line 99: Maybe you can get away with a `flat_map` and then you don't need the `flatten`. Not sure, please check...

```suggestion
    supported_benchmarks.flat_map do |bm|
```

app/models/policy.rb
```diff
@@ -94,8 +94,23 @@ def destroy_orphaned_business_objective
...
+    end.flatten.uniq
+  end
+
   private
 
+  def supported_benchmarks
+    Xccdf::Benchmark.joins(:profiles).where(profiles: { parent_profile_id: nil,
+                                                        ref_id: initial_profile.ref_id,
+                                                        upstream: false })
+  end
```
skateman, line 112: This is something we could extract into a scope, under the benchmarks, with profile as a parameter. Then it could be called above as:
```
Xccdf::Benchmark.including_profile(initial_profile)
```

skateman (changes requested):
Looks good, just minor code management nitpicks :+1:

app/models/policy_host.rb
```diff
@@ -8,10 +8,27 @@ class PolicyHost < ApplicationRecord
...
+
+  def host_supported?
+    if !host_major_supported?
+      errors.add(:host, 'Unsupported OS major version')
+    elsif !host_minor_supported?
+      errors.add(:host, 'Unsupported OS minor version')
+    end
+  end
+
+  def host_major_supported?
```
skateman, line 27:
```suggestion
  def os_major_supported?
```

app/models/policy_host.rb
```diff
@@ -8,10 +8,27 @@ class PolicyHost < ApplicationRecord
...
+    !host&.os_major_version.nil? && host&.os_major_version.to_s == policy&.initial_profile&.os_major_version.to_s
+  end
+
+  def host_minor_supported?
```
skateman, line 31:
```suggestion
  def is_minor_supported?
```

test/models/policy_host_test.rb
```diff
@@ -5,4 +5,65 @@
 class PolicyHostTest < ActiveSupport::TestCase
   should belong_to(:policy)
   should validate_presence_of(:host)
+
+  setup do
+    account1 = FactoryBot.create(:user).account
+    account2 = FactoryBot.create(:user).account
```
skateman, line 11: This is hinting me that you're not testing what you should be testing. You have 2 separate (user) accounts for 2 separate policies and are trying to tangle assignments across users. That will never work as per pundit authorization...

test/models/policy_host_test.rb
```diff
@@ -5,4 +5,65 @@
...
+    )
+
+    @policy1 = FactoryBot.create(:policy, account: account1)
+    @policy2 = FactoryBot.create(:policy, account: account2)
+    parent_profile1 = FactoryBot.create(:canonical_profile, upstream: false, benchmark: bm1)
+    parent_profile2 = FactoryBot.create(:canonical_profile, upstream: false, benchmark: bm2)
+    profile1 = FactoryBot.create(:profile, parent_profile: parent_profile1, policy: @policy1, account: account1)
+    profile2 = FactoryBot.create(:profile, parent_profile: parent_profile2, policy: @policy2, account: account2)
+    profile1.update!(ref_id: parent_profile1.ref_id)
+    profile2.update!(ref_id: profile2.ref_id)
```
skateman, line 33: You don't need to create and set the canonical profile, it all generates automatically. Just pass the benchmarks to the profiles below.

test/models/policy_host_test.rb
```diff
@@ -5,4 +5,65 @@
 class PolicyHostTest < ActiveSupport::TestCase
   should belong_to(:policy)
   should validate_presence_of(:host)
+
+  setup do
+    account1 = FactoryBot.create(:user).account
+    account2 = FactoryBot.create(:user).account
+    v1 = SupportedSsg.by_os_major['7'].first
```
skateman, line 12: This is unreliable, I think it should be rather hardcoded or mocked.

test/models/policy_host_test.rb
```diff
@@ -5,4 +5,65 @@
...
+    bm1 = FactoryBot.create(
+      :benchmark,
+      version: v1.version,
+      os_major_version: '7'
+    )
+
+    v2 = SupportedSsg.by_os_major['7'].last
```
skateman, line 19: Same here, I'd suggest you to mock the `SupportedSsg` completely. You'll find examples at other places.

skateman (changes requested):
Please think a bit more about the semantics of the tests you created. Ideally, they should be always failing without your other changes in the codebase.

app/models/policy_host.rb
```diff
@@ -8,10 +8,27 @@ class PolicyHost < ApplicationRecord
...
+    end
+  end
+
+  def os_major_supported?
+    !host&.os_major_version.nil? && host&.os_major_version.to_s == policy&.initial_profile&.os_major_version.to_s
+  end
+
+  def os_minor_supported?
+    policy&.supported_os_minor_versions&.include?(host&.os_minor_version.to_s)
+  end
```
skateman, line 33: These should be all `private`

skateman (approved):
The Seal of Approval

PR #1273, opened 2022-04-22
feat(tailoring): RHICOMPL-2554 Add conflict resolving for rules

marleystipich2:
Note: The XCCDF specifications says to only deselect the item that has a conflict and to not touch any other items in the Benchmark(so don't try to resolve the conflict by selecting or deselecting other items in the Benchmark)
```
Also, <xccdf:requires> and
<xccdf:conflicts> elements SHALL only change their parent item; they SHALL NOT modify
other items in the <xccdf:Benchmark>. 
```
```
Rules and Groups may contain any number of <xccdf:requires> and <xccdf:conflicts>
elements and if any of these elements do not evaluate to true, then that item SHALL become deselected. 
```

Here is an example of how conflicts are resolved: 
Screen Shot 2022-04-26 at 2 45 59 PM

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/concerns/profile_tailoring.rb
+++ b/app/models/concerns/profile_tailoring.rb
@@ -33,4 +33,13 @@ module ProfileTailoring
 
     update!(os_minor_version: version)
   end
+
+  def tailoring_valid?
+    conflicts, requires = %w[conflicts requires].map do |relationship|
+      with_relationships(relationship, rules, hierarchical_rule_groups).map(&:right)
+    end
+
+    rules_and_groups = rules | hierarchical_rule_groups
+    (requires - rules_and_groups).empty? && (conflicts & rules_and_groups).empty?
+  end
 end
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -30,8 +30,8 @@ class Rule < ApplicationRecord
 
   has_many :profile_rules, dependent: :delete_all
   has_many :profiles, through: :profile_rules, source: :profile
-  has_many :rule_group_rules, dependent: :delete_all
-  has_many :rule_groups, through: :rule_group_rules
+  has_one :rule_group_rule, dependent: :delete
+  has_one :rule_group, through: :rule_group_rule
   has_many :rule_results, dependent: :delete_all
   has_many :hosts, through: :rule_results, source: :host
   has_many :rule_references_rules, dependent: :delete_all
@@ -73,9 +73,7 @@ class Rule < ApplicationRecord
     joins(:profiles).where(profiles: { id: Profile.canonical })
   }
 
-  scope :without_rule_group_parent, lambda {
-    where.missing(:rule_groups)
-  }
+  scope :without_rule_group_parent, -> { where.missing(:rule_group) }
 
   scope :joins_identifier, lambda {
     left_outer_joins(:rule_identifier).select('rules.*', RuleIdentifier::AS_JSON.as('identifier'))
--- a/app/models/rule_group_relationship.rb
+++ b/app/models/rule_group_relationship.rb
@@ -11,6 +11,6 @@ class RuleGroupRelationship < ApplicationRecord
   validates :left_id, presence: true, uniqueness: { scope: %i[relationship right_id right_type left_type] }
 
   scope :with_relationships, lambda { |rules_or_groups, relationship|
-    where(relationship: relationship).includes(:left, :right).where(left: rules_or_groups)
+    where(relationship: relationship, left: rules_or_groups).includes(:left, :right)
   }
 end
--- a/app/models/concerns/rules_and_rule_groups.rb
+++ b/app/models/concerns/rules_and_rule_groups.rb
@@ -31,14 +31,19 @@ module RulesAndRuleGroups
 
     private
 
-    def with_relationships(relationship)
+    def hierarchical_rule_groups
+      RuleGroup.where(id: rules.includes(:rule_group)
+               .references(:rule_group).flat_map { |r| r.rule_group&.path_ids })
+    end
+
+    def with_relationships(relationship, rules, rule_groups)
       RuleGroupRelationship.with_relationships(rules, relationship).or(
         RuleGroupRelationship.with_relationships(rule_groups, relationship)
       )
     end
 
     def relationships_for(relationship)
-      with_relationships(relationship).each_with_object({}) do |rgr, relationships|
+      with_relationships(relationship, rules, rule_groups).each_with_object({}) do |rgr, relationships|
         relationships[rgr.left] ||= []
         relationships[rgr.left] << rgr.right
       end
--- a/test/models/profile_test.rb
+++ b/test/models/profile_test.rb
@@ -848,6 +848,27 @@ class ProfileTest < ActiveSupport::TestCase
       @profile.update! rules: [@rule2, @rule3]
       @rule2.update(precedence: 4)
       @rule3.update(precedence: 1)
+
+      @policy2 = FactoryBot.create(:policy, account: @account, hosts: [@host])
+
+      @profile2 = @parent.clone_to(
+        account: @account,
+        policy: @policy2
+      )
+
+      @rule4 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+      @rule5 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+      @rule6 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+      @rule7 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+
+      @rule_group1 = FactoryBot.create(:rule_group)
+      @rule_group2 = FactoryBot.create(:rule_group)
+      @rule_group3 = FactoryBot.create(:rule_group)
+
+      @rule_group1.update!(parent_id: @rule_group2.id)
+      @rule4.rule_group = @rule_group1
+
+      @profile2.update! rules: [@rule4, @rule5, @rule6, @rule7]
     end
 
     should 'send the correct rule ref ids to the tailoring file service' do
@@ -861,6 +882,101 @@ class ProfileTest < ActiveSupport::TestCase
 
     should 'properly detects removed_rules' do
       assert_equal [@rule1], @profile.removed_rules
+    end
+
+    should 'return tailoring_valid? false due to conflict between rule_group and rule' do
+      rgr = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_group_conflicts)
+      rgr.update!(left: @rule_group2, right: @rule7)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? false due to conflict between rule_groups' do
+      rule_group4 = FactoryBot.create(:rule_group)
+      @rule5.rule_group = rule_group4
+      rgr = FactoryBot.create(:rule_group_relationship, :for_rule_group_and_rule_group_conflicts)
+      rgr.update!(left: @rule_group2, right: rule_group4)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? false due to conflict between rules' do
+      rgr = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_conflicts)
+      rgr.update!(left: @rule4, right: @rule5)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? false due to missing requirement between rule_group and rule' do
+      rule8 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+      rgr = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_group_requires)
+      rgr.update!(left: @rule_group2, right: rule8)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? false due to missing requirement between rule_groups' do
+      rule_group4 = FactoryBot.create(:rule_group)
+      rgr = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_group_requires)
+      rgr.update!(left: @rule_group2, right: rule_group4)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? false due to missing requirement between rules' do
+      rule8 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+      rgr = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_requires)
+      rgr.update!(left: @rule4, right: rule8)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? false with duplicate requirement' do
+      rule8 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+      rgr1 = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_requires)
+      rgr2 = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_requires)
+      rgr1.update!(left: @rule4, right: rule8)
+      rgr2.update!(left: @rule5, right: rule8)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? false with duplicate conflict' do
+      rgr1 = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_conflicts)
+      rgr2 = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_conflicts)
+      rgr1.update!(left: @rule4, right: @rule5)
+      rgr2.update!(left: @rule6, right: @rule5)
+
+      assert_equal false, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? true when conflicts and requirements are satisfied' do
+      rule8 = FactoryBot.create(:rule, benchmark: @parent.benchmark)
+      rule_group4 = FactoryBot.create(:rule_group)
+      rule_group5 = FactoryBot.create(:rule_group)
+      @rule5.rule_group = rule_group4
+
+      rgr1 = FactoryBot.create(:rule_group_relationship, :for_rule_group_and_rule_conflicts)
+      rgr2 = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_requires)
+      rgr3 = FactoryBot.create(:rule_group_relationship, :for_rule_and_rule_group_requires)
+      rgr4 = FactoryBot.create(:rule_group_relationship, :for_rule_group_and_rule_group_requires)
+      rgr5 = FactoryBot.create(:rule_group_relationship, :for_rule_group_and_rule_group_conflicts)
+      rgr6 = FactoryBot.create(:rule_group_relationship, :for_rule_group_and_rule_group_requires)
+      rgr7 = FactoryBot.create(:rule_group_relationship, :for_rule_group_and_rule_group_conflicts)
+
+      rgr1.update!(left: @rule_group2, right: rule8)
+      rgr2.update!(left: @rule4, right: @rule6)
+      rgr3.update!(left: @rule4, right: @rule_group2)
+      rgr4.update!(left: @rule_group2, right: rule_group4)
+      rgr5.update!(left: @rule_group1, right: rule_group5)
+      rgr6.update!(left: @rule7, right: @rule6)
+      rgr7.update!(left: @rule7, right: rule8)
+
+      assert_equal true, @profile2.tailoring_valid?
+    end
+
+    should 'return tailoring_valid? true when there are no relationships' do
+      assert_equal true, @profile2.tailoring_valid?
     end
   end
 


InsightsDroid:
7687eb30-463c-4d23-9ade-c9a9ce3f7d08
b3bf17cd-21b6-4237-acb2-90065c954de0

InsightsDroid:
7c8b811a-401a-43ba-8e1e-4b9440906a6f
c4f7eedd-d99a-4903-a382-c0429d4c69d3

app/services/xccdf_tailoring_file.rb
```diff
@@ -46,11 +46,30 @@ def profile_builder(xml)
...
-    @rule_ref_ids.each do |rule_ref_id, selected|
+    resolved_rules.each do |rule_ref_id, selected|
       xml[XCCDF].select(idref: rule_ref_id, selected: selected)
     end
   end
 
+  def resolved_rules
+    @rule_ref_ids.select { |_id, selected| selected }.each do |rule_ref_id, _selected|
+      rule_for(rule_ref_id)&.left_rule_group_relationships&.each do |rgr|
+        @rule_ref_ids[rule_ref_id] = false if conflict?(rgr)
```
skateman, line 57: I think that the algorithm will be way more complex than this:
* where are you handling the requires?
* the text cited above mentions parent items deselection, but we don't keep track of parent items, so we need to populated that down into the siblings in the given group...

Anyway, this needs a bit more work.

skateman (changes requested):

app/services/xccdf_tailoring_file.rb
marleystipich2: The requires is handled in the `conflict?` method but I guess that naming for that method is confusing. By "conflict" I just meant any conflict that would cause issues for the Rule like a requirement isn't satisfied but I could rename it

When it said parent item in the description, I thought that it meant the parent of the `<requires>` or `<conflicts>` element which would just be the Rule? Not the parent of the Rule with the `conflicts` or `requires` element. At least that's what it sounds like to me... Because it mentions to never update any other item anywhere else in the Benchmark besides the one that's currently being evaluated

skateman: The parent item is the group I guess, if you look at the SCAP workbench, it will show the tree structure with the parent item.

marleystipich2: I put this in the description too but here's an example of how conflicts are resolved, from the xccdf specfications:
Screen Shot 2022-04-26 at 2 45 59 PM

Step 4 shows the process of evaluating `Rule2` which requires `Rule3`. Since `Rule3` isn't selected, `Rule2` is unselected and this is how the conflict is resolved. 

I think that when it says `Also <xccdf:requires> and <xccdf:conflicts> elements SHALL only change their parent item;` means the parent of the requires or conflicts element so in this example `<xccdf:Rule id="xccdf_org.example_rule_Rule2" selected="true"> ` is the parent of `<xccdf:requires idref="xccdf_org.example_rule_Rule3"> `

marleystipich2: Also, I don't think Groups can technically be selected yet? Like it isn't possible to select a Group on the front end? So until we add that functionality, is evaluating only Rules for now good enough? 
Or I can update it to evaluate Groups in the conflict resolving algortihm too but for now, selected will always be false for every Group.

Victoremepunto: /retest

InsightsDroid:
2e0b000b-85cd-4582-b2c1-7c5849fee0e6
b754286b-d13b-4c74-bed7-13bd0e629d70

app/services/xccdf_tailoring_file.rb
```diff
@@ -46,11 +46,30 @@ def profile_builder(xml)
...
-    @rule_ref_ids.each do |rule_ref_id, selected|
+    resolved_rules.each do |rule_ref_id, selected|
       xml[XCCDF].select(idref: rule_ref_id, selected: selected)
     end
   end
 
+  def resolved_rules
+    @rule_ref_ids.select { |_id, selected| selected }.each do |rule_ref_id, _selected|
+      rule_for(rule_ref_id)&.left_rule_group_relationships&.each do |rgr|
+        @rule_ref_ids[rule_ref_id] = false if deselect?(rgr)
```
romanblanco, line 57: Is `@rule_ref_ids[rule_ref_id]` always `true` by default?
Shouldn't we explicitly assign `@rule_ref_ids[rule_ref_id] = true` in case if `deselect?` is `false`?

marleystipich2: Sorry, I missed this yesterday! `@rule_ref_ids[rule_ref_id]` is initially set to either true or false when it's passed into the initialize method in this file so if we don't set it to false at this line, it'll just keep its original value.

marleystipich2:
@skateman I pushed what I have so far for this because it seems like it's working but I need to write tests and still want to see if there are ways that I can make it better. There aren't any N+1 queries and it's pretty fast but I'm sure there are still areas that can be improved.
cc @romanblanco

app/models/rule.rb
```diff
@@ -118,4 +118,9 @@ def latest_result(host, profile)
 
     test_result.rule_results.find_by(rule_id: id)
   end
+
+  def rule_group_ancestor_ids
+    # Rule should only have one rule_group parent
+    rule_groups.first.path_ids
+  end
```
marleystipich2, line 125: @skateman I think `rule_groups.first` should be good because it's impossible for a rule to have more than one rule_group parent which is what the `rule.rule_groups` association is

skateman: I'm not entirely sure about that, can't a rule belong to multiple rule groups across different profiles? You should verify in the XCCDF specification and if this is true, you can rewrite the `has_many :rule_groups` to `has_one :rule_groups` which would render the array into a single element, thus not needing the `#first`.

marleystipich2: The hierarchical structure of rules and rule groups is the same for the entire benchmark so a `rule` will always have the same parent `rule_group` even if it belongs to different profiles under the benchmark. This is because the `Rule`/`Group` structure is determined by how the `<xccdf-1.2:Rule/>` and `<xccdf-1.2:Group>` tags are nested under each other in the datastream files which is just defined once for the entire benchmark. The rules are just "selected" by a Profile with the line `<xccdf-1.2:select idref="xccdf_org.ssgproject.content_rule_selinux_policytype" selected="true"/>` and groups are unselected with this line `<xccdf-1.2:select idref="xccdf_org.ssgproject.content_group_bootloader-zipl" selected="false"/>`.

But yeah, that's a good idea. I'll update it from `has_many` to `has_one` and that'll work better

skateman: This might not work and would need simplification in the model as the RuleGroupRule table is obsolete. Try it and if it doesn't work, keep the `first`. We should surely do a remodel later, please create a ticket for it into the epic.

marleystipich2: Ah right, that's true. I just updated these 2 associations on Rule from `has_many` to `has_one` and it works correctly: 
```
has_one :rule_group_rule, dependent: :delete
has_one :rule_group, through: :rule_group_rule
```
But I can still create a ticket to do a remodel later if this is okay with you

skateman: Yes

marleystipich2: Just created a ticket for that and added it to the epic

app/models/rule.rb
```diff
@@ -30,8 +30,8 @@ class Rule < ApplicationRecord
 
   has_many :profile_rules, dependent: :delete_all
   has_many :profiles, through: :profile_rules, source: :profile
-  has_many :rule_group_rules, dependent: :delete_all
-  has_many :rule_groups, through: :rule_group_rules
+  has_one :rule_group_rule, dependent: :delete
+  has_one :rule_group, through: :rule_group_rule
```
skateman, line 34: Yes, this is the evil one that has to be remodeled :see_no_evil: `has_one through:` :laughing:

marleystipich2: Definitely 🙈 I already created the ticket 🙂

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,13 @@ def update_os_minor_version(version)
 
     update!(os_minor_version: version)
   end
+
+  def tailoring_valid?
+    rules_and_groups = rules + RuleGroup.where(id: rules.includes(:rule_group).flat_map(&:rule_group_ancestor_ids))
+
+    conflicts = RuleGroupRelationship.where(left: rules_and_groups, relationship: 'conflicts').map(&:right).uniq
+    requires = RuleGroupRelationship.where(left: rules_and_groups, relationship: 'requires').map(&:right).uniq
```
skateman, line 41: This can be definitely simplified, you can do `.select(:right).distinct.map(&:right)` which should be faster as the `uniq` is done in SQL. But I'd move it partially to a scope :thinking: except the map part maybe?

marleystipich2: Yeah, that should work! I think leaving the map part out of the scope is best too because it'll keep it more general so we can use it again in the future if we need.

app/models/rule_group_relationship.rb
```diff
@@ -13,4 +13,8 @@ class RuleGroupRelationship < ApplicationRecord
   scope :with_relationships, lambda { |rules_or_groups, relationship|
     where(relationship: relationship).includes(:left, :right).where(left: rules_or_groups)
   }
+
+  scope :with_distinct_relationships, lambda { |rules_or_groups, relationship|
+    where(left: rules_or_groups, relationship: relationship).select(:right_id, :right_type).distinct
```
marleystipich2, line 18: @skateman `.select(:right)` doesn't work so I did `.select(:right_id, :right_type)` but this is faster than using `uniq` 👍

skateman: No need to name it `_distinct`, enough to have it as `with_relationships` as it's shorter.

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,13 @@ def update_os_minor_version(version)
 
     update!(os_minor_version: version)
   end
+
+  def tailoring_valid?
+    rules_and_groups = rules + RuleGroup.where(id: rules.includes(:rule_group).flat_map(&:rule_group_ancestor_ids))
+
+    conflicts = RuleGroupRelationship.with_distinct_relationships(rules_and_groups, 'conflicts').map(&:right)
+    requires = RuleGroupRelationship.with_distinct_relationships(rules_and_groups, 'requires').map(&:right)
```
skateman, line 41: I still don't like this part, but will come up to make it nicer during the day

LightOfHeaven1994:
LGTM from my side

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,13 @@ def update_os_minor_version(version)
 
     update!(os_minor_version: version)
   end
+
+  def tailoring_valid?
+    rules_and_groups = rules + RuleGroup.where(id: rules.includes(:rule_group).flat_map(&:rule_group_ancestor_ids))
```
skateman, line 38: Thinking about converting this to a union, this way the `rules_and_groups` would be an SQL statement and not resolved to an array. Just not sure about the `rule_group_ancestor_ids`.

Why is the `flat_map` needed?

app/models/rule.rb
```diff
@@ -30,8 +30,8 @@ class Rule < ApplicationRecord
 
   has_many :profile_rules, dependent: :delete_all
   has_many :profiles, through: :profile_rules, source: :profile
-  has_many :rule_group_rules, dependent: :delete_all
-  has_many :rule_groups, through: :rule_group_rules
+  has_one :rule_group_rule, dependent: :delete
```
skateman, line 33: Not sure about this, so I created https://issues.redhat.com/browse/RHICOMPL-3401

app/models/rule_group_relationship.rb
marleystipich2: There already is a `with_relationships` scope :/ I was going to just add the `.distinct` part onto that one but we're already using `with_relationships` and we don't only want the distinct relationships so that's why I had to create a separate scope for this

skateman: As I see it's similar, so you could just do `with_relationship(...).distinct` or something like that instead. This way you don't need the separate scope.

app/models/concerns/profile_tailoring.rb
marleystipich2: Can you union Rule objects with RuleGroup objects using SQL? I thought that an array was the only way to combine the Rules and RuleGroups together since they're not the same type of object 

The `flat_map` is needed because `rule_group_ancestor_ids` returns an array of RuleGroup ids and it's being called for every `rule` in `rules` so if we did `rules.includes(:rule_group).map(&:rule_group_ancestor_ids)` instead of `flat_map`, it'd look something like this `[[id1, id2], [id3, id4, id5], [id6]] `

app/models/rule.rb
marleystipich2: Is this a duplicate of this ticket that I created? Maybe we can combine them? https://issues.redhat.com/browse/RHICOMPL-3384

app/models/rule_group_relationship.rb
marleystipich2: Ah right, I'll just do that instead

app/models/concerns/profile_fields.rb
```diff
@@ -62,10 +62,14 @@ def bm_versions
...
-    def relationships_for(relationship)
+    def relationships_for(rules, rule_groups, relationship)
       RuleGroupRelationship.with_relationships(rules, relationship).or(
         RuleGroupRelationship.with_relationships(rule_groups, relationship)
-      ).each_with_object({}) do |rgr, relationships|
+      )
+    end
+
+    def map_rule_group_relationships(relationship)
+      relationships_for(rules, rule_groups, relationship).each_with_object({}) do |rgr, relationships|
```
skateman, line 72: Is this because of rubocop complaining?

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,19 @@ def update_os_minor_version(version)
 
     update!(os_minor_version: version)
   end
+
+  def tailoring_valid?
+    rule_groups_for_rules = RuleGroup.where(id: rules.includes(:rule_group)
+                                     .references(:rule_group).flat_map(&:rule_group_ancestor_ids))
+
+    conflicts = relationships_for(rules, rule_groups_for_rules, 'conflicts').distinct.map(&:right)
+    requires = relationships_for(rules, rule_groups_for_rules, 'requires').distinct.map(&:right)
```
skateman, line 42: Move `distinct` inside, just keep the `map`

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,19 @@ def update_os_minor_version(version)
...
+
+    valid_relationships?(rules, rule_groups_for_rules, requires, conflicts)
+  end
+
+  def valid_relationships?(rules, rule_groups, requires, conflicts)
+    rules_and_groups = rules | rule_groups
```
skateman, line 48: What does the `|` operation do?

romanblanco: it concatenates the two arrays, https://ruby-doc.org/core-2.3.1/Array.html#method-i-7C

app/models/concerns/profile_fields.rb
marleystipich2: Yeah, but also because I wanted to re-use the ```
RuleGroupRelationship.with_relationships(rules, relationship).or(
        RuleGroupRelationship.with_relationships(rule_groups, relationship)
``` part of the `relationships_for` method in `tailoring_valid?` because that was exactly what we needed

Rubocop was complaining a lot when I tried using `RuleGroupRelationship.where(relationship: 'conflicts', left: rules).or(RuleGroupRelationship.where(relationship: 'conflicts', left: groups)).includes(:right).distinct.map(&:right)` in `tailoring_valid?`

File: app/models/concerns/profile_tailoring.rb (hunk 12 above)
Comment (marleystipich2) [reply app/models/concerns/profile_tailoring.rb:48]: It does a union because we still need to combine the rules and rule_groups to check them against `conflicts` and `requires`

Comment (skateman) [reply app/models/concerns/profile_tailoring.rb:48]: And why not ˙+`? I guess that's what's being used across the codebase.

File: app/models/concerns/profile_fields.rb (hunk 10 above)
Comment (skateman) [reply app/models/concerns/profile_fields.rb:72]: Okay, I'll need to think about this a bit more...will circle back to this tomorrow.

File: app/models/concerns/profile_tailoring.rb (hunk 11 above)
Comment (marleystipich2) [reply app/models/concerns/profile_tailoring.rb:42]: I just tried that but forgot that we actually don't want only distinct relationships when we use `relationships_for` for `rules_and_rule_groups`  so I can't add the distinct in that method. I could either just leave distinct where it is or create a `distinct_relationships_for` method for this situation?

File: app/models/concerns/profile_tailoring.rb (hunk 12 above)
Comment (marleystipich2) [reply app/models/concerns/profile_tailoring.rb:48]: Hmm I guess I just went with `|` because I'm also using the intersect `&` operator in that method but I can use `+` to stay consistent with the rest of the codebase because the result is the same

File: app/models/concerns/profile_fields.rb (hunk 10 above)
Comment (marleystipich2) [reply app/models/concerns/profile_fields.rb:72]: Okay, sounds good 👍  Btw, the rubocop error that I see if I use ```
conflicts = RuleGroupRelationship.where(relationship: 'conflicts', left: rules).or(
      RuleGroupRelationship.where(relationship: 'conflicts', left: rule_groups_for_rules)
    ).distinct.map(&:right)
``` directly in `tailoring_valid?` is 
`Assignment Branch Condition size for tailoring_valid? is too high.`

So that's why I had to move some of it into a different method.

File: app/models/concerns/profile_tailoring.rb (hunk 12 above)
Comment (skateman) [reply app/models/concerns/profile_tailoring.rb:48]: No, you should stick with `|`, it's not a simple concatenation:
```
[1] | [1,2] => [1,2]
```

File: app/models/concerns/profile_tailoring.rb (hunk 11 above)
Comment (skateman) [reply app/models/concerns/profile_tailoring.rb:42]: Why is it a problem to have the `distinct` in the `relationships_for`?

Comment (marleystipich2) [reply app/models/concerns/profile_tailoring.rb]: I actually was able to update `valid_relationships?` so that it works with duplicates in requires and conflicts so we don't need `distinct` anymore. I added tests to cover duplicate requires/conflicts to make sure it works correctly too.

File: app/models/concerns/profile_tailoring.rb [hunk 13]
```diff
@@ -33,4 +33,19 @@ def update_os_minor_version(version)
...
+
+    conflicts = relationships_for(rules, rule_groups_for_rules, 'conflicts').map(&:right)
+    requires = relationships_for(rules, rule_groups_for_rules, 'requires').map(&:right)
+
+    valid_relationships?(rules, rule_groups_for_rules, requires, conflicts)
+  end
+
+  def valid_relationships?(rules, rule_groups, requires, conflicts)
+    rules_and_groups = rules | rule_groups
+    (requires - rules_and_groups).empty? && (conflicts - rules_and_groups) == conflicts
```
marleystipich2, line 49: @skateman I adjusted this method a little so that it still works correctly even if there are duplicates in `requires` or `conflicts` so we don't need to worry about using `distinct` anymore

app/models/concerns/profile_fields.rb
```diff
@@ -39,8 +39,8 @@ def canonical?
     end
 
     def rules_and_rule_groups
-      conflicts = relationships_for('conflicts')
-      requires = relationships_for('requires')
+      conflicts = map_rule_group_relationships('conflicts')
+      requires = map_rule_group_relationships('requires')
```
skateman, line 43: Sometimes the solution is not to have an additional method, just to iterate :wink:
```suggestion
      conflicts, requires = %w[conflicts requires].map do |relationship|
        relationships_for(rules, rule_groups, relationship).each _with_object({}) do |rgr, relationships|
          relationships[rgr.left] ||= []
          relationships[rgr.left] << rgr.right
        end
      end
```

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,19 @@ def update_os_minor_version(version)
 
     update!(os_minor_version: version)
   end
+
+  def tailoring_valid?
+    rule_groups_for_rules = RuleGroup.where(id: rules.includes(:rule_group)
+                                     .references(:rule_group).flat_map(&:rule_group_ancestor_ids))
```
skateman, line 39: I guess it's safe to shadow here the `rule_groups` variable as the method is short enough.

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,19 @@ def update_os_minor_version(version)
...
+
+    conflicts = relationships_for(rules, rule_groups_for_rules, 'conflicts').map(&:right)
+    requires = relationships_for(rules, rule_groups_for_rules, 'requires').map(&:right)
```
skateman, line 42: Similar iteration could be done here as above

app/models/rule.rb
```diff
@@ -74,7 +74,7 @@ class Rule < ApplicationRecord
   }
 
   scope :without_rule_group_parent, lambda {
-    where.missing(:rule_groups)
+    where.missing(:rule_group)
   }
```
skateman, line 78: This can be rewritten as a one-liner

app/models/rule.rb
```diff
@@ -118,4 +118,8 @@ def latest_result(host, profile)
 
     test_result.rule_results.find_by(rule_id: id)
   end
+
+  def rule_group_ancestor_ids
+    rule_group&.path_ids
+  end
```
skateman, line 124: Do we really need a separate method for this?

app/models/rule_group_relationship.rb
```diff
@@ -11,6 +11,6 @@ class RuleGroupRelationship < ApplicationRecord
   validates :left_id, presence: true, uniqueness: { scope: %i[relationship right_id right_type left_type] }
 
   scope :with_relationships, lambda { |rules_or_groups, relationship|
-    where(relationship: relationship).includes(:left, :right).where(left: rules_or_groups)
+    where(relationship: relationship, left: rules_or_groups).includes(:left, :right)
```
skateman, line 14: Not sure about the `includes` here, especially if it does not have `references`.

skateman (changes requested):

app/models/rule.rb
marleystipich2: Probably not.. I can try to move it into the `tailoring_valid?` and see if rubocop complains about it

app/models/rule_group_relationship.rb
marleystipich2: Yeah, I only added the `includes` in this scope because we're calling `left` and `right` here:
```
def map_rule_group_relationships(relationship)
      relationships_for(rules, rule_groups, relationship).each_with_object({}) do |rgr, relationships|
        relationships[rgr.left] ||= []
        relationships[rgr.left] << rgr.right
     end
end
```
But I could do `relationships_for(rules, rule_groups, relationship).includes(:left, :right)` so that the `includes` isn't inside the scope? Or I could add `references` after the `includes` in the scope?

app/models/rule.rb
marleystipich2: I was able to move this line into `tailoring_valid?` and remove this method 👍

app/models/rule_group_relationship.rb
skateman: You can do two arguments to references, or no?

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,20 @@ def update_os_minor_version(version)
...
+
+  def tailoring_valid?
+    rule_groups = RuleGroup.where(id: rules.includes(:rule_group)
+                           .references(:rule_group).flat_map { |r| r.rule_group&.path_ids })
+
+    conflicts, requires = %w[conflicts requires].map do |relationship|
+      with_relationships(relationship, rules, rule_groups).map(&:right)
+    end
+
+    valid_relationships?(rules, rule_groups, requires, conflicts)
```
skateman, line 45: As I said earlier, extracting the end of a method chain into an external method is a bit complicated. You want a dataflow in the opposite direction, i.e. extract methods from the beginning and make the end the simplest as possible by just aggregating it there. Rubocop is just giving you hints and it's not always the best to start cleaning up a method from the end. So please try to extract the `rule_groups =` into `RulesAndRuleGroups#hierarchical_rule_groups` or `rule_groups_with_children`. I think you don't need to memoize it in a variable, as the SQL cache kicks in anyway. Then this `valid_relationships?` can be moved back here as line 38 and 39 will completely go away.

app/models/concerns/profile_tailoring.rb
```diff
@@ -33,4 +33,20 @@ def update_os_minor_version(version)
...
+  end
+
+  def valid_relationships?(rules, rule_groups, requires, conflicts)
+    rules_and_groups = rules | rule_groups
+    (requires - rules_and_groups).empty? && (conflicts - rules_and_groups) == conflicts
```
skateman, line 50: Not tested, but should be simpler:
```suggestion
    (requires - rules_and_groups).empty? && (conflicts & rules_and_groups).empty?
```

skateman (changes requested):
I did a rebase on top of my refactoring, so make sure you force-pull the changes before going forward.

app/models/rule_group_relationship.rb
marleystipich2: I tried to add references but it doesn't work with polymorphic associations apparently:
`ActiveRecord::EagerLoadPolymorphicError: Cannot eagerly load the polymorphic association :left`

skateman: Ok, let's keep it as it is...

app/models/concerns/profile_tailoring.rb
marleystipich2: Ahh, I misunderstood that previous comment about the `rule_groups =` part but I see what you mean now. I'll update this.

skateman (approved):
Looks much better and nicer :+1: if @romanblanco has no objections, this can be merged.

PR #1272, opened 2022-04-21
test(cgroups): Add test cgroups implementation difference POC script

Victoremepunto:
## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices

PR #1254, opened 2022-04-07
refactor(entrypoint): RHICOMPL-2858 rewrite cgroup-limits to bash

skateman:
Original implementation is [here](https://github.com/sclorg/container-common-scripts/blob/master/shared-scripts/core/usr/bin/cgroup-limits), this is intended for use in the ubi8-minimal container.

## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Authentication and Password Management
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Error Handling and Logging
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management
- [x] General Coding Practices

PR #1274, opened 2022-04-22
refactor(entrypoint): RHICOMPL-2858 move env var export from cgroups

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -36,7 +36,7 @@ if [ "$APPLICATION_TYPE" = "compliance-backend" ]; then
   fi
 
   if is_puma_installed; then
-    source scripts/set_cgroup_limits.sh
+    export_vars=$(./scripts/set_cgroup_limits.sh) ; export export_vars
 
     exec bundle exec "puma --config ../etc/puma.cfg -b tcp://0.0.0.0:${PORT}"
   else
--- a/scripts/set_cgroup_limits.sh
+++ b/scripts/set_cgroup_limits.sh
@@ -3,7 +3,7 @@
 # Script for parsing cgroup information
 # Inspired by https://github.com/sclorg/container-common-scripts/blob/master/shared-scripts/core/usr/bin/cgroup-limits
 
-export MAX_MEMORY_LIMIT_IN_BYTES=9223372036854775807
+# export MAX_MEMORY_LIMIT_IN_BYTES=9223372036854775807
 echo "MAX_MEMORY_LIMIT_IN_BYTES=${MAX_MEMORY_LIMIT_IN_BYTES}"
 
 # Determine memory limit from cgroups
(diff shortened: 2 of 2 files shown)


entrypoint.sh
```diff
@@ -36,7 +36,7 @@ if [ "$APPLICATION_TYPE" = "compliance-backend" ]; then
   fi
 
   if is_puma_installed; then
-    source scripts/set_cgroup_limits.sh
+    export_vars=$(./scripts/set_cgroup_limits.sh) ; export export_vars
```
vkrizan, line 39: This does not feel right, as this highly depends on what the script outputs. Sourcing the script and having it export necessary env vars feels much better.

skateman: https://github.com/RedHatInsights/compliance-backend/pull/1254/files#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2L39

vkrizan: I see that this is how it was. It was because it was an external script that could not do a bash env var export. This is much safer as it is.

skateman: I am aware but someone doesn't like it :see_no_evil:

skateman: I am aware, I wrote it intentionally like that :shrug: but QE wants it differently :thinking:

vkrizan: @Victoremepunto what's the issue here? Please read above. Thanks.

Victoremepunto: @vkrizan  I disagree, if you take a look at the [original implementation](https://github.com/sclorg/container-common-scripts/blob/326add3de53dd1bafd17099cfa786e0741972f6f/shared-scripts/core/usr/bin/cgroup-limits#L4-L12) of the `cgroup-limits` Python script this script tries to replace, this is actually faithful to that, and by miles far easier to blackbox test it to ensure there are no side effects from replacing the Python script and using our own script.

InsightsDroid:
1c2793c9-4b51-428b-9436-748dc270cb9e
a3ffc75a-57c0-4b45-ac16-7dd3b910240c

entrypoint.sh
vkrizan: How does this change affect black box testing of the script? Exported env vars should match what the original script has outputted, if the re-implementation is to be tested 1:1.

Victoremepunto: @vkrizan  it only makes it more difficult, as you have to isolate the environment, run the script, and filter out the variables it exports , to compare it to the ones the Python script yields to STDOUT. I hope you agree is far easier to compare the output of 2 commands than analyzing the environment changes. again, only more difficult, not impossible. I already provided a [POC script](https://github.com/RedHatInsights/compliance-backend/pull/1272/files#diff-07e6ade8d04cce19e6663fe5d24bc2137337aebe1df4a47f0de4ae7f7d560bb0R20) of how it can be tested with the current implementation.

skateman: The script also yields to STDOUT...

Victoremepunto: @vkrizan  I do something more specific, and I clean the environment with `env -i` which is exactly what we want (a clean as possible env)

your approach suggests to run the script twice in two separate Bash sessions and compare environments ? 

I think that would work too but I see it less efficient than the other approach I'm using.
 
And I must also note that I see the approach as a "workaround" to the fact the script does the exports itself, if it would just "print to stdout" , again, but that's just a small detail not a real "issue" .

PR #1278, opened 2022-04-26
chore(tests): RHICOMPL-2878 properly mock SupportedSsg in tests

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


skateman:
I'm 100% sure that the `SupportedSsg` is being called on way more places than this single one. Just try to add a debugging flag into all the `raw_supported` method and count how many times the debugger stops on a single-threaded test run...

InsightsDroid:
4769f744-bb43-4622-9171-b72c7e0e3642
f9cb7b90-010a-485d-9464-8de26683a354

skateman:
@marleystipich2 I'm pretty sure there are way more callsites. You can theoretically find them by removing all the public methods from the `SupportedSsg` superclass (line 28 til 84) and see what's failing.

marleystipich2:
@skateman To test if there were other places that I need to stub SupportedSsg, I commented out the method body of the `all` method and then printed out the value of `all` in the other SupportedSsg methods that call that method and also the printed out the method bodies of the others even if they don't directly call `all` and then I ran all of the tests 
There are a lot of places that the other methods in SupportedSsg are called like `for_os` is called a lot but since the `all` method is stubbed, the other methods are using the stubbed value of it. So their return value is based on the test values instead of the real Supportedssg data.

skateman:
You should not just comment the `all` method, but all the methods available in the module. Even better if you drop the content of the whole file...

marleystipich2:
@skateman I commented out all of the methods in SupportedSsg and this did trigger a failure at all of tests that weren't stubbing SupportedSsg but in a few areas this is intentional because certain tests rely on the value that's returned by one or more of the SupportedSsg methods. So stubbing the method wouldn't give an accurate result for the test. 
But even in these areas where we aren't stubbing the method directly, we are stubbing the methods that are used in those methods so real data isn't being used anywhere. Here's an example: 
```
context '#supported?' do
    should 'return true for OS/SSG matches' do
      SupportedSsg.stubs(:ssg_versions_for_os).returns(['0.0.1', '1.2.3'])
      assert SupportedSsg.supported?(ssg_version: '1.2.3',
                                     os_major_version: '7',
                                     os_minor_version: '4')
      assert_not SupportedSsg.supported?(ssg_version: '1.2.4',
                                         os_major_version: '7',
                                         os_minor_version: '4')
    end
  end
```
In this example, we don't want to stub `supported?` because we're actually testing that method. Right?

marleystipich2:
@skateman This is ready to be reviewed whenever you have time! We talked about your last comment in standup a couple days ago but just realized I never commented back on here

skateman (approved):
The Seal of Approval

test/test_helper.rb
```diff
@@ -90,6 +90,22 @@ def assert_equal_sets(arr1, arr2)
...
+
+        benchmark_versions.each do |bmv|
+          new_supported_ssgs = test_supported_ssgs(hosts, bmv)
+          supported_ssgs += new_supported_ssgs
+          # Stubbing by_ssg_version because some tests were failing due to
+          # the conditional assignment with the instance variable
+          SupportedSsg.stubs(:by_ssg_version).returns(bmv => new_supported_ssgs)
+        end
+        SupportedSsg.stubs(:all).returns(supported_ssgs)
+        SupportedSsg.stubs(:revision).returns('2022-04-20')
```
romanblanco, line 106: nitpick: In this case it's not making a difference, but in general, I'd avoid using hardcoded value and use random date instead.

romanblanco (approved):
LGTM :+1:

test/test_helper.rb
marleystipich2: Good point. Thanks!

PR #1284, opened 2022-05-06
test(scripts): RHICOMPL-2952 Add Shellcheck and reorganize scripts

Victoremepunto:
- Move all Bash scripts under 'scripts' directory, except pr_check.sh, build_deploy.sh and the entrypoint scripts
- Add ShellCheck to static analysis job
- Apply shellcheck suggestions to all scripts

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


Dockerfile
```diff
@@ -52,4 +52,4 @@ COPY --chown=1001:0 --from=build /opt/app-root/src/.bundle /opt/app-root/src/.bu
 
 ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true HOME=/opt/app-root/src DEV_DEPS=$devDeps
 
-CMD ["/opt/app-root/src/entrypoint.sh"]
+CMD ["/opt/app-root/src/scripts/entrypoint.sh"]
```
vkrizan, line 55: Doesn't `scripts/` folder contain process scripts as well? Entrypoint is part of the app, and should be always accessible within the bundle/container. I think that it is more common to have entrypoint at the app's root.

build_deploy.sh
```diff
@@ -2,10 +2,12 @@
...
 
-source 'deploy/build-deploy-common.sh'
+source 'scripts/deploy/build-deploy-common.sh'
 
 BACKWARDS_COMPATIBILITY_TAGS="latest qa"
 IMAGE_NAME="${IMAGE_NAME:-quay.io/cloudservices/compliance-backend}"
-IMAGE_TAG="${IMAGE_TAG}"
+IMAGE_TAG="${IMAGE_TAG:?}"
+
+git submodule update --init
```
vkrizan, line 11: Doesn't this need `--remote`?
From my own experience, git submodules are not that easy to work with. The value that is checked out is stored and kept, and usually it should be committed. Please check that this would pull new content from the remote and move the HEAD, esp. when this will be run on shared (test) environment. 

Configuration lacks branch tracking.

Dockerfile
Victoremepunto: > Doesn't scripts/ folder contain process scripts as well?
I'm sorry I don't understand what you mean with "process scripts" - This PR tried to put all the different Bash script files this repository maintains under one directory. That said, it is unfortunate that I didn't find a good way to remove `pr_check.sh` and `build_deploy.sh` from the root directory. 

> Entrypoint is part of the app, and should be always accessible within the bundle/container
I don't see how that is not achievable on the Dockerfile where we can configure the image as we want and put files where we feel they should be. However I would argue there's little value in adding an extra step while building the image when it's only a route we configure. 

Arranging the scripts under one directory is just me trying to do some housekeeping, we could decide where we want to put the files depending on their nature / role. But I'd like not to have them scattered around like we have them today. I'm very open to suggestions on how we should organize them, or to be referred to some best practices if anyone can share.

InsightsDroid:
7e54be24-d131-4a52-a17e-4db87baebf7b
a29afc8c-c3d2-4b57-ad8f-701dcf5b3487

Dockerfile
vkrizan: By "process" I've meant scripts that handle everything around the app, like build, ci, dev setup, support... `entrypoint` is not part of that. I agree that putting all scripts together on one place is a good idea. However, I would not want the entrypoint being mixed up with support scripts. Entrypoint is part of the app, support scripts usually are not, hence my suggestion to keep it on the root.

build_deploy.sh
vkrizan: If the intention is to have this only run on a clean pull and never have it updated after that, then this would work well. The problematic part would be when one would have existing repository pulled (i.e. wants to test the script), then the behavior of this might not be correct (at least not w/o `--remote`).

The ref of the submodule can be committed to the repository. That would pin it to specific commit. Then on pull (for test runs) there is an option to pull submodules as well. Configuration might need a tracking branch, just for the ease of updates.  ... What I found as the easiest way of updating the submodule, is to `cd` to its folder do a `pull` and then go back and commit the `ref` to the main repository.

Let's first clear out what do we want to achieve with this.

Dockerfile
Victoremepunto: @vkrizan  okay, fair enough, thank you for clarifying, `entrypoint.sh` may make sense to remain in the root dir. I believe the same applies to `devel.entrypoint.sh`  ?

InsightsDroid:
7cab677c-702a-46b5-a1f9-cfc23744c337
b7acdf11-1eb8-4f0a-a3ce-a9af58847efc

Dockerfile
vkrizan: I'd guess so, yes.

vkrizan: For these `pr_check.sh` and `build_deploy.sh` I guess the build options for Jenkins templates can be used, like `build_deploy_script_path` but I'm not fully sure if that would work.

Victoremepunto: @vkrizan  you're right, but , that comes with creating a custom template for Compliance, and I don't think that the trade-off of maintaining a custom JJB template for us alone is worth it (we did remove the one we had for `codecov` token for this reason)

build_deploy.sh
vkrizan: I'd suggest to try and test this deploy script with all cases in mind: having a clean repo, having the submodule already checked, having the submodule checked but with pending changes that needs to be pulled.

skateman:
What's the status on this? :thinking: :watch:

Victoremepunto:
> What's the status on this? thinking watch

I think I just need to address @vkrizan comments about testing how the git submodule works with different scenarios

build_deploy.sh
Victoremepunto: @vkrizan  tested:

- having a clean repo, no upstream changes: on a fresh repo clone (which is what happens on CI every time) - the `build_deploy` runs the line in build_deploy that does the submodule init, doing a checkout of the commit under version control that was added originally.

- having a clean repo, upstream changes: the SHA of the committed version of the submodule is the one used in the checkout by git on the init operation (as expected). Therefore, the latest changes won't be checked out . 

- having an existing repo with the submodule initialized, upstream changes: the SHA of the committed version of the submodule is the one used in the checkout by git on the init operation (again, as expected). if we want to pull the latest changes from upstream, we must add the `--remote` flag , and *add and commit* those changes (hence, proactively pulling the latest changes. I think this is expected, we may not want the "latest updates always", in the same way we have pinned versions for the other libraries we use.

@vkrizan  please let me know what's required to get this merged. we can meet if needed

Dockerfile
Victoremepunto: @vkrizan  since `entrypoint.sh` , `build_deploy.sh` and `pr_check.sh`  are left in the root dir, can be say it's good to resolve this ?

build_deploy.sh
vkrizan: @skateman thoughts on this approach? Technically it is a better solution that what we have today, but personally I'm bit hesitant as it might be harder to maintain (as git submodules are a bit advanced topic). I think that it could be brought up in a wider discussion with tenants and platform. Although, I guess it should not block this PR.

vkrizan:
@Victoremepunto I'd love to see it split to several commits. There is refactoring, shellcheck addition, and git submodule. If we're unable to get to a conclusion for the git submodule, then it can be moved to a separate PR to not block the refactoring.

Victoremepunto:
> @Victoremepunto I'd love to see it split to several commits. There is refactoring, shellcheck addition, and git submodule. If we're unable to get to a conclusion for the git submodule, then it can be moved to a separate PR to not block the refactoring.

Hey @vkrizan , @skateman  - I've removed the git-submodule for a follow-up PR , for when our discussion leads us to an agreement.

I've also split the remaining changes into three commits, hope's ready to be merged now.

vkrizan (approved):
LGTM. Thank you.

PR #1297, opened 2022-05-26
fix(profiles): RHICOMPL-2989 drop profiles without policy

skateman:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/services/orphaned_profiles_cleaner.rb
+++ b/app/services/orphaned_profiles_cleaner.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Service for cleaning up orphaned profiles
+class OrphanedProfilesCleaner
+  class << self
+    def run!
+      Profile.canonical(false).where(policy_id: nil, external: true).destroy_all
+    end
+  end
+end
--- a/db/migrate/20220525104102_cleanup_orphaned_profiles.rb
+++ b/db/migrate/20220525104102_cleanup_orphaned_profiles.rb
@@ -1,0 +1,9 @@
+class CleanupOrphanedProfiles < ActiveRecord::Migration[7.0]
+  def up
+    OrphanedProfilesCleaner.run!
+  end
+
+  def down
+    # nop
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_05_13_090623) do
+ActiveRecord::Schema[7.0].define(version: 2022_05_25_104102) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "dblink"
   enable_extension "pgcrypto"
--- a/test/services/orphaned_profiles_cleaner_test.rb
+++ b/test/services/orphaned_profiles_cleaner_test.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class OrphanedProfilesCleanerTest < ActiveSupport::TestCase
+  setup do
+    account = FactoryBot.create(:account)
+    host = FactoryBot.create(:host, account: account.account_number)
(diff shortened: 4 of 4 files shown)


InsightsDroid:
137b2861-aa4b-4608-a856-f23fdd4a834d
a957f71c-256e-40b9-8c1f-ba6222746d2a

app/services/orphaned_profiles_cleaner.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# Service for cleaning up orphaned profiles
+class OrphanedProfilesCleaner
+  class << self
+    def run!
+      # No need to run this in an empty DB
+      return if Profile.canonical(false).none?
```
vkrizan, line 8: What's the reason for this optimization? Ephemeral test run would not run the part of the code after this return.

app/services/orphaned_profiles_cleaner.rb
```diff
@@ -0,0 +1,28 @@
...
+      Profile.canonical(false).where(policy_id: nil).destroy_all # Remove remaining orphaned profiles
+    end
+
+    private
+
+    def reassign_by_parent
+      ActiveRecord::Base.connection.execute(%(
+        UPDATE "profiles" SET "policy_id" = "sq"."policy_id" FROM (
+          SELECT "profiles"."id", "xp"."policy_id" FROM "profiles"
+          INNER JOIN "profiles" as "xp" ON "xp"."parent_profile_id" = "profiles"."parent_profile_id"
```
vkrizan, line 20: Could this create duplicate profiles of same parent hence same benchmark being assigned to a policy?

app/services/orphaned_profiles_cleaner.rb
```diff
@@ -0,0 +1,28 @@
...
+                                        AND "xp"."account_id" = "profiles"."account_id"
+                                        AND "xp"."external" = false
```
vkrizan, line 22: Should there be a check for `"profiles".external = true` to avoid possible two profiles under same policy being marked as the "initial policy" (with `external = false`)?

vkrizan (commented):
I'm wondering if we need the `reassign_by_parent`. Could it do more harm than benefit?

app/services/orphaned_profiles_cleaner.rb
skateman: It's not really an optimization, the query below would probably fail on an empty DB + also it has no point to run it on an empty DB.

app/services/orphaned_profiles_cleaner.rb
skateman: Umm, I don't understand the question...

app/services/orphaned_profiles_cleaner.rb
skateman: Fixed

app/services/orphaned_profiles_cleaner.rb
vkrizan: Resolved in private chat. The duplicates might exist, however they could differ by OS minor version (which might not be the case for our db).

app/services/orphaned_profiles_cleaner.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Service for cleaning up orphaned profiles
+class OrphanedProfilesCleaner
+  class << self
+    def run!
+      Profile.canonical(false).where(policy_id: nil, external: true).destroy_all
```
vkrizan, line 7: Are there any Profiles that have `external: false` but do not have any policy assigned?

skateman: No

vkrizan (approved):

PR #1286, opened 2022-05-09
fix(RuleGroups): RHICOMPL-2956 update ancestry for rule groups correctly

marleystipich2:
## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices

PR #1298, opened 2022-05-26
Revert "fix(RuleGroups): RHICOMPL-2956 update ancestry for rule groups correctly"

marleystipich2:
Reverts RedHatInsights/compliance-backend#1286

### DIFF
--- a/db/migrate/20220511212358_clear_revisions_may_eleventh.rb
+++ b/db/migrate/20220511212358_clear_revisions_may_eleventh.rb
@@ -1,9 +1,0 @@
-class ClearRevisionsMayEleventh < ActiveRecord::Migration[7.0]
-  def up
-    Revision.find_by(name: 'datastreams')&.delete
-  end
-
-  def down
-    # nop
-  end
-end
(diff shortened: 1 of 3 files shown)


InsightsDroid:
b9e654ca-9fb9-4806-8c2d-c7cc77e6540d
e4f9a2a7-7682-4651-82ef-534f34de0209

db/migrate/20220511212358_clear_revisions_may_eleventh.rb
```diff
@@ -1,9 +0,0 @@
-class ClearRevisionsMayEleventh < ActiveRecord::Migration[7.0]
-  def up
-    Revision.find_by(name: 'datastreams')&.delete
-  end
-
-  def down
-    # nop
-  end
-end
```
skateman, line 9: First I wanted to say that we should keep the migration, just drop its body so we have the table revisions in sync, but actually as you didn't run `db:migrate` and the schema didn't change, it's fine to just drop this. Probably this never even ran on stage.

skateman (approved):

PR #1313, opened 2022-06-08
feat(Remediations): RHICOMPL-3018 Update revision comparison 

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/services/supported_remediations.rb
+++ b/app/services/supported_remediations.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+# Service for Supported Remediations
+class SupportedRemediations
+  class << self
+    def revision
+      SsgConfigDownloader.update_ssg_ansible_tasks
+      YAML.safe_load(SsgConfigDownloader.ssg_ansible_tasks)['revision']
+    end
+  end
+end
--- a/lib/tasks/import_datastream.rake
+++ b/lib/tasks/import_datastream.rake
@@ -10,11 +10,11 @@ namespace :ssg do
       "#{SupportedSsg.revision.inspect}\nDB revision: " \
       "#{Revision.datastreams.inspect}"
     remediations_message = "SSG remediations not synced\nRemediation config revision: " \
-      "#{SupportedSsg.revision.inspect}\nDB revision: " \
+      "#{SupportedRemediations.revision.inspect}\nDB revision: " \
       "#{Revision.remediations.inspect}"
     abort datastreams_message if Revision.datastreams != SupportedSsg.revision
-    abort remediations_message if Revision.remediations != SupportedSsg.revision
-    puts "Datastreams synced to revision: #{Revision.datastreams}"
+    abort remediations_message if Revision.remediations != SupportedRemediations.revision
+    puts "Datastreams and remediations synced to revision: #{Revision.datastreams} and #{Revision.remediations}"
   end
 
   desc 'Update supported SSGs fallback yaml'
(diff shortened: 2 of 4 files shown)


InsightsDroid:
20ebe3b8-226e-44d3-a9ba-b379557b8bc5
7e91d7b1-5d33-4739-bf9f-d1c58e048263

app/controllers/v1/supported_ssgs_controller.rb
```diff
@@ -17,7 +17,7 @@ def metadata(opts = {})
       {
         meta: {
           total: opts[:total],
-          revision: resource.revision
+          datastream_revision: resource.datastream_revision
```
skateman, line 20: Wouldn't this break the API?

skateman (changes requested):

app/controllers/v1/supported_ssgs_controller.rb
marleystipich2: Oh right, it would.. I could just leave the datastream revision method named `revision`? But still add the new `ansible_tasks_revision` method? That'd be easier and fewer updates.

skateman: The left side should stay as it is, for compatibility.

skateman (changes requested):
Instead of forcing the ansible revisions into the same file, you could just simply create a new one that loads the ansible revision only. It does not even need to be a structure, just a simple object as it doesn't even load any further information. It does not even need caching and anything around it as it is only being checked once in the init container. Any other cache-related operations like the checksum control are also unnecessary.

```
# app/services/supported_remediations.rb
class SupportedRemediations
  class << self
    def revision
      # ...
    end
  end
end
```

marleystipich2:
The test that's failing here is passing locally for me.. I'm not sure why it's failing

skateman (changes requested):

app/services/supported_remediations.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    private
+
+    def clear
+      @raw_supported_remediations = nil
+      @ansible_task_checksum = SsgConfigDownloader.ssg_ansible_tasks_checksum
+    end
+
+    def ssg_ansible_tasks_changed?
```
skateman, line 19: As I said earlier, here you don't really need caching as this is a one-time running task in a cronjob. Its pod gets deleted after each run, therefore, any cache will be invalidated...

marleystipich2: Right, sorry I should've moved this PR to draft, I'm trying to figure out why the PR check keeps failing. It's timing out even when I set up the SupportedRemediations `revision` method like this:
```
def revision
      YAML.safe_load(SsgConfigDownloader.ssg_ansible_tasks)
 end
```

It seems like this should be pretty fast though so I'm not sure why it's timing out?

app/services/supported_remediations.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+# Service for Supported Remediations
+class SupportedRemediations
+  class << self
+    def revision
+      SsgConfigDownloader.update_ssg_ansible_tasks
```
marleystipich2, line 9: @skateman I realized that this line was needed because without it, this method was returning the revision date in the fallback ansible tasks file `config/ssg-ansible-tasks.default.yaml`. I deployed to ephermal manually and made sure that the remediation revision date was syncing up and everything looked good

skateman (approved):

PR #1358, opened 2022-07-20
Make parameter handling for searching and includes more robust

skateman:
* Only string-based fields should have implicit search, anything else doesn't really make sense
* If search is not supported for a model, we should fail with 422 instead of 500
* Null characters in a string should be handled with 422 instead of 500
* Serializer library has a bug for validating includes when having 0 relationships

For reproducing these issues, see the JIRA ticket

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/app/controllers/concerns/collection.rb
+++ b/app/controllers/concerns/collection.rb
@@ -24,6 +24,11 @@ module Collection
     def search(data)
       return data if permitted_params[:search].blank?
 
+      # Fail if search is not supported for the given model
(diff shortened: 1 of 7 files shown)


vkrizan (approved):
Nice one!

PR #1376, opened 2022-07-29
chore(puma): RHICOMPL-3059 add Puma restart/shutdown hooks

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -47,3 +47,11 @@ after_worker_boot do
   require 'prometheus_exporter/instrumentation'
   PrometheusExporter::Instrumentation::Puma.start(type: 'puma')
 end
+
+on_worker_shutdown do
+  puts '-------------------------Puma worker shutdown-------------------------'
+end
+
+on_restart do
+  puts '-------------------------Puma worker restart-------------------------'
+end


config/puma.rb
```diff
@@ -47,3 +47,11 @@
   require 'prometheus_exporter/instrumentation'
   PrometheusExporter::Instrumentation::Puma.start(type: 'puma')
 end
+
+on_worker_shutdown do
+  puts '-------------------------Puma worker shutdown-------------------------'
```
vkrizan, line 52: Is there a better alternative using puma's logger?

skateman: We're not sure about the context here :thinking:

skateman: Also this is just temporary

marleystipich2: Yeah, this is just temporary so it won't be staying but if we wanted it to be permanent I could try to think of something better?

vkrizan: ack

PR #1399, opened 2022-08-09
perf(SupportedSsg): RHICOMPL-3200 drop Revision from cache key

skateman:
Dropping the revision from the cache key reduces the number of allocations by ~100k and the latency by 200ms for this query:
```
time curl -k -H "Content-Type: application/json" -H "X-RH-IDENTITY: $RHI" -XPOST -d '{
  "operationName": "supportedProfilesByOSMajor",
  "variables": {},
  "query": "query supportedProfilesByOSMajor {
      osMajorVersions {
          edges {
              node {
                  osMajorVersion
                  profiles {
                      id
                      name
                      refId
                      description
                      complianceThreshold
                      supportedOsVersions
                      benchmark {
                          id
                          refId
                          osMajorVersion
                      }
                  }
              }
          }
      }
      profiles(search: \"external = false and canonical = false\") {
          edges {
              node {
                  id
                  refId
                  benchmark {
                      refId
                  }
              }
          }
      }
  }"
}' "http://localhost:3000/api/compliance/graphql" | jq

```

@vkrizan any objections? 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -80,9 +80,7 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
     end
 
     def clear
-      %i[raw map by-ssg].each do |target|
-        Rails.cache.delete("SupportedSsg/datastreams/#{target}")
-      end
+      Rails.cache.delete_matched('SupportedSsg/datastreams/*')
     end
 
     private
--- a/app/models/supported_ssg.rb
+++ b/app/models/supported_ssg.rb
@@ -9,6 +9,8 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
   OS_NAME = 'RHEL'
   self::OS_NAME = OS_NAME
 
+  CACHE_PREFIX = 'SupportedSsg/datastreams'
+
   def ref_id
     'xccdf_org.ssgproject' \
     ".content_benchmark_#{OS_NAME}-#{os_major_version}"
@@ -80,7 +82,7 @@ SupportedSsg = Struct.new(:id, :package, :version, :profiles,
     end
 
     def clear
-      Rails.cache.delete_matched('SupportedSsg/datastreams/*')
+      Rails.cache.delete_matched("#{CACHE_PREFIX}/*")
     end
 
     private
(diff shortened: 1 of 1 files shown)


app/models/supported_ssg.rb
```diff
@@ -113,7 +113,7 @@ def build_latest_map
     end
 
     def cache(key, &block)
-      Rails.cache.fetch("SupportedSsg/datastreams/#{Revision.datastreams}/#{key}", expires_on: 1.day, &block)
+      Rails.cache.fetch("SupportedSsg/datastreams/#{key}", expires_on: 1.day, &block)
```
vkrizan, line 116: If this expires in a day, how do we ensure that each `key` that is cached is in sync with other keys?

skateman: As these variables depend on each other, invalidation will happen more or less the same time if it gets expired. The one day expiry is just a safeguard as I did not want to have a perpetual cache that can only be cleared when new content triggers an SSG import. This is only to ensure that if we change something, it prevails without manual intervention from the devprod team. The "standard" way of invalidation is ONLY via the SSG import that clears all three variables. 

I see your point with the programmer's error when adding a new cache key, we could come up with a custom cop for rubocop maybe, but I'd like to do it separately if that's okay.

skateman: Okay, I have an ugly idea for it without custom cops, give me some time

skateman:
@vkrizan I found the [delete_matched](https://apidock.com/rails/ActiveSupport/Cache/Store/delete_matched) method that clears the whole cache at once + extracted the prefix into a constant. So if someone creates a new cache key under the same prefix, it'll be cleared automagically.

vkrizan (approved):
Nice one!

I'm also adding this link https://redis.io/commands/keys/

PR #1411, opened 2022-08-16
fix(REST): RHICOMPL-1612 fix scope_search on profile name

marleystipich2:
Used this to test the time/efficiency of the old `name` scoped_search and the new version implemented here: 
`http --meta --verify no GET http://0.0.0.0:3000/api/compliance/v1/profiles search==name~"ANSSI" X-RH-IDENTITY:$ID_HEADER`

Old version: `Completed 200 OK in 361ms (Views: 253.5ms | ActiveRecord: 91.0ms | Allocations: 83467) `
Old version timed with httpie: 0.447715618s 

New version: `Completed 200 OK in 325ms (Views: 214.0ms | ActiveRecord: 90.5ms | Allocations: 83458)`
New version timed with httpie: 0.452380513s

Efficiency is about the same and this new `name` scoped_search for profiles is searching profiles and policies. It was only searching profiles before.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -8,7 +8,8 @@ module ProfileSearching
     scoped_search on: %i[id account_id compliance_threshold
                          external parent_profile_id],
                   only_explicit: true
-    scoped_search on: %i[name ref_id]
+    scoped_search on: %i[ref_id]
+    scoped_search on: :name, ext_method: 'filter_by_policy_and_profile_names'
     scoped_search relation: :assigned_hosts, on: :id, rename: :system_ids
     scoped_search relation: :assigned_hosts, on: :display_name,
                   rename: :system_names
@@ -147,6 +148,18 @@ module ProfileSearching
       {
         conditions: "profiles.id IN (#{any_child_profile.to_sql})"
       }
+    end
+
+    def filter_by_policy_and_profile_names(_filter, operator, value)
+      field = scoped_search_definition.field_by_name('name')
+      values = ScopedSearch::QueryBuilder.preprocess_parameters(scoped_search_definition, field, operator, value)
+
+      policy_cond = sanitize_sql_for_conditions(["policies.name #{operator} (?)", values])
+      profile_cond = sanitize_sql_for_conditions(["profiles.name #{operator} (?)", values])
+
+      profiles = left_outer_joins(:policy).where(policy_cond).or(Profile.where(profile_cond))
+
+      { conditions: 'profiles.id IN (?)', parameter: [profiles.select(:id)] }
     end
   end
 
--- a/config/initializers/override_scoped_search_escaping.rb
+++ b/config/initializers/override_scoped_search_escaping.rb
@@ -5,6 +5,19 @@
 module ScopedSearch
   class QueryBuilder
     alias preprocess_parameters_old preprocess_parameters
+
+    INVERTED_OPERATORS = SQL_OPERATORS.invert.merge('ILIKE' => :like, 'NOT ILIKE' => :unlike).freeze
+
+    # scoped_search gem does not preprocess values before passing to ext_method but sometimes it is necessary
+    # This method can be called from a scoped search ext_method to preprocess parameters
+    def self.preprocess_parameters(definition, field, operator, value)
+      # Passing empty block because nothing needs to happen on yield in preprocess_parameters
+      # The correct yield is invoked in to_ext_method_sql so all notification values are updated correctly
+      # See https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb
+      qb = ScopedSearch::QueryBuilder.new(definition, '', definition.profile)
+      qb.preprocess_parameters(field, INVERTED_OPERATORS[operator], value) { |_| }
+    end
+
     def preprocess_parameters(field, operator, value, &block)
       return preprocess_parameters_old(field, operator, value, &block) unless %i[like unlike].include?(operator)
       # Escaping the value part of the query
--- a/test/controllers/concerns/metadata_test.rb
+++ b/test/controllers/concerns/metadata_test.rb
@@ -88,13 +88,13 @@ class MetadataTest < ActionDispatch::IntegrationTest
       FactoryBot.create(:profile)
     end
 
-    search_query = 'name != ""'
+    search_query = "name != ''"
     get profiles_url, params: { search: search_query, limit: 1, offset: 2 }
     assert_response :success
-    assert_includes json_body['links']['first'], 'search=name+%21%3D+%22%22'
-    assert_includes json_body['links']['last'], 'search=name+%21%3D+%22%22'
-    assert_includes json_body['links']['next'], 'search=name+%21%3D+%22%22'
-    assert_includes json_body['links']['previous'], 'search=name+%21%3D+%22%22'
+    assert_includes json_body['links']['first'], 'search=name+%21%3D+%27%27'
+    assert_includes json_body['links']['last'], 'search=name+%21%3D+%27%27'
+    assert_includes json_body['links']['next'], 'search=name+%21%3D+%27%27'
+    assert_includes json_body['links']['previous'], 'search=name+%21%3D+%27%27'
   end
 
   test 'meta adds relationships param to JSON response' do
--- a/test/controllers/v1/profiles_controller_test.rb
+++ b/test/controllers/v1/profiles_controller_test.rb
@@ -261,6 +261,142 @@ module V1
         end)
       end
 
+      test 'search canonical profile name' do
+        canonical_profile = FactoryBot.create(:canonical_profile)
+
+        get v1_profiles_url, params: {
+          search: "name=\"#{canonical_profile.name}\""
+        }
+
+        assert_response :success
+
+        profiles = response.parsed_body
+
+        assert_equal 1, profiles['data'].count
+        assert_equal([canonical_profile.name], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
+      test 'search canonical profile name using like operator' do
+        canonical_profile = FactoryBot.create(:canonical_profile)
+
+        get v1_profiles_url, params: {
+          search: "name~\"#{canonical_profile.name}\""
+        }
+
+        assert_response :success
+
+        profiles = response.parsed_body
+
+        assert_equal 1, profiles['data'].count
+        assert_equal([canonical_profile.name], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
+      test 'search custom profile name with no spaces' do
+        FactoryBot.create(:profile, name: 'foo')
+        FactoryBot.create(
+          :profile,
+          name: 'foo',
+          policy: FactoryBot.create(:policy, name: 'bar')
+        )
+
+        get v1_profiles_url, params: {
+          search: 'canonical=false and name=bar'
+        }
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(%w[bar], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
+      test 'search custom profile name with spaces' do
+        FactoryBot.create(:profile, name: 'foo')
+        FactoryBot.create(
+          :profile,
+          policy: FactoryBot.create(:policy, name: 'Custom Name')
+        )
+
+        get v1_profiles_url, params: {
+          search: 'canonical=false and name="Custom Name"'
+        }
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(['Custom Name'], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
+      test 'search custom profile name with like operator' do
+        FactoryBot.create(:profile, name: 'foo')
+        FactoryBot.create(
+          :profile,
+          policy: FactoryBot.create(:policy, name: 'abc')
+        )
+        FactoryBot.create(
+          :profile,
+          policy: FactoryBot.create(:policy, name: 'bar')
+        )
+
+        get v1_profiles_url, params: {
+          search: 'canonical=false and name~a',
+          sort_by: 'name'
+        }
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(%w[abc bar], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
+      test 'search custom profile name with not eq operator' do
+        FactoryBot.create(:profile, name: 'foo')
+        FactoryBot.create(
+          :profile,
+          policy: FactoryBot.create(:policy, name: 'abc')
+        )
+        FactoryBot.create(
+          :profile,
+          policy: FactoryBot.create(:policy, name: 'bar')
+        )
+
+        get v1_profiles_url, params: {
+          search: "canonical=false and name != ''",
+          sort_by: 'name'
+        }
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(%w[abc bar foo], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
+      test 'search custom profile name with not like operator' do
+        FactoryBot.create(:profile, name: 'foo')
+        FactoryBot.create(
+          :profile,
+          policy: FactoryBot.create(:policy, name: 'bar')
+        )
+
+        get v1_profiles_url, params: {
+          search: 'canonical=false and name !~ a',
+          sort_by: 'name'
+        }
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(['foo'], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
+      end
+
       test 'fail if search contains null character' do
         get v1_profiles_url, params: {
           search: "foo\x00bar"
@@ -752,6 +888,17 @@ module V1
         assert_equal NAME, parsed_data.dig('attributes', 'name')
         assert_equal DESCRIPTION, parsed_data.dig('attributes', 'description')
         assert_audited 'Created policy'
+
+        get v1_profiles_url, params: {
+          search: 'canonical=false and name="A new name"',
+          sort_by: %w[name]
+        }
+
+        profiles = JSON.parse(response.body)
+
+        assert_equal(['A new name'], profiles['data'].map do |profile|
+          profile['attributes']['name']
+        end)
       end
 
       test 'create with all customized profile attributes' do


app/models/concerns/profile_searching.rb
```diff
@@ -9,6 +9,7 @@ module ProfileSearching
                          external parent_profile_id],
                   only_explicit: true
     scoped_search on: %i[name ref_id]
+    scoped_search relation: :policy, on: :name, rename: :policy_name
```
skateman, line 12: I'm not sure how introducing a new search param "fixes" a problem, the search on the `name` key should search in the name of the policy instead.

marleystipich2: Should the name key only search through policy names? If so, I could just make the name search
`scoped_search relation: :policy, on: :name` but this wouldn't search canonical profiles.

skateman: Well, you should to fix both issues then :grin: introducing a new query doesn't actually fix the existing one.

marleystipich2: Okay, the ext_method that I was working on was searching all profiles and policy names and this would be used for the `name` scoped search so I think this should solve the issue

marleystipich2:
@skateman This isn't finished yet and I'm sure there are ways to make it more efficient but just trying to get it working first.
Right now it's not working when you search using the `~` operator because it doesn't like when you use a `%` in the string inside the scoped_search `conditions` hash. I tried a lot of different things but I think this might be an issue with scoped_search. 
It also doesn't work when you search a profile name that contains a space because the `value` parameter of the `ext_method`  is always only the last word of the string. I haven't looked into this issue as much but searching by a profile name with spaces was working with the `policy_name` scoped search without the ext_method so I'm wondering if this is an issue with scoped search/ext_methods too.. 😞

skateman:
@marleystipich2 are the results in the description still accurate?

marleystipich2:
@skateman Yeah, they're accurate. I tested right before I pushed the last commit

skateman:
@romanblanco can you please take a look at this? I think the complexity is a bit high, but didn't really look into it deeper...

marleystipich2:
Can one of you try to test this locally searching a profile with spaces in the name? The tests pass but it's not working for me locally. I think it could be an issue with Mac though.
Something like this: 
`http --verify no GET http://0.0.0.0:3000/api/compliance/v1/profiles search==name="ANSSI-BP-028 (minimal)" X-RH-IDENTITY:$ID_HEADER`

skateman:
@marleystipich2 should work from somewhere in the frontend

skateman:
I think you can significantly reduce complexity if you follow the example [here](https://github.com/wvanbergen/scoped_search/wiki/Search-definition#external-search-methods). A lof of the preprocess and sanitization is redundant...

romanblanco:
> I think you can significantly reduce complexity if you follow the example [here](https://github.com/wvanbergen/scoped_search/wiki/Search-definition#external-search-methods). A lof of the preprocess and sanitization is redundant...

Agree. @marleystipich2, can you try if we can use the gem to simplify this as @skateman suggests, to something like


```ruby
scoped_search relation: policy, on: :name, ext_method: 'filter_by_policy_and_profile_names'

# ...

conditions = sanitize_sql_for_conditions(["policy.name #{operator} ?", value_to_sql(operator, value)])
profiles = Profile.joins(:policy).where(conditions).select('profile.name').map(&:id)

**Comment by marleystipich2** [comment]
I simplified this as much as I could and updated the description with the new times(which are about the same as before). 

The `preprocess_parameters` method in profile_searching.rb is needed because the value being passed into the `filter_by_policy_and_profile_names` ext method isn't being preprocessed in the `preprocess_parameters` method in override_scoped_search_escaping.rb. 

Also, `sanitize_sql_for_conditions` was needed because without it, there was a "Possible SQL injection" issue in the brakeman_output.codeclimate file when I ran all the tests. 

(I was able to test this using the 'like' operator with a name with spaces too and it did work correctly 👍 )

**Review by romanblanco — APPROVED** [review]
LGTM :+1:

**Comment by skateman** [comment]
I don't think you need preprocessing, you can specify what kind of operators you allow. Check the docs...

**Comment by marleystipich2** [comment]
@skateman Yeah, you can specify which operators that you want to allow but since all of the operators were allowed before, I figured we'd want to keep it consistent. But did we only want to only allow `=` and `!=`? Or everything besides like/unlike?

**Comment by skateman** [comment]
@marleystipich2 check what the frontend is supporting and go with that, but I think it's `=`, `!=`, `LIKE` and `NOT LIKE`. Anything else doesn't really make sense when comparing strings.

**Comment by marleystipich2** [comment]
@skateman I checked the frontend and it doesn't look like we're searching by profile/policy name anywhere.. Maybe I'm missing it or not looking in the right place. But the preprocessing does need to be done for LIKE and UNLIKE because it adds the `%...%` around the value which is needed for it to search correctly

**Comment by skateman** [comment]
> @skateman I checked the frontend and it doesn't look like we're searching by profile/policy name anywhere.. Maybe I'm missing it or not looking in the right place. But the preprocessing does need to be done for LIKE and UNLIKE because it adds the `%...%` around the value which is needed for it to search correctly

https://github.com/RedHatInsights/compliance-backend/blob/master/config/initializers/override_scoped_search_escaping.rb

File: app/models/concerns/profile_searching.rb [hunk 2]
```diff
@@ -148,6 +149,33 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
         conditions: "profiles.id IN (#{any_child_profile.to_sql})"
       }
     end
+
+    def filter_by_policy_and_profile_names(_filter, operator, value)
+      values = preprocess_parameters(operator, value)
```
skateman, line 154: As I wrote above this is not needed at all :see_no_evil:

app/models/concerns/profile_searching.rb
```diff
@@ -148,6 +149,33 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
...
+      cond = sanitize_sql_for_conditions(["policies.name #{operator} (?) OR profiles.name #{operator} (?)",
+                                          values, values])
```
skateman, line 156: Using raw SQL is always the worst possible option, you can write this nicer by using the [AR Query Interface](https://guides.rubyonrails.org/active_record_querying.html#or-conditions) at least partially. Or if that's not an option, there's always Arel. Something like this should work:

```
left_outer_joins(:policy).where("policies.name #{operator} ?", values).or(Profiles.where('profiles.name #{operator} ?', values))
```

app/models/concerns/profile_searching.rb
```diff
@@ -148,6 +149,33 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
...
+      search = left_outer_joins(:policy)
+      ids = search.where(cond)
+
+      { conditions: 'profiles.id IN (?)',
+        parameter: [ids.map(&:id)] }
```
skateman, line 161: A subquery should be always faster than a map. Haven't tried, but `ids.select(:id)` might work, or the sub-optimal `ids.pluck(:id)` will work for sure. Having a `map` makes Rails to do the iteration which is way more expensive than just sending everything for PG.

skateman (changes requested):

app/models/concerns/profile_searching.rb
marleystipich2: Referring to this comment: 
"According to [this](https://github.com/RedHatInsights/compliance-backend/blob/master/config/initializers/override_scoped_search_escaping.rb), the parameters are already preprocessed when you're getting them. The _ symbols are used to swap the % for like and unlike already. You don't really need to do any additional preprocessing."

The value that's passed to the ext_method `filter_by_policy_and_profile_names` isn't being updated by the preprocess_parameters [here](https://github.com/RedHatInsights/compliance-backend/blob/master/config/initializers/override_scoped_search_escaping.rb). The original value is always passed to the ext_method. I think it's just how the `ext_method`'s work or maybe it's a bug with search_scope

skateman: Maybe, but then you can just call the existing method :thinking:

marleystipich2: I thought about that too but there are two parameters that need to be passed into that one,`field` and `block`, and I'm not sure how to pass the correct values for those.. But I can try to figure it out!

marleystipich2: I wasn't able to call the `preprocess_parameters` method [here](https://github.com/RedHatInsights/compliance-backend/blob/master/config/initializers/override_scoped_search_escaping.rb) because the `field` param that's needed isn't being passed to the ext_method
The `field` param is an object of the class [ScopedSearch::Definition::Field](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/definition.rb) and I'm not sure if it's possible accessing this object from the `filter_by_policy_and_profile_names` :confused:

skateman: If I understand it [here](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb#L219) correctly, `this` might contain the `field` that you need. I'll give it a try and get back to you.

marleystipich2: Yeah, maybe I'm going to try now too

marleystipich2: My comment didn't post on Friday 😕

I figured out how to get the value of `field`. Based on [this](https://github.com/wvanbergen/scoped_search/blob/10d118690b6ba5a4941ac17e3d9157ddc817592f/lib/scoped_search/query_builder.rb#L425) `definition.klass` is `self` in the ext_method so I was able to get the `field` value by doing `self.scoped_search_definition.field_by_name('name')`
But now I need the ScopedSearch::QueryBuilder instance to call the `preprocess_parameters` method. Still trying to figure out if this is possible..

marleystipich2: I've looked through the scoped_search code and it doesn't look like there's any way of getting the `ScopedSearch::QueryBuilder::PostgreSQLAdapter` instance that's needed to call the `preprocess_parameters` method

skateman:
I dug through the [query builder](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb) and the documentation and maybe the best solution is to utilize the [value_translation](https://github.com/wvanbergen/scoped_search/wiki/Search-definition#value-translation). If this works out, we could get rid of the monkeypatch.

skateman:
I dug through the [query builder](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb) and the documentation and maybe the best solution is to utilize the [value_translation](https://github.com/wvanbergen/scoped_search/wiki/Search-definition#value-translation). If this works out, we could get rid of the monkeypatch.

marleystipich2:
> I dug through the [query builder](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb) and the documentation and maybe the best solution is to utilize the [value_translation](https://github.com/wvanbergen/scoped_search/wiki/Search-definition#value-translation). If this works out, we could get rid of the monkeypatch.

I tried to use value_translation and the value wasn't updating so I looked at the code and it looks like it's being used in the [map_value](https://github.com/wvanbergen/scoped_search/blob/10d118690b6ba5a4941ac17e3d9157ddc817592f/lib/scoped_search/query_builder.rb#L183) method which is being called in [preprocess_parameter](https://github.com/wvanbergen/scoped_search/blob/10d118690b6ba5a4941ac17e3d9157ddc817592f/lib/scoped_search/query_builder.rb#L249) which isn't being called when ext_methods are used :/

Roman and I were thinking we could try to update the gem itself so that the value is preprocessed before being passed to the ext_method. I tried updating [this if statement](https://github.com/wvanbergen/scoped_search/blob/10d118690b6ba5a4941ac17e3d9157ddc817592f/lib/scoped_search/query_builder.rb#L219) and adding a line to call `preprocess_parameter` before the value is sent to the ext_method but a lot of the tests were failing with an error like this when I did that: `ActiveRecord::PreparedStatementInvalid: wrong number of bind variables (4 for 2)`

Updating the gem might still work but I don't think it will be as simple as just calling preprocess_parameter before passing the value to the ext_method. Will keep trying tomorrow!

app/models/concerns/profile_searching.rb
```diff
@@ -148,6 +156,25 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
...
+
+    def filter_by_policy_and_profile_names(_filter, operator, value)
+      ss_qb = ScopedSearch::QueryBuilder.new(scoped_search_definition, '', scoped_search_definition.profile)
+      values = ss_qb.preprocess_parameters(
+        scoped_search_definition.field_by_name('name'), OPERATORS[operator], value
+      ) do |x|
+        # Passing empty block because nothing needs to happen on yield in preprocess_parameters
+        # The correct yield is invoked in to_ext_method_sql so all notification values are updated correctly
+        # See https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb
+      end
```
marleystipich2, line 168: This looks a little weird but I couldn't think of a better way to do this. A block needs to be passed to `preprocess_parameters` because of the yield. [This](https://github.com/wvanbergen/scoped_search/blob/10d118690b6ba5a4941ac17e3d9157ddc817592f/lib/scoped_search/query_builder.rb#L56) is the block that's normally passed so it's just updating the `notification` values. 
When ext_methods are used, this is being handled [here in the to_ext_method_sql method](https://github.com/wvanbergen/scoped_search/blob/10d118690b6ba5a4941ac17e3d9157ddc817592f/lib/scoped_search/query_builder.rb#L422). So nothing needs to happen in the block but because of the yield, passing a block isn't optional.

marleystipich2:
I was able to figure out a way to call preprocess_parameters from the ext_method. I added a comment [here](https://github.com/RedHatInsights/compliance-backend/pull/1411/files#r968563856) about the empty block that I passed to `preprocess_parameters` and explained why (I think) it's necessary.

app/models/concerns/profile_searching.rb
```diff
@@ -94,6 +100,8 @@ module ProfileSearching
 
   # class methods for profile searching
   module ClassMethods
+    include ScopedSearch
```
skateman, line 103: Is this needed because of the `scoped_search_definition`? I'm a bit worried about the side effects of including such a big module, isn't there any other way to refer these methods?

app/models/concerns/profile_searching.rb
```diff
@@ -4,11 +4,17 @@
 module ProfileSearching
   extend ActiveSupport::Concern
 
+  OPERATORS = { '=' => :eq, '<>' => :ne, 'LIKE' => :like, 'ILIKE' => :like,
+                'NOT LIKE' => :unlike, 'NOT ILIKE' => :unlike,
+                '>' => :gt, '<' => :lt, '<=' => :lte, '>=' => :gte,
+                'IN' => :in, 'NOT IN' => :notin }.freeze
```
skateman, line 10: Aren't these defined somewhere in scoped search? Then you don't need to define them twice, just refer the constant below from there and drop this.

app/models/concerns/profile_searching.rb
skateman: If you make an empty block, make it a one-liner and just add the comment explaining it before.

app/models/concerns/profile_searching.rb
```diff
@@ -148,6 +156,25 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
...
+        # See https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb
+      end
+
+      policy_cond = sanitize_sql_for_conditions(["policies.name #{operator} (?)", values])
+      profile_cond = sanitize_sql_for_conditions(["profiles.name #{operator} (?)", values])
+
+      profiles = left_outer_joins(:policy).where(policy_cond).or(Profile.where(profile_cond))
+
+      { conditions: 'profiles.id IN (?)',
+        parameter: [profiles.pluck(:id)] }
```
skateman, line 176: Do we really need the `pluck` here? Can't this be written with a simple select instead?

skateman (changes requested):

app/models/concerns/profile_searching.rb
marleystipich2: It actually worked without this. I'll remove it :slightly_smiling_face:

app/models/concerns/profile_searching.rb
marleystipich2: The constant in scoped search is the same as this but the keys/values are opposite. So it's `:eq => '='` instead of `'=' => :eq`. And that constant doesn't include `ILIKE` or `NOT ILIKE` :confused:

app/models/concerns/profile_searching.rb
marleystipich2: It did work with `select` instead of `pluck` 👍

app/models/concerns/profile_searching.rb
skateman: Let's extract these into the `ScopedSearch::QueryBuilder` monkeypatch we have in the initializers.

app/models/concerns/profile_searching.rb
```diff
@@ -148,6 +154,24 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
...
     end
+
+    def filter_by_policy_and_profile_names(_filter, operator, value)
+      ss_qb = ScopedSearch::QueryBuilder.new(scoped_search_definition, '', scoped_search_definition.profile)
+      # Passing empty block because nothing needs to happen on yield in preprocess_parameters
+      # The correct yield is invoked in to_ext_method_sql so all notification values are updated correctly
+      # See https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb
+      values = ss_qb.preprocess_parameters(
+        scoped_search_definition.field_by_name('name'), OPERATORS[operator], value
+      ) { |x| }
```
skateman, line 165: This could be extracted to a method, maybe into that monkeypatch (to collect bad stuff at one location) together with the `OPERATORS` like this:
```ruby
    def self.preprocess_parameters(definition, field, operator, value)
      # Long comment explaining dark magic
      qb = ScopedSearch::QueryBuilder.new(definition, '', definition.profile)
      qb.preprocess_parameters(definition.field_by_name(field), OPERATORS[operator], value) { |_| }
    end
```
And then you can call it as:
```ruby
ScopedSearch::QueryBuilder.preprocess_parameters(scoped_search_definition, 'name', operator, value)
```

app/models/concerns/profile_searching.rb
```diff
@@ -148,6 +154,24 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
...
+
+      policy_cond = sanitize_sql_for_conditions(["policies.name #{operator} (?)", values])
+      profile_cond = sanitize_sql_for_conditions(["profiles.name #{operator} (?)", values])
+
+      profiles = left_outer_joins(:policy).where(policy_cond).or(Profile.where(profile_cond))
+
+      { conditions: 'profiles.id IN (?)',
+        parameter: [profiles.select(:id)] }
```
skateman, line 173:
```suggestion
      { conditions: 'profiles.id IN (?)', parameter: [profiles.select(:id)] }
```

app/models/concerns/profile_searching.rb
```diff
@@ -148,6 +154,24 @@ def filter_by_policy(_filter, _operator, policy_or_profile_id)
...
+      ss_qb = ScopedSearch::QueryBuilder.new(scoped_search_definition, '', scoped_search_definition.profile)
+      # Passing empty block because nothing needs to happen on yield in preprocess_parameters
+      # The correct yield is invoked in to_ext_method_sql so all notification values are updated correctly
+      # See https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb
+      values = ss_qb.preprocess_parameters(
+        scoped_search_definition.field_by_name('name'), OPERATORS[operator], value
+      ) { |x| }
+
+      policy_cond = sanitize_sql_for_conditions(["policies.name #{operator} (?)", values])
+      profile_cond = sanitize_sql_for_conditions(["profiles.name #{operator} (?)", values])
```
skateman, line 168: Why are these arrays?

skateman (changes requested):

app/models/concerns/profile_searching.rb
marleystipich2: I thought that [sanitize_sql_for_conditions](https://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_for_conditions) only accepted arrays but I just checked and it also accepts strings. I'll update it to a string!

marleystipich2: Actually, it does work better passing the array into the [sanitize_sql_for_conditions](https://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_for_conditions) call since it takes the values parameter. `sanitize_sql_for_conditions` takes an array or a string.

config/initializers/override_scoped_search_escaping.rb
```diff
@@ -3,8 +3,22 @@
...
 module ScopedSearch
+  OPERATORS = { '=' => :eq, '<>' => :ne, 'LIKE' => :like, 'ILIKE' => :like,
+                'NOT LIKE' => :unlike, 'NOT ILIKE' => :unlike,
+                '>' => :gt, '<' => :lt, '<=' => :lte, '>=' => :gte,
+                'IN' => :in, 'NOT IN' => :notin }.freeze
+
   class QueryBuilder
     alias preprocess_parameters_old preprocess_parameters
+
+    def self.preprocess_parameters(definition, field, operator, value)
```
skateman, line 14: Please add some explanation why this method has been created, in which cases and how it should be used.

skateman (approved):
One small change and it's good to go :tada:

app/models/concerns/profile_searching.rb
skateman: @marleystipich2 please don't forget about this one

config/initializers/override_scoped_search_escaping.rb
marleystipich2: Okay, just added it 🙂

config/initializers/override_scoped_search_escaping.rb
```diff
@@ -5,6 +5,19 @@
 module ScopedSearch
   class QueryBuilder
     alias preprocess_parameters_old preprocess_parameters
+
+    # scoped_search gem does not preprocess values before passing to ext_method but sometimes it is necessary
+    # This method can be called from a scoped search ext_method to preprocess parameters
+    def self.preprocess_parameters(definition, field, operator, value)
+      inverted_operators = SQL_OPERATORS.invert.merge('ILIKE' => :like, 'NOT ILIKE' => :unlike)
```
skateman, line 12: This is not dynamic, so you can export it to a constant outside the method and freeze it.

skateman (changes requested):

skateman (approved):

PR #1422, opened 2022-08-18
fix(puma): RHICOMPL-3223 make the puma control socket path static

skateman:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -44,6 +44,6 @@ pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
 plugin :tmp_restart
 
 # Metrics
-activate_control_app
+activate_control_app("unix://#{File.expand_path(File.join(File.dirname(__FILE__), '../tmp/puma.sock'))}")
 plugin :yabeda
 plugin :yabeda_prometheus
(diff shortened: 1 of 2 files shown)


config/puma.rb
```diff
@@ -44,6 +44,6 @@
 plugin :tmp_restart
 
 # Metrics
-activate_control_app
+activate_control_app('unix://./tmp/sockets/puma.sock')
```
vkrizan, line 47: Would this work for forked workers?

skateman: Yes, puma has this solved for forked workers, you only control the parent process and not the children.

vkrizan: `./` bounds it to CWD. Shouldn't it be specific apps relative location?

skateman: CWD here is the specific app's relative location, Rails.root is not available in this context. Do you have any other suggestions how to put it into the `./tmp` folder?

vkrizan: Maybe from the file location?

skateman: Seems like an overkill, but okay

PR #1445, opened 2022-08-31
feat(GraphQL): RHICOMPL-3256 add inUse field for supportedProfiles

skateman:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


skateman:
@vkrizan this has some SQL stuff I'd like you to approve :pray:

app/graphql/types/profile.rb
```diff
@@ -118,13 +123,8 @@ def system_id(args)
...
     def latest_test_result_batch(args)
-      ::RecordLoader.for(
-        ::TestResult,
-        column: :host_id,
-        where: { profile_id: object.id },
-        includes: :rule_results, # TODO: check for lookaheads
-        order: 'created_at DESC'
-      ).load(system_id(args))
+      ::RecordLoader.for(::TestResult, column: :host_id, where: { profile_id: object.id },
+                                       includes: :rule_results, order: 'created_at DESC').load(system_id(args))
```
vkrizan, line 128: Seems like unrelated change.

app/models/os_major_version.rb
```diff
@@ -25,16 +25,20 @@ def self.aggregated_cast(column)
...
         ]
       ).as('int')
     ]
   ).as('os_major_version')
 
-  PROFILE_LAST_ID = Arel.sql("(#{OsMajorVersion.aggregated_cast(Profile.arel_table[:id]).to_sql})[1] as \"id\"")
-  PROFILE_BM_VERSIONS = OsMajorVersion.aggregated_cast(Xccdf::Benchmark.arel_table[:version]).as('bm_versions')
+  PROFILE_IN_USE = Arel::Nodes::True.new.as('in_use')
+  PROFILE_LAST_ID = Arel.sql("(#{aggregated_cast(Profile.arel_table[:id]).to_sql})[1] as \"id\"")
+  PROFILE_BM_VERSIONS = aggregated_cast(arel_table[:version]).as('bm_versions')
```
vkrizan, line 38: Refactoring, pls separate (all of the above).

app/models/os_major_version.rb
```diff
@@ -45,14 +49,25 @@ def self.aggregated_cast(column)
...
+                          .select(arel_table[:ref_id], OS_MAJOR_VERSION, PROFILE_IN_USE)
+
+    # Canonical profiles joined with profiles in use and grouped by supported benchmark versions
     supported_profiles = canonical.where(upstream: false)
-                                  .joins(:benchmark)
-                                  .select(PROFILE_LAST_ID, PROFILE_BM_VERSIONS, OS_MAJOR_VERSION)
-                                  .group(:ref_id, 'os_major_version')
+                                  .joins(PROFILE_BM_JOIN.join_sources)
+                                  .joins("LEFT OUTER JOIN (#{profiles_in_use.to_sql}) sq ON
+                                    sq.ref_id = profiles.ref_id AND
```
vkrizan, line 60: I feel like we are missing some condition. This pairs canonical profiles with all in use profiles for a current account just by `ref_id` and `os_major_version`. This compares only profiles and not benchmarks which parents. I'm wondering if we should add or replace `os_major_version` with benchmark `ref_id`.

app/graphql/types/profile.rb
skateman: It is a related change, we have some rubocop limits :grin:

app/models/os_major_version.rb
skateman: What do you mean?

app/models/os_major_version.rb
skateman: `os_major_version` IS (derived from) benchmark `ref_id`

vkrizan: Yes, it is, today. :-)

app/models/os_major_version.rb
vkrizan: Are there any changes here, because I only see changes like `OsMajorVersion.aggregated_cast` -> `aggregated_cast`

app/models/os_major_version.rb
skateman: We're in a big trouble if that changes :see_no_evil:

app/models/os_major_version.rb
skateman: Oh, you mean separate commit?

vkrizan: Yes please.

app/models/os_major_version.rb
vkrizan: Logically it would be better to match on benchmarks as well. Would that increase the SQL load? I believe that the `os_major_version` comes from benchmark nevertheless.

app/models/os_major_version.rb
```diff
@@ -45,14 +49,25 @@ def self.aggregated_cast(column)
...
+                                  .select(PROFILE_LAST_ID, PROFILE_BM_VERSIONS, OS_MAJOR_VERSION, 'sq.in_use')
+                                  .group(:ref_id, Xccdf::Benchmark.arel_table[:ref_id], 'in_use')
 
+    # Rails cannot consume grouped results as proper ActiveRecord classes, it has to be joined with itself
     canonical.where(upstream: false)
              .joins("INNER JOIN (#{supported_profiles.to_sql}) t ON t.id = profiles.id")
-             .select('"profiles".*, "t"."bm_versions" AS "bm_versions", "t"."os_major_version" AS "os_major_version"')
+             .select('"profiles".*, "t"."bm_versions" AS "bm_versions"',
+                     '"t"."os_major_version" AS "os_major_version"',
+                     '"t"."in_use" AS "in_use"')
```
vkrizan, line 70: What value would `in_use` have for profiles that are NOT in use? Is it `NULL`?

skateman: Yes

skateman: And it gets converted to boolean in GQL [here](https://github.com/RedHatInsights/compliance-backend/pull/1445/files#diff-55bef2bb2e7b33ab10c18422e60ddb05e02505244e59107f8ce099417e74b6d9R113)

app/models/os_major_version.rb
vkrizan: I had benchmark `ref_id` in mind which would be similar to OS major check, but seemingly logically more correct (?).

app/models/os_major_version.rb
vkrizan: I see. Wouldn't it be safer to do `COALESCE(t.in_use, FALSE)`? Now it is pushed downwards to GraphQL.

app/models/os_major_version.rb
skateman: And with 3x the complexity :sweat:

app/models/os_major_version.rb
skateman: Good idea!

app/models/os_major_version.rb
vkrizan: Check the `OS_MAJOR_VERSION` field definition :smile:, that is more complex compared to ref_id.

PR #1466, opened 2022-09-13
perf(Policy): RHICOMPL-3226 only update counters when updating hosts

skateman:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -27,8 +27,6 @@ class Policy < ApplicationRecord
   after_update :destroy_orphaned_business_objective
   after_rollback :destroy_orphaned_business_objective
 
-  after_save :update_counters!
-
   scope :with_hosts, lambda { |hosts|
     joins(:hosts).where(hosts: { id: hosts }).distinct
   }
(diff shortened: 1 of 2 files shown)


app/models/policy.rb
```diff
@@ -27,8 +27,6 @@ class Policy < ApplicationRecord
   after_update :destroy_orphaned_business_objective
   after_rollback :destroy_orphaned_business_objective
 
-  after_save :update_counters!
```
vkrizan, line 30: why?

vkrizan:
Can you please analyze all places where a save or update is called for a Policy?
This might break numbering of compliant hosts from test results.

app/models/policy.rb
skateman: Because this is only called via `Policy#update` and was only being used in tests.

skateman:
> Can you please analyze all places where a save or update is called for a Policy? This might break numbering of compliant hosts from test results.

If you do a fulltext search on `update_counters!`, you'll see that every aspect is being covered. The only place the `after_save` was being used is the `Policy#update` call that has been only used in tests.

vkrizan (approved):

PR #1507, opened 2022-11-10
refactor(RuleGroups): RHICOMPL-3402 make the mixin more readable

skateman:
@marleystipich2 please go through this, even if it's merged by the time you check it. Took your extraction from #1273 and moved it here + renamed it to `with_relationships` as it calls the same scope on the `RuleGroupRelationship` model. You don't need the `rules` and `rule_groups` as an argument as they are available in the scope of the `Benchmark` and the `Profile`. The usage of symbols and the `x: y` (seattle-style) syntax in hashes is preferred over the hash rockets and the strings `'x' => y` when possible. Also when a method is about building a hash (or hashes) with a lot of keys, it is allowed to disable the linelength cop and move them on separate lines as it is more readable. 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/concerns/rules_and_rule_groups.rb
+++ b/app/models/concerns/rules_and_rule_groups.rb
@@ -5,26 +5,40 @@ module RulesAndRuleGroups
   extend ActiveSupport::Concern
 
   included do
+    # rubocop:disable Metrics/MethodLength
     def rules_and_rule_groups
       conflicts = relationships_for('conflicts')
       requires = relationships_for('requires')
       arranged_rule_groups = rule_groups.includes(:rules).arrange_serializable do |parent, children|
-        { 'rule_group' => parent, 'group_children' => children,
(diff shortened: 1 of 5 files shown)


romanblanco (approved):
LGTM :+1:

marleystipich2:
This all makes sense to me! Thanks for making these changes 🙂 
That's good to know that we can disable rubocop when building hashes like this because it is a lot more readable. 
I remember you mentioning in the past to use `x: y` syntax for hashes but the only reason I used `'x' => y` is because I thought using the string key was correct JSON format but I think `x: y` should be good too.

PR #1534, opened 2022-11-29
feat: RHICOMPL-3349 add value_definitions database table

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


db/migrate/20221129094342_create_value_definitions.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Data model for storing value definitions
+class CreateValueDefinitions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :values, id: :uuid do |t|
```
marleystipich2, line 6: I think we'll need a column in this table for the `value` itself and the column type would be string. Unless @skateman told you otherwise?

skateman (changes requested):
+ a model test please

app/models/value_definition.rb
```diff
@@ -0,0 +1,15 @@
...
+  scoped_search on: :ref_id
+  include ShortRefId
+
+  belongs_to :benchmark, class_name: 'Xccdf::Benchmark'
+
+  validates :ref_id, uniqueness: { scope: %i[benchmark_id] }, presence: true
+  validates :description, presence: true
+  validates :benchmark_id, presence: true
+
+  POSSIBLE_VALUE_TYPES = %w[string number boolean]
```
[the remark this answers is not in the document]
romanblanco: bool, correcting.

db/migrate/20221129094342_create_value_definitions.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.string :ref_id
+      t.string :title
+      t.text :description
+      t.string :value_type
+      t.numeric :lower_bound, array: true, default: nil
+      t.numeric :upper_bound, array: true, default: nil
```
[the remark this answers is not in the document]
marleystipich2: According to the XCCDF specs, lower/upper bounds are "decimals" which is a float and since we're only worrying about the default upper/lower bounds for now I don't think this needs to be an array.

skateman (changes requested):

db/migrate/20221129094342_create_value_definitions.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    create_table :value_definitinos, id: :uuid do |t|
+      t.default_value :string
+      t.string :ref_id
+      t.uuid :benchmark_id, null: false
+      t.string :title
+      t.text :description
+      t.string :value_type
+      t.numeric :lower_bound, default: nil
+      t.numeric :upper_bound, default: nil
```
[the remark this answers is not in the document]
marleystipich2: Screenshot 2022-11-29 at 2 29 20 PM

It looks like `numeric` is similar to `decimal` or `float`. I think this should be fine because lower_bound and upper_bound can be floats/decimals.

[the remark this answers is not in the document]
skateman: Thanks!

app/models/value_definition.rb
```diff
@@ -0,0 +1,12 @@
...
+
+# Stores information about value definitions. This comes from SCAP.
+class ValueDefinition < ApplicationRecord
+  belongs_to :benchmark, class_name: 'Xccdf::Benchmark'
+
+  validates :ref_id, uniqueness: { scope: %i[benchmark_id] }, presence: true
+  validates :description, presence: true
+  validates :value_type, presence: true
+
+  POSSIBLE_VALUE_TYPES = %w[string number bool].freeze
```
skateman, line 11: Why to set this constant if we're not using it in a validation as requested?

db/migrate/20221129094342_create_value_definitions.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Data model for storing value definitions
+class CreateValueDefinitions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :value_definitions, id: :uuid do |t|
+      t.string :default_value
```
skateman, line 7: Logically this should be AFTER `value_type`

skateman (changes requested):

skateman:
You're missing the two fields defined on the rules and profiles from the architecture document. You can just simply put two `add_column` calls into the same migration. Also you need to implement some model tests as requested earlier.

test/models/value_definition_test.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ValueDefinitionTest < ActiveSupport::TestCase
+  should belong_to(:benchmark)
+  should validate_presence_of :ref_id
+  should validate_uniqueness_of(:ref_id).scoped_to(:benchmark_id)
+  should validate_presence_of(:value_type)
+  should validate_presence_of(:description)
```
skateman, line 10: Maybe a test case where you try to assign something else to the value type than allowed?

db/migrate/20221129094342_create_value_definitions.rb
```diff
@@ -0,0 +1,22 @@
...
+    create_table :value_definitions, id: :uuid do |t|
+      t.string :ref_id
+      t.string :title
+      t.text :description
+      t.string :value_type
+      t.string :default_value
+      t.numeric :lower_bound, default: nil
+      t.numeric :upper_bound, default: nil
+      t.uuid :benchmark_id
+      t.references :benchmark_id, type: :uuid, index: true
```
skateman, line 15: 1. Why you have them twice?
2. Why are you referencing to `benchmark_id`?

https://github.com/RedHatInsights/compliance-backend/blob/master/db/migrate/20220202002459_create_rule_groups.rb#L10

db/schema.rb
```diff
@@ -258,6 +258,20 @@
...
+    t.string "ref_id"
+    t.string "title"
+    t.text "description"
+    t.string "value_type"
+    t.string "default_value"
+    t.decimal "lower_bound"
+    t.decimal "upper_bound"
+    t.uuid "benchmark_id"
+    t.uuid "benchmark_id_id"
+    t.index ["benchmark_id_id"], name: "index_value_definitions_on_benchmark_id_id"
```
skateman, line 271: This does not make sense :confused: see above.

db/migrate/20221129094342_create_value_definitions.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    add_column :rules, :value_checks, :uuid, array: true
+    add_column :profiles, :value_overrides, :json
```
skateman, line 19: @marleystipich2 I'm not sure about the naming here. If we do the refinements/selectors, would we merge that information here into the column statically? In this case the naming is correct...otherwise we have the generic problem of starting a metal band.

skateman (changes requested):

db/migrate/20221201094342_create_value_definitions.rb
```diff
@@ -0,0 +1,21 @@
...
+      t.string :value_type
+      t.string :default_value
+      t.numeric :lower_bound, default: nil
+      t.numeric :upper_bound, default: nil
+      t.references :benchmark, type: :uuid, null: false, foreign_key: true
+      t.index [:ref_id, :benchmark_id], unique: true
+    end
+
+    add_column :rules, :value_checks, :uuid, array: true, default: []
+    add_column :profiles, :value_overrides, :json, defalt: {}
```
skateman, line 19:
```suggestion
    add_column :profiles, :value_overrides, :jsonb, defalt: {}
```

test/models/value_definition_test.rb
skateman: ^^^

skateman (changes requested):

db/migrate/20221201094342_create_value_definitions.rb
skateman: Tests are failing because you can't call distinct on a JSON column, only on JSONB

db/migrate/20221129094342_create_value_definitions.rb
marleystipich2: I think `value_overrides` would be okay for profile values that were refined but doesn't make sense for profile values that weren't refined because nothing was overwritten.

But I don't know if we need to track values for profiles that haven't been refined  anyway because if a value isn't in `value_overrides` then we'll know that the profile didn't refine it so they can get the default value from the `values` table?

marleystipich2: Talked about this over slack and decided `value_overrides` should be good 👍 Updated the architecture doc with more info about this column and what it will be used for

db/migrate/20221201094342_create_value_definitions.rb
```diff
@@ -0,0 +1,21 @@
...
+      t.string :value_type
+      t.string :default_value
+      t.numeric :lower_bound, default: nil
+      t.numeric :upper_bound, default: nil
+      t.references :benchmark, type: :uuid, null: false, foreign_key: true
+      t.index [:ref_id, :benchmark_id], unique: true
+    end
+
+    add_column :rules, :value_checks, :uuid, array: true, default: []
+    add_column :profiles, :value_overrides, :jsonb, defalt: {}
```
skateman, line 19:
```suggestion
    add_column :profiles, :value_overrides, :jsonb, default: {}
```

test/factories/value_definition.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :value_definition do
+    title { Faker::Lorem.sentence }
+    value_type { 'bool' }
+    ref_id { "foo_rule_#{SecureRandom.uuid}" }
```
skateman, line 7:
```suggestion
    ref_id { "foo_rule_group_#{SecureRandom.uuid}" }
```

skateman (changes requested):

skateman (approved):
The Seal of Approval

PR #1551, opened 2022-12-08
feat(Tailoring): RHICOMPL-3398 add set-value to tailoring file

marleystipich2:
Waiting to merge this until values are imported in the DB: https://issues.redhat.com/browse/RHICOMPL-3475

Example of what's added to the tailoring file:

`<set-value idref="foo_value_e61824be-c7a3-4701-96f6-55b21cdafd7b">testing</set-value>
<set-value idref="foo_value_62d90c6d-3cac-491d-a778-f236e7d9ddad">true</set-value>
<set-value idref="foo_value_0f821a89-4734-46d7-80d7-78dd459ad14a">3</set-value>`

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/services/xccdf_tailoring_file.rb
+++ b/app/services/xccdf_tailoring_file.rb
@@ -42,7 +42,7 @@ class XccdfTailoringFile
       xml[XCCDF].description(@profile.description,
                              'xmlns:xhtml' => 'http://www.w3.org/1999/xhtml',
                              'xml:lang' => 'en-US', override: true)
-      selections_builder(xml)
+      tailoring_builder(xml)
     end
   end
 
@@ -58,9 +58,17 @@ class XccdfTailoringFile
     end
   end
 
-  def selections_builder(xml)
+  def value_builder(xml)
+    @set_values.each do |value_ref_id, value|
+      # nokogiri maintainers recommend using send for tags with hyphens
+      xml[XCCDF].send('set-value', value, idref: value_ref_id)
+    end
+  end
+
+  def tailoring_builder(xml)
     rule_selections_builder(xml)
     rule_group_selections_builder(xml)
+    value_builder(xml)
   end
 
   def create_builder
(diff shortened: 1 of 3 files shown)


app/services/xccdf_tailoring_file.rb
```diff
@@ -58,9 +58,16 @@ def rule_group_selections_builder(xml)
     end
   end
 
-  def selections_builder(xml)
+  def value_builder(xml)
+    @set_values.each do |value_ref_id, value|
+      xml[XCCDF].send('set-value', value, idref: value_ref_id)
```
skateman, line 63: Calling `send` is a big red flag, especially with a fixed method name. Please look up the documentation of the class you're interacting with and find an alternative solution.

skateman (changes requested):
Looks good, but merging this before the values are in the DB can cause bugs for customers...

app/services/xccdf_tailoring_file.rb
marleystipich2: Besides `send`, the only other option I've seen is to use `tag!` like:
`xml[XCCDF].tag! 'set-value', value, idref: value_ref_id`

But I don't think this is supported anymore because it doesn't work correctly. I can look into this more after we finish importing values into the DB but just wanted to leave a comment here.

marleystipich2: @skateman What do you think about this? ^ 
Should I keep `send` and explain it with a comment or still try to find another option?

skateman: I guess it should be fine

marleystipich2: Okay, thanks I'll still add a comment explaining why we're using it

PR #1556, opened 2022-12-13
feat(Import): RHICOMPL-3475 import Values into the DB

marleystipich2:
Purpose of this PR:
- Update openscap_parser gem to version 1.6.0
- Import Values into the `value_definitions` table
- Import `value_overrides` that holds value refinements for a Profile
- Import `value_checks` for a Rule (these are the values that are used by the Rule)

Note: I imported the `lower_bound`s and `upper_bound`s of a ValueDefinition since we are parsing the info in openscap_parser but they are not used at all in our datastreams and the openscap team said they have no plans to use them in the future.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/rule.rb
+++ b/app/models/rule.rb
@@ -78,7 +78,9 @@ class Rule < ApplicationRecord
     (profiles & Profile.canonical).any?
   end
 
-  def self.from_openscap_parser(op_rule, existing: nil, rule_group_id: nil, benchmark_id: nil, precedence: nil)
+  # rubocop:disable Metrics/ParameterLists
+  def self.from_openscap_parser(op_rule, existing: nil, rule_group_id: nil,
+                                benchmark_id: nil, precedence: nil, value_checks: nil)
     rule = existing || new(ref_id: op_rule.id, benchmark_id: benchmark_id)
 
     rule.op_source = op_rule
@@ -87,10 +89,11 @@ class Rule < ApplicationRecord
                            rationale: op_rule.rationale, severity: op_rule.severity,
                            precedence: precedence, rule_group_id: rule_group_id,
                            upstream: false, slug: rule.normalize_friendly_id(op_rule.id),
-                           identifier: op_rule.identifier&.to_h)
+                           value_checks: value_checks, identifier: op_rule.identifier&.to_h)
 
     rule
   end
+  # rubocop:enable Metrics/ParameterLists
 
   def compliant?(host, profile)
     Rails.cache.fetch("#{id}/#{host.id}/compliant") do
--- a/app/models/value_definition.rb
+++ b/app/models/value_definition.rb
@@ -2,12 +2,25 @@
 
 # Stores information about value definitions. This comes from SCAP.
 class ValueDefinition < ApplicationRecord
+  include OpenscapParserDerived
+
   belongs_to :benchmark, class_name: 'Xccdf::Benchmark'
 
-  POSSIBLE_VALUE_TYPES = %w[string number bool].freeze
+  POSSIBLE_VALUE_TYPES = %w[string number boolean].freeze
 
   validates :benchmark_id, presence: true
   validates :ref_id, uniqueness: { scope: %i[benchmark_id] }, presence: true
   validates :description, presence: true
   validates :value_type, presence: true, inclusion: { in: POSSIBLE_VALUE_TYPES }
+
+  def self.from_openscap_parser(op_vd, existing: nil, benchmark_id: nil)
+    value_definition = existing || new(ref_id: op_vd.id, benchmark_id: benchmark_id)
+
+    value_definition.op_source = op_vd
+
+    value_definition.assign_attributes(title: op_vd.title, description: op_vd.description,
+                                       value_type: op_vd.type, default_value: op_vd.value)
+
+    value_definition
+  end
 end
--- a/app/services/concerns/xccdf/profiles.rb
+++ b/app/services/concerns/xccdf/profiles.rb
@@ -11,7 +11,8 @@ module Xccdf
           ::Profile.from_openscap_parser(
             op_profile,
             existing: old_profiles[op_profile.id],
-            benchmark_id: @benchmark&.id
+            benchmark_id: @benchmark&.id,
+            value_overrides: value_overrides(op_profile)
           )
         end
 
@@ -29,6 +30,13 @@ module Xccdf
         @old_profiles ||= ::Profile.where(
           ref_id: @op_profiles.map(&:id), benchmark: @benchmark&.id
         ).index_by(&:ref_id)
+      end
+
+      def value_overrides(op_profile)
+        op_profile.refined_values.each_with_object({}) do |(value_id, selector), value_map|
+          op_value = value_definition_for(ref_id: value_id).op_source
+          value_map[value_id] = op_value.value(selector)
+        end
       end
     end
   end
--- a/app/services/concerns/xccdf/rules.rb
+++ b/app/services/concerns/xccdf/rules.rb
@@ -9,12 +9,12 @@ module Xccdf
       def rules
         @rules ||= @op_rules.each_with_index.map do |op_rule, idx|
           rule_group = rule_group_for(ref_id: op_rule.parent_id)
+          value_checks = op_rule.values.map { |value_ref_id| value_definition_for(ref_id: value_ref_id).id }
 
           ::Rule.from_openscap_parser(
             op_rule,
-            existing: old_rules[op_rule.id],
-            precedence: idx,
-            rule_group_id: rule_group&.id,
+            existing: old_rules[op_rule.id], precedence: idx,
+            rule_group_id: rule_group&.id, value_checks: value_checks,
             benchmark_id: @benchmark&.id
           )
         end
@@ -28,7 +28,7 @@ module Xccdf
         ::Rule.import(old_rules.values,
                       on_duplicate_key_update: {
                         conflict_target: %i[ref_id benchmark_id],
-                        columns: %i[identifier description precedence rationale rule_group_id severity]
+                        columns: %i[identifier description precedence rationale rule_group_id severity value_checks]
                       }, validate: false)
       end
 
--- a/app/services/concerns/xccdf/value_definitions.rb
+++ b/app/services/concerns/xccdf/value_definitions.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Xccdf
+  # Methods related to saving value definitions
+  module ValueDefinitions
+    extend ActiveSupport::Concern
+
+    included do
+      def value_definitions
+        @value_definitions ||= @op_value_definitions.map do |op_value_definition|
+          ::ValueDefinition.from_openscap_parser(
+            op_value_definition,
+            existing: old_value_definitions[op_value_definition.id],
+            benchmark_id: @benchmark&.id
+          )
+        end
+      end
+
+      def save_value_definitions
+        # Import the new records first with validation
+        ::ValueDefinition.import!(new_value_definitions, ignore: true)
+
+        # Update the fields on existing value_definitions, validation is not necessary
+        ::ValueDefinition.import(old_value_definitions.values,
+                                 on_duplicate_key_update: {
+                                   conflict_target: %i[ref_id benchmark_id],
+                                   columns: %i[description default_value]
+                                 }, validate: false)
+      end
+
+      private
+
+      def new_value_definitions
+        @new_value_definitions ||= value_definitions.select(&:new_record?)
+      end
+
+      def old_value_definitions
+        @old_value_definitions ||= ::ValueDefinition.where(
+          ref_id: @op_value_definitions.map(&:id), benchmark_id: @benchmark&.id
+        ).index_by(&:ref_id)
+      end
+
+      def value_definition_for(ref_id:)
+        @cached_value_definitions ||= @value_definitions.index_by(&:ref_id)
+        @cached_value_definitions[ref_id]
+      end
+    end
+  end
+end
--- a/test/factories/value_definition.rb
+++ b/test/factories/value_definition.rb
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :value_definition do
     title { Faker::Lorem.sentence }
-    value_type { 'bool' }
+    value_type { 'boolean' }
     ref_id { "foo_rule_group_#{SecureRandom.uuid}" }
     description { Faker::Lorem.paragraph }
     benchmark
--- a/test/services/concerns/xccdf/profiles_test.rb
+++ b/test/services/concerns/xccdf/profiles_test.rb
@@ -15,26 +15,52 @@ module Xccdf
     end
 
     include Xccdf::Profiles
+    include Xccdf::ValueDefinitions
 
     setup do
       @account = FactoryBot.create(:account)
       @host = FactoryBot.create(:host, org_id: @account.org_id)
       @benchmark = FactoryBot.create(:canonical_profile, :with_rules).benchmark
       parser = OpenscapParser::TestResultFile.new(
-        file_fixture('xccdf_report.xml').read
+        file_fixture('rhel-xccdf-report.xml').read
       )
       @op_profiles = parser.benchmark.profiles
+      @op_value_definitions = parser.benchmark.values
       @op_test_result = parser.test_result
+      save_value_definitions
     end
 
     test 'save_profiles' do
-      assert_difference('Profile.count', 1) do
+      assert_difference('Profile.count', 10) do
         save_profiles
       end
 
       assert_no_difference('Profile.count') do
         save_profiles
       end
+    end
+
+    test 'correctly save value_overrides' do
+      save_profiles
+
+      profile1 = Profile.find_by(ref_id: 'xccdf_org.ssgproject.content_profile_pci-dss')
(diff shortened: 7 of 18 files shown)


marleystipich2:
@skateman 
I have a couple questions/comments about this: 
- I updated the `value_checks` column on the `rules` table to be an array of strings instead of an array of uuids. Storing the value ref_ids like this `Rule#value_checks:array -> [ref_id, ref_id, ref_id]` should be okay because the value ref_ids will always be unique under the benchmark. This simplifies the import because we don't need to go through `@op_rule.values` and map the `ref_ids` to `uuids`. It also keeps it consistent with the `value_overrides` column on the `profiles` table where we use ref_ids like this `Profile#value_overrides:json -> {ref_id: value, ref_id: value} `
- I think we're going to need to add a `lower_bound_overrides` and `upper_bound_overrides` column to Profile to track those overrides as well. Something like `Profile#lower_bound_overrides:json -> {ref_id: lower_bound, ref_id: lower_bound}` and `Profile#upper_bound_overrides:json -> {ref_id: upper_bound, ref_id: upper_bound}`

One other thing I wanted to mention that I just noticed.. lower and upper bounds aren't used at all in the datastreams but I guess it's good that we'll be supporting it anyway

skateman:
> I updated the value_checks column on the rules table to be an array of strings instead of an array of uuids.

A string is far more costly in storage than a UUID (basically a very long integer), storing an array of these makes it way more expensive. Whenever we call a DISTINCT query on profiles, we would have to compare strings with strings instead of UUIDs. I guess I don't have to go deeper into which one costs us more.

> Storing the value ref_ids like this Rule#value_checks:array -> [ref_id, ref_id, ref_id] should be okay because the value ref_ids will always be unique under the benchmark.

Instead of using the properties and performance of global uniqueness, you are proposing to have scoped uniqueness which is far more costly.

> This simplifies the import because we don't need to go through @op_rule.values and map the ref_ids to uuids.

Import runs once, simplifying it will have a cost when generating the tailoring file as you'd have to do a more expensive lookup based on `ref_id` and `benchmark_id` instead of the primary key. Having a bit more complex import saves us a lot of additional computation in the future. Also the complexity of a lookup from `ref_id` to `id` in the context of importing has a linear complexity as we already have a `ref_id -> id` mapping in the `rule_for` method.

> It also keeps it consistent with the value_overrides column on the profiles table where we use ref_ids like this Profile#value_overrides:json -> {ref_id: value, ref_id: value}

I'm not sure if the `value_overrides` should map `ref_id` to `value`, nor that this consistency would provide us any gain in performance. It really depends on how the values in the tailoring file look like...

>  lower and upper bounds aren't used at all in the datastreams

Let's ask the OpenSCAP folks before we do anything with our current DB schema.

marleystipich2:
I just messaged the openscap team to see if there are plans to use lower-bound and upper-bounds in the future. Waiting to hear back. 

But that all makes sense! I didn't think about the fact that a storing an array of strings is more expensive than uuids. I'll change it back to uuids and map the ref_ids in `op_rule.values` to uuids.

app/models/rule.rb
```diff
@@ -80,7 +80,9 @@ def canonical?
     (profiles & Profile.canonical).any?
   end
 
-  def self.from_openscap_parser(op_rule, existing: nil, rule_group_id: nil, benchmark_id: nil, precedence: nil)
+  # rubocop:disable Metrics/ParameterLists
+  def self.from_openscap_parser(op_rule, existing: nil, rule_group_id: nil,
+                                benchmark_id: nil, precedence: nil, value_checks: nil)
```
marleystipich2, line 85: @skateman Is it okay to disable this rubocop check for this method? We need to pass `value_checks` into this method but it brought the parameter count to 6 which rubocop complains about 🙁

app/models/value_definition.rb
```diff
@@ -4,10 +4,20 @@
 class ValueDefinition < ApplicationRecord
   belongs_to :benchmark, class_name: 'Xccdf::Benchmark'
 
-  POSSIBLE_VALUE_TYPES = %w[string number bool].freeze
+  POSSIBLE_VALUE_TYPES = %w[string number boolean].freeze
```
skateman, line 7: Why? How does it look like in the XCCDF specs? cc @romanblanco

app/models/rule.rb
skateman: Yeah, let's keep it like that

app/models/value_definition.rb
```diff
@@ -4,10 +4,20 @@
...
   validates :ref_id, uniqueness: { scope: %i[benchmark_id] }, presence: true
   validates :description, presence: true
   validates :value_type, presence: true, inclusion: { in: POSSIBLE_VALUE_TYPES }
+
+  def self.from_openscap_parser(op_vd, existing: nil, benchmark_id: nil)
+    value_definition = existing || new(ref_id: op_vd.id, benchmark_id: benchmark_id)
+
+    value_definition.assign_attributes(title: op_vd.title, description: op_vd.description,
+                                       value_type: op_vd.type, default_value: op_vd.value,
+                                       lower_bound: op_vd.lower_bound.to_f, upper_bound: op_vd.upper_bound.to_f)
```
skateman, line 19: Do we really need these last two?

app/services/concerns/xccdf/profiles.rb
```diff
@@ -30,6 +31,13 @@ def old_profiles
           ref_id: @op_profiles.map(&:id), benchmark: @benchmark&.id
         ).index_by(&:ref_id)
       end
+
+      def value_overrides(op_profile)
+        op_profile.refined_values.each_with_object({}) do |(value_id, selector), value_map|
+          value_definition = @op_value_definitions.find { |v| v.id == value_id }
```
skateman, line 37: This is `O(n*m)` and you could do this with `O(n*1)` if you follow the same pattern that we already have in other places with `index_by`.

skateman: And I see that you already do the indexing below...please use it.

app/services/concerns/xccdf/rules.rb
```diff
@@ -43,6 +43,11 @@ def rule_for(ref_id:)
         @cached_rules[ref_id]
       end
 
+      def value_ids_for(ref_ids:)
+        @cached_values ||= @value_definitions.index_by(&:ref_id)
+        ref_ids.map { |ref_id| @cached_values[ref_id].id }
+      end
```
skateman, line 49: Mixing variable names and memoized instance variables. Please don't do that and follow the existing patterns instead.

skateman: Also anything cached related to values should go where value definitions are declared. Again...follow existing patterns.

test/factories/value_definition.rb
```diff
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :value_definition do
     title { Faker::Lorem.sentence }
-    value_type { 'bool' }
+    value_type { 'boolean' }
```
skateman, line 6: See above.

test/services/concerns/xccdf/profiles_test.rb
```diff
@@ -36,5 +37,19 @@ def report_description
...
+      profile1 = Profile.find_by(ref_id: 'xccdf_org.ssgproject.content_profile_standard')
+      profile2 = Profile.where(value_overrides: {}).first
+      expected_overrides = { 'xccdf_org.ssgproject.content_value_var_auditd_num_logs' => '5',
+                             'xccdf_org.ssgproject.content_value_sshd_idle_timeout_value' => '300',
+                             'xccdf_org.ssgproject.content_value_var_accounts_maximum_age_login_defs' => '90',
+                             'xccdf_org.ssgproject.content_value_var_accounts_minimum_age_login_defs' => '7',
+                             'xccdf_org.ssgproject.content_value_var_accounts_password_minlen_login_defs' => '12',
+                             'xccdf_org.ssgproject.content_value_var_accounts_password_warn_age_login_defs' => '7' }
+      assert_equal expected_overrides, profile1.value_overrides
+      assert_equal 0, profile2.value_overrides.length
```
skateman, line 52: Please explain the significance of `profile2` in this test, because it does not make sense to me.

test/services/concerns/xccdf/rules_test.rb
```diff
@@ -47,6 +50,19 @@ def initialize(report_contents)
...
+    @mock.save_rules
+    rule1 = Rule.find_by(ref_id: 'xccdf_org.ssgproject.content_rule_network_ipv6_disable_interfaces')
+    rule2 = Rule.find_by(ref_id: 'xccdf_org.ssgproject.content_rule_sshd_set_idle_timeout')
+    value1 = ValueDefinition.find_by(ref_id: 'xccdf_org.ssgproject.content_value_sshd_required')
+    value2 = ValueDefinition.find_by(ref_id: 'xccdf_org.ssgproject.content_value_sshd_idle_timeout_value')
+
+    assert_includes rule2.value_checks, value1.id
+    assert_includes rule2.value_checks, value2.id
+    assert_equal 2, rule2.value_checks.length
+    assert_equal [], rule1.value_checks
```
skateman, line 63: You don't set up a single context for this test...how do you actually verify things here?

skateman (changes requested):

app/models/value_definition.rb
marleystipich2: It's `boolean` in the XCCDF specs. I just noticed it because during the values import, boolean types weren't being imported correctly since it wasn't a possible value type.

app/models/value_definition.rb
marleystipich2: No, not really. I was going to ask you about this. I think we can leave lower_bound and upper_bound off until openscap decides they want to start using them(which they said they aren't planning on doing)

test/services/concerns/xccdf/rules_test.rb
marleystipich2: `@mock.save_value_definitions` is being done in the setup which is making sure `@value_definitions` is defined because it's used in `@mock.save_rules`. 
The purpose of this test is just to make sure that the `value_checks` column is being set up correctly during `@mock.save_rules`. To verify, I used specific rule and value ref_ids from the test file that's being used for these tests: `xccdf_report.xml`

test/services/concerns/xccdf/profiles_test.rb
marleystipich2: It was supposed to test a profile that has no value refinements so `value_overrides` should be `{}` but I just realized it doesn't actually make sense the way that I defined `profile2` 🤔 But if I update it to `profile2 = Profile.find_by(ref_id: profile_with_no_refinements_ref_id)` that should work

test/services/concerns/xccdf/rules_test.rb
marleystipich2: `rule1` is a rule in `xccdf_report.xml` that has no value checks and `rule2` is a rule in `xccdf_report.xml` that has 2 value checks.

app/services/concerns/xccdf/rules.rb
marleystipich2: I don't understand this comment: 
`Mixing variable names and memoized instance variables. Please don't do that and follow the existing patterns instead.`

Is the issue with this line? 
`ref_ids.map { |ref_id| @cached_values[ref_id].id }`

app/models/value_definition.rb
romanblanco: @marleystipich2 is right, this was a mistake

image

app/services/concerns/xccdf/rules.rb
skateman: No, the issue is with the whole method, you are mixing existing patterns together. The `@cached_xy` is being used in the `xys_for` method and it is intended to be called multiple times. Here you are mixing it together with the `xy_ids_for` method that is being used with the `@xy` instance variable elsewhere. Also the method name is `value_ids_for` while you're working with `value_definitions`, these should be kept consistent across this codebase.

My suggestion is to extract the `value_definition_id_for` (with a cached instance variable) into a separate method in the `ValueDefinitions` module and use it here + rename this method as well to keep all consistent.

test/services/concerns/xccdf/profiles_test.rb
skateman: As far as I understand, you search for a profile that has no value overrides and then assert if the number of value overrides for that profile is zero. Sorry, but I fail to understand how does that make any sense.

test/services/concerns/xccdf/rules_test.rb
skateman: Okay, so this is just a simple assertion that the stuff is imported as it is in the fixture. You also need some tests for overriding existing stuff (see other files for examples).

test/services/concerns/xccdf/profiles_test.rb
marleystipich2: Right, it didn't make sense how I had it before. I updated it so that it's finding a specific Profile from the test file that has no `refine-value` elements:

`profile2 = Profile.find_by(ref_id: 'xccdf_org.ssgproject.content_profile_standard')`

And then verifying that `profile2.value_overrides = {} `

app/services/concerns/xccdf/value_definitions.rb
```diff
@@ -0,0 +1,49 @@
...
+        @new_value_definitions ||= value_definitions.select(&:new_record?)
+      end
+
+      def old_value_definitions
+        @old_value_definitions ||= ::ValueDefinition.where(
+          ref_id: @op_value_definitions.map(&:id), benchmark_id: @benchmark&.id
+        ).index_by(&:ref_id)
+      end
+
+      def value_definition_id_for(ref_id:)
```
skateman, line 43: You're breaking the pattern, this should be `value_definition_for` and should return the `ValueDefinition`

app/services/concerns/xccdf/profiles.rb
```diff
@@ -30,6 +31,14 @@ def old_profiles
           ref_id: @op_profiles.map(&:id), benchmark: @benchmark&.id
         ).index_by(&:ref_id)
       end
+
+      def value_overrides(op_profile)
+        @cached_value_overrides ||= @op_value_definitions.index_by(&:id)
```
skateman, line 36: You already have this cached if you look at `value_definition_for` (omitted the ID as I explained below). No need to cache it twice...

app/services/concerns/xccdf/profiles.rb
```diff
@@ -30,6 +31,14 @@ def old_profiles
...
+        op_profile.refined_values.each_with_object({}) do |(value_id, selector), value_map|
+          value_definition = @cached_value_overrides[value_id]
+          value_map[value_id] = value_definition.value(selector)
```
skateman, line 39:
```suggestion
          value_map[value_id] = value_definition_for(ref_id: value_id).value(selector)
```

skateman (changes requested):
Please use exisitng patterns that already work in other modules.

app/services/concerns/xccdf/profiles.rb
marleystipich2: I don't think this will work because `value_definition_for` returns a `ValueDefinition` but we need to call `.value(selector)` on the `OpenscapParser::Value`

app/services/concerns/xccdf/profiles.rb
marleystipich2: This cache is different though because we're indexing the `OpenscapParser::Value`s instead of `ValueDefinition`s

skateman: Isn't it the same after importing?

skateman: The key is to import the definitions first :wink:

app/services/concerns/xccdf/profiles.rb
skateman: You can look into how is `op_source` used in rules :wink:

app/services/concerns/xccdf/value_definitions.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      def save_value_definitions
+        # Import the new records first with validation
+        ::ValueDefinition.import!(new_value_definitions, ignore: true)
+
+        # Update the fields on existing value_definitions, validation is not necessary
+        ::ValueDefinition.import(old_value_definitions.values,
+                                 on_duplicate_key_update: {
+                                   conflict_target: %i[ref_id benchmark_id],
+                                   columns: %i[description default_value]
```
skateman, line 27: What about `title`? Is that guaranteed that won't change?

skateman (commented):
Looks good, just one minor thing.

app/services/concerns/xccdf/value_definitions.rb
marleystipich2: We didn't have `title` for rules or rule_groups so I assumed that it would be the same for value_definitions but I can add it just to be safe?

skateman: I think then we're fine, I'll approve :+1:

skateman (approved):
The Seal of Approval

marleystipich2:
@skateman Just remembered I need to reset the migrations so I'll add that first and then merge

PR #1562, opened 2022-12-14
fix(Hosts): RHICOMPL-3437 don't sort by SSG without policy scoping

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -7,6 +7,10 @@ class Host < ApplicationRecord
   OS_MAJOR_VERSION = Arel.sql("system_profile->'operating_system'->'major'")
   OS_VERSION = Arel.sql("system_profile->'operating_system'")
   TAGS = Arel.sql('jsonb_array_elements(tags)')
+  JOIN_NO_BENCHMARK = arel_table.join(
+    Xccdf::Benchmark.arel_table,
+    Arel::Nodes::OuterJoin
+  ).on(Arel::Nodes::False.new).join_sources
 
   sortable_by :name, :display_name
   sortable_by :os_major_version, OS_MAJOR_VERSION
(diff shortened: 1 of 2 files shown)


app/models/host.rb
```diff
@@ -7,6 +7,11 @@ class Host < ApplicationRecord
   OS_MAJOR_VERSION = Arel.sql("system_profile->'operating_system'->'major'")
   OS_VERSION = Arel.sql("system_profile->'operating_system'")
   TAGS = Arel.sql('jsonb_array_elements(tags)')
+  # NOTE: joining benchmark on falsey condition for query needs
```
skateman, line 10: I would move this note into the `with_benchmark` scope ... see below.

app/models/host.rb
```diff
@@ -38,6 +43,8 @@ class Host < ApplicationRecord
   scope :with_benchmark, lambda { |profile = nil|
     profile ||= RequestStore.store['scoped_search_context_profiles']
 
+    return joins(JOIN_NO_BENCHMARK) if profile.nil?
```
skateman, line 46: Please add a comment above, no need to prefix it with a NOTE. Something like:
```
# Join with nonexisting benchmarks if the hosts aren't scoped for a policy
```

skateman (changes requested):
The `hosts` in the commit message is redundant, if you drop it you can make the message not wrap. Also the `on a single policy` does not make sense, more like `without policy scoping`.

skateman (approved):
The Seal of Approval

PR #1568, opened 2022-12-16
chore(Profile): RHICOMPL-3503 copy value_overrides to non-canonical

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/services/copy_value_overrides_to_non_canonical.rb
+++ b/app/services/copy_value_overrides_to_non_canonical.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Service for copying value overrides from parent profiles to non-canonical profiles
+class CopyValueOverridesToNonCanonical
+  class << self
+    def run!
+      query = 'UPDATE "profiles"
+               SET "value_overrides" = "parent_profiles"."value_overrides"
+               FROM "profiles" AS "parent_profiles"
+               WHERE "profiles"."parent_profile_id" IS NOT NULL
+               AND "profiles"."parent_profile_id" = "parent_profiles"."id"'
+      ActiveRecord::Base.connection.execute(query)
+    end
+  end
+end
--- a/db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
+++ b/db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
@@ -1,0 +1,5 @@
+class ValueOverridesToNonCanonicalProfiles < ActiveRecord::Migration[7.0]
+  def up
+    CopyValueOverridesToNonCanonical.run!
+  end
+end
--- a/test/services/copy_value_overrides_to_non_canonical_test.rb
+++ b/test/services/copy_value_overrides_to_non_canonical_test.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class CopyValueOverridesToNonCanonicalTest < ActiveSupport::TestCase
+  setup do
+    account = FactoryBot.create(:account)
+    value1 = FactoryBot.create(:value_definition)
+    value2 = FactoryBot.create(:value_definition)
+
+    @parent1 = FactoryBot.create(:canonical_profile)
+    @parent1.update!(value_overrides: { value1.ref_id => 'test', value2.ref_id => 5 })
+    FactoryBot.create(:profile, parent_profile: @parent1, account: account)
+    FactoryBot.create(:profile, parent_profile: @parent1, account: account)
+
+    @parent2 = FactoryBot.create(:canonical_profile)
+    @parent2.update!(value_overrides: { value1.id => 'test2' })
+    FactoryBot.create(:profile, parent_profile: @parent2, account: account)
+    FactoryBot.create(:profile, parent_profile: @parent2, account: account)
+
+    @parent3 = FactoryBot.create(:canonical_profile)
+    @parent3.update!(value_overrides: {})
+    FactoryBot.create(:profile, parent_profile: @parent3, account: account)
+    FactoryBot.create(:profile, parent_profile: @parent3, account: account)
+  end
+
+  test 'correctly copies value overrides from parent profiles to non-canonical profiles' do
+    Profile.canonical(false).each do |p|
+      assert_equal p.value_overrides, {}
+    end
+
+    CopyValueOverridesToNonCanonical.run!
+
+    Profile.canonical(false).each do |p|
+      assert_equal p.parent_profile.value_overrides, p.value_overrides
+    end
+  end
+end


db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
```diff
@@ -0,0 +1,13 @@
+class ValueOverridesToNonCanonicalProfiles < ActiveRecord::Migration[7.0]
+  def up
+    Profile.canonical.find_each do |canonical_profile|
+      canonical_profile.child_profiles.update_all(
+        value_overrides: canonical_profile.value_overrides
+      )
+    end
```
skateman, line 7: You are iterating through canonical profiles and then querying their child profiles one-by-one. Try to catch this from the other end, request all non-canonical profiles, join them with their parent profile and use [`update_all`](https://stackoverflow.com/questions/5224881/update-one-column-to-value-of-another-in-rails-migration) to ask PG to do the heavy lifting for you. 

Also if you're using DB models in your migrations, you have to make sure those models are defined in the migration context itself. Right now it would work, but what if in 2 years we change the `Profile` model and someone wants to set up the DB by running migrations on an empty DB (or ephemeral).

db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def down
+    Profile.canonical(false).update_all(value_overrides: {})
+  end
```
skateman, line 12: I don't think this is really needed, but nothing goes wrong if you keep it I guess.

skateman (changes requested):

db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
marleystipich2: Yeah, I figured it wouldn't hurt

db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
```diff
@@ -0,0 +1,13 @@
+class ValueOverridesToNonCanonicalProfiles < ActiveRecord::Migration[7.0]
+  class Profile < ApplicationRecord; end
+
+  def up
+    sub_query = Profile.where('p.parent_profile_id = id').select(:value_overrides).to_sql
+    query = "UPDATE profiles AS p SET value_overrides = (#{sub_query}) WHERE parent_profile_id IS NOT NULL"
+    ActiveRecord::Base.connection.execute(query)
```
marleystipich2, line 7: This looks a little weird but I had to do it like this because `joins` and `update_all` don't work correctly together: https://github.com/rails/rails/issues/522

[This comment](https://github.com/rails/rails/issues/522#issuecomment-426037101) explains the issue and they came up with this workaround that I used

skateman: You should not write the whole query as a string, only put the string into `update_all` part of the original one.

marleystipich2: I've tried this: 
`Profile.canonical(false).joins(:parent_profile).update_all('value_overrides = parent_profiles_profiles.value_overrides')`

But it returns this error:
```
ERROR:  missing FROM-clause entry for table "parent_profiles_profiles"
LINE 1: UPDATE "profiles" SET value_overrides = parent_profiles_prof...
```

The error made sense after looking at the SQL query that's generated from this:
`UPDATE "profiles" 
SET value_overrides = parent_profiles_profiles.value_overrides
WHERE "profiles"."id" IN (SELECT "profiles"."id" FROM "profiles" INNER JOIN "profiles" "parent_profiles_profiles" ON "parent_profiles_profiles"."id" = "profiles"."parent_profile_id" WHERE "profiles"."parent_profile_id" IS NOT NULL)`

`parent_profiles_profiles` is in a nested query and can't be accessed from where the query is trying to use it. This discussion talks about the issue: https://github.com/rails/rails/issues/522

skateman: Try using an explicit join: `joins("parent_profile pp ON pp.id = profiles.parent_profile_id")` and then refer to it as `pp` in the update.

marleystipich2: I tried using an explicit join but had to update it slightly because `parent_profile` table doesn't exist so I used `profiles`: 
`Profile.canonical(false).joins("join profiles pp ON pp.id = profiles.parent_profile_id").update_all('value_overrides = pp.value_overrides')`

But got the same error 🙁 
```
Caused by PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "pp"
LINE 1: UPDATE "profiles" SET value_overrides = pp.value_overrides W...
```

The join is inside of a nested query again so that's why it's not working:
```
UPDATE "profiles" 
SET value_overrides = pp.value_overrides 
WHERE "profiles"."id" IN (SELECT "profiles"."id" FROM "profiles" join 
parent_profiles pp ON pp.id = profiles.parent_profile_id WHERE
 "profiles"."parent_profile_id" IS NOT NULL)
```

skateman: You'd need more explicit:
```suggestion
    update_all('profiles.value_overrides = pp.value_overrides')
```
But it might not work still...

skateman: Anyway, there's always arel ...

marleystipich2: `Profile.canonical(false).joins("join parent_profiles pp ON pp.id = profiles.parent_profile_id").update_all('profiles.value_overrides = pp.value_overrides')` still doesn't work 🙁 

I'll try to use arel instead!

db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
```diff
@@ -0,0 +1,16 @@
...
+    profiles = Arel::Table.new(:profiles)
+    parent_profiles = Arel::Table.new(:profiles, :as => "parent_profiles")
+
+    update_manager = Arel::UpdateManager.new(ActiveRecord::Base)
+    update_manager.table(profiles)
+    update_manager.set([[profiles[:value_overrides], parent_profiles[:value_overrides]]])
+
+    sub_query = profiles[:parent_profile_id].not_eq(nil).and(profiles[:parent_profile_id].eq(parent_profiles[:id]))
+    query = "#{update_manager.to_sql} FROM profiles AS parent_profiles WHERE #{sub_query.to_sql}"
+    ActiveRecord::Base.connection.execute(query)
```
marleystipich2, line 14: This is the PostgreSQL update/join syntax:
```
UPDATE t1
SET t1.c1 = new_value
FROM t2
WHERE t1.c2 = t2.c2;
```

and this is what this is the query generated by what I have here:
```
UPDATE "profiles" 
SET "value_overrides" = "parent_profiles"."value_overrides"
FROM profiles AS parent_profiles
WHERE "profiles"."parent_profile_id" IS NOT NULL AND "profiles"."parent_profile_id" = "parent_profiles"."id"
```

skateman:
I looked into this from all directions and it seems like the best course of action is to just run the query as raw SQL :disappointed: the complexity of making this string is way too high to have it generated by something that does not even support it. Thank you for your analysis @marleystipich2 and sorry for pulling it this far.

So please fall back to `ActiveRecord::Base.connection.execute`

marleystipich2:
> I looked into this from all directions and it seems like the best course of action is to just run the query as raw SQL 😞 the complexity of making this string is way too high to have it generated by something that does not even support it. Thank you for your analysis @marleystipich2 and sorry for pulling it this far.
> 
> So please fall back to `ActiveRecord::Base.connection.execute`

It's okay! I think that's the best option too. It seems like there should be a better way to do UPDATE with join but I guess not 🙁 I'll update it to raw SQL.

db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
```diff
@@ -0,0 +1,12 @@
+class ValueOverridesToNonCanonicalProfiles < ActiveRecord::Migration[7.0]
+  class Profile < ApplicationRecord; end
+
+  def up
+    query = 'UPDATE profiles AS p
+             SET value_overrides = (SELECT value_overrides
+                                    FROM profiles
+                                    WHERE (p.parent_profile_id = id))
+             WHERE parent_profile_id IS NOT NULL'
```
skateman, line 9: Isn't this a little bit too complicated? Also does it work as it should? What I found feasible (and shorter) is:
```
UPDATE "profiles" 
SET "value_overrides" = "parent_profiles"."value_overrides"
FROM profiles AS parent_profiles
WHERE "profiles"."parent_profile_id" IS NOT NULL AND "profiles"."parent_profile_id" = "parent_profiles"."id"
```

test/services/copy_value_overrides_to_non_canonical_test.rb
```diff
@@ -0,0 +1,39 @@
...
+    assert_equal 6, Profile.canonical(false).select { |p| p.value_overrides == {} }.count
+
+    CopyValueOverridesToNonCanonical.run!
+
+    assert_equal @parent1.value_overrides, @profile1.reload.value_overrides
+    assert_equal @parent1.value_overrides, @profile2.reload.value_overrides
+    assert_equal @parent2.value_overrides, @profile3.reload.value_overrides
+    assert_equal @parent2.value_overrides, @profile4.reload.value_overrides
+    assert_equal @parent3.value_overrides, @profile5.reload.value_overrides
+    assert_equal @parent3.value_overrides, @profile6.reload.value_overrides
```
skateman, line 37:
```suggestion
    Profile.canonical(false).each do |p|
      assert_equal p.parent_profile.value_overrides, p.value_overrides
    end
```

test/services/copy_value_overrides_to_non_canonical_test.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    @parent2 = FactoryBot.create(:canonical_profile)
+    @parent2.update!(value_overrides: { value1.id => 'test2' })
+    @profile3 = FactoryBot.create(:profile, parent_profile: @parent2, account: account)
+    @profile4 = FactoryBot.create(:profile, parent_profile: @parent2, account: account)
+
+    @parent3 = FactoryBot.create(:canonical_profile)
+    @parent3.update!(value_overrides: {})
+    @profile5 = FactoryBot.create(:profile, parent_profile: @parent3, account: account)
+    @profile6 = FactoryBot.create(:profile, parent_profile: @parent3, account: account)
```
skateman, line 24: No need for naming the child profiles I guess.

test/services/copy_value_overrides_to_non_canonical_test.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  test 'correctly copies value overrides from parent profiles to non-canonical profiles' do
+    assert_equal 6, Profile.canonical(false).select { |p| p.value_overrides == {} }.count
```
skateman, line 28:
```suggestion
    Profile.canonical(false).each do |p|
      assert_equal p.value_overrides, {}
    end    
```

db/migrate/20221216171738_value_overrides_to_non_canonical_profiles.rb
```diff
@@ -0,0 +1,7 @@
+class ValueOverridesToNonCanonicalProfiles < ActiveRecord::Migration[7.0]
+  class Profile < ApplicationRecord; end
```
skateman, line 2: I guess this is no longer needed.

app/services/copy_value_overrides_to_non_canonical.rb
```diff
@@ -0,0 +1,15 @@
...
+
+# Service for copying value overrides from parent profiles to non-canonical profiles
+class CopyValueOverridesToNonCanonical
+  class << self
+    def run!
+      query = 'UPDATE "profiles"
+             SET "value_overrides" = "parent_profiles"."value_overrides"
+             FROM profiles AS parent_profiles
+             WHERE "profiles"."parent_profile_id" IS NOT NULL
+             AND "profiles"."parent_profile_id" = "parent_profiles"."id"'
```
skateman, line 11: Please align the string or at least use HEREDOC

skateman (changes requested):

app/services/copy_value_overrides_to_non_canonical.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Service for copying value overrides from parent profiles to non-canonical profiles
+class CopyValueOverridesToNonCanonical
+  class << self
+    def run!
+      query = 'UPDATE "profiles"
+               SET "value_overrides" = "parent_profiles"."value_overrides"
+               FROM profiles AS parent_profiles
```
skateman, line 9: Nit: missing quotations here...for consistency

romanblanco (approved):
LGTM :+1:

skateman:
@romanblanco @marleystipich2 we will have to revert this and move it to in progress as the migration will not work without https://github.com/RedHatInsights/compliance-backend/pull/1577 being already in the release. We have no value definitions in production without https://github.com/RedHatInsights/compliance-backend/pull/1577 and migrations run before SSG import :disappointed:

PR #1569, opened 2022-12-20
feat(Values): RHICOMPL-3396 expose values and checks in APIs

skateman:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -44,6 +44,9 @@ class Profile < ApplicationRecord
   validates :benchmark_id, presence: true
   validates :account, presence: true, if: -> { parent_profile_id && hosts.any? }
   validates :policy, presence: true, if: -> { policy_id }
+  validate :value_uuids, if: -> { !canonical? }
+
+  alias_attribute :values, :value_overrides
 
   scope :canonical_for_os, lambda { |os_major_version, os_minor_version|
     benchmarks = Xccdf::Benchmark.latest_for_os(
@@ -69,5 +72,23 @@ class Profile < ApplicationRecord
 
       profile
     end
+
+    # Transforming value keys from ref_id to uuid
+    def prepare_values(values)
+      return nil unless values
+
+      value_definitions = ValueDefinition.where(ref_id: values.keys).index_by(&:ref_id)
+
+      raise ActiveRecord::RecordNotFound unless value_definitions.count == values.count
+
+      values.transform_keys { |key| value_definitions[key].id }
+    end
+  end
+
+  def value_uuids
+    keys = value_overrides.keys
+    return if ValueDefinition.where(id: keys).select(:id).count == keys.count
+
+    errors.add(:value_overrides, 'invalid value(s)')
   end
 end
--- a/spec/api/v1/schemas/profiles.rb
+++ b/spec/api/v1/schemas/profiles.rb
@@ -57,6 +57,13 @@ module Api
               type: 'string',
               example: '0.1.49'
             },
+            values: {
+              type: 'object',
+              example: {
+                'uuid1': 'value1',
+                'uuid2': 'value2'
+              }
+            },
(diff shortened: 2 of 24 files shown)


spec/api/v1/schemas/profiles.rb
```diff
@@ -57,6 +57,13 @@ module Profiles
               type: 'string',
               example: '0.1.49'
             },
+            values: {
+              type: 'object',
+              example: {
+                'uuid1': 'value1',
+                'uuid2': 'value2'
+              }
+            },
```
marleystipich2, line 66: We're storing `profile.value_overrides` as {value_ref_id: value} currently, not {uuid, value}
https://docs.google.com/document/d/1u9Yn4aYD3i0xrFNXbviJHRkXheyX5O0vCdYaDORkKT4/edit?usp=sharing

marleystipich2: We talked about it for `Rule#value_checks` where we decided to store that as `[uuid, uuid...]` and I mentioned that we used ref_ids for `Profile#value_overrides` and I think you said that we might need to change that in the future but we didn't decide on it yet

skateman: Yes, we need it to be consistent

marleystipich2: Yeah, that makes sense. I can create a ticket to update it

skateman: No need to create a ticket, I already pulled back the original one from testing to in progress. Just make a follow-up with the same JIRA number.

marleystipich2: Okay, thanks!

romanblanco (approved):

marleystipich2:
This looks good to me now that we have the `value_overrides` column set up correctly! Thanks for doing this! 🙂

skateman:
This is not yet ready, I need to figure out some stuff around the APIs.

app/models/profile.rb
```diff
@@ -69,5 +72,23 @@ def from_openscap_parser(op_profile, existing: nil, benchmark_id: nil, account_i
...
+    def prepare_values(values)
+      return nil unless values
+
+      value_definitions = ValueDefinition.where(ref_id: values.keys).index_by(&:ref_id)
+
+      raise ActiveRecord::RecordNotFound unless value_definitions.count == values.count
+
+      values.transform_keys { |key| value_definitions[key].id }
+    end
+  end
```
marleystipich2, line 86: @skateman I'm confused why we need this because we're already storing the `value_overrides` using value uuids like `{uuid, value}` 🤔

skateman: The frontend is working with `ref_id` instead of `uuid`, eventually I want both APIs to work with both (like tailoring).

marleystipich2: Ah okay, that makes sense now

marleystipich2 (approved):
This looks good to me! Thanks again for doing this 🙂

spec/api/v1/schemas/profiles.rb
```diff
@@ -57,6 +57,13 @@ module Profiles
               type: 'string',
               example: '0.1.49'
             },
+            values: {
+              type: 'object',
```
romanblanco, line 61: @skateman could this be a symbol, just like in `schemas/rules.rb`?

skateman: I think we should do a unification across in a separate step because all is tangled between strings/symbols.

PR #1586, opened 2023-02-02
chore(Dockerfile): RHICOMPL-3599 rollback base image to ubi8-minimal

marleystipich2:
All apps are to be on ubi8-minimal. Being on ubi8-minimal will make this transition easier. 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,17 @@ ARG deps="findutils hostname jq libpq openssl procps-ng ruby shared-mime-info tz
 ARG devDeps="gcc gcc-c++ gzip libffi-devel make openssl-devel patch postgresql postgresql-devel redhat-rpm-config ruby-devel tar util-linux xz"
 ARG extras=""
 ARG prod="true"
+ARG pgRepo="http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-stream-repos-8-4.el8.noarch.rpm"
+ARG pgRepoKey="http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-8-4.el8.noarch.rpm"
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS build
+FROM registry.access.redhat.com/ubi8/ubi-minimal AS build
 
 ARG deps
 ARG devDeps
 ARG extras
 ARG prod
+ARG pgRepo
+ARG pgRepoKey
 
 USER 0
 
@@ -17,7 +21,14 @@ WORKDIR /opt/app-root/src
 COPY ./.gemrc.prod /etc/gemrc
 COPY ./Gemfile.lock ./Gemfile /opt/app-root/src/
 
-RUN rpm -e --nodeps tzdata &>/dev/null                                          && \
+RUN FULL_RHEL=$(microdnf repolist --enabled | grep rhel-8);                        \
+    if [ -z "$FULL_RHEL" ] ; then                                                  \
+      rpm -Uvh $pgRepo $pgRepoKey                                               && \
+      sed -i 's/^\(enabled.*\)/\1\npriority=200/;' /etc/yum.repos.d/CentOS*.repo;  \
+    fi;                                                                            \
+    rpm -e --nodeps tzdata &>/dev/null                                          && \
+    microdnf module enable ruby:3.0                                             && \
+    microdnf module enable postgresql:13                                        && \
     microdnf install --nodocs -y $deps $devDeps $extras                         && \
     chmod +t /tmp                                                               && \
     gem update --system --install-dir=/usr/share/gems --bindir /usr/bin         && \
@@ -34,7 +45,7 @@ ENV prometheus_multiproc_dir=/opt/app-root/src/tmp
 
 #############################################################
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi-minimal
 
 ARG deps
 ARG devDeps
@@ -44,6 +55,7 @@ WORKDIR /opt/app-root/src
 USER 0
 
 RUN rpm -e --nodeps tzdata &>/dev/null                                  && \
+    microdnf module enable ruby:3.0                                     && \
     microdnf install --nodocs -y $deps                                  && \
     chmod +t /tmp                                                       && \
     gem update --system --install-dir=/usr/share/gems --bindir /usr/bin && \


Dockerfile
```diff
@@ -1,9 +1,9 @@
 ARG deps="findutils hostname jq libpq openssl procps-ng ruby shared-mime-info tzdata"
-ARG devDeps="gcc gcc-c++ gzip libffi-devel make openssl-devel patch postgresql postgresql-devel redhat-rpm-config ruby-devel tar util-linux xz"
+ARG devDeps="gcc gcc-c++ gzip libffi-devel make openssl-devel patch postgresql-devel redhat-rpm-config ruby-devel tar util-linux xz"
```
marleystipich2, line 2: The build wasn't working with postgresql but it works without it. Not sure if this is okay to remove or not though?

skateman:
Please create a separate dockerfile for fedramp purposes and keep the ubi9 for the main build. I'm sure that the build target on the fedramp side can specify a different dockerfile.

marleystipich2:
> Please create a separate dockerfile for fedramp purposes and keep the ubi9 for the main build. I'm sure that the build target on the fedramp side can specify a different dockerfile.

Okay, yeah I can do that. I think that should be fine.

Dockerfile
```diff
@@ -1,9 +1,9 @@
...
 ARG extras=""
 ARG prod="true"
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS build
+FROM registry.access.redhat.com/ubi8/ubi-minimal AS build
```
dkuc, line 6: This can remain UBI 9.  Only thing that needs to be ubi8 is the final result that gets uploaded to quay.

scripts/unit_test.sh
```diff
@@ -83,7 +83,7 @@ echo '===================================='
 echo '=== Installing Gem Dependencies ===='
 echo '===================================='
 set +e
-podman exec -u 0 "$TEST_CONTAINER_ID" /bin/bash -c 'microdnf install -y $DEV_DEPS'
```
romanblanco, line 86: @marleystipich2 is there a reason to drop the `-y` argument?

marleystipich2: Nope, not necessary actually.. Trying to get the PR check to pass 🙈

marleystipich2:
@romanblanco I think the build image might also need to be downgraded to ubi8-minimal because of the gcc differences between ubi8-minimal and ubi9-minimal.. The PR check passed when I updated both images

skateman:
Please make sure that the developer setup didn't break by changing the first stage. Also there might be a more sophisticated solution to this by keeping the ubi9 in the build stage. I bet there is an environment variable that if we set, the GCC things will be compatible with older versions.

Dockerfile
romanblanco: @marleystipich2 the `postgresql` package should be installed for development purposes. Looks like it's not in UBI8 repository though.

Looking at https://github.com/romanblanco/vulnerability-engine/blob/4b9e3e815ab5d8a953c7b384c1e69cfb2e2e20e3/Dockerfile, this could help getting the package:


```Dockerfile
# install postgresql from centos if not building on RHSM system
RUN FULL_RHEL=$(microdnf repolist --enabled | grep rhel-8) ; \
    if [ -z "$FULL_RHEL" ] ; then \
    rpm -Uvh http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-stream-repos-8-4.el8.noarch.rpm \
    http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-8-4.el8.noarch.rpm && \
    sed -i 's/^\(enabled.*\)/\1\npriority=200/;' /etc/yum.repos.d/CentOS*.repo ; \
    fi

RUN microdnf module enable postgresql:12 && \
    microdnf install --setopt=install_weak_deps=0 --setopt=tsflags=nodocs \
    postgresql && \
    microdnf clean all
```

marleystipich2: This did help! I added this to the Dockerfile and think this fixed the issue. Besides the weird `org_id` problem, the developer setup worked for me.

marleystipich2:
Putting this in draft while I try to figure out a way to get the PR check to pass using ubi9-minimal for the build stage.

Dockerfile
skateman: Postgresql is needed for the developer setup, it might worked for you if you already had the database working. But in general it will not allow us to use `bin/rails db`

skateman:
This is getting a little bit out of hand, why are we reverting code changes that steer us towards SQL unsafety?

romanblanco:
> This is getting a little bit out of hand, why are we reverting code changes that steer us towards SQL unsafety?

@skateman I've identified the changes in e3df2fd6 to cause issues when setting up development environment.

When trying to seed data using `podman-compose exec rails bundle exec rake dev:db:seed`, we're running into this error:

```
rake aborted!
ActiveRecord::StatementInvalid: PG::SyntaxError: ERROR:  syntax error at or near ":"
LINE 1: ... $42, $43, $44) AND (#<Arel::Nodes::InfixOperation:0x00007f1...
                                                             ^
```
Because this also occurs on `master` branch, I've proposed to revert the changes to move forward with rollback to ubi8 and create a new PR to address issues caused by e3df2fd6 aside.

skateman:
This is a bug that has to be fixed (not reverted) outside the scope of this PR

romanblanco:
@marleystipich2 let's remove the revert and address the bug in separate ticket (RHICOMPL-3672)

Dockerfile
marleystipich2: Right, this code here installed postgresql:
```
# install postgresql from centos if not building on RHSM system
RUN FULL_RHEL=$(microdnf repolist --enabled | grep rhel-8) ; \
    if [ -z "$FULL_RHEL" ] ; then \
    rpm -Uvh http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-stream-repos-8-4.el8.noarch.rpm \
    http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-8-4.el8.noarch.rpm && \
    sed -i 's/^\(enabled.*\)/\1\npriority=200/;' /etc/yum.repos.d/CentOS*.repo ; \
    fi

RUN microdnf module enable postgresql:13 && \
    microdnf install --setopt=install_weak_deps=0 --setopt=tsflags=nodocs \
    postgresql && \
    microdnf clean all
```

romanblanco:
> This is a bug that has to be fixed (not reverted) outside the scope of this PR

Addressed in https://github.com/RedHatInsights/compliance-backend/pull/1597

PR #1593, opened 2023-02-27
chore(Clowder): RHIENG-80 use TLS on http communications

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/config/initializers/0_clowder-config.rb
+++ b/config/initializers/0_clowder-config.rb
@@ -1,19 +1,28 @@
+# frozen_string_literal: true
+
+def build_endpoint_url(endpoint, config)
+  return URI::HTTPS.build(host: endpoint&.hostname, port: endpoint&.tlsPort).to_s if config.tlsCAPath
+
+  URI::HTTP.build(host: endpoint&.hostname, port: endpoint&.port).to_s
+end
 
 if ClowderCommonRuby::Config.clowder_enabled?
 
   config = ClowderCommonRuby::Config.load
 
+  ENV['SSL_CERT_FILE'] = config.tlsCAPath if config.tlsCAPath
+
   # compliance-ssg
   compliance_ssg_config = config.private_dependency_endpoints&.dig('compliance-ssg', 'service')
-  compliance_ssg_url = "http://#{compliance_ssg_config&.hostname}:#{compliance_ssg_config&.port}"
+  compliance_ssg_url = build_endpoint_url(compliance_ssg_config, config)
 
   # RBAC
   rbac_config = config.dependency_endpoints.dig('rbac', 'service')
-  rbac_url = "http://#{rbac_config&.hostname}:#{rbac_config&.port}"
+  rbac_url = build_endpoint_url(rbac_config, config)
 
   # Inventory
   host_inventory_config = config.dependency_endpoints&.dig('host-inventory', 'service')
-  host_inventory_url = "http://#{host_inventory_config&.hostname}:#{host_inventory_config&.port}"
+  host_inventory_url = build_endpoint_url(host_inventory_config, config)
 
   # Redis (in-memory db)
   redis_url = "#{config.dig('inMemoryDb', 'hostname')}:#{config.dig('inMemoryDb', 'port')}"


config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,28 @@
 
+def build_endpoint_url ep
+  config = ClowderCommonRuby::Config.load
+  protocol = config.tlsCAPath ? 'https' : 'http'
+  port = config.tlsCAPath ? ep&.tlsPort : ep&.port
+  return "#{protocol}://#{ep&.hostname}:#{port}"
```
skateman, line 6: Please use some standard solution like `URI` instead

config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,33 @@
 
+def build_endpoint_url ep
+  config = ClowderCommonRuby::Config.load
+  uri = if config.tls
+          # There is currently no tlsPort or anything similar in clowder-common-ruby
+          URI::HTTPS.build(host: ep&.hostname, port: ep&.tlsPort)
```
marleystipich2, line 6: [clowder-common-ruby](https://github.com/RedHatInsights/clowder-common-ruby) provides `.tls` which returns true or false so that we can check if TLS is being used or not but we need a field that provides the TLS port similar to how app-common-python provides `tlsPort`. This should be fine though as long as `config.tls` is false.

config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,33 @@
 
+def build_endpoint_url ep
+  config = ClowderCommonRuby::Config.load
```
skateman, line 3: You are loading the already loaded clowder config on each call of this method. Please pass the config as a parameter from the already loaded instance.

config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,33 @@
...
+  uri = if config.tls
```
skateman, line 4: You can just simply return from the condition, then you won't need the `else` and the variable allocation.

skateman (changes requested):

config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,27 @@
 
+def build_endpoint_url ep, config
+  # There is currently no tlsPort or anything similar in clowder-common-ruby
+  return URI::HTTPS.build(host: ep&.hostname, port: ep&.tlsPort).to_s if config.tls
+  return URI::HTTP.build(host: ep&.hostname, port: ep&.port).to_s
```
skateman, line 5: I don't think rubocop is working on this file :disappointed: can you please run it manually and at least fix the issues here? It will probably blow up on the rest of the code, but we can fix it in a followup, just please don't introduce more issues.

config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,29 @@
+# frozen_string_literal: true
+
+def build_endpoint_url(endpoint, config)
+  # There is currently no tlsPort or anything similar in clowder-common-ruby
```
skateman, line 4: Maybe it is a good time to implement it? :wink:

config/initializers/0_clowder-config.rb
marleystipich2: Yeah, I didn't notice rubocop wasn't checking this file 🙁 I fixed the issues for my changes though!

config/initializers/0_clowder-config.rb
marleystipich2: Yes, it is! Is this something we'd do or ask whoever maintains that repo 🤔 A lot of the other apps/services use `app-common-python` and that's being updated in a similar way

skateman:
This does not guarantee that rails will run with HTTPS, not sure if that it was a requirement.

config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,29 @@
...
+  return URI::HTTPS.build(host: endpoint&.hostname, port: endpoint&.tlsPort).to_s if config.tls
+
+  URI::HTTP.build(host: endpoint&.hostname, port: endpoint&.port).to_s
+end
 
 if ClowderCommonRuby::Config.clowder_enabled?
 
   config = ClowderCommonRuby::Config.load
 
+  ENV['SSL_CERT_FILE'] = '/cdapp/certs/service-ca.crt' if config.tls
```
skateman, line 14: This is also something that could be provided by the clowder ruby bindings.

marleystipich2: Do you mean the `'/cdapp/certs/service-ca.crt'` part? We could add a field called `tlsPath` or something to clowder-common-ruby so we can call `config.tlsPath`. That's how app-common-python works.

skateman:
@marleystipich2 if you look at the insights architecture, all incoming (external) requests go through 3scale which forwards them (internally) to rails. Right now this communication is unencrypted HTTP.

marleystipich2:
> @marleystipich2 if you look at the insights architecture, all incoming (external) requests go through 3scale which forwards them (internally) to rails. Right now this communication is unencrypted HTTP.

@skateman I see what you mean now. I think one way to configure rails to use HTTPS is by adding `config.force_ssl = true` in production.rb. Is it fine if this is always set to true or is it better if I find a way to only set it to true if TLS is used?

config/initializers/0_clowder-config.rb
marleystipich2: Petr opened [a PR](https://github.com/RedHatInsights/clowder-common-ruby/pull/29) in the clowder-common-ruby gem with the new fields that are needed 🙂  So when it's merged and the version is updated we'll have `tlsPort` and `tlsCAPath` like `app-common-python` already has

config/environments/production.rb
```diff
@@ -37,7 +37,7 @@
   # config.active_storage.service = :local
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
marleystipich2, line 40: Might want an if statement around this depending on if TLS is being used or not

skateman: This is a bad idea, because without an actual certificate (which you don't set up) this will fail. Also consider the situation when in non-fedramp environment the certificate is not available at all...

marleystipich2: For the certificate, I was originally setting that up in `0_clowder-config.rb` by assigning `SSL_CERT_FILE` to the cert file but apparently this breaks clowdwatch logging when tested in fedramp so I'm trying to figure out how to verify the certificate during runtime

I'll also find a different way to make sure rails uses HTTPS(when it needs to)

config/environments/production.rb
```diff
@@ -37,7 +37,12 @@
...
+  config.ssl_options = {
+    redirect: {
+      exclude: ->(request) { !ENV['TLS_CA_PATH'] }
+    }
+  }
```
marleystipich2, line 45: Something like this should make sure we're only forcing ssl when the `TLS_CA_PATH` is set(when TLS is being used)
Still need to set up the CA cert though.

config/environments/production.rb
skateman: There are 2 certificates, one is the CA for verifying other connections towards other services (rbac, ssg, etc) and one for being able to listen in an encrypted fashion in rails. Please make sure that you understand the difference between those two and know which is which.  If this is not clear in the clowder documentation, please reach out to Gábor or Pete.

config/environments/production.rb
skateman: Again, please make sure you know which cert is which.

config/environments/production.rb
```diff
@@ -38,6 +38,7 @@
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
+  config.ssl_options = { ca_file: ENV['TLS_CA_PATH'] }
```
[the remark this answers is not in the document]
skateman: This is for the listening part in rails and as you said this will be handled by `envoy`. We need the CA cert in all the other places we're calling the additional services like RBAC and inventory.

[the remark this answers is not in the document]
marleystipich2: That's what I thought 🤔 Is it possible to set up the CA cert for all the other places during runtime? It made the most sense to me to set up the `SSL_CERT_FILE` env variable in `0_clowder-config.rb` but apparently this causes issues with clowdwatch in fedramp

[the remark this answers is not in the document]
skateman: Well, we don't really have cloudwatch working :see_no_evil: for now. At least we never had the time to implement it properly.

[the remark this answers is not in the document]
marleystipich2: Hmm.. maybe it doesn't matter then 🤷‍♀️  Checking with Dan

config/initializers/0_clowder-config.rb
```diff
@@ -1,19 +1,28 @@
...
+  return URI::HTTPS.build(host: endpoint&.hostname, port: endpoint&.tlsPort).to_s if config.tlsCAPath
+
+  URI::HTTP.build(host: endpoint&.hostname, port: endpoint&.port).to_s
+end
 
 if ClowderCommonRuby::Config.clowder_enabled?
 
   config = ClowderCommonRuby::Config.load
 
+  ENV['SSL_CERT_FILE'] = config.tlsCAPath if config.tlsCAPath
```
marleystipich2, line 13: Possibly causes issues with cloudwatch logging but might not be a problem for us because we don't really have cloudwatch fully implemented anyway.

romanblanco (approved):
LGTM :+1:

skateman (approved):
The Seal of Approval

PR #1623, opened 2023-04-17
fix(ci): RHCLOUD-25345 add no-cache to builds

Victoremepunto:
- Adds the `DISABLE_BUILD_CACHE` to `build-deploy-common` 
- Sets the `DISABLE_BUILD_CACHE` to true to workaround: https://github.com/containers/buildah/issues/4632
- Disable requirement to log to registries on local build mode

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -6,5 +6,7 @@ source 'scripts/deploy/build-deploy-common.sh'
 
 IMAGE_NAME="${IMAGE_NAME:-quay.io/cloudservices/compliance-backend}"
 ADDITIONAL_TAGS="latest"
+REQUIRED_REGISTRIES_LOCAL=''
+DISABLE_BUILD_CACHE="true"
 
 build_deploy_main || exit 1
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -7,6 +7,5 @@ source 'scripts/deploy/build-deploy-common.sh'
 IMAGE_NAME="${IMAGE_NAME:-quay.io/cloudservices/compliance-backend}"
 ADDITIONAL_TAGS="latest"
 REQUIRED_REGISTRIES_LOCAL=''
-DISABLE_BUILD_CACHE="true"
 
 build_deploy_main || exit 1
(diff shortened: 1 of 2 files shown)


build_deploy.sh
```diff
@@ -6,5 +6,7 @@ source 'scripts/deploy/build-deploy-common.sh'
 
 IMAGE_NAME="${IMAGE_NAME:-quay.io/cloudservices/compliance-backend}"
 ADDITIONAL_TAGS="latest"
+REQUIRED_REGISTRIES_LOCAL=''
+DISABLE_BUILD_CACHE='true'
```
kwilczynski, line 10: A small suggestion to be consistent with other variables/options.

```suggestion
DISABLE_BUILD_CACHE="true"
```

kwilczynski: @Victoremepunto, do we want to add a small comment over this variable that explains why we are disabling the build cache? For some people, it might be a surprise as build-cache is often cited as a good practice.

I also wonder if having a small helper called `runtime_is_docker()` (or something along these lines) to detect Docker vs Podman and re-enable support for build-cache should be the former.

Victoremepunto: @kwilczynski  I pushed a draft version based on your suggestion - a detection mechanism that prevents explicit cache disabling configuration, but rather automatic detection if the engine is `podman` and the version is under `4.5.0` - please , take a look and let me know to move forward (I can undraft if it seems okay)

scripts/deploy/build-deploy-common.sh
```diff
@@ -223,18 +231,23 @@ build_image() {
...
     local LABEL_PARAMETER=''
-
+    local DISABLE_CACHE_PARAMETER=''
 
     if is_pr_or_mr_build; then
         LABEL_PARAMETER=$(get_expiry_label_parameter)
     fi
 
+    if _disable_build_cache || ([[ "$(get_container_engine_cmd)" == "podman" ]] && _podman_version_under_4_5_0); then
+        DISABLE_CACHE_PARAMETER='--no-cache'
```
kwilczynski, line 241: A warning for the end user:

```suggestion
        echo 'Old version of Podman detected. Build cache disabled.' >&2
        DISABLE_CACHE_PARAMETER='--no-cache'
```

The "Old" could be replaced with "Unsupported". Not sure which working is more correct, though.

kwilczynski (approved):
@Victoremepunto, looks good!

Thank you for working on this!

marleystipich2 (approved):
LGTM! Thanks, @Victoremepunto!

PR #1665, opened 2023-05-31
feat(Host): RHICOMPL-3732 sort systems by compliance score

marleystipich2:
Tests are passing now 🎉  
Here's the SQL generated when sorting by score: 
```
Host Load (9.7ms)  SELECT "inventory"."hosts".* FROM "inventory"."hosts" LEFT OUTER JOIN "test_results" ON 
"test_results"."host_id" = "inventory"."hosts"."id" LEFT OUTER JOIN "profiles" ON "profiles"."id" = "test_results"."profile_id" 
WHERE "inventory"."hosts"."org_id" = $1 ORDER BY COALESCE("test_results"."score") ASC, "inventory"."hosts"."id" ASC 
LIMIT $2 OFFSET $3
```

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -30,6 +30,10 @@ class Host < ApplicationRecord
   ).on(Arel::Nodes::False.new).join_sources
 
   sortable_by :name, :display_name
+  sortable_by :score, Arel::Nodes::NamedFunction.new(
+    'COALESCE',
+    [TestResult.arel_table[:score]]
+  ), scope: :joins_test_result_profiles
   sortable_by :os_major_version, OS_MAJOR_VERSION
   sortable_by :os_minor_version, OS_MINOR_VERSION
   sortable_by(
@@ -76,6 +80,10 @@ class Host < ApplicationRecord
     left_outer_joins(test_result_profiles: :benchmark).where(
       profiles: { id: profile.pluck(:id) }
     )
+  }
+
+  scope :joins_test_result_profiles, lambda {
+    left_outer_joins(:test_result_profiles)
   }
 
   def self.os_minor_versions(hosts)
--- a/test/graphql/queries/system_query_test.rb
+++ b/test/graphql/queries/system_query_test.rb
@@ -548,6 +548,78 @@ class SystemQueryTest < ActiveSupport::TestCase
       assert_equal @profile1.ssg_version,
                    returned_profiles.dig(0, 'ssgVersion')
     end
+
+    should 'properly sort by compliance score' do
+      host2 = FactoryBot.create(:host, org_id: @user.account.org_id)
+
+      # An outdated test result made in the past
+      FactoryBot.create(
+        :test_result,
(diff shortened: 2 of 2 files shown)


marleystipich2:
Not sure how to test this or if it's possible because sorting like this doesn't work because `score` is nested in `testResultProfiles`:

```
query getSystems($policyId: ID) {
          systems(limit: 50, offset: 1, sortBy: ["score"]) {
              edges {
                  node {
                      id
                      testResultProfiles(policyId: $policyId) {
                          id
                          compliant
                          score
                          supported
                          ssgVersion
                      }
                  }
              }
          }
      }
```

LightOfHeaven1994:
> Not sure how to test this or if it's possible.. Sorting like this doesn't work because `score` is nested in `testResultProfiles`:
> 
> ```
> query getSystems($policyId: ID) {
>           systems(limit: 50, offset: 1, sortBy: ["score"]) {
>               edges {
>                   node {
>                       id
>                       testResultProfiles(policyId: $policyId) {
>                           id
>                           compliant
>                           score
>                           supported
>                           ssgVersion
>                       }
>                   }
>               }
>           }
>       }
> ```

I'm just guessing here - maybe it's possible the same way as we sort systems by `ssgVersion`. As this attribute doesn't come from system itself but from profile (only when system reported so it is associated to some profile and we exactly know the version)

marleystipich2:
@LightOfHeaven1994 I thought the same thing but there isn't a test for sorting by `ssgVersion` so that's why I wasn't able to go off of that ): I can try sorting by SSG version in the UI and get the GQL request from there and see if that helps though

test/graphql/queries/system_query_test.rb
```diff
@@ -548,6 +548,70 @@ class SystemQueryTest < ActiveSupport::TestCase
       assert_equal @profile1.ssg_version,
                    returned_profiles.dig(0, 'ssgVersion')
     end
+
+    should 'properly sort by compliance score' do
```
marleystipich2, line 552: This test isn't working correctly right now even though the query should be correct. Still looking into why `sortBy` isn't working correctly when sorting anything nested under `testResultProfiles`.

app/models/host.rb
```diff
@@ -78,6 +82,10 @@ class Host < ApplicationRecord
     )
   }
 
+  scope :joins_test_results, lambda {
```
skateman, line 85: Semantically this is incorrect, should be:
```suggestion
  scope :joins_test_result_profiles, lambda {
```

skateman (approved):
The Seal of Approval

PR #1666, opened 2023-05-31
perf(GraphQL): RHICOMPL-3775 generalize the cache key for fragments

skateman:
This change adjusts the cache key generation to be independent from any other top-level parents than the actual object where the cached field is specified. This means that if in a query a given benchmark is wrapped in a profile, it would still resolve to the same cache key as requesting the benchmark directly. This generalization should speed up queries even more and also reduce the load on redis as each cached field should require just one entry per benchmark.

```
query Profile($id: String!) {
  profile(id: $id) {
    benchmark {
      ruleTree
    }
  }
}
```

```
query Benchmark($id: String!) {
  benchmark(id: $id) {
    ruleTree
  }
}
```
If the profile ID queried in the first query belongs to the benchmark that is queried in the second one, the cache key would resolve in both cases to the same thing, i.e. `graphql/rule_tree/#{SCHEMA_HASH}/benchmark-#{ID}` as before this change it would contain information about the top-level object in the query, this resolving into two separate cache keys.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/graphql/types/base_object.rb
+++ b/app/graphql/types/base_object.rb
@@ -34,6 +34,15 @@ module Types
       def enforce_rbac(*permissions)
         @rbac_permissions = permissions
       end
+
+      # This the field to be stored in the GraphQL fragment cache with a `title/` prefix
+      # Please note that only static and context-free fields, such as SSG content can be
+      # stored in this cache and that it gets cleared upon a successful SSG import.
+      def cached_static_field(title, type, **args)
+        field(title, type, **args.merge(
+          cache_fragment: { path_cache_key: title.to_s, cache_key: :object }
+        ))
+      end
     end
 
     protected
--- a/app/graphql/types/benchmark.rb
+++ b/app/graphql/types/benchmark.rb
@@ -14,10 +14,11 @@ module Types
     field :version, String, null: false
     field :os_major_version, String, null: false
     field :latest_supported_os_minor_versions, [String], null: false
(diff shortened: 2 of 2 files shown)


app/graphql/types/base_object.rb
```diff
@@ -34,6 +34,15 @@ def authorized?(object, context)
       def enforce_rbac(*permissions)
         @rbac_permissions = permissions
       end
+
+      # This the field to be stored in the GraphQL fragment cache with a `title/` prefix
+      # Please note that only static and context-free fields, such as SSG content can be
+      # stored in this cache and that it gets cleared upon a successful SSG import.
+      def cached_static_field(title, type, **args)
+        field(title, type, **args.merge(
+          cache_fragment: { path_cache_key: title.to_s, cache_key: :object }
```
vkrizan, line 43: Wouldn't the `cache_key` had a literal of `object`?

skateman: Nope, this is not in that context.

vkrizan: Okay, so this would use the object's `cache_key` implementation, in our case from ActiveRecord's models.

PR #1688, opened 2023-06-16
feat(Host): RHICOMPL-3731 sort systems by num of failed rules

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -53,6 +53,8 @@ class Host < ApplicationRecord
     scope: :with_benchmark
   )
 
+  sortable_by(:rules_failed, 'sq.rules_failed', scope: :with_failed_rules_count)
+
   self.table_name = 'inventory.hosts'
   self.primary_key = 'id'
 
@@ -84,6 +86,14 @@ class Host < ApplicationRecord
 
   scope :joins_test_result_profiles, lambda {
     left_outer_joins(:test_result_profiles)
+  }
+
+  scope :with_failed_rules_count, lambda {
+    sq = Host.left_outer_joins(test_result_profiles: :rule_results)
+             .where(rule_results: { result: RuleResult::FAILED })
+             .select(arel_table[:id].as('id'), RuleResult.arel_table[:result].count.as('rules_failed'))
+             .group('hosts.id')
+    joins("INNER JOIN (#{sq.to_sql}) sq ON sq.id = hosts.id")
   }
 
   def self.os_minor_versions(hosts)
--- a/test/graphql/queries/system_query_test.rb
+++ b/test/graphql/queries/system_query_test.rb
@@ -796,6 +796,79 @@ class SystemQueryTest < ActiveSupport::TestCase
       assert_equal 50, returned_profile3.dig(0, 'score')
       assert_equal 98, returned_profile4.dig(0, 'score')
     end
+
+    should 'properly sort by rules failed' do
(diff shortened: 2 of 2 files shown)


app/models/host.rb
```diff
@@ -86,6 +99,13 @@ class Host < ApplicationRecord
     left_outer_joins(:test_result_profiles)
   }
 
+  scope :joins_test_result_profiles_with_failed_rules, lambda {
+    # Needed to GROUP BY Host columns to work correctly with sortable_by :rules_failed
+    left_outer_joins(test_result_profiles: :rule_results).where(
+      rule_results: { result: RuleResult::FAILED }
+    ).group(HOST_COLUMNS)
```
marleystipich2, line 106: Grouping by the host columns was needed for `sortable_by :rules_failed` (like mentioned in the comment) but not sure if this is the best place the add it 🤔

skateman: A grouping can mess up the object in activerecord in ways that can not be handled by the serializer sometimes. What you could do is to "hide" it in a subquery. However, I would need a bit more context of why the grouping here is needed.

app/models/host.rb
```diff
@@ -53,6 +53,10 @@ class Host < ApplicationRecord
     scope: :with_benchmark
   )
 
+  sortable_by(
+    :rules_failed, 'sq.rules_failed', scope: :with_failed_rules_count
+  )
```
skateman, line 58: This is short enough to be a one-liner without the parentheses

app/models/host.rb
```diff
@@ -86,6 +90,14 @@ class Host < ApplicationRecord
     left_outer_joins(:test_result_profiles)
   }
 
+  scope :with_failed_rules_count, lambda {
+    sq = Host.left_outer_joins(test_result_profiles: :rule_results)
+             .where(rule_results: { result: RuleResult::FAILED })
+             .select(Host.arel_table[:id].as('id'), RuleResult.arel_table[:result].count.as('rules_failed'))
+             .group('hosts.id')
```
skateman, line 97: I think you can omit the `Host` as we are in its context. Maybe not the first one, but the second one for sure, so please try:
```suggestion
    sq = left_outer_joins(test_result_profiles: :rule_results)
             .where(rule_results: { result: RuleResult::FAILED })
             .select(arel_table[:id].as('id'), RuleResult.arel_table[:result].count.as('rules_failed'))
             .group('hosts.id')
```

app/models/host.rb
```diff
@@ -86,6 +90,14 @@ class Host < ApplicationRecord
...
+    joins("INNER JOIN (#{sq.to_sql}) sq ON sq.id = hosts.id")
```
skateman, line 98: So you don't need to select anything? NICE!

skateman (changes requested):
Looks good, just minor cosmetics please

app/models/host.rb
marleystipich2: Right, it works without it 👍

skateman (approved):
The Seal of Approval

PR #1698, opened 2023-06-23
test(DeleteHost): RHICOMPL-3844 rewrite DeleteHost test to rspec

RoamingNoMaD:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/delete_host_spec.rb
+++ b/spec/delete_host_spec.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+
+require 'rails_helper' # spec_helper?
+require 'sidekiq/testing'
+
+describe DeleteHost do
+  let(:user) { FactoryBot.create(:user) }
+  let(:host) { FactoryBot.create(:host, org_id: user.account.org_id) }
+  let(:profile) { FactoryBot.create(:profile, :with_rules, account: user.account) }
+  let(:message_id) { host.id }
+  let(:message) { { 'id' => message_id, 'type' => 'delete' } }
+  let!(:test_result) { FactoryBot.create(:test_result, profile: profile, host: host) }
+  let!(:rule_result) do
+    FactoryBot.create(:rule_result, rule: profile.rules.sample, test_result: test_result, host: host)
+  end
+
+  before do
+    logger = double(Rails.logger)
+    allow(logger).to receive(:info)
+    allow(logger).to receive(:debug)
+    allow(Sidekiq).to receive(:logger).and_return(logger)
+  end
+
+  after do
+    DeleteHost.clear
+  end
+
+  it 'performs and logs deletion' do
+    expect(profile.hosts.count).not_to eql(0)
+    expect(profile.score).not_to eql(0.0)
+
+    DeleteHost.perform_async(message)
+
+    expect(profile.test_results.count).to eql(1)
+    expect(profile.rule_results.count).to eql(1) # TODO: Deleteted
+
+    expect(Rails.logger).to receive(:audit_success).with("Deleteted related records for host #{message_id}")
+    DeleteHost.drain
+
+    expect(profile.test_results.count).to eql(0)
+    expect(profile.rule_results.count).to eql(0)
+    expect(profile.hosts.count).to eql(0)
+    expect(profile.reload.score).to eql(0.0)
+  end
+
+  context 'when multiple hosts exist under a profile' do
+    let(:message_id) { profile.hosts.first.id }
+
+    it 'deletes host and updates score and counters' do
+      test_result.destroy
+      rule_result.destroy
+
+      FactoryBot.create_list(:host, 3, org_id: user.account.org_id) do |host|
+        test_result = FactoryBot.create(:test_result, profile: profile, host: host)
+        FactoryBot.create(:rule_result, rule: profile.rules.sample, test_result: test_result, host: host)
+      end
+
+      expect(profile.score).not_to eql(0.0)
+      expect(profile.policy.test_result_host_count).to eql(3)
+
+      old_score = profile.score
+      DeleteHost.perform_async(message)
+      DeleteHost.drain
+
+      expect(profile.reload.score).not_to eql(old_score)
+      expect(profile.policy.test_result_host_count).to eql(2)
+    end
+  end
+
+  context 'when host to be deleted does not exist' do
+    let(:message_id) { 'invalid-id' }
+
+    it 'fails silently' do
+      expect(profile.hosts.count).to eql(1)
+      expect(profile.score).not_to eql(0.0)
+
+      DeleteHost.perform_async(message)
+      DeleteHost.drain
+
+      expect(profile.hosts.count).to eql(1)
+      expect(profile.reload.score).not_to eql(0.0)
+    end
+  end
+
+  context 'when deletion is not possible' do
+    let(:err_message) { "Failed to delete related records for host #{message_id}: StandardError" }
+
+    before do
+      DeleteHost.perform_async(message)
+      expect_any_instance_of(DeleteHost).to receive(:remove_related).and_raise(StandardError)
+    end
+
+    it 'logs failure and raises an error' do
+      expect(Rails.logger).to receive(:audit_fail).with(err_message)
+      expect { DeleteHost.drain }.to raise_error(StandardError)
+    end
+  end
+end
--- a/spec/delete_host_spec.rb
+++ b/spec/delete_host_spec.rb
@@ -32,9 +32,9 @@ describe DeleteHost do
     DeleteHost.perform_async(message)
 
     expect(profile.test_results.count).to eql(1)
-    expect(profile.rule_results.count).to eql(1) # TODO: Deleteted
+    expect(profile.rule_results.count).to eql(1)
 
-    expect(Rails.logger).to receive(:audit_success).with("Deleteted related records for host #{message_id}")
+    expect(Rails.logger).to receive(:audit_success).with("Deleted related records for host #{message_id}")
     DeleteHost.drain
 
     expect(profile.test_results.count).to eql(0)
--- a/app/jobs/delete_host.rb
+++ b/app/jobs/delete_host.rb
@@ -67,7 +67,7 @@ class DeleteHost
 
   def audit_success(host_id)
     Rails.logger.audit_success(
-      "Deleteted related records for host #{host_id}"
+      "Deleted related records for host #{host_id}"
     )
   end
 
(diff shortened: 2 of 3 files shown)


spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:user) { FactoryBot.create(:user) }
+  let(:host) { FactoryBot.create(:host, org_id: user.account.org_id) }
+  let(:profile) { FactoryBot.create(:profile, :with_rules, account: user.account) }
+  let(:message) { {'id' => host.id, 'type' => 'delete'} }
+
+  before do
+    logger = double(Rails.logger)
+    allow(logger).to receive(:info)
+    allow(logger).to receive(:debug)
+    allow(Sidekiq).to receive(:logger)
```
skateman, line 16:
```suggestion
    allow(Sidekiq).to receive(:logger).and_return(logger)
```

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  DeleteHost.clear
```
skateman, line 19: This should probably go to a `before` block

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  context 'when the deletion succeeds' do
```
skateman, line 21: This should be an implicit context, so drop it.

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+       .and change { rule_result }.by(-1)
+    end
+
+    it 'does not find a host anymore' do
+      expect(profile.reload.score).to equal(0)  
+      expect(profile.policy.reload.test_result_host_count).to equal(0)  
+    end    
+  end
+
+  context 'when the deletion fails' do
```
skateman, line 51:
```suggestion
  context 'deletion is not possible' do
```

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    before do
+      allow(DeleteHost).to receive(:remove_related).and_raise(StandardError)
```
skateman, line 53: I think this might have to be stricter:
```suggestion
      expect(DeleteHost).to receive(:remove_related).and_raise(StandardError)
```

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    allow(Sidekiq).to receive(:logger)
+  end
+
+  DeleteHost.clear
+
+  context 'when the deletion succeeds' do
+    let(:test_result) { FactoryBot.create(:test_result, profile: profile, host: host) }
+    let(:rule_result) { FactoryBot.create(:rule_result, rule: profile.rules.sample, test_result: test_result, host: host) } 
+
+    it 'finds a host to delete' do
```
skateman, line 25: This is great, but you might need a context when a host to be deleted doesn't exist.

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      DeleteHost.perform_async(message)
+      
+      expect(DeleteHost.jobs.size).to equal(1)  
+    end
+
+    it 'performs deletion' do
+      expect { 
+        DeleteHost.drain 
+      }.to change { test_result }.by(-1)
+       .and change { rule_result }.by(-1)
```
skateman, line 42: This is an interesting syntax, I would probably not go with it. Simplify:
```suggestion
      expect(test_result.count).to eq(n)
      expect(rule_result.count).to eq(n)
      DeleteHost.drain
      expect(test_result.count).to eq(n-1)
      expect(rule_result.count).to eq(n-1)
```

But IMO we can do much better tests and actually query the DB that ALL the test and rule results for the hosts are nonexistent afterwards.

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it 'does not find a host anymore' do
+      expect(profile.reload.score).to equal(0)  
+      expect(profile.policy.reload.test_result_host_count).to equal(0)  
+    end
```
skateman, line 48: This does not make sense, all tests should run isolated from each other. So this passes because the `let` blocks above are lazily created, i.e. they are only evaluated if there is a reference to them. You might want to consider using `let!` instead at some places (see Martin's PR for more examples).

skateman (changes requested):
Each `it` block should be independent from the others as they might run in random order.

spec/delete_host_spec.rb
RoamingNoMaD: When an invalid host is to be deleted, an exception is **not** being generated. Is that the behavior we want?

skateman: Yup, it should not fail with an error :)

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  it 'performs and logs deletion' do
+    expect(profile.hosts.count).not_to eql(0)
+    expect(profile.score).not_to eql(0.0)
+    DeleteHost.perform_async(message)
+
+    expect(profile.test_results.count).to eq(1)
+    expect(profile.rule_results.count).to eq(1) # TODO: Deleteted
+    expect(Rails.logger).to receive(:audit_success).with("Deleteted related records for host #{host.id}")
```
skateman, line 30: Deleteted? :rofl:

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      expect(profile.score).not_to eql(0.0)
+
+      DeleteHost.drain
+      expect(profile.hosts.count).to eql(1)
+      expect(profile.score).not_to eql(0.0)
+    end
+  end
+
+  context 'when deletion is not possible' do
+    let(:err_message) { "Failed to delete related records for host #{host.id}: StandardError" }
```
skateman, line 79: Let blocks are not regular variables, please move this into the expectation

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require 'rails_helper' # spec_helper?
+require 'sidekiq/testing'
+
+describe DeleteHost do
+  let(:user) { FactoryBot.create(:user) }
+  let(:host) { FactoryBot.create(:host, org_id: user.account.org_id) }
+  let(:profile) { FactoryBot.create(:profile, :with_rules, account: user.account) }
+  let(:message) { { 'id' => host.id, 'type' => 'delete' } }
```
skateman, line 10: I think you can extract the `id` part into a next `let` block under the name of `message_id` and then just set that in various contexts accordingly:
```
  let(:message) { { 'id' => message_id, 'type' => 'delete' } }
  let(:message_id) { 'something' }

  context 'something else' do
    let(:message_id) { 'something else' }
    ...
  end
```

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  it 'performs and logs deletion' do
+    expect(profile.hosts.count).not_to eql(0)
+    expect(profile.score).not_to eql(0.0)
+    DeleteHost.perform_async(message)
+
+    expect(profile.test_results.count).to eql(1)
+    expect(profile.rule_results.count).to eql(1)
+    expect(Rails.logger).to receive(:audit_success).with("Deleted related records for host #{host.id}")
+    DeleteHost.drain
```
skateman, line 33: Put a single empty line around these two, so they are logically separated from the expectations.

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    allow(logger).to receive(:debug)
+    allow(Sidekiq).to receive(:logger).and_return(logger)
+
+    DeleteHost.clear
+  end
+
+  it 'performs and logs deletion' do
+    expect(profile.hosts.count).not_to eql(0)
+    expect(profile.score).not_to eql(0.0)
+    DeleteHost.perform_async(message)
```
skateman, line 28: Single empty line above this, logical separation

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    expect(profile.test_results.count).to eql(1)
+    expect(profile.rule_results.count).to eql(1)
+    expect(Rails.logger).to receive(:audit_success).with("Deleted related records for host #{host.id}")
+    DeleteHost.drain
+    expect(profile.test_results.count).to eql(0)
+    expect(profile.rule_results.count).to eql(0)
+
```
skateman, line 36: No separation is needed here.

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+        test_result = FactoryBot.create(:test_result, profile: profile, host: host)
+        FactoryBot.create(:rule_result, rule: profile.rules.sample, test_result: test_result, host: host)
+      end
+
+      expect(profile.score).not_to eql(0.0)
+      old_score = profile.score
+      expect(profile.policy.test_result_host_count).to eql(3)
+      DeleteHost.perform_async(message)
+
+      DeleteHost.drain
```
skateman, line 58: Logical separation please

skateman (changes requested):
Looking good, just please try to logically separate expectations from operations:
```ruby
foo()

expect()
expect()

bar()

expect()
expect()

baz()

```

spec/delete_host_spec.rb
RoamingNoMaD: I had to extract the log message into a let due to RuboCop complaining about the length of the line. (over 120 chars)

spec/delete_host_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    FactoryBot.create(:rule_result, rule: profile.rules.sample, test_result: test_result, host: host)
+  end
+
+  before do
+    logger = double(Rails.logger)
+    allow(logger).to receive(:info)
+    allow(logger).to receive(:debug)
+    allow(Sidekiq).to receive(:logger).and_return(logger)
+
+    DeleteHost.clear
```
skateman, line 23: I think that this one should go into an `after` block instead if it is supposed to clean up stuff.

skateman (approved):

PR #1712, opened 2023-06-28
fix(Host): RHICOMPL-3731 sort all failed rules including nil

marleystipich2:
Still testing this but I think something like this will work.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -88,11 +88,17 @@ class Host < ApplicationRecord
     left_outer_joins(:test_result_profiles)
   }
 
-  scope :with_failed_rules_count, lambda {
-    sq = Host.left_outer_joins(test_result_profiles: :rule_results)
-             .where(rule_results: { result: RuleResult::FAILED })
+  scope :with_failed_rules_count, lambda { |profile = nil|
+    profile ||= RequestStore.store['scoped_search_context_profiles']
+    profile_ids = profile&.pluck(:id) || []
+
+    sq = Host.left_outer_joins(test_results: :rule_results)
+             .merge(TestResult.latest('LEFT OUTER JOIN'))
+             .where(test_results: { profile_id: profile_ids }, rule_results: { result: RuleResult::FAILED })
+             .or(Host.where(test_results: { profile_id: profile_ids }, rule_results: { id: nil }))
              .select(arel_table[:id].as('id'), RuleResult.arel_table[:result].count.as('rules_failed'))
              .group('hosts.id')
+
     joins("INNER JOIN (#{sq.to_sql}) sq ON sq.id = hosts.id")
   }
 
--- a/app/models/rule_result.rb
+++ b/app/models/rule_result.rb
@@ -34,7 +34,7 @@ class RuleResult < ApplicationRecord
   scope :failed, -> { where(result: FAILED) }
   scope :for_system, ->(host_id) { where(host_id: host_id) }
   scope :for_policy, ->(policy_id) { joins(:profile).where(profiles: ::Profile.in_policy(policy_id)) }
-  scope :latest, ->(policy_id) { for_policy(policy_id).joins(:test_result).joins(::TestResult.with_latest) }
+  scope :latest, ->(policy_id) { for_policy(policy_id).joins(:test_result).merge(TestResult.latest) }
 
   # When requesting rule results, the DB response is scoped down by org_id numbers on the joined inventory. By
   # using the same index to join hosts and count results, we can save a lot of time.
--- a/app/models/test_result.rb
+++ b/app/models/test_result.rb
@@ -30,7 +30,7 @@ class TestResult < ApplicationRecord
   after_save :update_cached_fields!
   after_destroy :update_cached_fields!
 
-  scope :latest, -> { joins(with_latest) }
+  scope :latest, ->(j = 'JOIN') { joins(with_latest(j)) }
 
   scope :supported, lambda { |supported = true|
     where(supported: supported)
@@ -46,11 +46,11 @@ class TestResult < ApplicationRecord
       .select(:profile_id, :host_id, 'MAX(end_time) as end_time')
   end
 
-  def self.with_latest
+  def self.with_latest(join)
     latest_without_ids = group(:profile_id, :host_id).select(:profile_id, :host_id, 'MAX(end_time) as end_time')
 
     <<-SQL
-      JOIN (#{latest_without_ids.to_sql}) as tr ON
+      #{join} (#{latest_without_ids.to_sql}) as tr ON
       test_results.profile_id = tr.profile_id AND
       test_results.host_id = tr.host_id AND
       test_results.end_time = tr.end_time
--- a/test/graphql/queries/system_query_test.rb
+++ b/test/graphql/queries/system_query_test.rb
@@ -887,14 +887,18 @@ class SystemQueryTest < ActiveSupport::TestCase
 
     should 'properly sort by rules failed' do
       host2 = FactoryBot.create(:host, org_id: @user.account.org_id)
+      host3 = FactoryBot.create(:host, org_id: @user.account.org_id)
+      # For testing sorting of nil failed_rules
+      host4 = FactoryBot.create(:host, org_id: @user.account.org_id)
+      [host4, host2, host3].each do |host|
+        host.policies << @profile1.policy
+      end
 
-      # An outdated test result made in the past
-      FactoryBot.create(
+      tr = FactoryBot.create(
         :test_result,
         profile: @profile1,
         host: host2,
-        end_time: 1.minute.ago,
-        score: 100
+        score: 50
       )
 
       FactoryBot.create(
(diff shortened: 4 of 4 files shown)


app/models/host.rb
```diff
@@ -90,7 +90,7 @@ class Host < ApplicationRecord
 
   scope :with_failed_rules_count, lambda {
     sq = Host.left_outer_joins(test_result_profiles: :rule_results)
-             .where(rule_results: { result: RuleResult::FAILED })
+             .where("rule_results.result IN (?) OR rule_results.id IS NULL", RuleResult::FAILED)
```
skateman, line 93: You can do this in SQL :pray: look at how `or` works

marleystipich2: Right, I considered that but it was pretty long but I can switch that 👍

skateman: I think you don't need more in the OR than `Host.where(rule_results: { id: nil })`

marleystipich2: I'll try that. I was going to do something like this:
```
join_trp_rr = Host.left_outer_joins(test_result_profiles: :rule_results)
    sq = join_trp_rr.where(rule_results: { result: RuleResult::FAILED })
                    .or(join_trp_rr.where(rule_results: { id: nil }))
                    .group('hosts.id')
```

marleystipich2: `Host.where(rule_results: { id: nil })` worked so I'll use that 👍

skateman: You don't need to overcomplicate the query inside the `or`, it is not a "real" query, just something that can be merged with the main one.

skateman:
CI failures are an env problem ... cc @LightOfHeaven1994

app/models/host.rb
```diff
@@ -91,6 +91,7 @@ class Host < ApplicationRecord
   scope :with_failed_rules_count, lambda {
     sq = Host.left_outer_joins(test_result_profiles: :rule_results)
              .where(rule_results: { result: RuleResult::FAILED })
+             .or(Host.where(rule_results: { id: nil }))
              .select(arel_table[:id].as('id'), RuleResult.arel_table[:result].count.as('rules_failed'))
              .group('hosts.id')
```
marleystipich2, line 96: Does the SQL around the `COUNT("rule_results"."result")` part look correct to you? @skateman 
```
SELECT "inventory"."hosts"."id" AS id, COUNT("rule_results"."result") AS rules_failed FROM "inventory"."hosts"
LEFT OUTER JOIN "test_results" ON "test_results"."host_id" = "inventory"."hosts"."id" LEFT OUTER JOIN 
"profiles" ON "profiles"."id" = "test_results"."profile_id" LEFT OUTER JOIN "test_results" 
"test_results_profiles_join" ON "test_results_profiles_join"."profile_id" = "profiles"."id" LEFT OUTER JOIN 
"rule_results" ON "rule_results"."test_result_id" = "test_results_profiles_join"."id" WHERE ("rule_results"."result" 
IN ($1, $2, $3, $4) OR "rule_results"."id" IS NULL) GROUP BY "hosts"."id"  [["result", "fail"], ["result", "error"], 
["result", "unknown"], ["result", "fixed"]]
```

I tested some more and realized that the rules_failed count is too high sometimes. At first I thought my test data wasn't right but I don't think that's the issue. 🤔

skateman: You are doing sorting on the host against ALL its profiles while you just need to do it for a specific one, at least that's what I think based on the ticket. You do join with `test_result_profiles` but not limit the scope against any `profile`. Also you don't select the `latest` test results for the given profile either :thinking:

skateman: Also the tests are wrong too, because you want to sort based on ALL the test results not just the profile, you need to also filter the GQL results using `with_results_for_policy_id`.

app/models/host.rb
```diff
@@ -88,11 +88,17 @@ class Host < ApplicationRecord
     left_outer_joins(:test_result_profiles)
   }
 
-  scope :with_failed_rules_count, lambda {
-    sq = Host.left_outer_joins(test_result_profiles: :rule_results)
-             .where(rule_results: { result: RuleResult::FAILED })
+  scope :with_failed_rules_count, lambda { |profile = nil|
+    profile ||= RequestStore.store['scoped_search_context_profiles']
+    profile_ids = profile&.pluck(:id) || []
```
skateman, line 93: We need the profile IDs from the context that is coming from a search.

app/models/host.rb
```diff
@@ -88,11 +88,17 @@ class Host < ApplicationRecord
...
+
+    sq = Host.left_outer_joins(test_results: :rule_results)
```
skateman, line 95: Only joining with `test_results` as profiles are not necessary for us, we already have their IDs.

app/models/host.rb
```diff
@@ -88,11 +88,17 @@ class Host < ApplicationRecord
...
+             .merge(TestResult.latest('LEFT OUTER JOIN'))
```
skateman, line 96: Scoping down the results to the latest one, but also taking into account of systems without test results (that's why the `LEFT OUTER JOIN`)

app/models/host.rb
```diff
@@ -88,11 +88,17 @@ class Host < ApplicationRecord
...
+             .where(test_results: { profile_id: profile_ids }, rule_results: { result: RuleResult::FAILED })
+             .or(Host.where(test_results: { profile_id: profile_ids }, rule_results: { id: nil }))
```
skateman, line 98: Almost the same as your query, but adding `test_results.profile_id` so other profiles that we are not looking for don't leak into our context.

app/models/rule_result.rb
```diff
@@ -34,7 +34,7 @@ class RuleResult < ApplicationRecord
   scope :failed, -> { where(result: FAILED) }
   scope :for_system, ->(host_id) { where(host_id: host_id) }
   scope :for_policy, ->(policy_id) { joins(:profile).where(profiles: ::Profile.in_policy(policy_id)) }
-  scope :latest, ->(policy_id) { for_policy(policy_id).joins(:test_result).joins(::TestResult.with_latest) }
+  scope :latest, ->(policy_id) { for_policy(policy_id).joins(:test_result).merge(TestResult.latest) }
```
skateman, line 37: This is a residual change, the time I wrote this, I did not know about the `merge` :grin:

app/models/test_result.rb
```diff
@@ -30,7 +30,7 @@ class TestResult < ApplicationRecord
   after_save :update_cached_fields!
   after_destroy :update_cached_fields!
 
-  scope :latest, -> { joins(with_latest) }
+  scope :latest, ->(j = 'JOIN') { joins(with_latest(j)) }
```
skateman, line 33: Making the JOIN parametrized here so we can do a `LEFT OUTER JOIN` in some cases, but keeping the default as it was.

test/graphql/queries/system_query_test.rb
```diff
@@ -887,32 +887,51 @@ class SystemQueryTest < ActiveSupport::TestCase
...
+        rule: @profile1.rules.first,
+        result: 'fail',
+        test_result: tr2
       )
 
       query = <<-GRAPHQL
-      query getSystems($policyId: ID, $sortBy: [String!]) {
-          systems(limit: 50, offset: 1, sortBy: $sortBy) {
+      query getSystems($policyId: ID, $filter: String, $sortBy: [String!]) {
+          systems(limit: 50, offset: 1, search: $filter, sortBy: $sortBy) {
```
skateman, line 934: Without searching for a given profile, we cannot really count failed rules properly as rules from other profiles can leak into (that's what happened for you)

skateman (commented):
@marleystipich2 please go through my changes thoroughly and try to understand them. This became a very complex query, but I think you can learn from it :wink:

app/models/host.rb
marleystipich2: This was a good catch 👍 I forgot there could be older test results in the database

test/graphql/queries/system_query_test.rb
marleystipich2: This makes sense 👌

test/graphql/queries/system_query_test.rb
```diff
@@ -935,27 +954,31 @@ class SystemQueryTest < ActiveSupport::TestCase
...
-
-      returned_profile3 = desc_result.dig(0, 'node', 'testResultProfiles')
-      returned_profile4 = desc_result.dig(1, 'node', 'testResultProfiles')
+      )['data']['systems']['edges'].map do |row|
+        row['node']['testResultProfiles'].first['rulesFailed']
+      end
 
-      assert_equal 1, returned_profile1.dig(0, 'rulesFailed')
-      assert_equal 0, returned_profile2.dig(0, 'rulesFailed')
+      assert_equal asc_result, [0, 1, 2]
```
marleystipich2, line 979: The tests look good but we should probably test a system with `nil` failed rules right?

skateman: Well, feel free to adjust the tests

marleystipich2: Yep, working on it! And then I think everything looks good 👍

test/graphql/queries/system_query_test.rb
```diff
@@ -935,27 +957,35 @@ class SystemQueryTest < ActiveSupport::TestCase
 
       asc_result = Schema.execute(
         query,
-        variables: { policyId: @profile1.id, sortBy: ['rulesFailed'] },
+        variables: {
+          policyId: @profile1.id,
+          sortBy: ['rulesFailed'],
+          filter: "policy_id = #{@profile1.id}"
```
marleystipich2, line 963: I updated the filter from `with_results_for_policy_id` to `policy_id` so that we could test the nil rules_failed and it wasn't being returned with `with_results_for_policy_id` (since it doesn't have any results)

test/graphql/queries/system_query_test.rb
```diff
@@ -935,27 +957,35 @@ class SystemQueryTest < ActiveSupport::TestCase
...
-      returned_profile4 = desc_result.dig(1, 'node', 'testResultProfiles')
+      )['data']['systems']['edges'].map do |row|
+        if row['node']['testResultProfiles'].first
+          row['node']['testResultProfiles'].first['rulesFailed']
+        end
+      end
 
-      assert_equal 1, returned_profile1.dig(0, 'rulesFailed')
-      assert_equal 0, returned_profile2.dig(0, 'rulesFailed')
+      assert_includes([[nil, 0, 1, 2], [0, nil, 1, 2]], asc_result)
```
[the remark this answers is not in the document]
skateman: We can't do anything about it

skateman (approved):
The Seal of Approval

PR #1734, opened 2023-07-03
feat(RBAC): RHICOMPL-3850 get IDs of permitted inventory groups

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -17,7 +17,34 @@ class User < ApplicationRecord
     end
   end
 
+  def inventory_groups
+    rbac_permissions.each_with_object([]) do |permission, ids|
+      next unless Rbac.verify(permission.permission, Rbac::INVENTORY_HOSTS_READ)
+      # Empty array on 'resource_definitions' symbolizes a global access to the permitted resource.
+      # In such case, the method returns Rbac::ANY and skips parsing of attributeFilter.
+      return Rbac::ANY if permission.resource_definitions == []
+
+      permission.resource_definitions.each do |filter|
+        next unless valid_inventory_groups_definition?(filter[:attributeFilter])
+
+        ids.append(*inventory_groups_definition_value(filter[:attributeFilter]))
+      end
+    end
+  end
+
   private
+
+  def valid_inventory_groups_definition?(definition)
+    definition[:value].instance_of?(Array) &&
+      definition[:operation] == 'in' &&
+      definition[:key] == 'group.id'
+  end
+
+  def inventory_groups_definition_value(definition)
+    # Received '[nil]' symbolizes access to ungrouped entries.
+    # In output represtented with an empty array.
+    definition[:value].map { |dv| dv || [] }
+  end
 
   def rbac_permissions
     @rbac_permissions ||= Rbac.load_user_permissions(account.identity_header.raw)
--- a/app/services/rbac.rb
+++ b/app/services/rbac.rb
@@ -9,10 +9,11 @@ class Rbac
   OPTS = { limit: 1000, auth_names: '' }.freeze
   ANY = '*'
 
+  INVENTORY_UNGROUPED_ENTRIES = [].freeze
+  INVENTORY_HOSTS_READ = 'inventory:hosts:read'
   INVENTORY_VIEWER = 'inventory:*:read'
   COMPLIANCE_VIEWER = 'compliance:policy:read' # universal read permission accross all roles
   COMPLIANCE_ADMIN = 'compliance:*:*'
-
   POLICY_READ = 'compliance:policy:read'
   POLICY_CREATE = 'compliance:policy:create'
   POLICY_DELETE = 'compliance:policy:delete'
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -15,11 +15,11 @@ class UserTest < ActiveSupport::TestCase
       data: [
         RBACApiClient::Access.new(
           permission: 'app:resource0:*',
-          resource_definitions: nil
+          resource_definitions: []
         ),
         RBACApiClient::Access.new(
           permission: 'app:resource1:write',
-          resource_definitions: nil
+          resource_definitions: []
         )
       ]
     )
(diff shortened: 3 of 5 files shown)


app/models/user.rb
```diff
@@ -17,6 +17,20 @@ def authorized_to?(access_request)
...
   end
 
+  def inventory_groups
+    inventory_groups_permission = rbac_permissions.select do |permission|
+      permission.permission == Rbac::INVENTORY_GROUPS_READ
+    end
+
+    return Rbac::INVENTORY_GROUPS_GLOBAL if inventory_groups_permission.any? do |permission|
+      permission.resource_definitions == Rbac::INVENTORY_GROUPS_GLOBAL
+    end
```
skateman, line 27: These two can be done together in a single iteration.

app/models/user.rb
```diff
@@ -17,6 +17,20 @@ def authorized_to?(access_request)
...
+
+    inventory_groups_permission.collect do |permission|
```
skateman, line 29: There are some edge-cases that this code would not handle well.

Access to all hosts even though there is a resourceDefinition:
```json
   {
     "resourceDefinitions": [
       {
         "attributeFilter": {
           "key": "group.id",
           "value": [],
           "operation": "in"
         }
       }
     ],
     "permission": "inventory:hosts:read"
   },
```

The key in the attribute filter is not verified:
```json
   {
     "resourceDefinitions": [
       {
         "attributeFilter": {
           "key": "foo.id",
           "value": [1,2,3,4,...],
           "operation": "in"
         }
       }
     ],
     "permission": "inventory:hosts:read"
   },
```

The operation can be `equal` as @coderbydesign said:
```json
   {
     "resourceDefinitions": [
       {
         "attributeFilter": {
           "key": "group.id",
           "value": 1234,
           "operation": "equal"
         }
       }
     ],
     "permission": "inventory:hosts:read"
   },
```

skateman (changes requested):

test/models/user_test.rb
```diff
@@ -15,11 +15,11 @@ class UserTest < ActiveSupport::TestCase
       data: [
         RBACApiClient::Access.new(
           permission: 'app:resource0:*',
-          resource_definitions: nil
```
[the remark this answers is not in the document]
romanblanco: No. Thanks, @lpichler!

app/models/user.rb
```diff
@@ -17,8 +17,34 @@ def authorized_to?(access_request)
     end
   end
 
+  def inventory_groups
+    inventory_groups_access = rbac_permissions.select do |permission|
+      next if permission.permission != Rbac::INVENTORY_GROUPS_READ
+      return Rbac::INVENTORY_GROUPS_GLOBAL if permission.resource_definitions == Rbac::INVENTORY_GROUPS_GLOBAL
+
+      true if supported_operation?(permission.resource_definitions.first[:attributeFilter])
```
skateman, line 25:
```suggestion
      supported_operation?(permission.resource_definitions.first[:attributeFilter])
```

app/models/user.rb
```diff
@@ -17,8 +17,34 @@ def authorized_to?(access_request)
...
+  end
+
+  def operation_value(attribute_filter)
+    if attribute_filter[:operation] == 'in' && attribute_filter[:key] == 'groups.id'
+      attribute_filter[:value]
+    elsif attribute_filter[:operation] == 'equal' && attribute_filter[:key] == 'group.id'
+      [attribute_filter[:value]]
+    end
+  end
+
```
skateman, line 47: Does it make sense to keep these in the `user` model? Could they be moved into the `Rbac` as static methods?

app/models/user.rb
```diff
@@ -17,8 +17,34 @@ def authorized_to?(access_request)
...
+      attribute_filter[:operation] == 'equal' && attribute_filter[:key] == 'group.id'
+  end
+
+  def operation_value(attribute_filter)
+    if attribute_filter[:operation] == 'in' && attribute_filter[:key] == 'groups.id'
+      attribute_filter[:value]
+    elsif attribute_filter[:operation] == 'equal' && attribute_filter[:key] == 'group.id'
+      [attribute_filter[:value]]
+    end
+  end
```
skateman, line 46: You're basically doing the same condition twice, once in the `select` once in the `map`. I think it makes sense to merge them somehow and do the condition just once.

skateman (changes requested):

app/services/rbac.rb
```diff
@@ -35,6 +37,19 @@ def load_user_permissions(identity)
...
+    end
+
+    def operation_value(attribute_filter)
+      if attribute_filter[:operation] == 'in' && attribute_filter[:key] == 'groups.id'
+        attribute_filter[:value] == [nil] ? {} : attribute_filter[:value]
+      elsif attribute_filter[:operation] == 'equal' && attribute_filter[:key] == 'group.id'
+        attribute_filter[:value]
+      end
+    end
+
```
skateman, line 52: Method names sound universal to me, but they are scoped for inventory groups. For now let's keep them, but I would put a comment in front of them about this.

app/models/user.rb
```diff
@@ -17,6 +17,18 @@ def authorized_to?(access_request)
...
+  def inventory_groups
+    rbac_permissions.map do |permission|
+      next if permission.permission != Rbac::INVENTORY_GROUPS_READ
+      return Rbac::GLOBAL_RESOURCE_PERMISSION if permission.resource_definitions == []
+
+      filter = permission.resource_definitions.first[:attributeFilter]
+      next unless Rbac.supported_operation?(filter)
+
+      Rbac.operation_value(filter)
+    end.flatten.compact
```
skateman, line 29: Code looks good, however, it should have some comments to explain things. Especially the jumps that need explanation.

skateman (changes requested):
Starting to look good :+1:

app/services/rbac.rb
```diff
@@ -35,6 +37,23 @@ def load_user_permissions(identity)
...
+    def supported_operation?(attribute_filter)
+      attribute_filter[:operation] == 'in' && attribute_filter[:key] == 'groups.id' ||
+        attribute_filter[:operation] == 'equal' && attribute_filter[:key] == 'group.id'
+    end
+
+    # Scoped for inventory groups
+    def operation_value(attribute_filter)
+      if attribute_filter[:operation] == 'in' && attribute_filter[:key] == 'groups.id'
+        # Received '[nil]' symbolizes access to ungrouped entries.
+        # In the output, the ungrouped entries are symbolized with '{}'.
```
skateman, line 50: I thought they are symbolized with empty array...

app/services/rbac.rb
```diff
@@ -35,6 +37,23 @@ def load_user_permissions(identity)
...
+        attribute_filter[:value] == [nil] ? {} : attribute_filter[:value]
+      elsif attribute_filter[:operation] == 'equal' && attribute_filter[:key] == 'group.id'
+        attribute_filter[:value]
```
skateman, line 53: What if the value is `nil` here?

test/controllers/graphql_controller_test.rb
```diff
@@ -20,7 +20,18 @@ class GraphqlControllerTest < ActionDispatch::IntegrationTest
...
+          {
+            permission: Rbac::COMPLIANCE_ADMIN,
+            resource_definitions: Rbac::GLOBAL_RESOURCE_PERMISSION
+          },
+          {
+            permission: Rbac::INVENTORY_VIEWER,
+            resource_definitions: Rbac::GLOBAL_RESOURCE_PERMISSION
+          }
+        ]
+      )
```
skateman, line 34: This seems very repetitive across the app. Could we move the second element into the `stub_rbac_permissions` method instead? At least in cases when it is unset? To have a default GLOBAL value...

skateman (changes requested):

app/services/rbac.rb
skateman: `flatten(1)` should do the trick

skateman: Or maybe use `flat_map` instead of the `map`

romanblanco:
```
[5] pry(#<User>)> a
=> [["78e3dc30-cec3-4b49-be2d-37482c74a9ac", "79e3dc30-cec3-4b49-be2d-37482c74a9ad"], "80e3dc30-cec3-4b49-be2d-37482c74a9ad", [], nil, nil]
[6] pry(#<User>)> a.flatten(1)
=> ["78e3dc30-cec3-4b49-be2d-37482c74a9ac", "79e3dc30-cec3-4b49-be2d-37482c74a9ad", "80e3dc30-cec3-4b49-be2d-37482c74a9ad", nil, nil]
```

Using `flatten(1)` for this would still remove the `[]` representing ungrouped hosts.

romanblanco: Will try look for alternative.

app/services/rbac.rb
```diff
@@ -35,6 +39,31 @@ def load_user_permissions(identity)
...
+    def operation_value(attribute_filter)
+      case attribute_filter[:operation]
+      when 'in'
+        # Received '[nil]' symbolizes access to ungrouped entries.
+        # In the output, the ungrouped entries are symbolized with '[[]]' return value.
+        attribute_filter[:value] == [nil] ? [INVENTORY_UNGROUPED_ENTRIES] : attribute_filter[:value]
+      when 'equal'
+        [attribute_filter[:value]]
+      end
+    end
```
skateman, line 65: This is getting a bit too complex :thinking:

skateman (changes requested):

test/controllers/concerns/authentication_test.rb
```diff
@@ -112,7 +112,12 @@ def raising
 
   context 'after successful login' do
     setup do
-      stub_rbac_permissions(Rbac::COMPLIANCE_VIEWER, Rbac::INVENTORY_VIEWER)
+      stub_rbac_permissions(
+        [
+          { permission: Rbac::COMPLIANCE_VIEWER },
+          { permission: Rbac::INVENTORY_VIEWER }
+        ]
+      )
```
skateman, line 120: The `stub_rbac_permission` method signature should not change. The old way of calling it with a simple array should work as before. In addition it could handle the complex values...

skateman (changes requested):

app/services/rbac.rb
```diff
@@ -35,6 +39,25 @@ def load_user_permissions(identity)
...
+        attribute_filter[:operation] == 'equal' &&
+        attribute_filter[:key] == 'group.id' &&
+        attribute_filter[:value].instance_of?(String)
+      )
+    end
+
+    # Scoped for inventory groups
+    def operation_value(attribute_filter)
+      # Received '[nil]' symbolizes access to ungrouped entries.
+      attribute_filter[:value] == [nil] ? [INVENTORY_UNGROUPED_ENTRIES] : attribute_filter[:value]
```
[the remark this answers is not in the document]
romanblanco: The unit test `'can request users access to inventory groups'` is testing this case, please take a look.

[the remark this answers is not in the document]
skateman: I don't think so, see my comment below.

test/models/user_test.rb
```diff
@@ -37,4 +37,235 @@ class UserTest < ActiveSupport::TestCase
...
+        ),
+        RBACApiClient::Access.new(
+          permission: Rbac::INVENTORY_GROUPS_READ,
+          resource_definitions: [
+            {
+              attributeFilter: {
+                key: 'groups.id',
+                value: [
+                  nil # ungrouped hosts
+                ],
```
skateman, line 81: What if you have more entities here than the `nil`?

skateman:
I think there's an additional uncovered edge-case when the user has access to `inventory:hosts:*`

app/models/user.rb
```diff
@@ -17,6 +17,20 @@ def authorized_to?(access_request)
...
 
+  def inventory_groups
+    rbac_permissions.each_with_object([]) do |permission, ids|
+      next if permission.permission != Rbac::INVENTORY_GROUPS_READ
+      # Empty array on 'resource_definitions' symbolizes a global access to the permitted resource.
+      # In such case, the method returns Rbac::GLOBAL_RESOURCE_ACCESS and skips parsing of attributeFilter.
+      return Rbac::GLOBAL_RESOURCE_ACCESS if permission.resource_definitions == []
+
+      filter = permission.resource_definitions.first[:attributeFilter]
+      next unless Rbac.supported_operation?(filter)
```
skateman, line 28:
```suggestion
      next unless Rbac.supported_definition?(filter, 'groups.in', 'group.in')
```

app/models/user.rb
```diff
@@ -17,6 +17,20 @@ def authorized_to?(access_request)
...
+
+      ids.append(*Rbac.operation_value(filter))
```
skateman, line 30:
```suggestion
      ids.append(*Rbac.definition_value(filter, 'groups.id', 'group.id'))
```

app/models/user.rb
skateman:
```suggestion
      next unless Rbac.supported_definition?(filter, group: :id)
      
      def supported_definition?(definition, selector)
        "#{selector.key.pluralize}.#{selector.value}"
        "#{selector.key}.#{selector.value}"
      end
```

test/models/user_test.rb
romanblanco: Created a test for this.

test/models/user_test.rb
```diff
@@ -37,4 +37,293 @@ class UserTest < ActiveSupport::TestCase
...
+      data: [
+        RBACApiClient::Access.new(
+          permission: Rbac::INVENTORY_HOSTS_READ,
+          resource_definitions: [
+            {
+              attributeFilter: {
+                key: 'group.id',
+                value: nil, # nil
+                operation: 'equal'
+              }
```
skateman, line 251: According to the specs, this represents access to only ungrouped hosts. So this test is not according to the specs :disappointed:

app/models/user.rb
```diff
@@ -17,6 +17,19 @@ def authorized_to?(access_request)
...
 
+  def inventory_groups
+    rbac_permissions.each_with_object([]) do |permission, ids|
+      next unless Rbac.verify(permission.permission, Rbac::INVENTORY_HOSTS_READ)
+      # Empty array on 'resource_definitions' symbolizes a global access to the permitted resource.
+      # In such case, the method returns Rbac::GLOBAL_RESOURCE_ACCESS and skips parsing of attributeFilter.
+      return Rbac::GLOBAL_RESOURCE_ACCESS if permission.resource_definitions == []
+
+      permission.resource_definitions.each do |filter|
+        ids.append(*Rbac.definition_value(filter[:attributeFilter], group: :id))
```
skateman, line 28: There's an edge case that this doesn't handle well...when the response is `nil` because the user only has `equal` access to ungrouped hosts.

app/services/rbac.rb
```diff
@@ -35,6 +39,21 @@ def load_user_permissions(identity)
...
+         definition[:key] == "#{selector.keys.first.to_s.pluralize}.#{selector.values.first}" &&
+         definition[:value].instance_of?(Array)
+        # Received '[nil]' symbolizes access to ungrouped entries.
+        # In output represtented with an empty array.
+        definition[:value].map { |dv| dv || [] }
+      elsif definition[:operation] == 'equal' &&
+            definition[:key] == "#{selector.keys.first}.#{selector.values.first}" &&
+            definition[:value].instance_of?(String)
+        definition[:value]
+      end
```
skateman, line 54: There might be other uncovered outputs that would result in `nil` which make the edge-case described above uncatchable. Please make sure if we REALLY need to support the `equal` syntax.

skateman (changes requested):

app/services/rbac.rb
romanblanco: We don't.

app/models/user.rb
```diff
@@ -17,6 +17,19 @@ def authorized_to?(access_request)
...
 
+  def inventory_groups
+    rbac_permissions.each_with_object([]) do |permission, ids|
+      next unless Rbac.verify(permission.permission, Rbac::INVENTORY_HOSTS_READ)
+      # Empty array on 'resource_definitions' symbolizes a global access to the permitted resource.
+      # In such case, the method returns Rbac::ANY and skips parsing of attributeFilter.
+      return Rbac::ANY if permission.resource_definitions == []
+
+      permission.resource_definitions.each do |filter|
+        ids.append(*Rbac.definition_value(filter[:attributeFilter], group: :id))
```
skateman, line 28: As we no longer need the juggling with pluralization, this can be simplified:
```suggestion
        ids.append(*Rbac.definition_value(filter[:attributeFilter], 'group.id'))
```

app/services/rbac.rb
```diff
@@ -35,6 +37,17 @@ def load_user_permissions(identity)
       end
     end
 
+    # Scoped for inventory groups
+    def definition_value(definition, selector)
+      if definition[:operation] == 'in' &&
+         definition[:key] == "#{selector.keys.first}.#{selector.values.first}" &&
```
skateman, line 43:
```suggestion
         definition[:key] == selector &&
```

app/services/rbac.rb
```diff
@@ -35,6 +37,17 @@ def load_user_permissions(identity)
       end
     end
 
+    # Scoped for inventory groups
+    def definition_value(definition, selector)
```
skateman, line 41: I would also parametrize the `operation` which would make the method universal and not just scoped for inventory groups.

skateman (changes requested):

app/services/rbac.rb
```diff
@@ -8,11 +8,13 @@ class Rbac
   APPS = 'compliance,inventory'
   OPTS = { limit: 1000, auth_names: '' }.freeze
   ANY = '*'
+  GLOBAL_RESOURCE_PERMISSION = [].freeze
```
skateman, line 11: Please drop this constant and just use the []

app/services/rbac.rb
```diff
@@ -35,6 +37,16 @@ def load_user_permissions(identity)
       end
     end
 
+    def definition_value(definition, operation, selector)
+      if definition[:operation] == operation &&
+          definition[:key] == selector &&
+          definition[:value].instance_of?(Array)
```
skateman, line 43:
```suggestion
      return unless definition[:operation] == operation && definition[:key] == selector && definition[:value].is_a?(Array)
```

app/models/user.rb
```diff
@@ -17,6 +17,19 @@ def authorized_to?(access_request)
...
 
+  def inventory_groups
+    rbac_permissions.each_with_object([]) do |permission, ids|
+      next unless Rbac.verify(permission.permission, Rbac::INVENTORY_HOSTS_READ)
+      # Empty array on 'resource_definitions' symbolizes a global access to the permitted resource.
+      # In such case, the method returns Rbac::ANY and skips parsing of attributeFilter.
+      return Rbac::ANY if permission.resource_definitions == []
+
+      permission.resource_definitions.each do |filter|
+        ids.append(*Rbac.definition_value(filter[:attributeFilter], 'in', 'group.id'))
```
skateman, line 28: Consider deoptimizing the function here with a `next` call as it is now just a single operation

app/services/rbac.rb
```diff
@@ -9,10 +9,11 @@ class Rbac
   OPTS = { limit: 1000, auth_names: '' }.freeze
   ANY = '*'
 
+  INVENTORY_UNGROUPED_ENTRIES = [].freeze
+  INVENTORY_HOSTS_READ = 'inventory:hosts:read'
```
skateman, line 13: What about `inventory:hosts:*` 🤔 isn't that a valid one?

romanblanco: Covered in test `'global access to hosts'`.

skateman (approved):
The Seal of Approval

PR #1739, opened 2023-07-11
fix(Host): RHICOMPL-3731 sort failed rules with 0 rules failed

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -53,7 +53,10 @@ class Host < ApplicationRecord
     scope: :with_benchmark
   )
 
-  sortable_by(:rules_failed, 'sq.rules_failed', scope: :with_failed_rules_count)
+  sortable_by :rules_failed, Arel::Nodes::NamedFunction.new(
+    'COALESCE',
+    [Arel::Nodes::SqlLiteral.new('sq.rules_failed'), 0]
+  ), scope: :with_failed_rules_count
 
   self.table_name = 'inventory.hosts'
   self.primary_key = 'id'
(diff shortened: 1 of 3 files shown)


skateman:
> The tests were already covering the 0 failed rules case so I didn't update them. Not sure why the tests were passing but it didn't work correctly for 0 failed rules for Egor.

If the tests are passing in both cases, it means that the tests aren't testing what they should be and should be extended with another testcase. This means that you should create a new test that fails in the old code and passes in the new.

app/models/host.rb
```diff
@@ -99,7 +99,7 @@ class Host < ApplicationRecord
              .select(arel_table[:id].as('id'), RuleResult.arel_table[:result].count.as('rules_failed'))
```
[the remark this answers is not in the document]
marleystipich2: Okay this brought back the disappearing 0 but it did change the ordering somehow. Now it treats 0 and nil as the highest values like `[0, nil, 2, 1]` for desc and [`1, 2, nil, 0]` for asc. Is this fine though? So `N/A` failed rules will be displayed as the 'highest' value failed rules.

But 0's are showing up now. I figured out what was wrong with the tests.

app/models/host.rb
```diff
@@ -92,14 +92,14 @@ class Host < ApplicationRecord
     profile ||= RequestStore.store['scoped_search_context_profiles']
     profile_ids = profile&.pluck(:id) || []
 
-    sq = Host.left_outer_joins(test_results: :rule_results)
-             .merge(TestResult.latest('LEFT OUTER JOIN'))
+    sq = Host.joins(test_results: :rule_results)
+             .merge(TestResult.latest)
```
skateman, line 96: I adjusted the `latest` to take an optional argument, that can be reverted I guess as it is no longer necessary.

test/graphql/queries/system_query_test.rb
```diff
@@ -886,11 +886,15 @@ class SystemQueryTest < ActiveSupport::TestCase
     end
 
     should 'properly sort by rules failed' do
+      TestResult.delete_all
+      RuleResult.delete_all
+      Host.delete_all
```
skateman, line 891:
```suggestion
      WHost.delete_all
```

skateman (approved):
The Seal of Approval

PR #1751, opened 2023-07-18
cleanup(deps): RHICOMPL-3853 replace ruby-kafka with waterdrop

RoamingNoMaD:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -115,8 +115,6 @@ GEM
     deep_merge (1.2.2)
     diff-lcs (1.3)
     digest (3.1.1)
-    digest-crc (0.6.4)
-      rake (>= 12.0.0, < 14.0.0)
     docile (1.4.0)
     dotenv (2.7.6)
     dotenv-rails (2.7.6)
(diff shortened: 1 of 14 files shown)


Gemfile.lock
```diff
@@ -207,6 +205,13 @@ GEM
       addressable (>= 2.8)
     jsonapi-serializer (2.2.0)
       activesupport (>= 4.2)
+    karafka-core (2.1.1)
+      concurrent-ruby (>= 1.1)
+      karafka-rdkafka (>= 0.13.1, < 0.14.0)
```
skateman, line 210: Hmm, this doesn't sound convincing 😞 seems like karafka has its own fork of rdkafka. One of the reasons to use waterdrop was that it will use the same rdkafka gem for producing and consuming messages as well, but it seems that it is not true. 

I am not saying that we should throw away your PR, because eventually I want to migrate us towards a single producer/consumer, but as long as we don't have an alternative for racecar, we should probably stick with a single library. 

Do you think you could stash this PR somewhere and for now just revive mine as a substitution for the ruby-kafka producer we are using at the moment?

mensfeld: @skateman I "own" both karafka-rdkafka and rdkafka. The only reason why they are not merged is :clock1: 

karafka-rdkafka is compatible with rdkafka (though not the other way around). I am actively working on merging them into one but I cannot say when will it happen. All new development is added to both as long as it is possible.

PR #1757, opened 2023-07-21
feat(CI): RHICOMPL-3902 ephemeral ui nightly script

LightOfHeaven1994:
We need this script to be able to use in Jenkins pipeline to check compliance ephemeral health nightly

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/scripts/nightly_smoke_ui_tests.sh
+++ b/scripts/nightly_smoke_ui_tests.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# This job is used for checking compliance frontend health in ephemeral environment
+
+set -ex
+
+export IMAGE="quay.io/cloudservices/compliance-backend"
+export APP_NAME="compliance"  # name of app-sre "application" folder this component lives in
+export COMPONENT_NAME="compliance-backend"  # name of app-sre "resourceTemplate" in deploy.yaml for this component
+export IQE_PLUGINS="compliance"
+export IQE_MARKER_EXPRESSION="ephemeral_ui_nightly"
+export IQE_FILTER_EXPRESSION=""
+export IQE_ENV="ephemeral"
+export IQE_SELENIUM="true"
+export IQE_CJI_TIMEOUT="30m"
+export DEPLOY_TIMEOUT="900"
+export DEPLOY_FRONTENDS="true"
+export COMPONENTS_W_RESOURCES="compliance"
+
+# Get bonfire helper scripts
+CICD_URL="https://raw.githubusercontent.com/RedHatInsights/cicd-tools/main"
+rm -f .cicd_bootstrap.sh
+# shellcheck source=/dev/null
+curl -s $CICD_URL/bootstrap.sh > .cicd_bootstrap.sh && source .cicd_bootstrap.sh
+
+# Smoke tests
+# shellcheck source=/dev/null
+source "$CICD_ROOT/deploy_ephemeral_env.sh"
+# shellcheck source=/dev/null
+source "$CICD_ROOT/cji_smoke_test.sh"


skateman:
cc @vkrizan

LightOfHeaven1994:
This script is going to be used in ci jenkins. You can related MR here https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/74820

scripts/nightly_smoke_ui_tests.sh
```diff
@@ -0,0 +1,30 @@
...
+export COMPONENT_NAME="compliance-backend"  # name of app-sre "resourceTemplate" in deploy.yaml for this component
+export IQE_PLUGINS="compliance"
+export IQE_MARKER_EXPRESSION="ephemeral_ui_nightly"
+export IQE_FILTER_EXPRESSION=""
+export IQE_ENV="ephemeral"
+export IQE_SELENIUM="true"
+export IQE_CJI_TIMEOUT="30m"
+export DEPLOY_TIMEOUT="900"
+export DEPLOY_FRONTENDS="true"
+export COMPONENTS_W_RESOURCES="compliance"
```
vkrizan, line 18: We're duplicating a lots of constants that we already have in https://github.com/RedHatInsights/compliance-backend/blob/master/Jenkinsfile

scripts/nightly_smoke_ui_tests.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
vkrizan, line 1: Can this be a Jenkins file? cc @Victoremepunto

Victoremepunto: I believe you could add a condition for this to run on schedule against a fixed branch, with a cron schedule directly on the pipeline. I have not tested this myself. This pipeline is mimicking what Advisor does for their smoke test builds . 

Similarly checking FE availability I must say, perhaps this pipeline is not required if relying on Advisor's smoke test pipeline is enough ? This will not cover Compliance-specific health checks, of course, but checking the a FE deployment "works" should be covered - where "works" means: we are able to deploy a CRC aaplication on EE and we are able to successfully navigate to one of the views and interact with parts of if through a Selenium browser (effectively validating a FE application is able to run on EE).

scripts/nightly_smoke_ui_tests.sh
Victoremepunto: Unfortunately, by leveraging Bash scripts from the main CICD tooling, one has to set these variables on the environment. 

Being this a Bash script and the PR verification process a Jenkinsfile file, duplication removal comes at the cost of obscurity (you could add these to a variables-only script you source in both the Jenkinsfile and this smoke tests script, but at the cost of having a dependency between them, which is IMHO way worse).

scripts/nightly_smoke_ui_tests.sh
vkrizan: So what I hear is that a parametrization in a Jenkins file and a regular schedules for a specific branch would do it.

Victoremepunto: Yes that is correct. I have not tried this myself (on our current app-interface context, that is, it does work on "vanilla" Jenkins) but that should do the trick. 

@LightOfHeaven1994  LMK if you need help testing this ^

LightOfHeaven1994: @Victoremepunto I will, let's speak about this on Friday meeting if possible. Thanks!

PR #1769, opened 2023-08-03
feat(V2::Parameters): RHICOMPL-3891 allow params for nested routes

skateman:
I am splitting out the parameter handling for APIv2 as we need to have different behavior for nested routes and I don't want to allow these parameters in APIv1. Rails routes allows the specification of custom parameters for resources, which I am planning to use for setting up a list of parent models in a nested resource scenario:
```ruby
resources :security_guides do
  resources :profiles, parents: [SecurityGuide] do
    resources :rules, parents: [SecurityGuide, Profile]
  end
end
```
This list of parents will be useful for many other scenarios, but for now we will stick to using them for parameter validation only. The `StrongerParameters` gem that we use for parameter validation does not like the implicit parameters (`profile_id` and `security_guide_id` in case of the example above) of parent resources. Therefore, I adjusted the parameter validation to incorporate both the `parents` parameter coming from the router and all the possible `_id` parameters generated from these parents.

As we have no nested controllers yet, tests for this feature will arrive later...

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/concerns/v2/parameters.rb
+++ b/app/controllers/concerns/v2/parameters.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+module V2
+  # Reusable parameter checking for all controllers
+  module Parameters
+    extend ActiveSupport::Concern
+
+    # Constraint validating model classes passed via params[:parents]
+    class ModelConstraint < StrongerParameters::Constraint
+      def value(val)
+        return val if val.is_a?(Class) && val < ::ApplicationRecord
+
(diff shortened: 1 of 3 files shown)


romanblanco (approved):
LGTM :+1:

PR #1785, opened 2023-08-22
feat(RSpec): helper wrapper for nested route `params[:parents]` mocking

skateman:
RSpec automatically coerces all passed params to string by calling the `to_s` method on them. This is unfortunate as we are using route-level parameters to determine the nesting of a given route. To avoid this, I created a wrapper that can temporarily mock the `.to_s` method on a given list of parent classes that would be passed as a route-level parameter. After the request is called, the mocks are reset back to normal so there should be no side effects. The method yields a list of mocked parents that can be passed with the `params` in the block:

```ruby
nested(Model1, Model2) do |parents|
  get :index, params: { model_1_id: 1, model_2_id: 2, parents: parents }
end
```

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -36,6 +36,18 @@ def stub_rbac_permissions(*arr, **hsh)
 end
 # rubocop:enable Metrics/MethodLength
 
+def nested_route(*parents)
+  # Mock the .to_s method on each parent in the array to avoid RSpec converting it
+  parents.each { |parent| allow(parent).to receive(:to_s).and_return(parent) }
+
+  yield(parents)
+ensure # Make sure that the mock is just for the time of the HTTP request
+  parents.each do |parent|
+    klass = RSpec::Mocks.space.proxy_for(parent)
+    klass.instance_variable_get(:@method_doubles)[:to_s].reset
+  end
+end
+
 def response_body_data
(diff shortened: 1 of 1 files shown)


xmicha82 (dismissed):
LGTM

xmicha82 (approved):
works well, thank you

vkrizan:
@skateman any reference material for the RSpec? I'd wanna know why is it coerced and whether it can be contributed back to RSpec.

skateman:
> @skateman any reference material for the RSpec? I'd wanna know why is it coerced and whether it can be contributed back to RSpec.

RSpec request simulation is weird, it doesn't go through the routes directly, just looks up the controllerfrom the router config and calls the method directly. Because of this, the parents need to be passed explicitly to match the routing table, but then they are treated as a regular parameter from the outside world. The type conversion is a feature and not a bug, when the request is text/plain, this HAS to be string...the only exception is for JSON, but our classes don't comply with the JSON specs either.

PR #1811, opened 2023-09-01
RHICOMPL-3959 make the controllers usable for multiple routes

skateman:
* V2 model reflections and foreign keys now match the parent IDs in nested routes
* Simplified relationships on the V2 models, less to clean up later after the remodel
* Instead of parent resources, routes now define parent reflections
* Resources are automatically joined & scoped with their parent resources in nested routes

Note that the `parents` configuration expects the reflection (`has_many`, `belongs_to`, etc) name, which is in some cases pluralized, e.g.:
```ruby
# app/models/v2/rules.rb
belongs_to :security_guide
has_many :profiles, through: :profile_rules, source: :profile, class_name: 'V2::Profile'


# config/routes.rb
resources :security_guides, only: [:index, :show] do
  resources :profiles, only: [:index, :show], parents: [:security_guide] do
    resources :rules, only: [:index, :show], parents: [:security_guide, :profiles] # SG is singular, profiles is plural
  end
end

``` 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/concerns/v2/collection.rb
+++ b/app/controllers/concerns/v2/collection.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module V2
+  # Generic methods to be used when calling resolve_collection on our models
+  module Collection
+    extend ActiveSupport::Concern
+
+    include ::TagFiltering
+
+    included do
+      private
+
+      def resolve_collection
+        result = filter_by_tags(sort(search(policy_scope(expand_resource))))
+        result.paginate(page: pagination_offset, per_page: pagination_limit)
+      end
+
+      # :nocov:
+      def filter_by_tags(data)
+        unless TagFiltering.tags_supported?(resource) && permitted_params[:tags]&.any?
+          return data
+        end
+
+        tags = parse_tags(permitted_params[:tags])
+        data.where('tags @> ?', tags.to_json)
+      end
+      # :nocov:
+
+      # rubocop:disable Metrics/AbcSize
+      # :nocov:
+      def search(data)
+        return data if permitted_params[self.class::SEARCH].blank?
+
+        # Fail if search is not supported for the given model
+        if !data.respond_to?(:search_for) || permitted_params[self.class::SEARCH].match(/\x00/)
+          raise ActionController::UnpermittedParameters.new(self.class::SEARCH => permitted_params[self.class::SEARCH])
+        end
+
+        data.search_for(permitted_params[self.class::SEARCH])
+      end
+      # :nocov:
+      # rubocop:enable Metrics/AbcSize
+
+      # :nocov:
+      def sort(data)
+        order_hash, extra_scopes = data.klass.build_order_by(permitted_params[:sort_by])
+
+        extra_scopes.inject(data.order(order_hash)) do |result, scope|
+          result.send(scope)
+        end
+      end
+      # :nocov:
+    end
+  end
+end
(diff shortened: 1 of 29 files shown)


app/controllers/concerns/v2/collection.rb
```diff
@@ -0,0 +1,57 @@
...
+        extra_scopes.inject(data.order(order_hash)) do |result, scope|
+          result.send(scope)
+        end
+      end
+
+      def expand_resource
+        # Iterate through all the parents of the resource and join them one-by-one to the scope
+        permitted_params[:parents].to_a.reduce(resource) do |scope, parent|
+          ref = scope.reflect_on_association(parent)
+          scope.joins(parent).where(parent => { ref.klass.primary_key => permitted_params[ref.foreign_key] })
```
skateman, line 52: @petrblaho you're going to love this 😁

app/controllers/v2/application_controller.rb
```diff
@@ -60,5 +60,17 @@ def rbac_allowed?
...
+    def expand_resource
+      permitted_params[:parents].to_a.reduce(resource) do |scope, parent|
+        ref = scope.reflect_on_association(parent)
+        klass = ref.klass
+
+        scope.joins(parent)
+             .where(parent => { klass.primary_key => permitted_params[ref.foreign_key] })
+             .merge(Pundit.policy_scope(current_user, klass))
+      end
+    end
```
skateman, line 74: @petrblaho you're going to love this 😉

RoamingNoMaD (approved):
lgtm

app/controllers/v2/application_controller.rb
```diff
@@ -60,5 +60,17 @@ def rbac_allowed?
       permission = self.class.instance_variable_get(:@action_permissions)[action_name.to_sym]
       user.authorized_to?(Rbac::INVENTORY_HOSTS_READ) && user.authorized_to?(permission)
     end
+
+    # Reduce through all the parents of the resource and join+scope them on the resource
+    def expand_resource
+      permitted_params[:parents].to_a.reduce(resource) do |scope, parent|
```
vkrizan, line 66: Can this check for existence of `:parents` first? Or it isn't needed?

app/controllers/v2/application_controller.rb
```diff
@@ -60,5 +60,17 @@ def rbac_allowed?
...
+        ref = scope.reflect_on_association(parent)
+        klass = ref.klass
+
+        scope.joins(parent)
+             .where(parent => { klass.primary_key => permitted_params[ref.foreign_key] })
```
vkrizan, line 71: Similar here, can `permitted_params[ref.foreign_key]` be `nil`?

app/controllers/v2/application_controller.rb
vkrizan: `nil.to_a` returns `[]` empty array. Can you please add a comment for this case, that it would return `resource` "non-reduced"?

app/controllers/v2/application_controller.rb
skateman: This won't be called if the ID is `nil` also this behavior is given by rails.

vkrizan (approved):

petrblaho (approved):
LGTM!

romanblanco (approved):
LGTM :+1:

PR #1828, opened 2023-10-02
fix(Host): RHINENG-2329 return empty scope as arel node for groups

skateman:
The `SqlLiteral` arel object has no `or` method defined, so the ungrouped hosts only condition needs to be converted to a different node that returns with an empty collection.
 
## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -135,7 +135,7 @@ class Host < ApplicationRecord
   def self.arel_inventory_groups(groups, key)
     jsons = groups.map { |group| [{ key => group }].to_json.dump }
 
-    return Arel.sql('1 = 0') if jsons.empty?
+    return AN::InfixOperation.new('=', Arel.sql('1'), Arel.sql('0')) if jsons.empty?
 
     AN::InfixOperation.new(
       '@>', arel_table[:groups],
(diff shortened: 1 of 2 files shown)


LightOfHeaven1994 (approved):
lgtm!

app/models/host.rb
```diff
@@ -135,7 +135,7 @@ def self.available_os_versions
   def self.arel_inventory_groups(groups, key)
     jsons = groups.map { |group| [{ key => group }].to_json.dump }
 
-    return Arel.sql('1 = 0') if jsons.empty?
+    return AN::InfixOperation.new('=', Arel.sql('1'), Arel.sql('0')) if jsons.empty?
```
vkrizan, line 138: Can this `1=0` be a (global) constant? 
Just curious, isn't there a way go without `1=0`?

skateman: In any other case the `scope.none` should be sufficient, but in the arel domain we need the arel version of it. So in every other place we stick to the `none`, but here as the `or` needs to be also in the arel domain, we cannot avoid it. So long story short: this will be the only place to keep it, everywhere else we are using the ActiveRecord way.

PR #1829, opened 2023-10-03
refactor(tests): introduce collection and resource shared examples

RoamingNoMaD:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/controllers/v2/concerns/collection_spec.rb
+++ b/spec/controllers/v2/concerns/collection_spec.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+# The `parents` parameter is required when testing nested controllers, and it should contain
+# an ordered list of reflection symbols similarly to how they are defined in the routes. It is
+# also required to set the `extra_params` variable in a let block and pass all the parent IDs
+# there as a hash. For example:
+# ```
+# let(:extra_params) { { security_guide_id: 123, profile_id: 456 } }
+#
+# it_behaves_like 'collection', :security_guide, :profile
+# ```
+#
+# In a non-nested case the let block should contain an empty hash and the `parents` parameter
+# should be empty, like so:
+# ```
+# let(:extra_params) { {} }
+#
+# it_behaves_like 'collection'
+# ```
+#
+RSpec.shared_examples 'collection' do |*parents|
+  it 'returns base fields for each result' do
+    collection = items.map do |item|
+      hash_including(
+        'id' => item.id,
+        'type' => described_class.controller_name.chop,
+        **attributes.each_with_object({}) do |(key, value), obj|
+          obj[key.to_s] = item.send(value)
+        end
+      )
+    end
+
+    get :index, params: extra_params.merge(parents: parents)
+
+    expect(response).to have_http_status :ok
+    expect(response_body_data).to match_array(collection)
+
+    response_body_data.each do |item|
+      expect(item.keys.count).to eq(attributes.keys.count + 2)
+    end
+  end
+
+  context 'under incorrect parent', if: parents.present? do
+    let(:parent) { FactoryBot.create(:"v2_#{parents.first}") }
+
+    it 'returns not_found' do
+      get :index, params: extra_params.merge(parents: parents)
+
+      expect(response_body_data).to be_empty
+    end
+  end
+
+  context 'Unathorized' do
+    let(:rbac_allowed?) { false }
+
+    it 'responds with unauthorized status' do
+      get :index, params: extra_params.merge(parents: parents)
+
+      expect(response).to have_http_status :forbidden
+    end
+  end
+end
--- a/spec/controllers/v2/profiles_controller_spec.rb
+++ b/spec/controllers/v2/profiles_controller_spec.rb
@@ -40,36 +40,10 @@ describe V2::ProfilesController do
   end
 
   describe 'GET show' do
-    let(:profile) { FactoryBot.create(:v2_profile) }
-    let(:security_guide) { profile.security_guide }
+    let(:item) { FactoryBot.create(:v2_profile) }
+    let(:parent) { item.security_guide }
+    let(:extra_params) { { security_guide_id: parent.id, id: item.id } }
 
-    context 'Authorized' do
-      let(:rbac_allowed?) { true }
-
-      it 'returns profile by id' do
-        item = hash_including('data' => {
-                                'id' => profile.id,
-                                'type' => 'profile',
-                                **attributes.each_with_object({}) do |(key, value), obj|
-                                  obj[key.to_s] = profile.send(value)
-                                end
-                              })
-
-        get :show, params: { security_guide_id: security_guide.id, id: profile.id, parents: %i[security_guide] }
-
-        expect(response.parsed_body).to match(item)
-      end
-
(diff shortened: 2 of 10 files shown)


spec/controllers/v2/concerns/collection_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+#   let(:extra_params) { { security_guide_id: 123, profile_id: 456 } }
+# end
+# ```
+#
+# In a non-nested case this example can be simply included as:
+# ```
+# it_behaves_like 'collection'
+# ```
+#
+RSpec.shared_examples 'collection' do |*parents|
```
skateman, line 18: Look at how the `with metadata` deals with `extra_prams`

spec/controllers/v2/concerns/collection_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+          obj[key.to_s] = item.send(value)
+        end
+      )
+    end
+
+    if parents
+      get :index, params: { parents: parents }.merge(extra_params)
+    else
+      get :index
+    end
```
skateman, line 34: I think this logic can be omitted if you look how existing examples deal with default values.

spec/controllers/v2/concerns/collection_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+#   let(:extra_params) { { security_guide_id: 123, profile_id: 456 } }
+# end
+# ```
+#
+# In a non-nested case this example can be simply included as:
+# ```
+# it_behaves_like 'collection'
+# ```
+#
+RSpec.shared_examples 'collection' do |*parents|
```
skateman, line 18: Tha `parents` is no longer needed here and please adjust the documentation above.

spec/controllers/v2/concerns/index_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
```
skateman, line 1: Let's name this file `collection_spec`, the name of the example can stay as it is

spec/controllers/v2/concerns/show_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
```
skateman, line 1: WIth the same logic, this should be `resource_soec`

spec/controllers/v2/rules_controller_spec.rb
```diff
@@ -97,32 +80,21 @@
...
       context 'under incorrect parent security guide' do
-        let(:item) { FactoryBot.create(:v2_rule) }
         let(:security_guide) { FactoryBot.create(:v2_security_guide) }
 
         it 'returns not_found' do
-          get :show, params: { security_guide_id: security_guide.id, id: rule.id, parents: %i[security_guide] }
+          get :show, params: { security_guide_id: security_guide.id, id: item.id, parents: %i[security_guide] }
 
           expect(response).to have_http_status :not_found
         end
```
skateman, line 100: These seem to be almost the same to me, can we move it also to a shared example into the right (collection/resource) file? Please do the same thing with the rest, ideally we should not have any specs that are not shared, the whole thing was written in a shareable way...

skateman (changes requested):

spec/controllers/v2/profiles_controller_spec.rb
```diff
@@ -33,25 +33,8 @@
...
-
-        expect(response).to have_http_status :ok
-        expect(response_body_data).to match_array(collection)
-
-        response_body_data.each do |pf|
-          expect(pf['attributes'].keys.count).to eq(attributes.keys.count)
-        end
+      it_behaves_like 'index', :security_guide do
+        let(:extra_params) { { security_guide_id: security_guide.id } }
       end
```
skateman, line 38: If all these examples use the same `extra_params` in the same context, you can extract the `let` block to the same level to have it just once. This way you might be able to eliminate all the blocks on shared example calls ...

RoamingNoMaD: If I extract the `let` blocks to a level above, the parameter `parents:` is not passed to the examples correctly, causing all tests to fail. Maybe there is a way to do this correctly, but I do not think this is it.

spec/controllers/v2/concerns/collection_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    expect(response).to have_http_status :ok
+    expect(response_body_data).to match_array(collection)
+
+    response_body_data.each do |item|
+      expect(item.keys.count).to eq(attributes.keys.count + 2)
+    end
+  end
+
+  context 'under incorrect parent security guide', if: parents.present? do
```
skateman, line 42: I didn't even know this syntax existed 😁 not sure if this would look better, I'll think about it
```suggestion
  if parents.present?
    context 'under incorrect parent security guide' do
```

spec/controllers/v2/concerns/collection_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    let(:security_guide) { FactoryBot.create(:v2_security_guide) }
```
skateman, line 43: This is probably not okay, you don't want to make this bound to a security guide as other parents can nest as well. Maybe you could look up the first parent and find its factory and apply it in a generic way...

spec/controllers/v2/concerns/collection_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    it 'returns not_found' do
+      get :index, params: extra_params.merge(parents: parents)
+
+      expect(response_body_data).to be_empty
+    end
+  end
+
+  context 'unathorized' do
```
skateman, line 52:
```suggestion
  context 'Unathorized' do
```

spec/controllers/v2/value_definitions_controller_spec.rb
```diff
@@ -32,108 +33,33 @@
...
-            parents: %i[security_guide]
-          }
+    let(:item) { FactoryBot.create(:v2_value_definition) }
+    let(:security_guide) { item.security_guide }
 
-          expect(response).to have_http_status :not_found
-        end
-      end
+    it_behaves_like 'resource', :security_guide do
+      let(:extra_params) { { security_guide_id: security_guide.id, id: item.id } }
```
skateman, line 62: What happens if you extract this one a level above?

skateman (changes requested):

spec/controllers/v2/value_definitions_controller_spec.rb
RoamingNoMaD: I get an error complaining that no route matches my request.

RoamingNoMaD: When changing the definition order of `extra_params`, it is a viable solution. I've changed the comment in the shared example to reflect this.

skateman (approved):
The Seal of Approval

PR #1838, opened 2023-10-09
feat(APIv2): RHINENG-2198 introduce the derived_attribute serializer

skateman:
This change introduces a `fields` method on the serializer(s) that allow the user to determine what fields are required from a given model to construct the JSON serialized result returned by the controller. Furthermore, a new `derived_attribute` is introduced to the serializer's DSL for declaring fields that depend on other models. The logic deciding what fields should be shown has been implemented using Panko's [`.filter_for`](https://panko.dev/docs/attributes#filters-for) method. Finally, the rendering is adjusted to only select & alias fields that are really required for serialization. 

The new `derived_attribute` should be used to declare a field of an ActiveRecord object that is conditionally constructed by data from the parent (joined) objects if they are available. A good example of this is the `Rule#remediation_issue_id` which should be only available when the `Rule` objects are joined with both `Profile` and `SecurityGuide` objects, i.e. under `/security_guides/:id/profiles/:id/rules`.

```
derived_attribute :os_major_version, :has_remediation, profiles: [:ref_id], security_guide: [:ref_id]
```

The first argument is the name of the method, next is an optional array of dependent fields on the current model, then a hash of dependent `associations: [fields]`. The generated attribute is automatically delegated to a method with the same name defined in the model, so in the case of the example a `V2::Rule#remediation_issue_id` methods needs to exist.

Based on what is declared in the serializer, we can construct a list of fields to be queried, therefore, this also improves DB performance. The dependencies make sure that the field is only rendered in cases when the required auto-joined dependencies are available. 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/serializers/v2/rule_serializer.rb
+++ b/app/serializers/v2/rule_serializer.rb
@@ -4,5 +4,7 @@ module V2
   # JSON serialization for an OpenSCAP Rule
   class RuleSerializer < V2::ApplicationSerializer
     attributes :ref_id, :title, :rationale, :description, :severity, :precedence
+
+    derived_attribute :remediation_issue_id, :remediation_available, profiles: [:ref_id], security_guide: [:ref_id]
   end
 end
--- a/app/controllers/concerns/v2/rendering.rb
+++ b/app/controllers/concerns/v2/rendering.rb
@@ -30,7 +30,7 @@ module V2
       def serialize_individual(model, **args)
         Panko::Response.create do |r|
           {
-            data: r.serializer(model, serializer),
+            data: r.serializer(model, serializer, context: serialization_context),
             **args
           }
         end
@@ -38,10 +38,15 @@ module V2
 
       def serialize_collection(model, **args)
         Panko::Response.new(
-          data: Panko::ArraySerializer.new(model, each_serializer: serializer),
+          data: Panko::ArraySerializer.new(model, each_serializer: serializer, context: serialization_context),
           **metadata,
           **args
         )
+      end
+
+      # The serializer expects the list of parents to determine which `derived_attribute` should be skipped
+      def serialization_context
+        { parents: permitted_params[:parents] }
       end
 
       def index?
--- a/app/controllers/v2/application_controller.rb
+++ b/app/controllers/v2/application_controller.rb
@@ -61,9 +61,16 @@ module V2
       user.authorized_to?(Rbac::INVENTORY_HOSTS_READ) && user.authorized_to?(permission)
     end
 
+    def expand_resource
+      # Get the list of fields to be selected from the serializer
+      fields = serializer.fields(permitted_params)
+      # Select only the fields that are really necessary
+      join_parents.select(*select_fields(fields))
+    end
+
     # Reduce through all the parents of the resource and join+scope them on the resource
     # or return with the resource untouched if not nested under other resources
-    def expand_resource
+    def join_parents
       permitted_params[:parents].to_a.reduce(resource) do |scope, parent|
         ref = scope.reflect_on_association(parent)
         klass = ref.klass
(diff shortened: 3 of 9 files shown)


skateman:
@vkrizan I am not fully done with the feature (the implementation of the `remediation_issue_id` is still missing), but I am trying to stabilize the API of the DSL, so I'd like your input on it.

app/serializers/v2/rule_serializer.rb
```diff
@@ -4,5 +4,7 @@ module V2
   # JSON serialization for an OpenSCAP Rule
   class RuleSerializer < V2::ApplicationSerializer
     attributes :ref_id, :title, :rationale, :description, :severity, :precedence
+
+    complex_attribute :remediation_issue_id, profiles: [:ref_id], security_guide: [:ref_id]
```
vkrizan, line 8: What about naming it `dependent_attribute`?

vkrizan (commented):
I'd need a bit more time to go through it and understand the need to do it this way.

My concern is that we're creating something very custom, which adds complexity and future maintenance costs. If there would be a different way of resolving the problem that would be simpler, I'd choose that.

For example I'd like to know why this pushes the problem to the sterilizer. My other solution was to implement such computed fields (like `remediation_issue_id`) directly on the model. I think that model can query the scope to find out whether data that it requires to compute it are available. This seems a bit more simpler to me, as it does not crosses the boundary of a model (which is an ActiveRecord). However, it would need to be done for each computed field that would need to have performance problem improved.

app/serializers/v2/rule_serializer.rb
skateman: Longer 😁

skateman:
The only place we are concerned with performance is during serialization. The idea is that you can (and should be able to) call `remediation_issue_id` on the model itself without the dependencies available and then the model would do the querying for you. This is standard behavior in rails, what we are doing is making this fast when we ~~sterilize~~ serialize stuff.

vkrizan:
> The only place we are concerned with performance is during serialization.

Sure. The same is true to say about API responses. But I hope that this is not the reason why to fix it on the serializer.

> The idea is that you can (and should be able to) call remediation_issue_id on the model itself without the dependencies available and then the model would do the querying for you. This is standard behavior in rails, ...

The idea is that the model already knows about querying, hence it can know whether or not it has the required data, and stop itself from (auto)querying it to avoid the N+1 query problem.

skateman:
The model is not aware of this and I am not sure if it is a good idea to put this there, the query interface is separated from the models on purpose and its intended place to be called is in controllers. From this perspective the serializer is a helper service in the controller, where it makes more sense to hold such logic.

On the other hand, even if the model deals with this logic, the serializer still has to be aware of the absence of this field. Setting it to `nil` on the model and hiding it is not an option as in some cases we do want to show empty field values and it is not a good idea to tangle these two. Even though the serializer has some kind of special value that we can set if the value should be hidden, I don't like the idea of tangling the serializer code into the model as we might request this data in other places than the serializer.

app/models/application_record.rb
```diff
@@ -27,4 +27,10 @@ def self.taggable?
   def self.count_by
     primary_key.to_sym
   end
+
+  private
+
+  def attributes_available?(*attrs)
```
romanblanco, line 33: @skateman this method is not used

app/serializers/v2/rule_serializer.rb
romanblanco: I'd also lean towards `dependent_attribute`. It better describes the use.

skateman: What about `derived_`?

romanblanco: `derived_` works for me :+1:

vkrizan: :+1:

PR #1857, opened 2023-10-26
fix(APIv2): RHINENG-2720 return not_found when parent ID is invalid

RoamingNoMaD:
When requesting a nested `/index` with an existing but different root parent ID, the query returns an empty array instead of failing with a 404 error.

By joining the parents during parent validation and calling `find` we can trigger a not_found error when any of the parent hierarchy is invalid.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/controllers/v2/concerns/rendering_spec.rb
+++ b/spec/controllers/v2/concerns/rendering_spec.rb
@@ -40,7 +40,7 @@ RSpec.shared_examples 'collection' do |*parents|
     end
   end
 
-  context 'under incorrect parent', if: parents.present? do
+  context 'under a single incorrect parent', if: parents.present? do
     parents.each do |parent|
       context "#{parent} is incorrect" do
         it 'returns not_found' do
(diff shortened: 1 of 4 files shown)


skateman:
/retest
cc @LightOfHeaven1994

spec/controllers/v2/rules_controller_spec.rb
```diff
@@ -78,6 +78,10 @@
           profile_id: parent.id
         ).sort_by(&:id)
       end
+      let(:invalid_params) do
+        parent = FactoryBot.create(:v2_profile)
+        { security_guide_id: parent.security_guide.id, profile_id: parent.id }
```
skateman, line 83: Hm, I am wondering if the `security_guide_id` should be from the same hierarchy as the `profile`. I think it has a better coverage if just one of them is inconsistent.

RoamingNoMaD: Just for clarity: this test only exchanged one ID at a time. 
This will be changed in the next push.

spec/controllers/v2/concerns/rendering_spec.rb
```diff
@@ -54,6 +54,26 @@
     end
   end
 
+  context 'under invalid parent hierarchy', if: parents.count > 1 do
+    (1..parents.length - 1).each do |count|
+      parents.combination(count).each do |missing_parents|
```
romanblanco, line 59: @RoamingNoMaD would `invalid_parents` be better name?

```suggestion
      parents.combination(count).each do |invalid_parents|
```

skateman (approved):
The Seal of Approval

PR #1872, opened 2023-11-14
test(APIv2): remove AR relationships from extra_params in requests

skateman:
The `extra_params` in searchable/sortable concerns were used both in dynamically generated factories and in the params to be passed to the final testing request. However, there are cases when the factories need these entities, but the params do not. The simplest example could be the `account_id` that is never passed to any request as a param, it is always assumed from the context, however, it is essential for the creation of some records such as policies.

In order to keep both the controller and the factories happy, I am proposing to pass whole `ActiveRecord` objects in `extra_params` to be omitted from the request, but kept when creating a factory. Internally this is done by the `passable_params` variable that contains only the non-ActiveRecord objects, while the `extra_params` keeps them. These are passed to the request to be tested and the factories correspondingly in each shared example.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/controllers/v2/concerns/paginable_spec.rb
+++ b/spec/controllers/v2/concerns/paginable_spec.rb
@@ -18,13 +18,26 @@
 # it_behaves_like 'paginable'
 # ```
 #
+# In some cases, however, additional ActiveRecord objects are required for invoking a factory.
+# Therefore, if you don't want these objects to be passed to the `params` of the request, you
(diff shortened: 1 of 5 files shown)


romanblanco (approved):
LGTM :+1:

PR #1884, opened 2023-11-30
feat(APIv2): RHINENG-5354 implement the supported_profiles endpoint

skateman:
The idea here is to extract the complex `OsMinorVersions` query from the GQL implementation to a view, so it can be abstracted away from ActiveRecord. This might change in the future, but as long as we are in the remodelling process, we are fine like this.

The query joins `canonical_profiles` with `security_guides` and the with new `profile_os_minor_versions` table to group results by the `ref_id` and `os_major_version`. This results in a lot of redundant aggregated data, from which w need just the first entity for `id`,  `title` and `security_guide_version`, which are sorted all by the `security_guide_version` always selecting the oldest one. However, we do need all the `os_minor_versions` aggregated into an array of integers so we can construct the support matrix for the create policy page.

The rest of the code is just a classic controller with an index-only approach as we have other workflows to retrieve profiles.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/controllers/v2/supported_profiles_controller_spec.rb
+++ b/spec/controllers/v2/supported_profiles_controller_spec.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe V2::SupportedProfilesController do
+  let(:attributes) do
+    {
+      ref_id: :ref_id,
+      title: :title,
+      security_guide_version: :security_guide_version,
+      os_major_version: :os_major_version,
+      os_minor_versions: :os_minor_versions
+    }
+  end
+
+  let(:current_user) { FactoryBot.create(:v2_user) }
+  let(:rbac_allowed?) { true }
+
+  before do
+    request.headers['X-RH-IDENTITY'] = current_user.account.identity_header.raw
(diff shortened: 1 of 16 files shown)


RoamingNoMaD (approved):
lgtm!

spec/controllers/v2/supported_profiles_controller_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    it_behaves_like 'collection' do
+      let(:extra_params) { { limit: 20 } }
+    end
+    include_examples 'with metadata' do
+      let(:item_count) { 20 }
+    end
+    it_behaves_like 'paginable'
+    it_behaves_like 'sortable'
+    # it_behaves_like 'searchable'
```
romanblanco, line 82: @skateman Is this comment intended to stay?

skateman: Fixed it

romanblanco (approved):
LGTM :+1:

PR #1901, opened 2024-01-11
test(sorting): improve experience and deduplicate definitions

skateman:
From now on the `sortable` shared example in APIv2 specs does not return the mismatching database IDs of entities, but the indexes that we actually compare against in the YAML:
```
       expected: [3, 0, 1, 5, 4, 2]
            got: [0, 3, 1, 5, 4, 2]
```
It does not deal with the nested arrays, that has to be still figured out by the developer, but it definitely improves the experience when writing these YAML files.

Additionally, I dropped the explicit `:asc` and `:desc` tests that have been replaced with automatically generated tests. So from now on if you specify a sorting without a direction, both the explicit `:asc` and reverse `:desc` sorting will be automatically done without specifying it.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/controllers/v2/concerns/sortable_spec.rb
+++ b/spec/controllers/v2/concerns/sortable_spec.rb
@@ -66,18 +66,52 @@ RSpec.shared_examples 'sortable' do |*parents|
   end
 
   tests[:queries].each do |test_case|
-    it "sorts by #{test_case[:sort_by].join(', ')}" do
-      result = test_case[:result].flat_map do |item|
-        if item.is_a?(Array)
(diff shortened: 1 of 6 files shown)


romanblanco (approved):
LGTM :+1:

PR #1903, opened 2024-01-16
feat(APIv2): RHINENG-1986 tailoring rule model

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/v2/tailoring.rb
+++ b/app/models/v2/tailoring.rb
@@ -11,6 +11,8 @@ module V2
     belongs_to :profile, class_name: 'V2::Profile'
     has_one :security_guide, through: :profile, class_name: 'V2::SecurityGuide'
     has_one :account, through: :policy, class_name: 'V2::Account'
+    has_many :tailoring_rules, class_name: 'V2::TailoringRule', dependent: :destroy
+    has_many :rules, class_name: 'V2::Rule', through: :tailoring_rules
 
     validates :policy, presence: true
     validates :profile, presence: true
--- a/app/models/v2/tailoring_rule.rb
+++ b/app/models/v2/tailoring_rule.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module V2
+  # Model link between (Profile) Tailoring and Rule
+  class TailoringRule < ApplicationRecord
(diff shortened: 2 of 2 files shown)


skateman (approved):
The Seal of Approval
@jvasik please also take a look before merging

app/models/v2/tailoring.rb
```diff
@@ -11,6 +11,8 @@ class Tailoring < ApplicationRecord
     belongs_to :profile, class_name: 'V2::Profile'
     has_one :security_guide, through: :profile, class_name: 'V2::SecurityGuide'
     has_one :account, through: :policy, class_name: 'V2::Account'
+    has_many :tailoring_rules, class_name: 'V2::TailoringRule', dependent: :destroy
```
RoamingNoMaD, line 14: Do we need to use `dependent: :destroy` here? 
The callbacks in the case of a many-to-many helper table are not needed afaik. We're only deleting associations after all.
Also, `dependent: :delete_all` is supposed to be more performant being only one query, this could be faster in case of a tailoring with many rules.

Please correct me if I'm wrong.

RoamingNoMaD (changes requested):

app/models/v2/tailoring.rb
skateman: We should NEVER use `delete_all` as it does not run validations 🙈 the performance increase is exactly because of validations. The `dependent` is only added to make rubocop happy, but if we actually drop a tailoring, it should do destroy for sure.

RoamingNoMaD: Thanks for clarifying.

RoamingNoMaD (approved):
LGTM!

app/models/v2/tailoring.rb
skateman: We might want to change this later, but let's keep it consistent

PR #1968, opened 2024-03-07
refactor(specs): allow marking scalars to be skipped from URL params

skateman:
The `extra_params` in controller spec shared examples has a dual use:
* additional attributes for the factories
* additional parameters for the URL as `passable_params`

While the factories have no problems with consuming any kind of parameters, the `passable_params` it a bit more picky as routing often depend on what params are passed to a controller. Furthermore, sometimes we want to pass values to the factory that we don't want to see in the URL. As an initial solution, there was a simple logic implemented in the `passable_params` to omit any object that is the instance of `ActiveRecord::Base`, which allowed us to pass objects to the `extra_params` without polluting the URL. 

This solution isn't flexible enough for our needs and we already had to do some hacks when testing `V2::PoliciesController` and `V2::SystemsController`. So as a simplification, I am proposing a `pw` (parameter-wrapper) method that can mark scalar values such as integers or strings to be skipped from `passable_params` while consuming them for factories.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/controllers/v2/concerns/indexable_by_ref_id_spec.rb
+++ b/spec/controllers/v2/concerns/indexable_by_ref_id_spec.rb
@@ -7,7 +7,7 @@
 # ```
 # let(:extra_params) { { security_guide_id: 123, profile_id: 456, id: 789 } }
 #
-# it_behaves_like 'individual', :security_guide, :profile
+# it_behaves_like 'indexable by ref_id', :security_guide, :profile
 # ```
 #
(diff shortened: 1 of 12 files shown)


romanblanco (approved):
LGTM :+1:

PR #1978, opened 2024-03-13
fix(Tailoring): RHINENG-8782 do not restrict tailoring_file to cert_auth

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/concerns/v2/authentication.rb
+++ b/app/controllers/concerns/v2/authentication.rb
@@ -1,0 +1,96 @@
+# frozen_string_literal: true
+
+module V2
+  # Authentication logic for all controllers
+  module Authentication
+    extend ActiveSupport::Concern
+
+    ALLOWED_CERT_BASED_RBAC_ACTIONS = [
+      { controller: 'policies', action: 'index' },
+      { controller: 'tailorings', action: 'tailoring_file' }
+    ].freeze
+
+    included do
+      around_action :authenticate_user
+    end
+
+    def authenticate_user
+      if identity_header.blank?
+        return unauthenticated(error: 'Error parsing the X-RH-IDENTITY header')
+      end
+      return unauthenticated unless identity_header.valid?
+      return unauthorized unless rbac_allowed?
+
+      return if performed?
+
+      set_authenticated_user
+      yield
+    ensure
+      User.current = nil
+    end
+
+    def current_user
+      User.current
+    end
+
+    # :nocov:
+    def unauthorized(error: 'User is not authorized to view this page')
+      render(
+        json: { error: "Authorization error: #{error}" },
+        status: :forbidden
+      )
+      false
+    end
+    # :nocov:
+
+    # :nocov:
+    def unauthenticated(error: 'X-RH-IDENTITY header should be provided')
+      render(
+        json: { error: "Authentication error: #{error}" },
+        status: :unauthorized
+      )
+      false
+    end
+    # :nocov:
+
+    def user
+      @user ||= User.new(account: Account.from_identity_header(identity_header))
+    end
+
+    private
+
+    def set_authenticated_user
+      User.current = user
+    end
+
+    def valid_cert_endpoint?
+      ALLOWED_CERT_BASED_RBAC_ACTIONS.include?(
+        controller: controller_name, action: action_name
+      )
+    end
+
+    # :nocov:
+    def any_inventory_hosts?
+      Insights::Api::Common::HostInventory.new(b64_identity: raw_identity_header).hosts.dig('results').present?
+    end
+    # :nocov:
+
+    # :nocov:
+    def valid_cert_auth?
+      valid_cert_endpoint? && any_inventory_hosts?
+    rescue Faraday::Error => e
+      Rails.logger.error(e.full_message)
+
+      false
+    end
+    # :nocov:
+
+    def raw_identity_header
+      request.headers['X-RH-IDENTITY']
+    end
+
+    def identity_header
+      @identity_header ||= Insights::Api::Common::IdentityHeader.new(raw_identity_header)
+    end
+  end
+end
--- a/spec/controllers/v2/tailorings_controller_spec.rb
+++ b/spec/controllers/v2/tailorings_controller_spec.rb
@@ -70,179 +70,153 @@ describe V2::TailoringsController do
   end
 
   context '/policies/:id/tailorings/:id/tailoring_file' do
-    before do
-      header = JSON.parse(Base64.decode64(current_user.account.identity_header.raw))
-      header['identity']['auth_type'] = Insights::Api::Common::IdentityHeader::CERT_AUTH
-      request.headers['X-RH-IDENTITY'] = Base64.encode64(header.to_json)
-    end
-
-    let(:parent) do
-      FactoryBot.create(:v2_policy, account: current_user.account, profile: canonical_profile)
-    end
-
-    let(:extra_params) { { policy_id: parent.id, id: item.id } }
-
-    context 'with no rules and no values' do
-      let(:canonical_profile) do
+    RSpec.shared_examples 'tailoring_file' do
+      let(:parent) do
         FactoryBot.create(
-          :v2_profile,
-          rule_count: 5,
-          value_count: 5,
-          security_guide: FactoryBot.create(:v2_security_guide, rule_count: 3, os_major_version: 9),
-          os_major_version: 9,
-          ref_id_suffix: 'bar',
+          :v2_policy,
+          :for_tailoring,
+          account: current_user.account,
           supports_minors: [8]
         )
       end
 
-      let(:item) do
-        FactoryBot.create(:v2_tailoring, rules: [], value_overrides: [], policy: parent, os_minor_version: 8)
+      let(:extra_params) { { policy_id: parent.id, id: item.id } }
+
+      context 'with no tailored rules and no values' do
+        let(:item) do
+          FactoryBot.create(
+            :v2_tailoring,
+            rules: [], # no tailored rules
+            value_overrides: [], # no tailored values
+            policy: parent,
+            os_minor_version: 8
+          )
+        end
+
+        it 'returns XCCDF tailoring file with unselected rules' do
+          get :tailoring_file, params: extra_params.merge(parents: [:policy], format: :xml)
+
+          tailoring_file = Nokogiri::XML(response.body).remove_namespaces!
+          deselected, selected = tailoring_file.xpath('//Profile/select')
+                                               .partition { |sel| sel.attributes['selected'].value == 'false' }
+
+          expect(response).to have_http_status :ok
+          expect(deselected.map { |r| r.attributes['idref'].value }).to match_array(item.profile.rules.map(&:ref_id))
+          expect(selected).to match_array(item.rules.map(&:ref_id))
+        end
       end
 
-      it 'returns XCCDF tailoring file with unselected rules' do
-        get :tailoring_file, params: extra_params.merge(parents: [:policy], format: :xml)
+      context 'with default rules and values' do
+        let(:item) do
+          FactoryBot.create(
+            :v2_tailoring,
+            :with_default_rules, # no tailored rules
+            policy: parent,
+            os_minor_version: 8
+          )
+        end
 
-        tailoring_file = Nokogiri::XML(response.body).remove_namespaces!
-        deselected, selected = tailoring_file.xpath('//Profile/select')
-                                             .partition { |sel| sel.attributes['selected'].value == 'false' }
+        it 'returns empty response' do
+          get :tailoring_file, params: extra_params.merge(parents: [:policy], format: :xml)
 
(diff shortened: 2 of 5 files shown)


spec/factories/tailoring.rb
```diff
@@ -41,7 +41,7 @@
 
           result = sec_guide_rules + profile_rules
           # Make sure that the randomized result doesn't match the profile tailoring
-          break if result.to_set(&:id) != profile.rules.to_set(&:id)
+          break if result.to_set(&:id) != profile.rules.to_set(&:id) || profile.rules == []
```
skateman, line 44: This is a bit reversed logic, `with_mixed_rules` should fail if there are no rules to mix...

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -253,6 +247,32 @@
       end
     end
 
+    context 'with authorization only via cert_auth' do
+      before do
+        header = JSON.parse(Base64.decode64(current_user.account.identity_header.raw))
+        header['identity']['auth_type'] = Insights::Api::Common::IdentityHeader::CERT_AUTH
+        request.headers['X-RH-IDENTITY'] = Base64.encode64(header.to_json)
+      end
```
skateman, line 255: Simply setting up the context won't work, there's logic in the authentication concern ensuring a passthrough over RBAC is cert_auth is valid for some endpoints. All these endpoints are defined for V1, in V2 we have different ones. If his test is passing, there are bigger issues as well...

skateman (changes requested):

app/controllers/concerns/v2/authentication.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+module V2
+  # Authentication logic for all controllers
+  module Authentication
+    extend ActiveSupport::Concern
+
+    ALLOWED_CERT_BASED_RBAC_ACTIONS = [
+      { controller: 'tailorings', action: 'index' },
```
skateman, line 9: Why is the index needed?

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -253,6 +247,42 @@
       end
     end
 
+    context 'with authorization only via cert_auth' do
+      before do
+        allow_any_instance_of(V2::ApplicationController).to receive(:rbac_allowed?).and_call_original
+        allow_any_instance_of(Insights::Api::Common::HostInventory).to receive(:hosts).and_return('results' => [:foo])
```
skateman, line 253: In the first case you definitely not need the `_any_instance_of`, look at the mocking example from above. For the second one, you can extract the host inventory API lookup into a separate method and mock that. Calling a global mock is both dangerous and slow.

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -253,6 +247,42 @@
...
+        header = JSON.parse(Base64.decode64(current_user.account.identity_header.raw))
```
skateman, line 254: Maybe we can have a trait in the `v2_account` factory that creates an account with a CERT_AUTH identity header.

spec/factories/tailoring.rb
```diff
@@ -32,16 +32,18 @@
...
-          # sample of rules under profile
-          profile_rules = profile.rules.sample(profile.rules.count / 2)
-          # sample of rules under security guide, outside of the profile
-          sec_guide_rules_avail = security_guide.rules.reject { |rule| profile_rules.include?(rule) }
-          sec_guide_rules = sec_guide_rules_avail.sample(sec_guide_rules_avail.count / 2)
-
-          result = sec_guide_rules + profile_rules
-          # Make sure that the randomized result doesn't match the profile tailoring
-          break if result.to_set(&:id) != profile.rules.to_set(&:id)
+        if profile.rules.any?
```
skateman, line 35: Instead of trying to hack around the factory, maybe it could be called properly at its callsite.

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -253,6 +247,42 @@
...
+      end
+      let(:extra_params) { { policy_id: parent.id, id: item.id } }
+      let(:canonical_profile) { FactoryBot.create(:v2_profile, :with_rules) }
+      let(:item) do
+        FactoryBot.create(
+          :v2_tailoring,
+          :with_mixed_rules,
+          :with_tailored_values,
+          policy: parent,
+          profile: canonical_profile,
```
skateman, line 274: You definitely don't need to specify both, policy should be always enough as the profile can be looked up automatically if everything is valid.

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -253,6 +247,42 @@
...
+      before do
+        allow_any_instance_of(V2::ApplicationController).to receive(:rbac_allowed?).and_call_original
+        allow_any_instance_of(Insights::Api::Common::HostInventory).to receive(:hosts).and_return('results' => [:foo])
+        header = JSON.parse(Base64.decode64(current_user.account.identity_header.raw))
+        header['identity']['auth_type'] = Insights::Api::Common::IdentityHeader::CERT_AUTH
+        request.headers['X-RH-IDENTITY'] = Base64.encode64(header.to_json)
+      end
+
+      let(:parent) do
+        FactoryBot.create(
```
skateman, line 260: Consider using the `for_tailoring` trait, there are examples on how to use it already.

skateman (changes requested):
I think it should have tests both for with and without CERT_AUTH 🙏

spec/factories/tailoring.rb
romanblanco: Here, I'd actually prefer to keep the condition in the place, altough the callsite has been corrected.

In future, someone can still try to generate tailoring while having no rules under profile, and the condition is a clear hint for why the loop shouldn't be executed.

skateman: If the thing will not fail under the wrong circumstances, people will not see that what they are doing is wrong...

spec/controllers/v2/concerns/tailoring_file_spec.rb
```diff
@@ -0,0 +1,208 @@
+# frozen_string_literal: true
```
skateman, line 1: Tailoring file is not a concern, so this is semantically not correct. You can have a shared example inside a regular test. Then I think you can deduplicate a bunch of things, like you would not even need to specify the `parent`

spec/controllers/v2/concerns/tailoring_file_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+      FactoryBot.create(
+        :v2_profile,
+        rule_count: 5,
+        value_count: 5,
+        security_guide: FactoryBot.create(:v2_security_guide, rule_count: 3, os_major_version: 9),
+        os_major_version: 9,
+        ref_id_suffix: 'bar',
+        supports_minors: [8]
+      )
+    end
```
skateman, line 26: Seems like this is getting duplicated across multiple examples.

spec/controllers/v2/concerns/tailoring_file_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+
+    let(:item) do
+      FactoryBot.create(
```
skateman, line 29: Same here, you can maybe extract the changing params into `let` blocks and define just them across contexts.

spec/controllers/v2/concerns/tailoring_file_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+      expect(tailoring_values).to eq(tailored_values.map(&:value))
+      expect(rules_only).to match_array(item.rules_added.map(&:ref_id))
+      expect(rules_and_groups - rules_only).to match_array(item.rule_group_ref_ids)
+    end
+  end
+
+  context 'with unauthorized policy' do
+    let(:canonical_profile) do
+      FactoryBot.create(:v2_profile)
+    end
```
skateman, line 167: Short enough to be a one-liner

```suggestion
    let(:canonical_profile) { FactoryBot.create(:v2_profile) }
```

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -92,210 +92,19 @@
...
-      let(:extra_params) do
-        {
-          policy_id: FactoryBot.create(:v2_policy, account: FactoryBot.create(:v2_account)),
-          id: Faker::Internet.uuid
-        }
-      end
-
-      it 'results in 404 error' do
-        get :tailoring_file, params: extra_params.merge(parents: [:policy], format: :xml)
+    it_behaves_like 'tailoring_file'
```
skateman, line 104:
```suggestion
    include_examples 'tailoring_file'
```

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -92,210 +92,19 @@
...
+  end
 
-        expect(response).to have_http_status :not_found
-      end
-    end
+  context 'unauthorized via cert_auth' do
```
skateman, line 107: I think it should be semantically:
```suggestion
  context 'authorized via regular authentication' do
```
But probably you don't need this context explicitly and can just unwrap the include.

spec/factories/account.rb
```diff
@@ -17,5 +18,13 @@
         }.to_json)
       )
     end
+
+    transient do
+      auth_type { nil }
+    end
```
skateman, line 24: Short-enough to be a one-liner:
```suggestion
    transient { auth_type { nil } }
```

skateman (changes requested):
Please take a look at https://github.com/RedHatInsights/compliance-backend/pull/1990 for inspiration on how to simplify things regarding policies/profiles.

app/controllers/concerns/v2/authentication.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    def valid_cert_endpoint?
+      ALLOWED_CERT_BASED_RBAC_ACTIONS.include?(
+        controller: controller_name, action: action_name
+      )
+    end
+
+    def any_inventory_hosts?
+      Insights::Api::Common::HostInventory.new(b64_identity: raw_identity_header).hosts.dig('results').present?
```
skateman, line 69: This is an external service that we cannot test, so please wrap `# :nocov:` around it

app/controllers/concerns/v2/authentication.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    def valid_cert_auth?
+      valid_cert_endpoint? && any_inventory_hosts?
+    rescue Faraday::Error => e
+      Rails.logger.error(e.full_message)
+
+      false
```
skateman, line 77: I think this can also be `# :nocov:` as Faraday handles errors anyway.

app/controllers/concerns/v2/authentication.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    def unauthenticated(error: 'X-RH-IDENTITY header should be provided')
+      render(
+        json: { error: "Authentication error: #{error}" },
+        status: :unauthorized
+      )
+      false
+    end
```
skateman, line 50: This should probably be covered in `application_controller_spec`

app/controllers/concerns/v2/authentication.rb
```diff
@@ -0,0 +1,88 @@
...
+      { controller: 'tailorings', action: 'tailoring_file' }
+    ].freeze
+
+    included do
+      around_action :authenticate_user
+    end
+
+    def authenticate_user
+      if identity_header.blank?
+        return unauthenticated(error: 'Error parsing the X-RH-IDENTITY header')
```
skateman, line 19: This should be covered by `application_controller_spec`

spec/controllers/v2/tailorings_controller_spec.rb
```diff
@@ -70,179 +70,154 @@
...
-        }
+    context 'authorized via cert_auth' do
+      before do
+        allow(controller).to receive(:rbac_allowed?).and_call_original
+        allow(controller).to receive(:any_inventory_hosts?).and_return(true)
       end
 
-      it 'results in 404 error' do
-        get :tailoring_file, params: extra_params.merge(parents: [:policy], format: :xml)
+      let(:account) { FactoryBot.create(:v2_account, :cert_authorized) }
```
skateman, line 215: Do we need the account to be defined as a separate `let`? Can't we merge it into the block beneath?

skateman (changes requested):
Starting to look good

skateman (approved):
The Seal of Approval

PR #2023, opened 2024-03-27
feat(APIv2): implemented the rules#index under tailorings + cleanups

skateman:
The `index` endpoint under `/policies/:id/tailorings/:id/rules` was never implemented and it is
necessary for the frontend to function correctly. Therefore, I added it, which uncovered additional
problems in the `RulesController` tests.

The `validate_parents!` method was assuming that the `parents` on routes are incremental. However,
as they depend on the model relationships, it is not always possible to move upwards in the routing
table by removing the last item of the `parents` array. Furthermore, the `pundit_subquery` called
the reflection lookup on the wrong `relation` in the same method chain.

We had tests for various combinations of broken parent hierarchy, which were only running when testing
`rules` under `profiles`. This required additional dynamic parameters to be configured and its added
value was minimal, because we already had individual tests for each parent being broken. Therefore,
I decided to drop this additional complexity and get rid of a few unnecessary tests.

There were also some redundancies in factories for `rule` assignment under `tailorings` and the dynamic
`extra_params` was improperly used in the tests.  Finally, the openapi specs were in the tailorings spec instead of the rules + also used bloated factories so I cleaned that up as well.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


romanblanco (approved):
LGTM :+1:

PR #2050, opened 2024-04-19
perf(APIv2): RHINENG-8352 harmonize aliases when merging with pundit

skateman:
There was an issue with merging pundit scopes with autojoined parents as they automatically get aliased based on the reflection name. I created an `Arel::Visitor` implementing BFS that can be used for both retrieving aliases from one scope and adjusting them in another. 

There were some errors around adjusting an arel AST of a cached query, so I made sure that we can opt-in construct a new AST while visiting nodes. This is ensured by the `copy` argument of the visitor and the `dup` call on each node.

If a new query shows up with which the visitor cannot deal, an exception describing this problem is displayed that should be available when running tests.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/concerns/v2/merge_with_alias.rb
+++ b/app/models/concerns/v2/merge_with_alias.rb
@@ -1,0 +1,96 @@
+# frozen_string_literal: true
+
+module V2
+  # Extended merge method that can harmonize table aliases, mainly for pundit scoping
+  module MergeWithAlias
+    extend ActiveSupport::Concern
+
+    def merge_with_alias(other)
+      other = visitor_for_merge(aliases_for_merge).accept(other.where_clause.ast)
+      other.to_sql == '' ? self : where(other)
+    end
+
+    private
+
+    def aliases_for_merge
+      aliases = {}
+
+      # Look up all the defined table aliases in the current scope and feed them into the `aliases` hash
+      visitor = ArelVisitor.new do |node|
+        aliases[node.table_name] = node.name if node.is_a?(Arel::Nodes::TableAlias)
+      end
+
+      visitor.accept(where_clause.ast)
+      aliases
+    end
+
+    # Iterate over the `other` scope and harmonize its table aliases
+    def visitor_for_merge(aliases)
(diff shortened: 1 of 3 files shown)


app/models/concerns/v2/merge_with_alias.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      private
+
+      # :nocov:
+      def visit(node, *args)
+        node = node.dup if @copy
+        super
+        @block.call(node)
+      rescue TypeError => e
+        raise [e.message, 'You should implement an alias for the missing method'].join(': ')
```
vkrizan, line 52:
```
"... in the format of #{"visit_#{(node.class.name || "").gsub("::", "_")}"}"
```

vkrizan (commented):
Interesting...

Do you think you can add some (isolated) test cases, or do you think that those that exist would be suffice?

app/models/concerns/v2/merge_with_alias.rb
skateman: That's already in the exception message.

app/models/concerns/v2/merge_with_alias.rb
```diff
@@ -0,0 +1,92 @@
...
+        left = visit(node.left)
+        right = visit(node.right)
+
+        node.left = left if @copy
+        node.right = right if @copy
+      end
+
+      def nary(node)
+        children = node.children.map { |child| visit(child) }
+        node.instance_variable_set(:@children, children) if @copy
```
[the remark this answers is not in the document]
skateman: Well yes and no 😁

vkrizan (approved):

skateman:
@vkrizan the current test coverage should be sufficient, if this would not work, 80% of our tests would fail.

PR #2062, opened 2024-04-30
fix(APIv2): RHINENG-9808 optimize multiple aggregations within relation

skateman:
The current implementation of aggregations in the `V2::Resolver` supports one aggregation per reflection. When trying to define multiple ones, the resulting `[lambda, alias]` pairs get flattened with other pairs. Furthermore, the joining for a given reflection happens multiple times, once per aggregation, which is inefficient and would require aliasing. 

I had to adjust the `V2::ApplicationSerializer#merge_dependencies` to not flatten the `[lambda, alias]` pairs when doing the aggregation. The flattening still happens in any other case (joining parents and 1:1s) as it should. I also extracted the field evaluation/aliasing + selection in the `V2::Resolver#join_aggregated` method into a single step per reflection. 

Now that different aggregations overlap and use the same `JOIN` fragment, it is necessary to do any scoping of aggregations using `FILTER` and the `WHERE` clause for such purposes should be avoided.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/concerns/v2/resolver.rb
+++ b/app/controllers/concerns/v2/resolver.rb
@@ -38,30 +38,32 @@ module V2
     end
 
     # Self-join with the requested aggregations built using 1:n associations, also select
-    # the aggregated field from the self-joined subquery.
+    # the aggregated/evaluated/aliased fields from the self-joined subquery.
     def join_aggregated(relation)
-      aggregations.reduce(relation) do |scope, (association, (aggregation, column))|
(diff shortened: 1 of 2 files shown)


romanblanco (approved):
Tested, works.

LGTM :+1:

PR #2076, opened 2024-05-12
fix(APIv2): bulk deletion should delete link not the entity

skateman:
The bulk assignment was always deleting the right side from the database when unassigning. This means that when unassigning a system from a tailoring, we didn't delete the `TailoringSystem` record, but the `System` itself. Same happened for `TailoringRule` unassignment. Furthermore, due to the dynamic evaluation of the entities to be unassigned, the unassignment itself interfered with the assignment.

So I adjusted the unassignments to deal with the linking entities and reordered the deletion to happen before the assignment. This is something we aren't really able to test in our codebase though 😞 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/v2/rules_controller.rb
+++ b/app/controllers/v2/rules_controller.rb
@@ -14,10 +14,7 @@ module V2
     permission_for_action :show, Rbac::COMPLIANCE_VIEWER
 
     def create
-      inserts, deletes = V2::TailoringRule.bulk_assign(
-        new_tailoring_rules,
-        tailoring.rules.where.not(lookup_key => permitted_params[:ids])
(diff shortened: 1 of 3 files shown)


LightOfHeaven1994 (approved):
Tested in ephemeral, works great!
Ignore CI failures as there is some issue on qe api client auth side

PR #2079, opened 2024-05-14
fix(APIv2): RHINENG-10037 clone rules/values from a supported profile

skateman:
When assigning a system to a policy, the tailoring creation is triggered if there is no already tailoring created. In case there is a rule support mismatch between the policy's parent profile and the one that the system supports, the validation on the rule assignment fails. This caused issues when assigning systems to older policies with no longer supported parent profile. The fix is trivial, the rule cloning should always happen from the tailoring's parent and not the policy's.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/v2/policy_system.rb
+++ b/app/models/v2/policy_system.rb
@@ -19,8 +19,8 @@ module V2
         # Look up the latest Profile supporting the given OS minor version
         tailoring.profile = policy.profile.variant_for_minor(system.os_minor_version)
 
-        tailoring.rules = policy.profile.rules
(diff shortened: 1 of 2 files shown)


LightOfHeaven1994 (approved):
tested in ephemeral, lgtm

PR #2093, opened 2024-05-22
fix(APIv2): RHINENG-10041 do not join with parents multiple times

skateman:
When joining with multiple parents, sometimes it happens that due to a `through` association more than one parent is joined in a single step of the iteration. This can cause multiple joins with the same association and aliasing them with weird names. These joins might result in duplications, so I am adding a mechanism to prevent them.

We can retrieve a list of already joined associations using the new `already_joined` method that lists not just the directly joined associations coming from `joins_values` but the whole chain of `through` models. Then we can decide to join or not based on this computed list.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/concerns/v2/resolver.rb
+++ b/app/controllers/concerns/v2/resolver.rb
@@ -21,11 +21,12 @@ module V2
     def join_parents(relation, associations)
       associations.to_a.reduce(relation) do |scope, association|
         ref = scope.reflect_on_association(association)
-        klass = ref.klass
 
-        scope.joins(association)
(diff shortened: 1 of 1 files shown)


LightOfHeaven1994 (approved):
LGTM in ephemeral, so far test that checks all profiles haven't found any issues

PR #2145, opened 2024-06-21
feat(APIv2): RHINENG-1986 branch out historical test results

skateman:
The `V2::TestResult` model now returns only what `::TestResult.latest` is returning and the rest of test results are available under `V2::HistoricalTestResults`. For now we don't have multiple test results for a single tailoring/system combo, but the codebase always allowed it. 

This should allow us in the future to compare historical test results in a similar fashion as Drift does with system profiles.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_06_07_103908) do
+ActiveRecord::Schema[7.1].define(version: 2024_06_21_081235) do
   create_schema "inventory"
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -408,6 +408,26 @@
...
+      test_results.supported,
+      test_results.failed_rule_count,
+      test_results.created_at,
+      test_results.updated_at
+     FROM ((test_results
+       JOIN profiles ON ((profiles.id = test_results.profile_id)))
+       JOIN ( SELECT test_results_1.profile_id,
+              test_results_1.host_id,
+              max(test_results_1.end_time) AS end_time
+             FROM test_results test_results_1
```
vkrizan, line 428: add `AS` in-between to make the syntax consistent.

db/schema.rb
```diff
@@ -408,6 +408,26 @@
...
+            GROUP BY test_results_1.profile_id, test_results_1.host_id) tr ON (((test_results.profile_id = tr.profile_id) AND (test_results.host_id = tr.host_id) AND (test_results.end_time = tr.end_time))));
```
vkrizan, line 429: The same here for the `tr`.

Why not join on primary key here?

skateman: This is generated by `pg_dump`, look at the SQL files lower.

PR #2150, opened 2024-06-27
feat(APIv2): RHINENG-283 allow the creation of sibling policies

skateman:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/v2/systems_controller.rb
+++ b/app/controllers/v2/systems_controller.rb
@@ -75,8 +75,10 @@ module V2
         major = policy.os_major_version
         minors = policy.os_minor_versions
         # Filter the passed systems based on what OS versions the policy supports
+        # and also drop any system that is assigned to a sibling policy
         items = pundit_scope.where(id: permitted_params[:ids])
                             .os_major_versions(major).os_minor_versions(minors)
+                            .where.not(id: systems_with_sibling_policies)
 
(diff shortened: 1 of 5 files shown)


app/controllers/v2/systems_controller.rb
```diff
@@ -94,6 +96,13 @@ def build_tailorings!
       end
     end
 
+    def systems_with_sibling_policies
+      V2::PolicySystem.joins(policy: :profile)
+                      .where(profile: { ref_id: policy.ref_id }, system_id: permitted_params[:ids])
+                      .where.not(policy_id: policy.id)
+                      .select(:system_id)
```
vkrizan, line 103: Technically this can return more than it is within in pundit scope if the right combination of ref_id and system_id is provided. At this use-case it is not an issue, but I'd harden it nevertheless (unless it is already).

skateman: Nice catch, but for that I would have to join with the systems endpoint and that makes the querying even more expensive. This is only being used in an already scoped-down environment, so would you be okay with just adding a comment above it?

vkrizan (approved):

PR #2159, opened 2024-07-09
feat(APIv2): RHINENG-10187 sorting reports by compliant percentage

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/v2/report.rb
+++ b/app/models/v2/report.rb
@@ -7,10 +7,6 @@ module V2
     self.table_name = :v2_policies
     self.primary_key = :id
 
-    def percent_compliant
-      0
-    end
-
     SYSTEM_COUNT = lambda do
       AN::NamedFunction.new('COUNT', [V2::System.arel_table[:id]]).filter(
         Pundit.policy_scope(User.current, V2::System).where_clause.ast
@@ -33,6 +29,44 @@ module V2
       )
     end
 
+    # rubocop:disable Metrics/BlockLength
+    PERCENT_COMPLIANT = lambda do
+      AN::NamedFunction.new(
+        'CAST',
+        [
+          AN::NamedFunction.new(
+            'FLOOR',
+            [
+              AN::Multiplication.new(
+                AN::NamedFunction.new(
+                  'COALESCE',
+                  [
+                    AN::Division.new(
+                      AN::NamedFunction.new(
+                        'CAST',
+                        [
+                          COMPLIANT_SYSTEM_COUNT.call.as('FLOAT')
+                        ]
+                      ),
+                      AN::NamedFunction.new(
+                        'NULLIF',
+                        [
+                          AN::NamedFunction.new('CAST', [SYSTEM_COUNT.call.as('FLOAT')]),
+                          Arel.sql('0')
+                        ]
+                      )
+                    ), Arel.sql('0')
+                  ]
+                ),
+                Arel.sql('100')
+              )
+            ]
+          ).as('INTEGER')
+        ]
+      )
+    end
+    # rubocop:enable Metrics/BlockLength
+
     # To prevent an autojoin with itself, there should not be an inverse relationship specified
     belongs_to :policy, class_name: 'V2::Policy', foreign_key: :id # rubocop:disable Rails/InverseOf
     belongs_to :account
@@ -44,12 +78,27 @@ module V2
     has_many :report_systems, class_name: 'V2::ReportSystem', dependent: nil # rubocop:disable Rails/InverseOf
     has_many :systems, class_name: 'V2::System', through: :report_systems
     has_many :reported_systems, class_name: 'V2::System', through: :test_results, source: :system, dependent: nil
+    has_many :reporting_and_non_reporting_systems, lambda {
+      # joining TestResult and System to correctly count the system with content to report
+      system_test_results = arel_table.join(V2::System.arel_table, AN::InnerJoin)
+                                      .on(V2::System.arel_table[:id].eq(V2::ReportSystem.arel_table[:system_id]))
+                                      .join(V2::TestResult.arel_table, AN::OuterJoin)
+                                      .on(V2::TestResult.arel_table[:system_id].eq(V2::System.arel_table[:id]))
+                                      .join_sources
+
+      # aggregation to prevent optimizer (possibly buggy) filtering out join
+      agg = V2::TestResult.arel_table[:id]
+                          .not_eq(nil)
+                          .or(V2::TestResult.arel_table[:id].eq(nil))
+
+      joins(system_test_results).where(agg)
+    }, class_name: 'V2::ReportSystem', dependent: nil, inverse_of: false
 
     sortable_by :title
     sortable_by :os_major_version
     sortable_by :business_objective
     sortable_by :compliance_threshold
-    sortable_by :compliance_percentage, Arel.sql('0')
+    sortable_by :percent_compliant, 'aggregate_percent_compliant'
 
     searchable_by :title, %i[like unlike eq ne in notin]
     searchable_by :os_major_version, %i[eq ne in notin] do |_key, op, val|
--- a/spec/controllers/v2/reports_controller_spec.rb
+++ b/spec/controllers/v2/reports_controller_spec.rb
@@ -13,10 +13,10 @@ describe V2::ReportsController do
       business_objective: :business_objective,
       all_systems_exposed: -> { true },
       compliance_threshold: :compliance_threshold,
-      percent_compliant: :percent_compliant,
-      assigned_system_count: -> { 0 },
-      reported_system_count: -> { 0 },
-      compliant_system_count: -> { 0 },
+      percent_compliant: -> { 50 },
+      assigned_system_count: -> { 2 },
+      reported_system_count: -> { 2 },
+      compliant_system_count: -> { 1 },
       unsupported_system_count: -> { 0 }
     }
   end
@@ -44,7 +44,9 @@ describe V2::ReportsController do
       let(:items) do
         FactoryBot.create_list(
           :v2_report, item_count,
-          assigned_system_count: 0,
+          assigned_system_count: 2,
+          compliant_system_count: 1,
+          unsupported_system_count: 0,
           os_major_version: 8,
           supports_minors: [0, 1],
           account: current_user.account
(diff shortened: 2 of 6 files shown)


app/models/v2/report.rb
```diff
@@ -7,6 +7,19 @@ class Report < ApplicationRecord
...
     self.primary_key = :id
 
+    def self.percent_compliant
+      compliant = Arel.sql('aggregate_compliant_system_count::float')
+      supported = Arel.sql('(aggregate_assigned_system_count - aggregate_unsupported_system_count)::float')
+
+      Arel::Nodes::Division.new(compliant, supported).as('percentage')
+    end
+
+    REPORTED_SYSTEM_COUNT = lambda do
```
romanblanco, line 17: not usefull

app/models/v2/report.rb
```diff
@@ -45,6 +58,7 @@ class Report < ApplicationRecord
     sortable_by :os_major_version
     sortable_by :business_objective
     sortable_by :compliance_threshold
+    sortable_by :percentage, percent_compliant
```
romanblanco, line 61:
```suggestion
    sortable_by :percentage, 'aggregate_percent_compliant'
```

app/policies/v2/report_policy.rb
```diff
@@ -24,7 +24,7 @@ class Scope < V2::ApplicationPolicy::Scope
       def resolve
         return scope.where('1=0') if user&.account_id.blank?
 
-        scope.where(account_id: user.account_id)
+        scope.where(account_id: user.account_id).select(scope.percent_compliant)
```
romanblanco, line 27: don't change this here

app/models/v2/report.rb
```diff
@@ -44,12 +86,22 @@ def percent_compliant
...
+      )
+    }, class_name: 'V2::ReportSystem', dependent: nil
+    # rubocop:enable Rails/InverseOf
 
     sortable_by :title
     sortable_by :os_major_version
     sortable_by :business_objective
     sortable_by :compliance_threshold
-    sortable_by :compliance_percentage, Arel.sql('0')
+    sortable_by :compliance_percentage, 'aggregate_percent_compliant'
```
skateman, line 104: Maybe we should make these consistent? The sort key should match the field name returned by the API.

app/models/v2/report.rb
```diff
@@ -44,12 +86,22 @@ def percent_compliant
     has_many :report_systems, class_name: 'V2::ReportSystem', dependent: nil # rubocop:disable Rails/InverseOf
     has_many :systems, class_name: 'V2::System', through: :report_systems
     has_many :reported_systems, class_name: 'V2::System', through: :test_results, source: :system, dependent: nil
+    # rubocop:disable Rails/InverseOf
+    has_many :reporting_and_non_reporting_systems, lambda {
+      joins(
+        "INNER JOIN inventory.hosts ON inventory.hosts.id = report_systems.system_id \
+          LEFT OUTER JOIN v2_test_results ON v2_test_results.system_id = report_systems.system_id"
```
skateman, line 93: This should be done in arel.

app/models/v2/report.rb
```diff
@@ -44,12 +86,22 @@ def percent_compliant
...
+      ).where(
+        Pundit.policy_scope(User.current, V2::System).where_clause.ast.to_sql
```
skateman, line 95: Is this really necessary? Aren't you already doing system scoping on the aggregation level?

skateman (changes requested):
We'd need some extra test coverage that the aggregations don't leak data from other inventory groups and users to make sure that the scoping I mentioned is safe to remove.

app/models/v2/report.rb
romanblanco: It does seem like. Investigating...

skateman: I would need a bit more details here ... because you are scoping down systems to users but then in the aggregation there is additional scoping happening.

app/models/v2/report.rb
```diff
@@ -44,12 +86,25 @@ def percent_compliant
...
+      joins(
+        arel_table.join(V2::System.arel_table, AN::InnerJoin)
+                  .on(V2::System.arel_table[:id].eq(V2::ReportSystem.arel_table[:system_id]))
+                  .join(V2::TestResult.arel_table, AN::OuterJoin)
+                  .on(V2::TestResult.arel_table[:system_id].eq(V2::System.arel_table[:id]))
+                  .join_sources
+      ).where(
+        Pundit.policy_scope(User.current, V2::System).where_clause.ast.to_sql
+      )
+    }, class_name: 'V2::ReportSystem', dependent: nil
```
skateman, line 100: Why is this pointing to the `ReportSystem` model? What are we exactly joining here? Not the `System` and the `TestResult` model? We should pick one of those, but the semantically correct is the `System` as not all records will have a `TestResult`.

app/models/v2/report.rb
```diff
@@ -44,12 +86,25 @@ def percent_compliant
     has_many :report_systems, class_name: 'V2::ReportSystem', dependent: nil # rubocop:disable Rails/InverseOf
     has_many :systems, class_name: 'V2::System', through: :report_systems
     has_many :reported_systems, class_name: 'V2::System', through: :test_results, source: :system, dependent: nil
+    # rubocop:disable Rails/InverseOf
+    has_many :reporting_and_non_reporting_systems, lambda {
+      joins(
+        arel_table.join(V2::System.arel_table, AN::InnerJoin)
+                  .on(V2::System.arel_table[:id].eq(V2::ReportSystem.arel_table[:system_id]))
```
skateman, line 93: If you pick the `class_name` correctly (see below) then you don't need to specify the join with the systems, it will be assumed automatically.

app/models/v2/report.rb
```diff
@@ -44,12 +86,25 @@ def percent_compliant
...
+                  .join(V2::TestResult.arel_table, AN::OuterJoin)
+                  .on(V2::TestResult.arel_table[:system_id].eq(V2::System.arel_table[:id]))
+                  .join_sources
+      ).where(
+        Pundit.policy_scope(User.current, V2::System).where_clause.ast.to_sql
```
skateman, line 98: If you don't utilize a `where` clause over the explicitly joined `TestResult` model, it will be omitted by ActiveRecord. In my opinion this is a bug and will take a look at how this could be fixed in the future. This policy scoping is unnecessary, you just need ANY `where` condition involving the `TestResult` model to make this work instead. 

Be careful that you can't use `where.not(test_results: { id: nil })` as it could render the `LEFT OUTER JOIN` useless, instead you can create a complex arel-style (OR is hard in pure ActiveRecord) condition that allows both `NULL` AND `NOT NULL` for the field.

skateman (changes requested):

skateman (changes requested):

spec/controllers/v2/reports_controller_spec.rb
```diff
@@ -240,6 +262,64 @@
...
+
+            # change groups for some of the non-compliant systems
+            item
+            non_compliant_systems = V2::TestResult.all.map do |tr|
+              tr.system if tr.score.present? &&
+                           tr.score < (V2::Report.find(tr.report_id).compliance_threshold)
+            end.compact
+            non_compliant_systems.first(2).each do |system|
+              system.update!(groups: [Faker::Internet.uuid])
+            end
```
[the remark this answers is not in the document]
romanblanco: Changing inventory group is only applied to some of the systems filtered in the first iteration.

spec/controllers/v2/reports_controller_spec.rb
```diff
@@ -240,6 +262,64 @@
...
+                attribute_filter: {
+                  key: 'group.id',
+                  operation: 'in',
+                  value: [nil] # access to ungrouped hosts
+                }
+              }]
+            )
+
+            # change groups for some of the non-compliant systems
+            item
```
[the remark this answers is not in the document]
skateman: `let!`?

spec/controllers/v2/reports_controller_spec.rb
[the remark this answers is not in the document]
skateman: Still, can't you do a single iteration with a `reject`?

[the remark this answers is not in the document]
skateman: Also going through all test results is not safe in a test, you should ideally use a factory remnant.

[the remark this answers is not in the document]
romanblanco: Updated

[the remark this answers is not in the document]
skateman: > Also going through all test results is not safe in a test, you should ideally use a factory remnant.

This is still valid

spec/controllers/v2/reports_controller_spec.rb
```diff
@@ -240,6 +262,65 @@
...
+                  key: 'group.id',
+                  operation: 'in',
+                  value: [nil] # access to ungrouped hosts
+                }
+              }]
+            )
+
+            # change groups for some of the non-compliant systems
+            non_compliant_systems = []
+            item.test_results.each do |tr|
```
skateman, line 310: You can call a `select` on the `test_results` and call `take(2)` on it instead.

spec/controllers/v2/reports_controller_spec.rb
```diff
@@ -240,6 +262,65 @@
...
+              non_compliant_system = tr.score.present? && tr.score < V2::Report.find(tr.report_id).compliance_threshold
```
skateman, line 311: `tr.report` doesn't work?

skateman: Also you can play with `try(:<, threshold)` to do safe comparison

spec/controllers/v2/reports_controller_spec.rb
skateman:
```
item.test_results.select { condition }.take(2).map { |tr| tr.update() }
```

skateman (approved):
The Seal of Approval

PR #2190, opened 2024-08-26
feat(APIv2): RHINENG-11773 filter failed rules by severity

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/v2/test_result.rb
+++ b/app/models/v2/test_result.rb
@@ -70,6 +70,15 @@ module V2
       { conditions: systems.arel.where_sql.gsub(/^where /i, '') }
     end
 
+    searchable_by :failed_rule_severity, %i[eq in] do |_key, _op, val|
+      values = val.split
+      ids = ::V2::RuleResult.joins(:rule)
+                            .where(v2_rules: { severity: values }, rule_results: { result: 'fail' })
+                            .select(:test_result_id)
+
+      { conditions: "v2_test_results.id IN (#{ids.to_sql})" }
+    end
+
     scope :with_groups, lambda { |groups, table = arel_table, key = :id|
       # Skip the [] representing ungrouped hosts from the array when generating the query
       grouped = V2::System.arel_inventory_groups(groups.flatten, key, table)
@@ -77,8 +86,6 @@ module V2
       # The OR is inside of Arel in order to prevent pollution of already applied scopes
       where(groups.include?([]) ? grouped.or(ungrouped) : grouped)
     }
-
-    # TODO: searchable_by :failed_rule_severity
 
     def display_name
       attributes['system__display_name'] || try(:system)&.display_name
--- a/spec/factories/test_result.rb
+++ b/spec/factories/test_result.rb
@@ -34,7 +34,23 @@ FactoryBot.define do
       display_name { nil }
       os_major_version { nil }
       os_minor_version { nil }
+      additional_rule_results { [] }
       groups { nil }
+    end
+
+    after(:create) do |tr, ev|
+      ev.additional_rule_results.each do |rr|
+        FactoryBot.create(
+          :v2_rule_result,
+          rule: FactoryBot.create(
+            :v2_rule,
+            security_guide: tr.tailoring.security_guide
+          ),
+          test_result_id: tr.id,
+          severity: rr[:severity],
+          result: rr[:result]
+        )
+      end
     end
   end
 end
--- a/spec/fixtures/files/searchable/test_results_controller.yaml
+++ b/spec/fixtures/files/searchable/test_results_controller.yaml
@@ -294,3 +294,82 @@
         :account: ${account}
         :policy_id: ${policy.id}
   :query: (group_name ^ "one,two")
+
+- :name: "in search by failed_rule_severity"
+  :entities:
+    :found:
+      - :factory: :v2_test_result
+        :account: ${account}
+        :policy_id: ${policy.id}
+        :additional_rule_results:
+          - :severity: 'high'
+            :result: 'fail'
+          - :severity: 'low'
+            :result: 'fail'
+          - :severity: 'medium'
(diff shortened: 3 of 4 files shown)


spec/controllers/v2/test_results_controller_spec.rb
```diff
@@ -72,6 +73,14 @@
         end.sort_by(&:id)
       end
 
+      let(:rule_result_data) do
+        %w(low medium high).each_with_object({}) do |level, obj|
+          obj["#{level}_severity_failing_rule".to_sym] = pw({
+            Faker::Internet.uuid => [level, 'fail']
```
skateman: AFAIK you don't need the left side, so you can get rid of these objects completely. SImply define everything in the YAML.

skateman (changes requested):
We can get to something like this:
```suggestion
        :additional_rule_results:
          - severity: 'high'
             result: 'pass'
          - severity: 'high'
             result: 'fail'
          - severity: 'low'
             result: 'fail'
```

app/models/v2/test_result.rb
```diff
@@ -70,6 +70,15 @@ class TestResult < ApplicationRecord
       { conditions: systems.arel.where_sql.gsub(/^where /i, '') }
     end
 
+    searchable_by :failed_rule_severity, %i[in notin] do |_key, op, val|
+      values = val.split(',').map(&:strip)
```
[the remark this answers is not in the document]
romanblanco: We do need to separate the values here.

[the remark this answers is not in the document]
skateman: Yes, but do we need a `,` or isn't space enough? The rule of thumb is if there's space in the strings that you're searching for (see other examples).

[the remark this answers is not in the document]
romanblanco: There is no space in the string to split by.

app/models/v2/test_result.rb
```diff
@@ -70,6 +70,15 @@ class TestResult < ApplicationRecord
       { conditions: systems.arel.where_sql.gsub(/^where /i, '') }
     end
 
+    searchable_by :failed_rule_severity, %i[in] do |_key, _op, val|
+      values = val.split(',').map(&:strip)
+      ids = ::V2::TestResult.joins(rule_results: :rule)
+                            .where(v2_rules: { severity: values }, rule_results: { result: 'fail' })
+                            .reselect(:id)
+
+      { conditions: "v2_test_results.id IN (#{ids.to_sql})" }
```
skateman, line 79: I know why you're bumping into the select cache: you don't need to select from the test results table at all. You can directly go to rule_results and join it with rules + select by `test_result_id`. That way you're joining one less table and you won't need the `reselect`

spec/fixtures/files/searchable/test_results_controller.yaml
```diff
@@ -294,3 +294,33 @@
...
+
+- :name: "in search by failed_rule_severity"
+  :entities:
+    :found:
+      - :factory: :v2_test_result
+        :account: ${account}
+        :policy_id: ${policy.id}
+        :additional_rule_results:
+          - severity: 'high'
+            result: 'fail'
```
skateman, line 306: Please add more than just a single rule result here + combine them with passing tests with higher severity to properly cover the DB lookup.

app/models/v2/test_result.rb
```diff
@@ -70,6 +70,15 @@ class TestResult < ApplicationRecord
       { conditions: systems.arel.where_sql.gsub(/^where /i, '') }
     end
 
+    searchable_by :failed_rule_severity, %i[in] do |_key, _op, val|
```
skateman, line 73: If you do `in` then you'd also have to do `eq` for semantic compatibility.

app/models/v2/test_result.rb
```diff
@@ -70,6 +70,15 @@ class TestResult < ApplicationRecord
       { conditions: systems.arel.where_sql.gsub(/^where /i, '') }
     end
 
+    searchable_by :failed_rule_severity, %i[eq in] do |_key, op, val|
+      values = ['IN'].include?(op) ? val.split : val
```
skateman, line 74: You don't need to deal with the operator at all, the `split` returns a single element array when no separator is detected. So you can always call `split`, see other existing examples.

skateman (changes requested):
One minor issue.

spec/fixtures/files/searchable/test_results_controller.yaml
```diff
@@ -294,3 +294,82 @@
...
+        :policy_id: ${policy.id}
+        :additional_rule_results:
+          - severity: 'low'
+            result: 'fail'
+      - :factory: :v2_test_result
+        :account: ${account}
+        :policy_id: ${policy.id}
+        :additional_rule_results:
+          - severity: 'high'
+            result: 'pass'
```
skateman, line 333: Nit: use symbols

spec/factories/test_result.rb
```diff
@@ -34,7 +34,23 @@
...
+      ev.additional_rule_results.each do |rr|
+        FactoryBot.create(
+          :v2_rule_result,
+          rule: FactoryBot.create(
+            :v2_rule,
+            security_guide: tr.tailoring.security_guide
+          ),
+          test_result_id: tr.id,
+          severity: rr['severity'],
+          result: rr['result']
```
skateman, line 51: Nit: use symbols

skateman (commented):
One last thing as you need to rebase anyway

skateman (approved):
The Seal of Approval

PR #2230, opened 2024-09-30
feat(RBAC): RHINENG-13334 allow reading SSG with any read permission

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/controllers/v2/security_guides_controller.rb
```diff
@@ -6,23 +6,23 @@ class SecurityGuidesController < ApplicationController
     def index
       render_json security_guides
     end
-    permission_for_action :index, Rbac::COMPLIANCE_VIEWER
```
skateman, line 9: What about reusing this constant instead? I think that should lead to a cleaner solution, semantically as well.

romanblanco: Good point.

skateman:
Is there a way to make sure it doesn't break stuff in v1/GQL?

romanblanco:
I can add another constant specific for APIv1 with the current permissions, so that the `COMPLIANCE_VIEWER` is only used in APIv2.

Will think of cleaner solution.

spec/controllers/v2/security_guides_controller_spec.rb
```diff
@@ -32,6 +32,12 @@
     it_behaves_like 'paginable'
     it_behaves_like 'sortable'
     it_behaves_like 'searchable'
+
+    context 'with any read permission' do
+      before { stub_rbac_permissions(Rbac::REPORT_READ) }
+
+      it_behaves_like 'collection'
+    end
```
skateman, line 40: Is there any point in having this test?

romanblanco: It can be removed

skateman: IMO for testing RBAC we either need separate tests independent from controllers, but I think it's better to keep as it is as it's fully declarative and I'd rather rely on QE to test declarative stuff.

PR #2232, opened 2024-10-02
fix(APIv2): RHINENG-13402 enable multidimensional sort in OpenAPI spec

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/spec/swagger_helper.rb
+++ b/spec/swagger_helper.rb
@@ -158,7 +158,7 @@ def sort_params(model = nil)
 end
 
 def sort_params_v2(model = nil, except: [])
-  parameter name: :sort_by, in: :query, required: false,
+  parameter name: :sort_by, in: :query, required: false, style: :form, explode: true,
             description: 'Attribute and direction to sort the items by. ' \
               'Represented by an array of fields with an optional direction ' \
               '(`<key>:asc` or `<key>:desc`).<br><br>' \
--- a/swagger/v2/openapi.json
+++ b/swagger/v2/openapi.json
@@ -64,6 +64,8 @@
             "name": "sort_by",
             "in": "query",
             "required": false,
+            "style": "form",
+            "explode": true,
             "description": "Attribute and direction to sort the items by. Represented by an array of fields with an optional direction (`<key>:asc` or `<key>:desc`).<br><br>If no direction is selected, `<key>:asc` is used by default.",
             "schema": {
               "type": "array",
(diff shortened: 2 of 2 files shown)


spec/swagger_helper.rb
```diff
@@ -158,7 +158,7 @@ def sort_params(model = nil)
 end
 
 def sort_params_v2(model = nil, except: [])
-  parameter name: :sort_by, in: :query, required: false,
+  parameter name: :sort_by, in: :query, required: false, style: :form, explode: true,
```
skateman, line 161: Why is the `style: :form` needed?

romanblanco: Following rswag documentation: https://github.com/rswag/rswag/wiki/FAQ#for-openapi-3 

https://swagger.io/docs/specification/v3_0/serialization/#query-parameters

skateman: Okay, still don't understand why is this needed.

romanblanco:
```
form – (default) ampersand-separated values, also known as form-style query expansion. Corresponds to the {?param_name} URI template.
```

I believe this is the option we want to use

skateman: If it's default, why do we need this now? The format checks out, we only need the `explode` as far as I understand.

romanblanco: Technically, we should not need it, as it is supposed to be the default option.

I suggest the change to explicitly state the expansion options to rule out that the problem is occuring in the generated OpenAPI spec.

skateman: And why just here be explicit? If it's a default value, it should be consistent across all parameters in the whole specification. So either change it everywhere to `style: form` or please delete this.

romanblanco: We want this to be applied on those parameters of `type: :array`. Where else do you see this missing?

skateman: If this is default (which is), it means it is applied to everything, not just the `type: :array` so I am asking again, why is this case so special that you want to have an explicit setting that is already implicitly set on it? Why and how is this one single case different from all the other parameters that may or may not have a `type: :array` which is not connected to the formatting that you are trying to set now explicitly?

romanblanco: The referenced ticket answers your question.

skateman: No it doesn't, you have 2 independent attributes that you are for some reason try to bind into a single context. One of them is default, so its setting is irrelevant and to maintain consistency across the OpenAPI spec, it makes no sense to explicitly specify in a single location.

PR #2449, opened 2025-05-20
revert(Kafka): return Karafka back in stage

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/consumers/application_consumer.rb
```diff
@@ -1,13 +1,33 @@
...
+                "#{messages.metadata.first_offset}-#{messages.metadata.last_offset}"
+    messages.each do |message|
+      @message = message
+
+      if retrying?
+        logger.debug 'Retrying message'
+
+        if attempt > 3
+          logger.error 'Discarded message'
+          mark_as_consumed(message)
```
RoamingNoMaD, line 19: Note: `the #mark_as_consumed method is designed to be asynchronous, meaning it doesn't immediately commit the offset but schedules it to be committed later`

As [stated in the Karafka FAQ](https://github.com/karafka/karafka/wiki/FAQ#does-running-mark_as_consumed-increase-the-processing-time)

By default, commits happen each 5s (in Racecar this was 10s)

karafka.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    # Wait for at least 1 seconds after an error
+    config.pause_timeout = 1_000
+
+    config.consumer_persistence = !Rails.env.development?
+  end
+
+  Karafka.monitor.subscribe(
+    Karafka::Instrumentation::LoggerListener.new(
+      log_polling: true
```
RoamingNoMaD, line 42: This is extra verbosity we probably don't need. We have our own INFO level logging in `ApplicationConsumer.rb#consume` that essentially does the same, but is more tailored to our needs.

RoamingNoMaD (changes requested):
Other important things to note:

1. Karafka consumes in batches of 100 - Racecar did in batches of 1000.
2. Karafka implements concurrent processing (5 threads by deafult), Racecar did not. This might make a difference in CPU/memory usage and we might have to adjust our pod replicas.
3. Karafka by default uses `exponential_backoff`, Racecar did not. 
4. To mitigate infinite wait time caused by the previous point, the `max_timeout` is by default 30s, but we won't reach it as after 3 retries we call `mark_as_consumed` and stop trying. (after ~3-4s)

We could potentially enhance the way we access the message payloads with [default JSON de-serialization](https://karafka.io/docs/Deserialization/#default-deserializers). So instead of using a method to do this, we can just tell Karafka to call the de-serializer for us.

app/consumers/compliance_consumer.rb
```diff
@@ -0,0 +1,36 @@
...
+    else
+      logger.debug "Skipped message of type '#{message_type}'"
+    end
+  end
+  # rubocop:enable Metrics/AbcSize
+
+  private
+
+  def payload
+    JSON.parse(@message.raw_payload)
```
RoamingNoMaD, line 22: We could potentially enhance the way we access the message payloads with [default JSON de-serialization](https://karafka.io/docs/Deserialization/#default-deserializers). So instead of using a method to do this, we can just tell Karafka to call the de-serializer for us.

karafka.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+# Karafka configuration
+class KarafkaApp < Karafka::App
+  CLIENT_ID = 'compliance-backend'
```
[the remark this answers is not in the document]
romanblanco: Seems that we were using both.

`compliance-backend` in [producer](https://github.com/RedHatInsights/compliance-backend/blob/e41eea8759c00a23449f67a2e0e912b94a0cbfa0/app/producers/application_producer.rb#L9) and `compliance_backend` in [consumer](https://github.com/RedHatInsights/compliance-backend/blob/e41eea8759c00a23449f67a2e0e912b94a0cbfa0/config/racecar.yml#L3).

Let's keep using the consuming id as suggested for both.

app/producers/application_producer.rb
```diff
@@ -1,13 +1,13 @@
...
-require 'rdkafka'
+require 'karafka'
 
 # Common Kafka producer client
 class ApplicationProducer
   BROKERS = Settings.kafka.brokers.split(',').freeze
   EXCEPTIONS = [Rdkafka::RdkafkaError, Rdkafka::AbstractHandle::WaitTimeoutError].freeze
-  CLIENT_ID = 'compliance-backend'
   SERVICE = 'compliance'
+  CLIENT_ID = 'compliance-backend'
```
romanblanco, line 10: in `karafka.rb` we're setting `CLIENT_ID` as `compliance_backend`

karafka.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+# Karafka configuration
+class KarafkaApp < Karafka::App
+  CLIENT_ID = 'compliance_backend'
+
+  # librdkafka config creation
+  security_protocol = Settings.kafka.security_protocol.downcase
+
+  sasl_config = if security_protocol == 'sasl_ssl'
```
romanblanco, line 10: This could be extracted into a method for better readability

app/producers/report_validation.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# A Kafka producer client for upload-compliance
+class ReportValidation < ApplicationProducer
+  TOPIC = Settings.kafka.topics.upload_compliance
+  DATE_FORMAT = :rfc3339
```
romanblanco, line 6: The `rfc3339` was only set to fix PayloadTracker and might need to be removed from here to return to the default ISO 8601.

See https://github.com/RedHatInsights/compliance-backend/commit/43082eea6b82e08e39ed408bfac4726caec2bbb1

app/services/kafka/report_parser.rb
```diff
@@ -0,0 +1,144 @@
...
+
+    def initialize(message, logger)
+      @message = message
+      @logger = logger
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def parse_reports
+      # Map successfuly parsed (validated) reports by scanned profile
+      parsed_reports = validated_reports.map do |xml|
```
romanblanco, line 19:
```suggestion
      parsed_reports = downloaded_reports.map do |xml|
```

spec/producers/payload_tracker_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe PayloadTracker do
+  let(:request_id) { '1' }
```
romanblanco, line 6:
```suggestion
  let(:request_id) { Faker.random_int }
```

This could be something else than a hard-coded id, unless we really need to specify it in the unit test

RoamingNoMaD (approved):

PR #2460, opened 2025-05-28
fix(Rules): RHINENG-16918 search rules by identifier

LightOfHeaven1994:
### What does it do?
This PR gets back filtering rules/rule_results by identifier as it was before. See https://issues.redhat.com/browse/RHINENG-14846 or https://issues.redhat.com/browse/RHINENG-16918

### How to test?

1. Check if any rules endpoint and rule_results endpoint can be filtered by:
-  `identifier_label=CEE-...`
- `identifier_label !=CEE-...`
- `identifier_label ~ 'CEE-...'`
-  `identifier_label !~ 'CEE-...`

So we have separate search attribute specifically for identifier.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/v2/rule_result.rb
+++ b/app/models/v2/rule_result.rb
@@ -2,6 +2,7 @@
 
 module V2
   # Class representing individual rule results unter a test result
+  # rubocop:disable Metrics/ClassLength
   class RuleResult < ApplicationRecord
     # FIXME: clean up after the remodel
     self.table_name = :rule_results
@@ -76,6 +77,15 @@ module V2
       }
     end
 
+    searchable_by :identifier_label, %i[eq neq like unlike] do |_key, op, val|
+      val = "%#{val}%" if ['ILIKE', 'NOT ILIKE'].include?(op)
+
+      {
+        conditions: "rule.identifier->>'label' #{op} ?",
+        parameter: [val]
+      }
+    end
+
     def ref_id
       attributes['rule__ref_id'] || try(:rule)&.ref_id
     end
@@ -135,4 +145,5 @@ module V2
     end
     # :nocov:
   end
+  # rubocop:enable Metrics/ClassLength
 end
--- a/app/models/v2/rule.rb
+++ b/app/models/v2/rule.rb
@@ -50,6 +50,14 @@ module V2
     searchable_by :severity, %i[eq ne in notin]
     searchable_by :remediation_available, %i[eq]
     searchable_by :rule_group_id, %i[eq]
+    searchable_by :identifier_label, %i[eq neq like unlike] do |_key, op, val|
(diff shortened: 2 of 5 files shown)


app/models/v2/rule_result.rb
```diff
@@ -43,14 +43,32 @@ class RuleResult < ApplicationRecord
 
     searchable_by :result, %i[eq ne in notin]
 
+    # Searching rule results by title should also consider rule result identifier label as it was before
     searchable_by :title, %i[eq neq like unlike] do |_key, op, val|
       val = "%#{val}%" if ['ILIKE', 'NOT ILIKE'].include?(op)
-      bind = ['IN', 'NOT IN'].include?(op) ? '(?)' : '?'
```
LightOfHeaven1994, line 48: I believe this has been added here by an accident. I don't remember any place where you were able to filter rules by multiple titles. API doc also doesn't mention it. I don't remember any FE place that uses this either so removed it.

LightOfHeaven1994:
We can leave searching by indentifier inside `title` or move it out and have it separately like `identifier_label` like:
```
    searchable_by :identifier_label, %i[eq, like] do |_key, op, val|
      val = "%#{val}%" if op == 'ILIKE'

      {
        conditions: "v2_rules.identifier->>'label' #{op} ?",
        parameter: [val]
      }
    end
```

RoamingNoMaD (changes requested):
The changes LGTM, but could you please get the commits cleaned up a bit? 
Squash the ones containing changes to the same files and get rid of the merge commit at the end.

LightOfHeaven1994:
@RoamingNoMaD done, thanks!

app/models/v2/rule_result.rb
```diff
@@ -76,6 +77,15 @@ class RuleResult < ApplicationRecord
       }
     end
 
+    searchable_by :identifier_label, %i[eq neq like unlike] do |_key, op, val|
+      val = "%#{val}%" if ['ILIKE', 'NOT ILIKE'].include?(op)
+
+      {
+        conditions: "rule.identifier->>'label' #{op} ?",
```
RoamingNoMaD, line 84: Didn't catch this yesterday, shouldn't this use v2?

RoamingNoMaD (changes requested):

app/models/v2/rule_result.rb
LightOfHeaven1994: @RoamingNoMaD hm, maybe. I see other filters do the same - using `rule`

PR #2708, opened 2026-04-29
refactor(DB): remodel DB

RoamingNoMaD:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/concerns/profile_searching.rb
+++ b/app/models/concerns/profile_searching.rb
@@ -76,20 +76,12 @@ module ProfileSearching
     scope :in_policy, lambda { |policy_or_profile_id|
       return none unless ::UUID.validate(policy_or_profile_id)
 
-      policy_cond = { policy_id: policy_or_profile_id }
-      profile_cond = {
-        policy: {
-          profiles_policies: {
-            id: policy_or_profile_id
-          }
-        }
-      }
+      parent_policy_id = where(id: policy_or_profile_id).select(:policy_id)
 
-      search = left_outer_joins(policy: :profiles)
-      search.where(id: policy_or_profile_id)
-            .or(search.where(policy_cond))
-            .or(search.where(profile_cond))
-            .distinct
+      where(id: policy_or_profile_id)
+        .or(where(policy_id: policy_or_profile_id))
+        .or(where(policy_id: parent_policy_id))
+        .distinct
     }
   end
 
--- a/app/models/v2/rule_result.rb
+++ b/app/models/v2/rule_result.rb
@@ -5,8 +5,7 @@ module V2
   # rubocop:disable Metrics/ClassLength
   class RuleResult < ApplicationRecord
     # FIXME: clean up after the remodel
-    self.table_name = :rule_results
-    self.ignored_columns += %w[account host_id]
+    self.table_name = :rule_results_v2
 
     belongs_to :test_result, class_name: 'V2::TestResult'
     belongs_to :historical_test_result, class_name: 'V2::HistoricalTestResult', foreign_key: :test_result_id,
--- a/db/functions/v1_policies_insert_v01.sql
+++ b/db/functions/v1_policies_insert_v01.sql
@@ -1,0 +1,33 @@
+CREATE OR REPLACE FUNCTION v1_policies_insert() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+DECLARE result_id uuid;
+DECLARE bo_title varchar;
+BEGIN
+    SELECT "business_objectives"."title" INTO bo_title
+    FROM "business_objectives"
+    WHERE "business_objectives"."id" = NEW."business_objective_id";
+
+    INSERT INTO "policies_v2" (
+      "title",
+      "description",
+      "compliance_threshold",
+      "business_objective",
+      "profile_id",
+      "account_id",
+      "created_at",
+      "updated_at"
+    ) VALUES (
+      NEW."name",
+      NEW."description",
+      COALESCE(NEW."compliance_threshold", 100.0),
+      bo_title,
+      NEW."profile_id",
+      NEW."account_id",
+      NOW(),
+      NOW()
+    ) RETURNING "id" INTO "result_id";
+
+    NEW."id" := "result_id";
+    RETURN NEW;
+END
+$func$;
--- a/db/functions/v1_profile_rules_update_v03.sql
+++ b/db/functions/v1_profile_rules_update_v03.sql
@@ -1,0 +1,27 @@
+CREATE OR REPLACE FUNCTION v1_profile_rules_update() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+DECLARE
+    is_tailoring boolean;
+BEGIN
+    SELECT EXISTS(
+        SELECT 1 FROM "tailorings_v2"
+        WHERE "id" = NEW."profile_id"
+    ) INTO is_tailoring;
+
+    IF is_tailoring THEN
+        UPDATE "tailoring_rules_v2" SET
+          "tailoring_id" = NEW."profile_id",
+          "rule_id" = NEW."rule_id",
+          "updated_at" = COALESCE(NEW."updated_at", NOW())
+        WHERE "id" = OLD."id";
+    ELSE
+        UPDATE "profile_rules_v2" SET
+          "profile_id" = NEW."profile_id",
+          "rule_id" = NEW."rule_id",
+          "updated_at" = COALESCE(NEW."updated_at", NOW())
+        WHERE "id" = OLD."id";
+    END IF;
+
+    RETURN NEW;
+END
+$func$;
--- a/db/functions/v1_profiles_delete_v02.sql
+++ b/db/functions/v1_profiles_delete_v02.sql
@@ -1,0 +1,10 @@
+CREATE OR REPLACE FUNCTION v1_profiles_delete() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+BEGIN
+    IF OLD."parent_profile_id" IS NOT NULL THEN
+        DELETE FROM "tailorings_v2" WHERE "id" = OLD."id";
+    END IF;
+
+    RETURN OLD;
+END
+$func$;
--- a/db/functions/v1_profiles_update_v02.sql
+++ b/db/functions/v1_profiles_update_v02.sql
@@ -1,0 +1,26 @@
+CREATE OR REPLACE FUNCTION v1_profiles_update() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+BEGIN
+    IF OLD."parent_profile_id" IS NULL THEN
+        UPDATE "canonical_profiles_v2" SET
+          "title" = NEW."name",
+          "ref_id" = NEW."ref_id",
+          "description" = NEW."description",
+          "security_guide_id" = NEW."benchmark_id",
+          "upstream" = COALESCE(NEW."upstream", FALSE),
+          "value_overrides" = COALESCE(NEW."value_overrides", '{}'),
+          "updated_at" = COALESCE(NEW."updated_at", NOW())
+        WHERE "id" = OLD."id";
+    ELSE
+        UPDATE "tailorings_v2" SET
+          "policy_id" = NEW."policy_id",
+          "profile_id" = NEW."parent_profile_id",
+          "value_overrides" = COALESCE(NEW."value_overrides", '{}'),
+          "os_minor_version" = CAST(NULLIF(NEW."os_minor_version", '') AS int),
+          "updated_at" = COALESCE(NEW."updated_at", NOW())
+        WHERE "id" = OLD."id";
+    END IF;
+
+    RETURN NEW;
+END
+$func$;
--- a/db/functions/v1_test_results_insert_v01.sql
+++ b/db/functions/v1_test_results_insert_v01.sql
@@ -1,0 +1,37 @@
+CREATE OR REPLACE FUNCTION v1_test_results_insert() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+DECLARE result_id uuid;
+DECLARE v_report_id uuid;
+BEGIN
+    SELECT "tailorings_v2"."policy_id" INTO v_report_id
+    FROM "tailorings_v2"
+    WHERE "tailorings_v2"."id" = NEW."profile_id";
+
+    INSERT INTO "historical_test_results_v2" (
+      "tailoring_id",
+      "report_id",
+      "system_id",
+      "start_time",
+      "end_time",
+      "score",
+      "supported",
+      "failed_rule_count",
+      "created_at",
+      "updated_at"
+    ) VALUES (
+      NEW."profile_id",
+      v_report_id,
+      NEW."host_id",
+      NEW."start_time",
+      NEW."end_time",
+      NEW."score",
+      COALESCE(NEW."supported", TRUE),
+      COALESCE(NEW."failed_rule_count", 0),
+      COALESCE(NEW."created_at", NOW()),
+      COALESCE(NEW."updated_at", NOW())
+    ) RETURNING "id" INTO "result_id";
+
+    NEW."id" := "result_id";
+    RETURN NEW;
+END
+$func$;
--- a/db/functions/v1_test_results_insert_v01.sql
+++ b/db/functions/v1_test_results_insert_v01.sql
@@ -1,0 +1,38 @@
+CREATE OR REPLACE FUNCTION v1_test_results_insert() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+DECLARE
+    result_id uuid;
+    v_report_id uuid;
+BEGIN
+    SELECT "tailorings_v2"."policy_id" INTO v_report_id
+    FROM "tailorings_v2"
+    WHERE "tailorings_v2"."id" = NEW."profile_id";
+
+    INSERT INTO "historical_test_results_v2" (
+      "tailoring_id",
+      "report_id",
+      "system_id",
+      "start_time",
+      "end_time",
+      "score",
+      "supported",
+      "failed_rule_count",
+      "created_at",
+      "updated_at"
+    ) VALUES (
+      NEW."profile_id",
+      v_report_id,
+      NEW."host_id",
+      NEW."start_time",
+      NEW."end_time",
+      NEW."score",
+      COALESCE(NEW."supported", TRUE),
+      COALESCE(NEW."failed_rule_count", 0),
+      COALESCE(NEW."created_at", NOW()),
+      COALESCE(NEW."updated_at", NOW())
+    ) RETURNING "id" INTO "result_id";
+
+    NEW."id" := "result_id";
+    RETURN NEW;
+END
+$func$;
--- a/db/functions/v2_test_results_delete_v02.sql
+++ b/db/functions/v2_test_results_delete_v02.sql
@@ -1,0 +1,7 @@
+CREATE OR REPLACE FUNCTION v2_test_results_delete() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+BEGIN
+    DELETE FROM "historical_test_results_v2" WHERE "id" = OLD."id";
+    RETURN OLD;
+END
+$func$;
--- a/db/migrate/20260528151023_create_policies_v2_table.rb
+++ b/db/migrate/20260528151023_create_policies_v2_table.rb
@@ -1,0 +1,36 @@
+class CreatePoliciesV2Table < ActiveRecord::Migration[8.0]
+  def up
+    create_table :policies_v2, id: :uuid do |t|
+      t.string :title, null: true
+      t.string :description
+      t.float :compliance_threshold, default: 100.0, null: true
+      t.string :business_objective
+      t.references :profile, type: :uuid, index: false, null: true
+      t.references :account, type: :uuid, index: false, null: true
+
+      t.timestamps precision: nil, null: true
+    end
+
+    execute <<-SQL
+      INSERT INTO policies_v2 (id, title, description, compliance_threshold, business_objective,
+        profile_id, account_id, created_at, updated_at)
+      SELECT id, title, description, compliance_threshold, business_objective,
+        profile_id, account_id, NOW(), NOW()
+      FROM v2_policies;
+    SQL
+
+    change_column_null :policies_v2, :title, false
+    change_column_null :policies_v2, :compliance_threshold, false
+    change_column_null :policies_v2, :profile_id, false
+    change_column_null :policies_v2, :account_id, false
+    change_column_null :policies_v2, :created_at, false
+    change_column_null :policies_v2, :updated_at, false
+
+    add_index :policies_v2, [:account_id], name: 'index_policies_v2_on_account_id'
+    add_index :policies_v2, [:profile_id], name: 'index_policies_v2_on_profile_id'
+  end
+
+  def down
+    drop_table :policies_v2
+  end
+end
--- a/db/migrate/20260528151153_create_tailorings_v2_table.rb
+++ b/db/migrate/20260528151153_create_tailorings_v2_table.rb
@@ -1,0 +1,30 @@
+class CreateTailoringsV2Table < ActiveRecord::Migration[8.0]
+  def up
+    create_table :tailorings_v2, id: :uuid do |t|
+      t.integer :os_minor_version, null: true
+      t.jsonb :value_overrides, default: {}
+      t.references :policy, type: :uuid, index: false, null: true
+      t.references :profile, type: :uuid, index: false, null: true
+
+      t.timestamps precision: nil, null: true
+    end
+
+    execute <<-SQL
+      INSERT INTO tailorings_v2 (id, policy_id, profile_id, value_overrides, os_minor_version, created_at, updated_at)
+      SELECT id, policy_id, profile_id, value_overrides, os_minor_version, created_at, updated_at
+      FROM tailorings;
+    SQL
+
+    change_column_null :tailorings_v2, :policy_id, false
+    change_column_null :tailorings_v2, :profile_id, false
+    change_column_null :tailorings_v2, :created_at, false
+    change_column_null :tailorings_v2, :updated_at, false
+
+    add_index :tailorings_v2, [:policy_id], name: 'index_tailorings_v2_on_policy_id'
+    add_index :tailorings_v2, [:profile_id], name: 'index_tailorings_v2_on_profile_id'
+  end
+
+  def down
+    drop_table :tailorings_v2
+  end
+end
--- a/db/migrate/20260528151307_create_rule_results_v2_table.rb
+++ b/db/migrate/20260528151307_create_rule_results_v2_table.rb
@@ -1,0 +1,31 @@
+class CreateRuleResultsV2Table < ActiveRecord::Migration[8.0]
+  def up
+    create_table :rule_results_v2, id: :uuid do |t|
+      t.string :result
+      t.references :rule, type: :uuid, index: false, null: true
+      t.references :test_result, type: :uuid, index: false, null: true
+
+      t.timestamps precision: nil, null: true
+    end
+
+    execute <<-SQL
+      INSERT INTO rule_results_v2 (id, result, rule_id, test_result_id, created_at, updated_at)
+      SELECT id, result, rule_id, test_result_id, created_at, updated_at
+      FROM rule_results;
+    SQL
+
+    change_column_null :rule_results_v2, :rule_id, false
+    change_column_null :rule_results_v2, :test_result_id, false
+    change_column_null :rule_results_v2, :created_at, false
+    change_column_null :rule_results_v2, :updated_at, false
+
+    add_index :rule_results_v2, [:rule_id], name: 'index_rule_results_v2_on_rule_id'
+    add_index :rule_results_v2, [:test_result_id], name: 'index_rule_results_v2_on_test_result_id'
+    add_index :rule_results_v2, [:test_result_id, :rule_id], unique: true, name: 'index_rule_results_v2_on_test_result_id_and_rule_id'
+    add_index :rule_results_v2, [:test_result_id, :result], name: 'index_rule_results_v2_on_test_result_id_and_result'
+  end
+
+  def down
+    drop_table :rule_results_v2
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_04_29_180435) do
+ActiveRecord::Schema[8.1].define(version: 2026_05_03_165833) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "dblink"
   enable_extension "pgcrypto"
@@ -489,10 +489,10 @@ ActiveRecord::Schema[8.1].define(version: 2026_04_29_180435) do
      FROM policy_hosts;
   SQL
   create_view "report_systems", sql_definition: <<-SQL
-      SELECT id,
-      policy_id AS report_id,
-      host_id AS system_id
-     FROM policy_hosts;
+      SELECT policy_systems_v2.id,
+      policy_systems_v2.policy_id AS report_id,
+      policy_systems_v2.system_id AS system_id
+     FROM policy_systems_v2;
   SQL
   create_view "security_guides", sql_definition: <<-SQL
       SELECT benchmarks.id,
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_04_28_150507) do
+ActiveRecord::Schema[8.1].define(version: 2026_04_29_103626) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "dblink"
   enable_extension "pgcrypto"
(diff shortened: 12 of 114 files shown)


romanblanco (changes requested):
`test_results_v2` table figures in queries, but is not being created.

Looks like a mistake in renaming.

RoamingNoMaD:
> `test_results_v2` table figures in queries, but is not being created.
> 
> Looks like a mistake in renaming.

Right, forgot to change the references in our FloorPlans...

RoamingNoMaD:
I've reverted some extra constraints I added because our DB state is not ready for this to happen.
In the future I'd like to be in a more stable and constrained situation tho.

db/functions/v1_test_results_insert_v01.sql
```diff
@@ -0,0 +1,37 @@
+CREATE OR REPLACE FUNCTION v1_test_results_insert() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+DECLARE result_id uuid;
+DECLARE v_report_id uuid;
```
romanblanco: Why is there the `v_` prefix?

romanblanco: Ah, I guess it's `virtual`, because it's actually policy.

app/models/v2/rule_result.rb
```diff
@@ -5,8 +5,7 @@ module V2
   # rubocop:disable Metrics/ClassLength
   class RuleResult < ApplicationRecord
     # FIXME: clean up after the remodel
-    self.table_name = :rule_results
-    self.ignored_columns += %w[account host_id]
```
romanblanco, line 9: This should be obsolete info by now.

romanblanco: It's a dead code. The field is not present in the model anymore.

db/views/v1_profiles_v02.sql
```diff
@@ -0,0 +1,39 @@
...
+  "canonical_profiles_v2"."title" AS "name",
+  "canonical_profiles_v2"."ref_id",
+  "canonical_profiles_v2"."created_at",
+  "canonical_profiles_v2"."updated_at",
+  "canonical_profiles_v2"."description",
+  NULL::uuid AS "account_id",
+  "canonical_profiles_v2"."security_guide_id" AS "benchmark_id",
+  NULL::uuid AS "parent_profile_id",
+  FALSE AS "external",
+  NULL::uuid AS "policy_id",
```
romanblanco, line 12: For checking that report parsing still works correctly, this would be great to verify before we merge it.

(trigger profile cloning mechanisms, etc.)

deploy/clowdapp.yaml
```diff
@@ -553,31 +553,32 @@ objects:
...
-          "policies"."id",
-          "profiles"."ref_id",
-          "profiles"."name",
+          "policies_v2"."id",
+          "canonical_profiles_v2"."ref_id",
+          "canonical_profiles_v2"."title" AS "name",
           "accounts"."org_id",
-          REGEXP_REPLACE("benchmarks"."ref_id", '.*RHEL-(\d)$', '\1') AS "os_major_version"
-          FROM "policies"
-          INNER JOIN "profiles" ON "profiles"."policy_id" = "policies"."id"
```
romanblanco:
```
ON "profiles"."policy_id" = "policies"."id"
```

This should probably use `tailoring_v2` instead of `canonical_profiles_v2`.
Please, review.

deploy/clowdapp.yaml
```diff
@@ -644,13 +645,13 @@ objects:
...
-        FROM "v2_policies"
-        INNER JOIN "policy_systems" ON "policy_systems"."policy_id" = "v2_policies"."id"
-        INNER JOIN "v2_test_results" ON "v2_test_results"."system_id" = "policy_systems"."system_id"
-        GROUP BY "v2_policies"."account_id", "policy_systems"."system_id";
+          "policies_v2"."account_id" AS "org_id",
+          "policy_systems_v2"."system_id"::TEXT AS "system_id",
+          MAX("historical_test_results_v2"."updated_at") AS "updated_at"
+        FROM "policies_v2"
+        INNER JOIN "policy_systems_v2" ON "policy_systems_v2"."policy_id" = "policies_v2"."id"
+        INNER JOIN "historical_test_results_v2" ON "historical_test_results_v2"."system_id" = "policy_systems_v2"."system_id"
```
romanblanco, line 656: `historical_test_results_v2` is used here instead of `v2_test_results`, so we prepare the DB for the opportunity of expanding amount of collected test results over the time.

If we'll wish to preserve `historical_test_results` in the future, this may produce different kind of results.

spec/factories/test_result.rb
```diff
@@ -13,14 +13,15 @@
...
-        policy_id: policy_id,
+        policy_id: report_id,
         os_major_version: os_major_version,
         os_minor_version: os_minor_version,
         groups: groups
       }.compact
       association(:system, account: account, **args)
     end
 
+    report_id { nil }
```
romanblanco, line 24: It's been moved here correctly, but `test_result` can not have `report_id` as `nil`.

Shouldn't this have some factory generated?

romanblanco: It may be that we only need to decide based on `report_id` and so it's not needed for the testing of `test_result` itself.

app/models/concerns/profile_searching.rb
```diff
@@ -76,27 +76,19 @@ module ProfileSearching
     scope :in_policy, lambda { |policy_or_profile_id|
       return none unless ::UUID.validate(policy_or_profile_id)
 
-      policy_cond = { policy_id: policy_or_profile_id }
-      profile_cond = {
-        policy: {
-          profiles_policies: {
```
romanblanco, line 82: Rails magic 🪄

app/models/concerns/profile_searching.rb
```diff
@@ -76,27 +76,19 @@ module ProfileSearching
...
-            id: policy_or_profile_id
-          }
-        }
-      }
+      policy_ids = where(id: policy_or_profile_id).select(:policy_id)
```
romanblanco: `policy_ids` could be renamed to represent better both values.

app/models/concerns/profile_searching.rb
```diff
@@ -76,27 +76,19 @@ module ProfileSearching
...
+        .or(where(policy_id: policy_ids))
+        .distinct
     }
   end
 
   # class methods for profile searching
   module ClassMethods
     def first_by_os_minor_version_preferred(os_minor_version)
-      where(os_minor_version: ['', os_minor_version])
+      where(os_minor_version: ['', '0', os_minor_version])
```
romanblanco: Selfnote: If tailoring gets created without `os_minor_version` (due to a bug), it is created with emtpy string (`''`), and for the rewrite, we're changing `os_minor_version` to integer. Hence the added `'0'`.

romanblanco: This workaround could be extracted into a separate PR and added together with adding `notnull` constraints.

app/models/concerns/profile_tailoring.rb
```diff
@@ -51,7 +51,7 @@ def removed_rules
   end
 
   def update_os_minor_version(version)
-    return unless version && os_minor_version.blank?
+    return unless version && (os_minor_version.blank? || os_minor_version == '0')
```
romanblanco: Same case as in https://github.com/RedHatInsights/compliance-backend/pull/2708/changes/730c984724c2614332f364f30e69b79bb7e56d31#r3219252091

test/factories/policy.rb
```diff
@@ -4,5 +4,7 @@
   factory :policy do
     name { Faker::Lorem.sentence }
     account { User.current.account }
+    compliance_threshold { Policy::DEFAULT_COMPLIANCE_THRESHOLD }
+    profile_id { association(:canonical_profile).id }
```
romanblanco, line 8: Can not use `canonical_profile` for V1 model

test/models/host_test.rb
```diff
@@ -310,7 +310,8 @@ class HostTest < ActiveSupport::TestCase
         account: @account,
         external: true,
         parent_profile: @profile,
-        rules: @profile.rules
+        rules: @profile.rules,
+        os_minor_version: '1'
```
romanblanco: Why not `'0'` as in https://github.com/RedHatInsights/compliance-backend/pull/2708/changes/730c984724c2614332f364f30e69b79bb7e56d31#r3219215203

test/models/policy_test.rb
```diff
@@ -55,22 +55,23 @@ class PolicyTest < ActiveSupport::TestCase
     context '#with_ref_ids' do
       setup do
         @p1 = FactoryBot.create(:profile, account: @account, policy: @policy)
-        @p2 = FactoryBot.create(:profile, account: @account, policy: nil)
+        @p2_policy = FactoryBot.create(:policy, account: @account)
```
romanblanco: Selfnote: Creating factory, since tailoring even in V1 model can not be `nil`.

test/models/policy_test.rb
```diff
@@ -55,22 +55,23 @@ class PolicyTest < ActiveSupport::TestCase
...
+        @p2 = FactoryBot.create(:profile, account: @account, policy: @p2_policy, os_minor_version: '1')
       end
 
       should '#with_ref_ids accepts multiple ref_ids' do
-        assert_empty Policy.with_ref_ids([@p2.ref_id])
+        assert_empty Policy.with_ref_ids([@p2.ref_id]).where(id: @policy.id)
```
romanblanco: Why are the `.where(id: @policy.id)` added here?

test/models/policy_test.rb
```diff
@@ -242,11 +243,12 @@ class PolicyTest < ActiveSupport::TestCase
...
   context 'score' do
     should 'return the associated profile score' do
-      test_results = 2.times.map do
+      test_results = 2.times.map do |i|
         profile = FactoryBot.create(
           :profile,
           account: @account,
-          policy: @policy
+          policy: @policy,
+          os_minor_version: i.to_s
```
romanblanco: Same as earlier. No need for adding `os_minor_version` in this PR.

test/models/policy_test.rb
```diff
@@ -452,64 +442,9 @@ class PolicyTest < ActiveSupport::TestCase
           @profile.policy.update_os_minor_versions
 
           assert_equal @profile.reload.os_minor_version, '9'
-          assert_equal child_profile.reload.os_minor_version, ''
+          assert_equal child_profile.reload.os_minor_version, '0'
```
romanblanco: Same as earlier.

test/models/profile_test.rb
```diff
@@ -222,7 +183,7 @@ class ProfileTest < ActiveSupport::TestCase
     end
 
     should 'host is compliant if it is compliant on some policy profile' do
-      p2 = FactoryBot.create(:profile, policy: @policy, account: @account)
+      p2 = FactoryBot.create(:profile, policy: @policy, account: @account, os_minor_version: '1')
```
romanblanco: Same as earlier.

test/models/test_result_test.rb
```diff
@@ -3,10 +3,16 @@
 require 'test_helper'
 
 class TestResultTest < ActiveSupport::TestCase
+  setup do
+    User.current = FactoryBot.create(:user)
```
romanblanco, line 8: Why does the unit test need `user` defined?

test/policies/test_result_policy_test.rb
```diff
@@ -45,30 +45,4 @@ class TestResultPolicyTest < ActiveSupport::TestCase
 
     stub_rbac_permissions(Rbac::INVENTORY_HOSTS_READ)
   end
-
-  test 'only test results within visible profiles are accessible' do
```
romanblanco, line 49: This test seems like it could be passing with the current changes.

Is it failing because of the `os_minor_version` adaption? Does it still need to be removed, if we'll move it into a separate PR?

test/tasks/cleanup_db_test.rb
```diff
@@ -8,28 +8,6 @@ class CleanupDbTest < ActiveSupport::TestCase
...
   end
 
-  test 'cleanup_db removes dangling records' do
-    account = FactoryBot.create(:account)
-    FactoryBot.create(:host, org_id: account.org_id)
-    policy = FactoryBot.create(:policy, account: account)
-    account.dup.update!(org_id: '8675309')
-
-    TestResult.new(host_id: UUID.generate).save(validate: false)
-    RuleResult.new(host_id: UUID.generate).save(validate: false)
```
romanblanco: Selfnote: Removing `RuleResults` and `TestResults` are no longer having same constraints. Hence, removing whole test is more meaningful than preserving it.

test/services/xccdf_tailoring_file_test.rb
```diff
@@ -44,7 +46,9 @@ class XccdfTailoringFileTest < ActiveSupport::TestCase
         ref_id: 'test',
         name: 'test profile',
         benchmark_id: @parent.benchmark.id,
-        parent_profile_id: @parent.id
+        parent_profile_id: @parent.id,
+        policy_id: @policy.id,
+        os_minor_version: '1'
```
romanblanco: Same as earlier.

deploy/clowdapp.yaml
RoamingNoMaD: Looking at it again, none of the columns we `SELECT` are coming from `Tailoring`s.  
We could get to the same results through `Tailoring`s too, but that just means `policies_v2 → tailorings_v2 → canonical_profiles_v2 → security_guides_v2` instead of `policies_v2 → canonical_profiles_v2 → security_guides_v2`. 

This could stay as it is, with one caveat. Now we'd also create metrics about Policies which have no Tailorings. Not a huge issue, but the data going to Floorist will be a bit different. 
Should I change this back and limit us to Policies with Tailorings?

spec/factories/test_result.rb
RoamingNoMaD: If `report_id` is `nil`, `association :v2_tailoring` creates a tailoring, which creates a Policy, which is essentially a Report... This could be better but I've tried to keep my changes minimal.

app/models/concerns/profile_searching.rb
RoamingNoMaD: Since we're trying to get rid of V1, I'd just leave it as it is. 

Otherwise I agree that I've named this quite poorly. There shouldn't be more than 1 Policy tied to a Tailoring (profile here)

app/models/concerns/profile_searching.rb
RoamingNoMaD: Yeah we can't enforce not null yet as DB state is improper.

test/factories/policy.rb
RoamingNoMaD: Here the `canonical_profile` is a sub-factory of the profile factory:
```
    factory :canonical_profile do
      parent_profile { nil }
      account { nil }
      policy { nil }
    end
```

test/models/host_test.rb
RoamingNoMaD: Random hardcoded value. 
I'll get rid of it anyways as `os_minor` will be nullable still.

test/models/policy_test.rb
RoamingNoMaD: It's a mess due to me trying to monkey patch tests after enforcing `os_minor` not being null.
I'll revert it.

app/models/concerns/profile_searching.rb
RoamingNoMaD: I've renamed it to singular.

test/models/test_result_test.rb
RoamingNoMaD: The old factories are missing the chain of related records being created. This is a must have in V2 and this is easier to do than to try and fix V1 tests/factories.

test/policies/test_result_policy_test.rb
RoamingNoMaD: This is true, but at the same time we do not use the V1 API anymore. We do test visibility (Pundit scoping) in V2 and the API works fine.

This is only an API consideration regarding RBAC.

test/tasks/cleanup_db_test.rb
RoamingNoMaD: The reason for this is that the compatibility view has joins that would filter out the dangling record so we cannot test this from V1 anymore. 
I'll create a PR right away so we do not forget to rewrite this rake task and its test, but it's out of scope in this PR.

deploy/clowdapp.yaml
```diff
@@ -553,31 +553,32 @@ objects:
...
+          "policies_v2"."id",
+          "canonical_profiles_v2"."ref_id",
+          "canonical_profiles_v2"."title" AS "name",
           "accounts"."org_id",
-          REGEXP_REPLACE("benchmarks"."ref_id", '.*RHEL-(\d)$', '\1') AS "os_major_version"
-          FROM "policies"
-          INNER JOIN "profiles" ON "profiles"."policy_id" = "policies"."id"
-          INNER JOIN "accounts" ON "accounts"."id" = "policies"."account_id"
-          INNER JOIN "benchmarks" ON "benchmarks"."id" = "profiles"."benchmark_id";
+          "security_guides_v2"."os_major_version"
```
RoamingNoMaD: Note: To match the earlier output this should be cast to TEXT.

deploy/clowdapp.yaml
RoamingNoMaD: Tailorings do not have names, so the canonical_profile JOIN is necessary. 
If need be, we'll change it after the fact.

db/views/v1_profiles_v02.sql
RoamingNoMaD: I'm not sure what we were getting at here. A canonical profile will always have these values.

db/functions/v1_policies_insert_v01.sql
```diff
@@ -0,0 +1,33 @@
...
+    ) VALUES (
+      NEW."name",
+      NEW."description",
+      COALESCE(NEW."compliance_threshold", 100.0),
+      bo_title,
+      NEW."profile_id",
+      NEW."account_id",
+      NOW(),
+      NOW()
+    ) RETURNING "id" INTO "result_id";
```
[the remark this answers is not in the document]
RoamingNoMaD: Shouldn't be an issue since only Report parsing depends on V1 now. 
This will also get rewritten ASAP.

db/functions/v1_profile_rules_update_v03.sql
```diff
@@ -0,0 +1,27 @@
...
+          "rule_id" = NEW."rule_id",
+          "updated_at" = COALESCE(NEW."updated_at", NOW())
+        WHERE "id" = OLD."id";
+    ELSE
+        UPDATE "profile_rules_v2" SET
+          "profile_id" = NEW."profile_id",
+          "rule_id" = NEW."rule_id",
+          "updated_at" = COALESCE(NEW."updated_at", NOW())
+        WHERE "id" = OLD."id";
+    END IF;
```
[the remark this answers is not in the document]
RoamingNoMaD: There's absolutely no way for anyone to move a rule from a canonical profile to a tailoring.

db/functions/v1_profiles_delete_v02.sql
```diff
@@ -0,0 +1,10 @@
+CREATE OR REPLACE FUNCTION v1_profiles_delete() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+BEGIN
+    IF OLD."parent_profile_id" IS NOT NULL THEN
+        DELETE FROM "tailorings_v2" WHERE "id" = OLD."id";
+    END IF;
```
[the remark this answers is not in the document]
RoamingNoMaD: And that is correct, we never delete canonical profiles.

db/functions/v1_profiles_update_v02.sql
```diff
@@ -0,0 +1,26 @@
...
+        WHERE "id" = OLD."id";
+    ELSE
+        UPDATE "tailorings_v2" SET
+          "policy_id" = NEW."policy_id",
+          "profile_id" = NEW."parent_profile_id",
+          "value_overrides" = COALESCE(NEW."value_overrides", '{}'),
+          "os_minor_version" = CAST(NULLIF(NEW."os_minor_version", '') AS int),
+          "updated_at" = COALESCE(NEW."updated_at", NOW())
+        WHERE "id" = OLD."id";
+    END IF;
```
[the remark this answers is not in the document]
RoamingNoMaD: Same as above

db/migrate/20260528151023_create_policies_v2_table.rb
```diff
@@ -0,0 +1,36 @@
...
+      t.timestamps precision: nil, null: true
+    end
+
+    execute <<-SQL
+      INSERT INTO policies_v2 (id, title, description, compliance_threshold, business_objective,
+        profile_id, account_id, created_at, updated_at)
+      SELECT id, title, description, compliance_threshold, business_objective,
+        profile_id, account_id, NOW(), NOW()
+      FROM v2_policies;
+    SQL
```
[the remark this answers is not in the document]
RoamingNoMaD: There's no timestamps in the old table under that view.

db/migrate/20260528151153_create_tailorings_v2_table.rb
```diff
@@ -0,0 +1,30 @@
...
+      FROM tailorings;
+    SQL
+
+    change_column_null :tailorings_v2, :policy_id, false
+    change_column_null :tailorings_v2, :profile_id, false
+    change_column_null :tailorings_v2, :created_at, false
+    change_column_null :tailorings_v2, :updated_at, false
+
+    add_index :tailorings_v2, [:policy_id], name: 'index_tailorings_v2_on_policy_id'
+    add_index :tailorings_v2, [:profile_id], name: 'index_tailorings_v2_on_profile_id'
```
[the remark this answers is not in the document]
RoamingNoMaD: We have a bug that needs fixing before this can happen. These duplicates share a null value in os_minor_version.

db/migrate/20260528151307_create_rule_results_v2_table.rb
```diff
@@ -0,0 +1,31 @@
...
+    execute <<-SQL
+      INSERT INTO rule_results_v2 (id, result, rule_id, test_result_id, created_at, updated_at)
+      SELECT id, result, rule_id, test_result_id, created_at, updated_at
+      FROM rule_results;
+    SQL
+
+    change_column_null :rule_results_v2, :rule_id, false
+    change_column_null :rule_results_v2, :test_result_id, false
+    change_column_null :rule_results_v2, :created_at, false
+    change_column_null :rule_results_v2, :updated_at, false
```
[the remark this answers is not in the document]
RoamingNoMaD: They won't be null, I've checked.

db/schema.rb
```diff
@@ -508,24 +631,23 @@
...
+      historical_test_results_v2.supported,
+      historical_test_results_v2.failed_rule_count,
+      historical_test_results_v2.created_at,
+      historical_test_results_v2.updated_at
+     FROM (historical_test_results_v2
+       JOIN ( SELECT historical_test_results_v2_1.tailoring_id,
+              historical_test_results_v2_1.system_id,
+              max(historical_test_results_v2_1.end_time) AS end_time
+             FROM historical_test_results_v2 historical_test_results_v2_1
+            GROUP BY historical_test_results_v2_1.tailoring_id, historical_test_results_v2_1.system_id) tr ON (((historical_test_results_v2.tailoring_id = tr.tailoring_id) AND (historical_test_results_v2.system_id = tr.system_id) AND (historical_test_results_v2.end_time = tr.end_time))));
```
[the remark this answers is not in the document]
RoamingNoMaD: This is current functionality, I'm trying to not change what I do not need to.

db/schema.rb
```diff
@@ -1221,54 +1195,410 @@
...
+            v_report_id,
+            NEW."host_id",
+            NEW."start_time",
+            NEW."end_time",
+            NEW."score",
+            COALESCE(NEW."supported", TRUE),
+            COALESCE(NEW."failed_rule_count", 0),
+            COALESCE(NEW."created_at", NOW()),
+            COALESCE(NEW."updated_at", NOW())
+          ) RETURNING "id" INTO "result_id";
```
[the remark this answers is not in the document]
RoamingNoMaD: So with tailorings_v2 and canonical_profiles_v2 as the two separate tables now, the trigger lookup via tailorings_v2.id = NEW.profile_id is actually correct — V1 test_results.profile_id would always point at what is now a tailorings_v2 record, not a canonical one. The concern about canonical profiles is a non-issue because test results were never scoped to canonical profiles in V1 either.

db/views/v1_policies_v01.sql
```diff
@@ -0,0 +1,18 @@
...
+  "policies_v2"."profile_id",
+  0 AS "test_result_host_count",
+  COALESCE("sq"."total_host_count", 0) AS "total_host_count",
+  0 AS "unsupported_host_count"
+FROM "policies_v2"
+LEFT OUTER JOIN "business_objectives" ON "business_objectives"."title" = "policies_v2"."business_objective"
+LEFT OUTER JOIN (
+  SELECT COUNT("policy_systems_v2"."id") AS "total_host_count", "policy_systems_v2"."policy_id"
+  FROM "policy_systems_v2" GROUP BY "policy_systems_v2"."policy_id"
+) "sq" ON "sq"."policy_id" = "policies_v2"."id";
```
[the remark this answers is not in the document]
RoamingNoMaD: Again, this would affect Report parsing only, but that doesn't touch business objectives.

db/views/v1_profile_rules_v03.sql
```diff
@@ -0,0 +1,17 @@
...
+
+UNION ALL
+
+SELECT
+  "tailoring_rules_v2"."id",
+  "tailoring_rules_v2"."tailoring_id" AS "profile_id",
+  "tailoring_rules_v2"."rule_id",
+  "tailoring_rules_v2"."created_at",
+  "tailoring_rules_v2"."updated_at"
+FROM "tailoring_rules_v2";
```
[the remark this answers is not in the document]
RoamingNoMaD: They do not overlap, they've been filtered at table creation time.

deploy/clowdapp.yaml
```diff
@@ -556,30 +556,31 @@ objects:
...
         INNER JOIN (
-          SELECT "test_results"."profile_id", "test_results"."host_id", MAX("test_results"."end_time") AS "end_time" FROM "test_results"
-          GROUP BY "test_results"."profile_id", "test_results"."host_id"
-        ) AS "tr" ON ("test_results"."profile_id" = "tr"."profile_id" AND "test_results"."host_id" = "tr"."host_id" AND "test_results"."end_time" = "tr"."end_time")
-        WHERE "rule_results"."result" IN ('fail', 'error', 'unknown', 'fixed')
+          SELECT "historical_test_results_v2"."tailoring_id", "historical_test_results_v2"."system_id", MAX("historical_test_results_v2"."end_time") AS "end_time" FROM "historical_test_results_v2"
+          GROUP BY "historical_test_results_v2"."tailoring_id", "historical_test_results_v2"."system_id"
+        ) AS "tr" ON ("historical_test_results_v2"."tailoring_id" = "tr"."tailoring_id" AND "historical_test_results_v2"."system_id" = "tr"."system_id" AND "historical_test_results_v2"."end_time" = "tr"."end_time")
+        WHERE rule_results_v2.result IN ('fail', 'error', 'unknown', 'fixed')
         GROUP BY "rules_v2"."ref_id", "accounts"."org_id";
```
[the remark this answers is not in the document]
RoamingNoMaD: There's always just one test result stored as of now, this is just a design decision to allow historical data in the future

deploy/clowdapp.yaml
```diff
@@ -647,13 +648,13 @@ objects:
...
-        INNER JOIN "policy_systems" ON "policy_systems"."policy_id" = "v2_policies"."id"
-        INNER JOIN "v2_test_results" ON "v2_test_results"."system_id" = "policy_systems"."system_id"
-        GROUP BY "v2_policies"."account_id", "policy_systems"."system_id";
+          "policies_v2"."account_id" AS "org_id",
+          "policy_systems_v2"."system_id"::TEXT AS "system_id",
+          MAX("historical_test_results_v2"."updated_at") AS "updated_at"
+        FROM "policies_v2"
+        INNER JOIN "policy_systems_v2" ON "policy_systems_v2"."policy_id" = "policies_v2"."id"
+        INNER JOIN "historical_test_results_v2" ON "historical_test_results_v2"."system_id" = "policy_systems_v2"."system_id"
+        GROUP BY "policies_v2"."account_id", "policy_systems_v2"."system_id";
```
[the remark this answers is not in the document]
RoamingNoMaD: Yeah this is weird but has always been like this. It's out of scope of this PR.

test/models/policy_test.rb
```diff
@@ -419,20 +419,8 @@ class PolicyTest < ActiveSupport::TestCase
...
-          account: @account,
-          policy: @policy
-        )
-
-        @profile.policy.update_os_minor_versions
-
-        assert_equal @profile.reload.os_minor_version, ''
-        assert_equal child_profile.reload.os_minor_version, '4'
+        new_profile = @profile.policy.profiles.where.not(id: @profile.id).first
+        assert_equal new_profile.os_minor_version, '4'
```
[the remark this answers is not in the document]
RoamingNoMaD: minitests are not important anymore. They'll be cleaned up later.

mtclinton:
Hey Janos, a few things my robot friend found this reviewing:

**Missing FK constraints on new V2 child tables**

`rule_results_v2`, `historical_test_results_v2`, and `tailoring_rules_v2` don't have foreign key constraints. The migrations create indexes but no `add_foreign_key` calls.

`delete_all` patterns in `Kafka::DeletedSystemCleaner` and `V2::Policy#delete_associated` can orphan rows in these tables (callbacks are skipped by `delete_all`). The `cleanup_db` rake task cleans up some of these, but `ON DELETE CASCADE` FKs on the DB side would be a more reliable safety net.

Specifically:
- `rule_results_v2` → `rules_v2`, `historical_test_results_v2`
- `historical_test_results_v2` → `tailorings_v2`, `policies_v2`
- `tailoring_rules_v2` → `tailorings_v2`, `rules_v2`

Everything else in the PR looks solid — the V1 compatibility layer is well-structured and the V2 table/index design is clean.

RoamingNoMaD:
Thanks for the review! 
I do agree that the structural integrity of the DB is sub-par, but I didn't want to inflate the already huge scope of this PR. 

We should add them later, separately.

db/schema.rb
```diff
@@ -371,6 +444,33 @@
...
+      0 AS test_result_host_count,
+      COALESCE(sq.total_host_count, (0)::bigint) AS total_host_count,
+      0 AS unsupported_host_count
+     FROM ((policies_v2
+       LEFT JOIN business_objectives ON (((business_objectives.title)::text = (policies_v2.business_objective)::text)))
+       LEFT JOIN ( SELECT count(policy_systems_v2.id) AS total_host_count,
+              policy_systems_v2.policy_id
+             FROM policy_systems_v2
+            GROUP BY policy_systems_v2.policy_id) sq ON ((sq.policy_id = policies_v2.id)));
+  SQL
```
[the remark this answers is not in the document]
RoamingNoMaD: This is just for testing purposes. Shouldn't cause bugs as no API or Report parsing code relies on the business objective.

db/schema.rb
```diff
@@ -1221,54 +1195,411 @@
...
-  create_trigger :historical_test_results_delete, sql_definition: <<-SQL
-      CREATE TRIGGER historical_test_results_delete INSTEAD OF DELETE ON public.historical_test_results FOR EACH ROW EXECUTE FUNCTION v2_test_results_delete()
-  SQL
+  create_function :v1_policies_insert, sql_definition: <<-'SQL'
+      CREATE OR REPLACE FUNCTION public.v1_policies_insert()
+       RETURNS trigger
+       LANGUAGE plpgsql
+      AS $function$
+      DECLARE result_id uuid;
+      DECLARE bo_title varchar;
```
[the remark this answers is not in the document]
RoamingNoMaD: Untrue, we've had such declares for a long time and there's no issues with it.

db/functions/v2_test_results_delete_v02.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE FUNCTION v2_test_results_delete() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+BEGIN
+    DELETE FROM "historical_test_results_v2" WHERE "id" = OLD."id";
+    RETURN OLD;
```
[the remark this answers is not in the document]
RoamingNoMaD: The broken trigger is effectively dead code right now. If no model ever issues a DELETE FROM historical_test_results, the mismatch is harmless.

mtclinton:
i know this might be obvious, but isn't this a really massive migration to do all at once?

romanblanco:
> i know this might be obvious, but isn't this a really massive migration to do all at once?

It is a lot, but to make sure we won't miss any data, we haven't found a better way than switching the flow at once.

romanblanco (approved):
LGTM

petrblaho (approved):
LGTM

RoamingNoMaD:
The work overhead that separate migrations would create is just not worth it. 
This is complex as it is. I'm not sure that it'd help if we moved step-by-step and had to pay attention to any weird interactions between the V1 and V2 models/tables.

PR #2713, opened 2026-05-04
cleanup(DB): drop unused content related V1 tables

RoamingNoMaD:
Also dropping a table that is empty in all of our envs. (`[friendly_id_slugs`)

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models/rule_group.rb
+++ b/app/models/rule_group.rb
@@ -17,7 +17,7 @@ class RuleGroup < ApplicationRecord
 
   has_many :rules_with_references, lambda {
     left_outer_joins(:rule_references_container).select(
-      'rules.*', 'rule_references_containers.rule_references AS references'
+      'v1_rules.*', 'v1_rule_references_containers.rule_references AS references'
     )
   }, inverse_of: :rule_group, dependent: :nullify, class_name: 'Rule'
 
--- a/app/models/rule_references_container.rb
+++ b/app/models/rule_references_container.rb
@@ -2,6 +2,10 @@
 
 # Model that is holding the additional rule_references field for each Rule
 class RuleReferencesContainer < ApplicationRecord
+  # FIXME: V2 compatibility - clean up after V2 report parsing refactor
+  self.table_name = :v1_rule_references_containers
+  self.primary_key = :id
+
   belongs_to :rule
 
   validates :rule_id, uniqueness: true
--- a/app/models/rule_references_container.rb
+++ b/app/models/rule_references_container.rb
@@ -8,8 +8,6 @@ class RuleReferencesContainer < ApplicationRecord
 
   belongs_to :rule
 
-  validates :rule_id, uniqueness: true
-
   def self.from_openscap_parser(op_rule, rule_id:, existing: nil)
     container = existing || new(rule_id: rule_id)
 
--- a/db/migrate/20260504122645_drop_v1_content_tables.rb
+++ b/db/migrate/20260504122645_drop_v1_content_tables.rb
@@ -1,0 +1,24 @@
+class DropV1ContentTables < ActiveRecord::Migration[8.1]
+  def change
+    drop_table(:rule_group_relationships) {}
+
+    drop_trigger :v2_rules_insert, on: :v2_rules, revert_to_version: 1
+    drop_trigger :v2_rules_update, on: :v2_rules, revert_to_version: 1
+    drop_trigger :v2_rules_delete, on: :v2_rules, revert_to_version: 1
+    drop_function :v2_rules_insert, revert_to_version: 2
+    drop_function :v2_rules_update, revert_to_version: 2
+    drop_function :v2_rules_delete, revert_to_version: 1
+    drop_view :v2_rules, revert_to_version: 2
+    drop_table(:rule_references_containers) {}
+    drop_table(:rules) {}
+
+    drop_view :v2_rule_groups, revert_to_version: 1
+    drop_table(:rule_groups) {}
+
+    drop_view :v2_value_definitions, revert_to_version: 1
+    drop_table(:value_definitions) {}
+
+    drop_view :security_guides, revert_to_version: 2
+    drop_table(:benchmarks) {}
+  end
+end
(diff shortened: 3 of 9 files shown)


app/models/rule_references_container.rb
```diff
@@ -2,6 +2,10 @@
 
 # Model that is holding the additional rule_references field for each Rule
 class RuleReferencesContainer < ApplicationRecord
+  # FIXME: V2 compatibility - clean up after V2 report parsing refactor
+  self.table_name = :v1_rule_references_containers
+  self.primary_key = :id
```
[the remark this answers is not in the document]
RoamingNoMaD: V1 models will be removed very soon, I do not want to waste time tending to them. 
SSG import (which creates all content related records) is V2 native at this point, there's no need for write functions/triggers for this view.

db/migrate/20260504122645_drop_v1_content_tables.rb
```diff
@@ -0,0 +1,24 @@
...
+    drop_function :v2_rules_update, revert_to_version: 2
+    drop_function :v2_rules_delete, revert_to_version: 1
+    drop_view :v2_rules, revert_to_version: 2
+    drop_table(:rules) {}
+
+    drop_view :v2_rule_groups, revert_to_version: 1
+    drop_table(:rule_groups) {}
+
+    drop_table(:rule_group_relationships) {}
+    drop_table(:rule_references_containers) {}
```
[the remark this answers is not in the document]
RoamingNoMaD: Even if we rollback, the tables will be empty. The migration to the new tables happened months ago, so I do not see value in making these migrations rollbackable. We're fine from a functional standpoint and won't be returning the old tables.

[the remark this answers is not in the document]
RoamingNoMaD: Our service is on hold until migrations finished. This won't break anything as all migrations happen "at the same time", before the puma server starts.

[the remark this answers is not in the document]
RoamingNoMaD: They're not empty but they store an old state which has been superseeded since. It's just redundant data at this point.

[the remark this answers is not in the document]
RoamingNoMaD: Fair enough, fixed.

RoamingNoMaD:
Only the floorist tests are failing now, I'll fix them separately.

RoamingNoMaD:
AI comment is nonsense. Profile Rules never change `profile_id` between tailorings and canonical. Neither do we delete canonical profiles.

test/factories/rule.rb
```diff
@@ -18,7 +18,10 @@
       end
 
       after(:create) do |rule, evaluator|
-        FactoryBot.create(:rule_references_container, rule_id: rule.id, reference_count: evaluator.reference_count)
+        references = evaluator.reference_count.times.map do
+          { href: Faker::Internet.url, label: Faker::Lorem.sentence }
+        end
+        rule.rule_references_container.update!(rule_references: references)
```
opacut, line 24: This calls `update!` on the `v1_rule_references_containers`, if I understand correctly. 
Is that view able to handle the update with the above migration changes?

db/views/v1_rule_references_containers_v01.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+  "rules_v2"."id",
+  "rules_v2"."id" AS "rule_id",
```
opacut, line 3: why is this attribute used twice under different names?

app/models/rule_group.rb
```diff
@@ -17,7 +17,7 @@ class RuleGroup < ApplicationRecord
 
   has_many :rules_with_references, lambda {
     left_outer_joins(:rule_references_container).select(
-      'rules.*', 'rule_references_containers.rule_references AS references'
+      'v1_rules.*', 'v1_rule_references_containers.rule_references AS references'
```
opacut, line 20: is this `rules_v2` joining on `rules_v2`?

test/factories/rule.rb
RoamingNoMaD: Yeah the view won't handle this. :thinking: 

On a second thought this code path is only used for V1 tests at the moment, so I think it's safe to leave as long as tests pass.

SSG import (rule creation) is using V2 already, so this is fine since the view is read-only.

db/views/v1_rule_references_containers_v01.sql
RoamingNoMaD: So in V1 we had a separate table for RuleReferences. This proved to be useless for us, since we do not work with those and just need to show them on the API. 

In V2 we've merged the separate references table into the rules table and now it's just a column. 
The line you're commenting on is a view that fakes two tables (which are in reality one) so that the remaining V1 code continues to function. Here we're essentially treating one record as two, so the ID is both the Rule's ID that it's References are tied to and the References' ID.

app/models/rule_group.rb
RoamingNoMaD: Yes and I agree that it looks terrible. :D

This is the same case as two comments above. It's a temporary placeholder that supports unused code.
V1 API is not live anymore and we do not import SSG content using V1 models either.

I'm trying to deflate these PRs, so even though I could just delete these, I want to do it later as a separate commit.

opacut: oh, so I suppose there's like a list of things to be deleted later, to keep track?

db/views/v1_rule_references_containers_v01.sql
opacut: And do we have that captured in a comment somewhere? Or that's not really how we do things around here

test/factories/rule.rb
opacut: I ran all the tests (`podman-compose exec rails bundle exec rake spec`) and no failures found. I think lgtm?

app/models/rule_group.rb
RoamingNoMaD: Yes, we'll do the cleanup later.

opacut (approved):

PR #2728, opened 2026-05-13
perf(Report): reduce amount of subqueries when selecting Report aggregates

RoamingNoMaD:
## Summary
- Consolidate 3 separate report aggregate subqueries into 1, eliminating redundant `inventory.hosts` table scans and RBAC filter duplication
- Add `REPORTED_SYSTEM_COUNT` lambda to distinguish reported systems via `test_result.id IS NOT NULL` in the LEFT OUTER JOIN context
- Include `assigned_system_count` in `/systems/:id/reports` responses (previously excluded as a side effect), fixing `all_systems_exposed` accuracy

## Test plan
- [ ] `spec/integration/v2/reports_spec.rb` passes
- [ ] `spec/controllers/v2/reports_controller_spec.rb` passes (including nested systems route)
- [ ] Verify generated SQL has a single aggregate subquery instead of three

### DIFF
--- a/app/models/v2/report.rb
+++ b/app/models/v2/report.rb
@@ -13,6 +13,14 @@ module V2
       )
     end
 
+    REPORTED_SYSTEM_COUNT = lambda do
+      AN::NamedFunction.new('COUNT', [V2::System.arel_table[:id]]).filter(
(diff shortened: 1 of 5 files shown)


petrblaho (approved):
LGTM!

PR #2729, opened 2026-05-13
perf(DB): optimize rule_results COUNT query and policy counter N+1

mtclinton:
RHINENG-26237

- ProfilePolicyScoring N+1 fix that replaces the per-host compliant?(host) loop with a single SQL COUNT
- Adds index for test_results(profile_id, host_id, end_time DESC) INCLUDE (id) for index-only scans on the v2_test_results view

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/controllers/v2/rule_results_controller.rb
+++ b/app/controllers/v2/rule_results_controller.rb
@@ -27,5 +27,16 @@ module V2
       scope = join_parents(pundit_scope, permitted_params[:parents])
       scope.with_serializer_data
     end
+
+    # Use a lighter scope for counting that avoids the full serializer data
+    # JOINs (profile, security_guide) that aren't needed just to count rows.
+    def count_collection(_scope)
+      @count_collection ||= begin
+        count_scope = join_parents(pundit_scope, permitted_params[:parents])
+                      .with_count_data
+        count_scope = search(filter_by_tags(count_scope))
+        count_scope.reselect(resource.base_class.count_by).count
+      end
+    end
   end
 end
--- a/app/models/concerns/profile_policy_scoring.rb
+++ b/app/models/concerns/profile_policy_scoring.rb
@@ -17,14 +17,22 @@ module ProfilePolicyScoring
     update_columns(
       total_host_count: hosts.count,
       test_result_host_count: latest_supported_policy_test_result_hosts.count,
-      compliant_host_count: latest_supported_policy_test_result_hosts
-                            .count { |host| compliant?(host) },
+      compliant_host_count: compliant_host_count_sql,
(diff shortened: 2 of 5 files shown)


app/controllers/v2/rule_results_controller.rb
```diff
@@ -27,5 +27,16 @@ def expand_resource
...
+    # Use a lighter scope for counting that avoids the full serializer data
+    # JOINs (profile, security_guide) that aren't needed just to count rows.
+    def count_collection(_scope)
+      @count_collection ||= begin
+        count_scope = join_parents(pundit_scope, permitted_params[:parents])
+                      .with_count_data
+        count_scope = search(filter_by_tags(count_scope))
+        count_scope.reselect(resource.base_class.count_by).count
+      end
+    end
```
RoamingNoMaD: This is exactly what we need but for the sake of DRY, I would like to have it in the [magic autojoin concerns](https://github.com/RedHatInsights/compliance-backend/tree/master/app/controllers/concerns/v2). Reports could also benefit from such an approach.

There's also the overridden `expand_resource` method above, that one should also be incorporated inside those concerns. I'd like to have this as an integral part of David's original architecture, since these are valid use cases which are sub-optimal ATM. 

We do kind of have a plan for the future where we could extract this whole framework for nested APIs and have it as an open source gem. In Kraków, when he presented the idea there were quite a few people interested in this, I think the Ruby community could benefit from having such a gem.

RoamingNoMaD (changes requested):
Just the one comment. Otherwise awesome work, thank you <3

app/controllers/v2/rule_results_controller.rb
mtclinton: sounds good, thanks for the info, I'll start refining and taking a look at the docs you put in knowledge transfer presentation

RoamingNoMaD: https://www.youtube.com/watch?v=y6QigtMXGGM is a good place to start :D

RoamingNoMaD: Sorry to go around you, but this was a good chance for me to do a refactoring and our pipelines are blocked anyways.

Please leave this out from this PR, I've addressed it in [this PR](https://github.com/RedHatInsights/compliance-backend/pull/2738).

RoamingNoMaD (approved):

PR #2775, opened 2026-06-15
refactor(ReportParsing): eliminate double download and validate in consumer

mtclinton:
RHINENG-18501. Builds on the merged #2705 (V2 models) + DB remodel. Scoped to the ParseReport simplification the ticket calls for — "services calling services… most validation should happen in the Kafka consumer, not during parsing."

## What

- **Eliminate the double download.** The consumer validates each report and hands it to the job as a packed payload (`ReportArtifact` — gzip + base64) instead of the job re-downloading the same artifact. Compression keeps the job argument small and JSON-safe.
- **Split `XccdfReportParser#save_all` into `validate!` and `persist!`.** `validate!` runs in the Kafka consumer, so invalid reports are rejected before a job is enqueued; `persist!` runs in the job and only writes. The validation logic itself (OS version, external, benchmark/profile/rules) is unchanged — just relocated.
- **Specs** updated for the new interface (`perform(report_blob, …)`, `parser.persist!`, consumer enqueues a packed payload) plus a `ReportArtifact` unit spec.

## Out of scope / follow-ups

The SSG-version `mismatched` flag and the deterministic-lookup tweak are intentionally not here. Most of the SSG-mismatch correctness — correct-tailoring attribution, dropping unattributable rules, the nil-tailoring guard, real deletion of superseded results — already landed on master via #2705's follow-ups. The remaining piece is the customer-facing `mismatched` signal (RHINENG-26234), which is FE-coupled and worth confirming scope on first.

> Draft: opened for early review — CI is the first full test run on this branch.

## Secure Coding Checklist
- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

- Refactored Kafka report parsing to validate each downloaded XML in the consumer and enqueue a job with a packed, gzip+base64 `ReportArtifact` payload (removes the double-download path).
- Split `XccdfReportParser` into `validate!` (runs unchanged validation checks) and `persist!` (does persistence only, wrapping `save_all_test_result_info` in a transaction).
- Updated `ParseReportJob` to accept `report_blob`, unpack it to recover the XML, and call `persist!` (replacing `save_all` / removed download-by-index logic).
- Reworked `Kafka::ReportParser#parse_reports` to enqueue per report using `profile_id`, issue per-report failure validation on `XccdfReportParser::ERRORS`, and send a batched success audit via `notify_enqueued_reports`.
- Updated specs: added `ReportArtifact` pack/unpack unit tests and adjusted consumer/job/parser specs to assert the new payload/enqueue flow.

### DIFF
--- a/app/jobs/parse_report_job.rb
+++ b/app/jobs/parse_report_job.rb
@@ -4,8 +4,10 @@
 class ParseReportJob < ApplicationJob
   include Notifications
 
+  self.log_arguments = false
+
   # https://github.com/RoamingNoMaD/yabeda-activejob#custom-tags
-  def yabeda_tags(_idx, message)
+  def yabeda_tags(_report_blob, message)
     { qe: OpenshiftEnvironment.qe_account?(message['org_id']) }
   end
 
@@ -14,7 +16,7 @@ class ParseReportJob < ApplicationJob
     provider_job_id || job_id
   end
 
-  def perform(idx, message)
+  def perform(report_blob, message)
     return if cancelled?
 
     @msg_value = message
@@ -23,7 +25,8 @@ class ParseReportJob < ApplicationJob
       "system #{@msg_value['id']}"
     )
 
-    @file = retrieve_file(idx)
+    @file = resolve_report(report_blob)
+    return unless @file
 
     parse_and_save_report
   end
@@ -38,18 +41,22 @@ class ParseReportJob < ApplicationJob
 
   private
 
-  def retrieve_file(idx)
-    @file = SafeDownloader.download_reports(
-      @msg_value['url'],
-      ssl_only: Settings.report_download_ssl_only
-    )[idx]
-  rescue SafeDownloader::DownloadError => e
+  def resolve_report(report_blob)
+    if report_blob.is_a?(Integer)
+      SafeDownloader.download_reports(
+        @msg_value['url'], ssl_only: Settings.report_download_ssl_only
+      )[report_blob]
+    else
+      ReportArtifact.unpack(report_blob)
+    end
+  rescue ArgumentError, NoMethodError, Zlib::GzipFile::Error, SafeDownloader::DownloadError => e
     handle_error(e)
+    nil
   end
 
   def parse_and_save_report
     notify_payload_tracker(:processing, "Job #{jid} is now processing")
-    compliance_notification_wrapper { parser.save_all }
+    compliance_notification_wrapper { parser.persist! }
     notify_remediation
     audit_success
     notify_payload_tracker(:success, "Job #{jid} has completed successfully")
--- a/app/services/kafka/report_parser.rb
+++ b/app/services/kafka/report_parser.rb
@@ -13,30 +13,35 @@ module Kafka
       @logger = logger
     end
 
-    # rubocop:disable Metrics/MethodLength
     def parse_reports
-      # Map successfuly parsed (validated) reports by scanned profile
-      parsed_reports = downloaded_reports.map do |xml|
-        [parse(xml).test_result_file.test_result.profile_id, xml]
-      end
-      # Evaluate each report individually and notify about the result
-      parsed_reports.each_with_index do |(profile_id, _report), idx|
-        job = enqueue_parsing(profile_id, idx)
-        notify_report_success(profile_id, job.jid)
-      end
-      produce_validation_message('success')
+      @enqueued = []
+      @failed = false
+      downloaded_reports.each { |xml| validate_and_enqueue(xml) }
+      notify_enqueued_reports
+      produce_validation_message('success') unless @failed
     rescue EntitlementError, ReportParseError => e
       parse_error(e)
     rescue SafeDownloader::DownloadError => e
       parse_error(e)
       raise
     end
-    # rubocop:enable Metrics/MethodLength
 
     private
 
-    def enqueue_parsing(profile_id, idx)
-      ParseReportJob.perform_later(idx, metadata) or
+    def validate_and_enqueue(xml)
+      parser = XccdfReportParser.new(xml, metadata)
+      parser.validate!
+      profile_id = parser.test_result_file.test_result.profile_id
+      job = enqueue_parsing(profile_id, xml)
+      @enqueued << "#{profile_id}:#{job.jid}"
+      notify_payload_tracker(:received, "File of #{profile_id} is valid. Job #{job.jid} enqueued")
+    rescue *XccdfReportParser::ERRORS => e
+      @failed = true
+      parse_error(e, notify_tracker: true)
+    end
+
+    def enqueue_parsing(profile_id, xml)
+      ParseReportJob.perform_later(ReportArtifact.pack(xml), metadata) or
         raise ReportParseError, "Failed to enqueue parsing of #{profile_id}"
     end
 
@@ -47,10 +52,11 @@ module Kafka
       reports
     end
 
-    def notify_report_success(profile_id, job)
-      msg = "Enqueued report parsing of #{profile_id} from request #{request_id} as a job #{job}"
+    def notify_enqueued_reports
+      return if @enqueued.empty?
+
+      msg = "Enqueued #{@enqueued.size} report(s) from request #{request_id}: #{@enqueued.join(', ')}"
       @logger.audit_success("[#{org_id}] #{msg}")
-      notify_payload_tracker(:received, "File of #{profile_id} is valid. Job #{job} enqueued")
     end
 
     def notify_payload_tracker(status, status_msg = '')
@@ -73,7 +79,6 @@ module Kafka
       (@message.dig('platform_metadata', 'metadata') || {}).merge(
         'id' => id,
         'b64_identity' => b64_identity,
-        'url' => url,
         'request_id' => request_id,
         'org_id' => org_id
       )
@@ -103,14 +108,6 @@ module Kafka
       @message.dig('platform_metadata', 'url')
     end
 
-    def parse(xml)
-      XccdfReportParser.new(xml, metadata)
-    rescue PG::Error, ActiveRecord::StatementInvalid => e
-      parse_error(e)
-    rescue StandardError
-      raise ReportParseError, 'Report parsing failed'
-    end
-
     def produce_validation_message(result)
       return if Settings.kafka.topics.upload_compliance.blank?
 
@@ -121,17 +118,15 @@ module Kafka
       )
     end
 
-    def parse_error(exception)
+    def parse_error(exception, notify_tracker: false)
       msg = "[#{org_id}] #{exception_message(exception)}"
-      @logger.error msg
-      @logger.audit_fail msg
-
+      @logger.error(msg)
+      @logger.audit_fail(msg)
       ReportUploadFailed.deliver(
         system: V2::System.find_by(id: id, org_id: org_id),
-        request_id: request_id,
-        error: exception_message(exception),
-        org_id: org_id
+        request_id: request_id, error: exception_message(exception), org_id: org_id
       )
+      notify_payload_tracker(:error, msg) if notify_tracker
       produce_validation_message('failure')
     end
 
@@ -144,7 +139,7 @@ module Kafka
       when 'Kafka::ReportParser::ReportParseError'
         "Invalid report: #{exception.message}"
       else
-        "Error parsing report: #{request_id} - #{exception.message}"
+        "Error parsing report: #{request_id} - #{exception.class.to_s.demodulize}"
       end
     end
   end
--- a/app/services/report_artifact.rb
+++ b/app/services/report_artifact.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# Packs/unpacks a report XML so it can travel as a background-job argument
+# instead of being re-downloaded by the job (eliminating the double download).
+#
+# * gzip  -> XCCDF compresses ~10-20x, keeping the job argument small
+# * base64 -> keeps the payload a JSON-safe ASCII string for the job backend
+#             (Redis today, PostgreSQL once we move to GoodJob — RHINENG-24397)
+module ReportArtifact
+  module_function
+
+  def pack(xml)
+    Base64.strict_encode64(ActiveSupport::Gzip.compress(xml))
+  end
+
+  def unpack(blob)
+    ActiveSupport::Gzip.decompress(Base64.strict_decode64(blob))
+  end
+end
--- a/app/services/xccdf_report_parser.rb
+++ b/app/services/xccdf_report_parser.rb
@@ -125,11 +125,13 @@ class XccdfReportParser
     check_for_missing_rules
   end
 
-  def save_all
+  def validate!
     check_os_version
     check_for_external_reports
     check_for_missing_benchmark_info
+  end
 
+  def persist!
     V2::System.transaction do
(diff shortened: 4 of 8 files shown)


app/services/kafka/report_parser.rb
```diff
@@ -13,30 +13,38 @@ def initialize(message, logger)
...
       raise
     end
-    # rubocop:enable Metrics/MethodLength
 
     private
 
-    def enqueue_parsing(profile_id, idx)
-      ParseReportJob.perform_later(idx, metadata) or
+    def validate_and_enqueue(xml)
+      parser = XccdfReportParser.new(xml, metadata)
```
RoamingNoMaD, line 32: I think we could improve this further later.
Ideally we could use the XML push parser described in RHINENG-1998, which was briefly tested in [David's PR](https://github.com/RedHatInsights/compliance-backend/pull/1737) so we only parse what we need for validation, thus not polluting the memory.

I'm fine with this implementation. We can and should be gradual about changes here.

app/services/kafka/report_parser.rb
```diff
@@ -13,30 +13,38 @@ def initialize(message, logger)
...
+      parser.validate!
+      profile_id = parser.test_result_file.test_result.profile_id
+      job = enqueue_parsing(profile_id, xml)
+      notify_report_success(profile_id, job.jid)
+    rescue *XccdfReportParser::ERRORS, PG::Error, ActiveRecord::StatementInvalid => e
+      # A single invalid report is rejected on its own; sibling reports in
+      # the same payload are still processed.
+      parse_error(e)
+    rescue StandardError
+      raise ReportParseError, 'Report parsing failed'
```
RoamingNoMaD:
```ruby
    rescue StandardError => e
      raise ReportParseError, "Report parsing failed: #{e.message}"
```

I think this would be nice for debug purposes.

app/services/report_artifact.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# Packs/unpacks a report XML so it can travel as a background-job argument
+# instead of being re-downloaded by the job (eliminating the double download).
+#
+# * gzip  -> XCCDF compresses ~10-20x, keeping the job argument small
+# * base64 -> keeps the payload a JSON-safe ASCII string for the job backend
+#             (Redis today, PostgreSQL once we move to GoodJob — RHINENG-24397)
+module ReportArtifact
+  module_function
```
RoamingNoMaD, line 10: This could be a regular class (service object) so we eliminate `module_function`. 
It's a little cryptic IMO and personally I'm not a big fan of the old `mixing mixins into mixins` style of V1.

RoamingNoMaD: Also if this would be a service object, we could name it something like `ReportPacker` or similar.

These two comments are bit opinionated, but when we were doing the Kafka rewrite, I had to choose the style of the refactor and since this is a debate as old as Ruby itself, there's no right answers.

You can search online about this topic, I found it interesting. I can't find the articles I've read back then, but essentially there's three main ways the community does this:
1. Modules (mixins in other languages)
2. Service Objects (a little hated on due to not being OOP to the T)
3. Plain Old Ruby Objects (I like this one too, but I do not know where would I put them)

RoamingNoMaD (commented):
I like this very much, it's a great first step. Thanks :)

megaslurpie:
thx, will address tomorrow

app/services/report_artifact.rb
megaslurpie: I don't feel strongly about either option, but I thought service objects/classes would be more useful if there was initialization and data/state usage?

megaslurpie:
@coderabbitai review

megaslurpie:
@coderabbitai review

app/services/kafka/report_parser.rb
```diff
@@ -13,30 +13,38 @@ def initialize(message, logger)
       @logger = logger
     end
 
-    # rubocop:disable Metrics/MethodLength
+    # Validate each report up front (RHINENG-18501): invalid reports are
+    # rejected here, in the consumer, so the job never has to re-validate
+    # or re-download. Valid reports are handed to the job as a packed
+    # payload, eliminating the second download.
```
RoamingNoMaD: This is nice for development, but I kind of dislike long AI generated comments that are context specific.
The method is nice, short and readable. Ruby is supposed to be self documenting so we should not comment if there's nothing too magical happening.

app/services/report_artifact.rb
RoamingNoMaD: I completely agree.

spec/jobs/parse_report_job_spec.rb
```diff
@@ -171,14 +172,66 @@
...
+      end
+
+      it 'handles the unpack error gracefully' do
+        expect(ReportUploadFailed).to receive(:deliver)
+
+        job.perform(bad_blob, msg_value)
+      end
+    end
+
+    context 'with an Integer report_blob (rolling deploy backward compat)' do
```
RoamingNoMaD:
```suggestion
    context 'with a legacy report_blob' do
```

spec/services/kafka/report_parser_spec.rb
```diff
@@ -100,29 +100,141 @@
...
+
+      expect(enqueued_jobs.size).to eq(1)
+    end
+
+    it 'does not emit a per-report failure validation message' do
+      expect(ReportValidation).not_to receive(:deliver).with(hash_including(validation: 'failure'))
+      allow(ReportValidation).to receive(:deliver)
+
+      service.parse_reports
+    end
```
RoamingNoMaD: The old behavior was that if one fails, all fail. This is not very nice. 
I see you changed this and now we do enqueue all we can even if one report fails validation.
Couldn't we log failures just so we know why they failed? It'd be nice in the future. The others that pass can be summed up in one `audit_success` IMO.

spec/services/report_artifact_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe ReportArtifact do
+  let(:xml) { "<xml><TestResult id='abc'/></xml>" * 200 }
```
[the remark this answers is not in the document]
RoamingNoMaD: Sorry I have to agree with CodeRabbit here. If faker can generate XML, lets use that. We don't care about the contents, it's just compression tests.

app/services/kafka/report_parser.rb
megaslurpie: created these as well:
https://redhat.atlassian.net/browse/RHINENG-28323
https://redhat.atlassian.net/browse/RHINENG-28324

megaslurpie:
@coderabbit review

spec/services/kafka/report_parser_spec.rb
megaslurpie: updated

app/services/kafka/report_parser.rb
```diff
@@ -47,10 +52,11 @@ def downloaded_reports
...
 
-    def notify_report_success(profile_id, job)
-      msg = "Enqueued report parsing of #{profile_id} from request #{request_id} as a job #{job}"
+    def notify_enqueued_reports
+      return if @enqueued.empty?
+
+      msg = "Enqueued #{@enqueued.size} report(s) from request #{request_id}: #{@enqueued.join(', ')}"
       @logger.audit_success("[#{org_id}] #{msg}")
-      notify_payload_tracker(:received, "File of #{profile_id} is valid. Job #{job} enqueued")
     end
```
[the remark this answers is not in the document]
megaslurpie: i think this is fine

spec/services/kafka/report_parser_spec.rb
```diff
@@ -106,34 +106,46 @@
...
-            /\A\[#{org_id}\] Enqueued report parsing of #{profile_id} from request #{request_id} as a job \S+\z/
+            /\A\[#{org_id}\] Enqueued 1 report\(s\) from request #{request_id}: #{profile_id}:\S+\z/
           )
         )
 
       service.parse_reports
 
       expect(enqueued_jobs.size).to eq(1)
+      expect(ReportArtifact.unpack(enqueued_jobs.first[:args].first)).to eq(xml)
     end
```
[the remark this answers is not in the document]
megaslurpie: not sure how much this really matters

app/services/kafka/report_parser.rb
```diff
@@ -13,30 +13,35 @@ def initialize(message, logger)
...
       raise
     end
-    # rubocop:enable Metrics/MethodLength
 
     private
 
-    def enqueue_parsing(profile_id, idx)
-      ParseReportJob.perform_later(idx, metadata) or
+    def validate_and_enqueue(xml)
+      parser = XccdfReportParser.new(xml, metadata)
```
romanblanco, line 32: This might be missing the exception handling it previously had.

`rescue PG::Error, ActiveRecord::StatementInvalid => e`

app/services/kafka/report_parser.rb
```diff
@@ -144,7 +139,7 @@ def exception_message(exception)
       when 'Kafka::ReportParser::ReportParseError'
         "Invalid report: #{exception.message}"
       else
-        "Error parsing report: #{request_id} - #{exception.message}"
+        "Error parsing report: #{request_id} - #{exception.class.to_s.demodulize}"
```
romanblanco, line 142: Was this changed on purpose? Commit message says the opposite.

app/services/kafka/report_parser.rb
megaslurpie: the old code had a rescue StandardError that swallowed DB errors and treated them as permanent parse failures. now they propagate up to karafka for retry which I think is the correct behavior. we only catch XccdfReportParser::ERRORS here since those are real validation failures that shouldnt be retried

app/services/kafka/report_parser.rb
megaslurpie: yea good catch on the commit msg, will fix that. the change itself is intentional tho this else branch hits for xccdf validation errors that go through ReportUploadFailed.deliver which can reach the customer. using the class name (WrongFormatError) instead of raw exception.message avoids leaking internal details

app/services/kafka/report_parser.rb
```diff
@@ -13,30 +13,35 @@ def initialize(message, logger)
...
-      parsed_reports.each_with_index do |(profile_id, _report), idx|
-        job = enqueue_parsing(profile_id, idx)
-        notify_report_success(profile_id, job.jid)
-      end
-      produce_validation_message('success')
+      @enqueued = []
+      @failed = false
+      downloaded_reports.each { |xml| validate_and_enqueue(xml) }
+      notify_enqueued_reports
+      produce_validation_message('success') unless @failed
```
opacut, line 21: if I understand correctly, if we have for example 3 reports, and 2 of them succeed, then we want to report a batch failure?

app/jobs/parse_report_job.rb
```diff
@@ -14,7 +16,9 @@ def jid
     provider_job_id || job_id
   end
 
-  def perform(idx, message)
+  # `report_blob` is the validated report packed by Kafka::ReportParser
+  # (see ReportArtifact) — no second download.
```
RoamingNoMaD: I don't think this comment is necessary. 🤔

app/services/kafka/report_parser.rb
```diff
@@ -121,17 +118,15 @@ def produce_validation_message(result)
...
-      @logger.audit_fail msg
-
+      @logger.error(msg)
+      @logger.audit_fail(msg)
       ReportUploadFailed.deliver(
         system: V2::System.find_by(id: id, org_id: org_id),
-        request_id: request_id,
-        error: exception_message(exception),
-        org_id: org_id
+        request_id: request_id, error: exception_message(exception), org_id: org_id
```
RoamingNoMaD, line 127: Was the method too long? 😆

app/jobs/parse_report_job.rb
```diff
@@ -23,9 +27,10 @@ def perform(idx, message)
       "system #{@msg_value['id']}"
     )
 
-    @file = retrieve_file(idx)
-
+    @file = ReportArtifact.unpack(report_blob)
```
RoamingNoMaD: Switching these two at once would mean that enqueued jobs fail, right?
Could we have a fallback for the time being so we do not lose any reports when we deploy these changes?

RoamingNoMaD (changes requested):
Please get rid of the comments. I've also added a few questions and observations. 
I'd almost merge this, thanks for the great work :)

app/services/kafka/report_parser.rb
megaslurpie: lol

app/services/kafka/report_parser.rb
megaslurpie: yep, this is a more granular way of logging and reporting individual reports but if 1 fails we report whole batch as failure

app/services/kafka/report_parser.rb
megaslurpie: i guess RuboCop wanted it

app/jobs/parse_report_job.rb
megaslurpie: done

app/services/xccdf_report_parser.rb
```diff
@@ -125,11 +125,18 @@ def check_for_missing_benchmark_info
     check_for_missing_rules
   end
 
-  def save_all
+  # Validation that must pass before a report is accepted. Runs in the Kafka
+  # consumer (Kafka::ReportParser) so invalid reports are rejected before a
+  # job is enqueued — the job no longer re-validates.
```
RoamingNoMaD: Can you please get rid of all these comments? IMO they're stating the obvious.

app/services/xccdf_report_parser.rb
```diff
@@ -125,11 +125,18 @@ def check_for_missing_benchmark_info
...
+  def validate!
     check_os_version
     check_for_external_reports
     check_for_missing_benchmark_info
+  end
 
+  # Persistence only. Runs in ParseReportJob; trusts the consumer's
+  # validation and just writes the results inside a transaction.
```
RoamingNoMaD: Same here, this can be assumed from the method name.

spec/services/kafka/report_parser_spec.rb
```diff
@@ -106,34 +106,48 @@
...
+          "[#{org_id}] Error parsing report: #{request_id} - WrongFormatError"
         )
       service.parse_reports
       expect(enqueued_jobs.size).to eq(0)
     end
   end
 
   context 'with valid reports' do
+    # Validation now happens in the consumer; stub the parser so this isolates
+    # the consumer's job: validate, then enqueue the *packed* report (no index).
```
RoamingNoMaD: Same here, we shouldn't describe implementation in a unit test.

RoamingNoMaD (changes requested):

app/jobs/parse_report_job.rb
megaslurpie: added fallback

RoamingNoMaD (approved):
LGTM

PR #2800, opened 2026-06-23
fix(CI): add rake dependency explicitly

romanblanco:
## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices

- Added an explicit top-level `rake` dependency to the Gemfile to ensure CI has rake available.

PR #2753, opened 2026-06-01
feat(CI): enable hermetic builds

romanblanco:
RHINENG-23444

## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices

## Release Notes

- **CI hermetic mode (Tekton)**: Set `hermetic: "true"`, provided `prefetch-input`, and passed `build-args: HERMETIC=true` to the build task in Tekton `PipelineRun` manifests.
- **Hermetic Docker builds**: Added `ARG HERMETIC` behavior to conditionally skip network-dependent Postgres repo setup and tzdata removal; removed RubyGems/Bundler system bootstrap steps from the Docker build/runtime flow.
- **Hermetic build documentation + tooling**: Added `.hermetic_builds/` README and a Dockerfile parser to generate deterministic RPM dependency inputs for Konflux hermetic pipelines.
- **RPM dependency pinning automation**: Introduced root `Makefile` targets to generate `ubi.repo`, `rpms.in.yaml`, and `rpms.lock.yaml` from the hermetic workflow inputs.
- **Rust crates vendoring for hermeticity**: Added `.hermetic_builds/cargo/` Cargo config to redirect `crates.io` to a vendored directory and added the `fast_mmaped_file_rs` crate metadata used for hermetic builds.
- **Hermetic helper artifacts cleanup**: Updated `.gitignore` to exclude generated hermetic artifacts (`redhat-uep.pem`, `entitlement/`, `redhat.repo`).
- **Secure coding practices checklist**: Added the secure coding practices checklist reference/link with only **“General Coding Practices”** checked.

PR #2822, opened 2026-06-24
fix(Dockerfile): restore gem/bundler setup in runtime stage

megaslurpie:
## Summary

- Restores `gem update --system` and `gem install bundler` to the runtime stage of the Dockerfile
- These lines were removed in PR #2753 (hermetic builds), which broke the runtime bundler environment
- Without them, init containers fail with `bundler: command not found: rake` because the system bundler cannot resolve gems in the deployment `.bundle/` path
- This is causing all new pods in `compliance-stage` to crash (CrashLoopBackOff / Init:Error), while old pods on the previous image (`3a5ed60`) remain healthy

## Root Cause

PR #2753 removed `gem update --system` and `gem install bundler` from both the build and runtime Dockerfile stages. The build stage works without them because `$devDeps` pulls in sufficient Ruby tooling. However, the runtime stage only installs `$deps`, which lacks a properly configured gem/bundler environment to find gems installed in `.bundle/`.

PR #2800 attempted to fix this by adding `gem 'rake'` to the Gemfile, but the issue is not a missing gem — it's that the runtime bundler cannot locate *any* gems in the deployment path.

## Test plan

- [ ] Build the Docker image and verify the init container (`check_migration_status_and_ssg_synced.sh`) completes successfully
- [ ] Verify `bundle exec rake --version` works inside the runtime container
- [ ] Deploy to stage and confirm pods start without Init:Error

## Secure Coding Checklist
- [x] System Configuration

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- Restored runtime Ruby tooling setup in the Dockerfile by running `gem update --system` and `gem install bundler` in the final image stage.
- Kept the rest of the runtime stage unchanged: install dependencies, adjust permissions, clean package manager cache, and `chown` the app directory.
- Goal: fix init container failures where the system bundler could not resolve gems from the deployed `.bundle/` path, causing `bundler: command not found: rake` and CrashLoopBackOffs in new pods.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,12 +61,14 @@ WORKDIR /opt/app-root/src
 
 USER 0
 
-RUN ( [[ $HERMETIC == "true" ]] || rpm -e --nodeps tzdata &>/dev/null ) && \
-    microdnf module enable -y ruby:3.3                                  && \
-    microdnf install --nodocs -y $deps                     && \
-    chmod +t /tmp                                          && \
-    microdnf clean all -y                                  && \
-    chown 1001:root ./                                     && \
+RUN ( [[ $HERMETIC == "true" ]] || rpm -e --nodeps tzdata &>/dev/null )                && \
+    microdnf module enable -y ruby:3.3                                                 && \
+    microdnf install --nodocs -y $deps                                                 && \
+    chmod +t /tmp                                                                      && \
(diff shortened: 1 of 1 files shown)


petrblaho (approved):
LGTM

romanblanco (approved):

RoamingNoMaD (approved):

PR #2825, opened 2026-06-25
fix(Dockerfile): set BUNDLE_PATH env var in final image for hermetic build

RoamingNoMaD:
After the hermetic build was introduced, pods fail to start because bundler
cannot find any gems at runtime. Inspection of the running image shows gems
present in `.bundle/ruby/` but `.bundle/config` missing entirely, so bundler
falls back to the empty system GEM_PATH.

The root cause is unknown — the most likely explanation is that Cachi2 sets
`BUNDLE_APP_CONFIG` during the build, redirecting `bundle config set --local`
writes away from the project directory, but this was not directly verified.

Fix by setting `BUNDLE_PATH=".bundle"` in the image ENV, which is sufficient
regardless of why the config file is absent.

- Fix Docker hermetic runtime image: set `BUNDLE_PATH=".bundle"` in the final image `ENV` so Bundler can find installed gems at pod startup even if `.bundle/config` is missing.
- Clean up production runtime Dockerfile `ENV` by consolidating Rails/logging/home/dev-deps and Prometheus settings into a single block.
- Update `yabeda-activejob` Gemfile entry to use the published gem version instead of a pinned GitHub repo/branch.
- Backport custom ActiveJob metric tag support into `yabeda-activejob` 0.6.0 via a Rails initializer (custom tag extraction + label merge for success/failure/executed/runtime counters).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,6 +76,7 @@ USER 1001
 COPY --chown=1001:0 . /opt/app-root/src
 COPY --chown=1001:0 --from=build /opt/app-root/src/.bundle /opt/app-root/src/.bundle
 
-ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true HOME=/opt/app-root/src DEV_DEPS=$devDeps prometheus_multiproc_dir=/opt/app-root/src/tmp prometheus_rust_mmaped_file=false
+ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true HOME=/opt/app-root/src DEV_DEPS=$devDeps prometheus_multiproc_dir=/opt/app-root/src/tmp prometheus_rust_mmaped_file=false \
+    BUNDLE_PATH=".bundle" BUNDLE_WITHOUT="development:test" BUNDLE_DEPLOYMENT="true" BUNDLE_VERSION="system" BUNDLE_DISABLE_VERSION_CHECK="true"
 
 CMD ["/opt/app-root/src/entrypoint.sh"]
--- a/Gemfile
+++ b/Gemfile
@@ -68,7 +68,7 @@ gem 'yabeda-prometheus-mmap'
 gem 'yabeda-puma-plugin'
 gem 'yabeda-rails'
 gem 'yabeda-sidekiq'
-gem 'yabeda-activejob', github: 'RoamingNoMaD/yabeda-activejob', branch: 'add-custom-tagging'
+gem 'yabeda-activejob', '= 0.6.0'
 
 # Nokogiri
 gem 'nokogiri', force_ruby_platform: true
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -76,6 +76,7 @@ USER 1001
 COPY --chown=1001:0 . /opt/app-root/src
 COPY --chown=1001:0 --from=build /opt/app-root/src/.bundle /opt/app-root/src/.bundle
 
-ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true HOME=/opt/app-root/src DEV_DEPS=$devDeps prometheus_multiproc_dir=/opt/app-root/src/tmp prometheus_rust_mmaped_file=false
+ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=true HOME=/opt/app-root/src DEV_DEPS=$devDeps prometheus_multiproc_dir=/opt/app-root/src/tmp prometheus_rust_mmaped_file=false \
+    BUNDLE_PATH=".bundle"
```
romanblanco:
```suggestion
    BUNDLE_PATH="./.bundle" BUNDLE_WITHOUT="development:test" BUNDLE_DEPLOYMENT="true" BUNDLE_FORCE_RUBY_PLATFORM="true"
```

RoamingNoMaD: Why would we need to duplicate this setting?

romanblanco: Why would you install development and test packages when you set RAILS_ENV to production?

RoamingNoMaD: This is not the build phase of the Dockerfile. We're not installing gems here.

You can see that none of those are present if you inspect the resulting image.

romanblanco (approved):

megaslurpie (commented):
**The prepend+unsubscribe pattern won't work — `Yabeda.configure` defers block execution.**

`super` calls `Yabeda.configure { ... subscribe ... }`, but that block isn't executed immediately — it's pushed onto `Yabeda.configurators` and only runs later when `configure!` is called (via `config.after_initialize` in `yabeda/railtie.rb`). So when `listeners_for('perform.active_job')` runs right after `super`, no yabeda listeners exist yet — the unsubscribe is a no-op. The patched listener is then added directly. Later, when `configure!` runs the deferred block, the original listener is also registered. Both fire, double-counting all perform metrics.

Yabeda source (`lib/yabeda/dsl/class_methods.rb`):
```ruby
def configure(&block)
  Yabeda.configurators.push([@group, block])
  class_exec(&block) if Yabeda.configured?  # false at install! time
end
```

Minor note: the fork also injected custom_tags into `enqueue`/`enqueue_at`/`perform_start`/`enqueue_all` — this patch only covers `perform`.

RoamingNoMaD:
Nice find @megaslurpie. This didn't work either so I'll just revert the hermetic builds and ask @romanblanco to work on it in a PR without blocking others.

PR #2827, opened 2026-06-25
feat(jobs): install GoodJob and migrate from Sidekiq

megaslurpie:
RHINENG-24398. Adds [GoodJob](https://github.com/bensheldon/good_job) (PostgreSQL-backed ActiveJob adapter) alongside Sidekiq, following GoodJob's recommended per-job migration pattern.

## What

- **GoodJob gem + migration**: Adds `good_job ~> 4` and the database migration for job tables (`good_jobs`, `good_job_executions`, `good_job_batches`, `good_job_processes`, `good_job_settings`).
- **Per-job adapter opt-in**: `ParseReportJob` declares `self.queue_adapter = :good_job`. Global adapter stays `:sidekiq` — no existing Sidekiq behavior is changed.
- **Job retention**: Only jobs with unhandled errors are preserved (lambda form), cleaned up after 14 days. Matches Sidekiq's existing behavior where completed jobs are discarded.
- **New GoodJob pod**: `compliance-goodjob` application type in `entrypoint.sh` and a matching deployment in `clowdapp.yaml` with parameterized health probes (`GOOD_JOB_PROBE_PORT`).
- **Sidekiq untouched**: All Sidekiq-specific code in `ParseReportJob` (`cancelled?`, `cancel!`, `jid`, `Sidekiq.logger`) remains intact for the drain period. TODO comments added for the follow-up cleanup. `cancelled?` is guarded to skip Redis when running under GoodJob.
- **Memory limits**: GoodJob pod defaults to 4000Mi/2000Mi (matching Sidekiq stage config). GoodJob uses more memory than Sidekiq at idle (~470Mi vs ~200Mi) due to its scheduler, LISTEN/NOTIFY connection, and probe server. During report parsing, memory spikes require the higher limit. Tunable per-environment via app-interface.

## How it works during transition

- New jobs enqueued by the Kafka consumer go to GoodJob (PostgreSQL) because `ParseReportJob.queue_adapter` is `:good_job`
- The GoodJob pod (`compliance-goodjob`) picks up and processes these jobs
- Any jobs already in Redis are drained by the existing Sidekiq pod
- Once Redis is empty, Sidekiq can be removed in a follow-up

## Out of scope / follow-ups

- Remove Sidekiq gem, Redis gem, `gitlab-sidekiq-fetcher`, and the Sidekiq deployment once Redis is fully drained
- Remove `inMemoryDb: true` from ClowdApp
- Clean up Sidekiq-specific code in `ParseReportJob` (`cancelled?`, `cancel!`, `jid`, `Sidekiq.logger`)
- Remove or migrate Redis cache configuration
- GoodJob web dashboard
- `yabeda-sidekiq` metrics cleanup
- app-interface updates for the new GoodJob deployment

## Secure Coding Checklist
- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/jobs/parse_report_job.rb
+++ b/app/jobs/parse_report_job.rb
@@ -4,6 +4,7 @@
 class ParseReportJob < ApplicationJob
   include Notifications
 
+  self.queue_adapter = :good_job
   self.log_arguments = false
 
   # https://github.com/RoamingNoMaD/yabeda-activejob#custom-tags
@@ -32,6 +33,8 @@ class ParseReportJob < ApplicationJob
   end
 
   def cancelled?
+    return false unless self.class.queue_adapter_name == 'sidekiq'
+
     Sidekiq.redis { |c| c.exists?("cancelled-#{jid}") }
   end
 
--- a/config/initializers/good_job.rb
+++ b/config/initializers/good_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+Rails.application.configure do
+  config.good_job.execution_mode = :external
+  config.good_job.max_threads = ENV.fetch('GOOD_JOB_MAX_THREADS', 5).to_i
+  config.good_job.shutdown_timeout = 25
+  config.good_job.preserve_job_records = lambda { |_job, _error, error_event|
+    error_event == :unhandled
+  }
+  config.good_job.cleanup_preserved_jobs_before_seconds_ago = 14.days.to_i
+  config.good_job.on_thread_error = ->(exception) { Rails.error.report(exception) }
+end
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -580,6 +580,103 @@ objects:
           requests:
             cpu: ${CPU_REQUEST_CONSUMER}
             memory: ${MEMORY_REQUEST_CONSUMER}
+    - name: goodjob
+      minReplicas: ${{REPLICAS_GOODJOB}}
+      deploymentStrategy:
+        privateStrategy: Recreate
+      podSpec:
+        image: ${IMAGE}:${IMAGE_TAG}
+        initContainers:
+        - command: ["/bin/sh"]
+          args: ["-c", "$HOME/scripts/abort_if_pending_migrations.sh"]
+          inheritEnv: true
+        resources:
+          limits:
+            cpu: ${CPU_LIMIT_GOODJOB}
+            memory: ${MEMORY_LIMIT_GOODJOB}
+          requests:
+            cpu: ${CPU_REQUEST_GOODJOB}
+            memory: ${MEMORY_REQUEST_GOODJOB}
+        livenessProbe:
+          httpGet:
+            path: /status
+            port: ${{GOOD_JOB_PROBE_PORT}}
+          timeoutSeconds: 5
+          periodSeconds: 30
+        readinessProbe:
+          httpGet:
+            path: /status
+            port: ${{GOOD_JOB_PROBE_PORT}}
+          timeoutSeconds: 5
+          periodSeconds: 30
+        env:
+        - name: APPLICATION_TYPE
+          value: compliance-goodjob
+        - name: GOOD_JOB_PROBE_PORT
+          value: "${GOOD_JOB_PROBE_PORT}"
+        - name: RAILS_ENV
+          value: "${RAILS_ENV}"
+        - name: RAILS_LOG_TO_STDOUT
+          value: "${RAILS_LOG_TO_STDOUT}"
+        - name: PATH_PREFIX
+          value: "${PATH_PREFIX}"
+        - name: APP_NAME
+          value: "${APP_NAME}"
+        - name: GOOD_JOB_MAX_THREADS
+          value: "${GOOD_JOB_MAX_THREADS}"
+        - name: POSTGRESQL_MAX_CONNECTIONS
+          value: "${POSTGRESQL_MAX_CONNECTIONS}"
+        - name: MALLOC_ARENA_MAX
+          value: '2'
+        - name: RUBY_YJIT_ENABLE
+          value: "${RUBY_YJIT_ENABLE}"
+        - name: SECRET_KEY_BASE
+          valueFrom:
+            secretKeyRef:
+              name: compliance-backend
+              key: secret_key_base
+        - name: SETTINGS__SLACK_WEBHOOK
+          valueFrom:
+            secretKeyRef:
+              name: compliance-backend
+              key: slack_webhook
+              optional: true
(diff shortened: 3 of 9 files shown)


config/initializers/good_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+GoodJob.configure do |config|
```
romanblanco: `GoodJob.configure` does not exist in GoodJob v4

https://github.com/bensheldon/good_job/blob/befffb0c126162f51a98fa705cc1479408551a39/README.md#configuration-options

```suggestion
Rails.application.configure do
  config.good_job.execution_mode = Rails.env.test? ? :inline : :external
  config.good_job.max_threads = ENV.fetch('GOOD_JOB_MAX_THREADS', 5).to_i
  # ...
```

deploy/clowdapp.yaml
```diff
@@ -497,6 +497,101 @@ objects:
...
+        - name: RAILS_LOG_TO_STDOUT
+          value: "${RAILS_LOG_TO_STDOUT}"
+        - name: PATH_PREFIX
+          value: "${PATH_PREFIX}"
+        - name: APP_NAME
+          value: "${APP_NAME}"
+        - name: GOOD_JOB_MAX_THREADS
+          value: "${GOOD_JOB_MAX_THREADS}"
+        - name: POSTGRESQL_MAX_CONNECTIONS
+          value: "${POSTGRESQL_MAX_CONNECTIONS}"
```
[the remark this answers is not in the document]
megaslurpie: postgres connections is set to 10 in stage and prod so goodjob should be fine for now

spec/jobs/parse_report_job_spec.rb
```diff
@@ -25,15 +25,11 @@
...
-  let(:sidekiq_logger) { instance_double(Logger, error: nil, info: nil) }
-
   before do
     allow(SafeDownloader).to receive(:download_reports)
       .with('', ssl_only: Settings.report_download_ssl_only)
       .and_return([Faker::Lorem.word])
-    allow(Sidekiq).to receive(:redis).and_return(false)
-    allow(job).to receive(:jid).and_return('1')
-    allow(Sidekiq).to receive(:logger).and_return(sidekiq_logger)
+    allow(job).to receive(:job_id).and_return('1')
```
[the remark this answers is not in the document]
megaslurpie: think this is minor

config/application.rb
```diff
@@ -30,18 +31,12 @@ module ComplianceBackend
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 8.0
+    config.autoload_lib(ignore: %w[assets tasks])
```
RoamingNoMaD: Why was this change added? 🤔 

Anyways I'd leave the comment that was above the method call. At least the `Insights is also required explicitly...`. As it's something that is app-specific.

config/application.rb
```diff
@@ -15,6 +15,7 @@
 # require 'action_view/railtie'
 # require 'action_cable/engine'
 # require 'sprockets/railtie'
+require 'rails/test_unit/railtie'
```
RoamingNoMaD: Is this truly necessary for this PR? 🤔 

If I am not mistaken this is for Minitests which we're deleting soon. (We only use Rspec now)

app/jobs/parse_report_job.rb
```diff
@@ -11,14 +11,7 @@ def yabeda_tags(_report_blob, message)
     { qe: OpenshiftEnvironment.qe_account?(message['org_id']) }
   end
 
-  # FIXME: compatibility method between Sidekiq and ActiveJob, remove after migration to GoodJob
-  def jid
-    provider_job_id || job_id
-  end
-
```
RoamingNoMaD, line 18: If we change the code that is used by Sidekiq at the moment will it not stop functioning? 

I'd feel safer with some duplicate code for the time being so Sidekiq is untouched until Redis is empty.
We can always rename classes after we see that Sidekiq has stopped processing.

If you agree on this, please create a ticket in the epic of Redis removal so we don't forget to excise Sidekiq when the time comes. Sorry for not creating it when I planned this process.

deploy/clowdapp.yaml
```diff
@@ -497,6 +497,101 @@ objects:
...
+        livenessProbe:
+          httpGet:
+            path: /status
+            port: 7001
+          timeoutSeconds: 5
+          periodSeconds: 30
+        readinessProbe:
+          httpGet:
+            path: /status
+            port: 7001
```
RoamingNoMaD: Could we parameterize and set these in `clowdapp.yaml`? 
You already have an env var in `entrypoint.sh`, it's just not being set anywhere. (`GOOD_JOB_PROBE_PORT`)

spec/jobs/parse_report_job_spec.rb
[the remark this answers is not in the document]
RoamingNoMaD: It's truly minor, but unlike Sidekiq GoodJob uses a job ID format of UUID.
We should try to match reality with `Faker::Internet.uuid` as that is what GoodJob will produce.

RoamingNoMaD (changes requested):
I think we should look at changing the code that influences Sidekiq a bit more as it could break already scheduled jobs in Redis. 

Also there's the consideration of the retention table size, since now we're passing report blobs as parameters instead of passing the URLs to them. 

As we've decided on Slack, we should just configure GoodJob to mimic our config of Sidekiq. Only retain failed jobs. 
As described on the Slack thread, we should use the lambda form for this, as the other one is potentially buggy as described in https://github.com/bensheldon/good_job/issues/1706.

The second option might have to be reverted and it's due to a bad decision on my part. I am sorry if it turns out I wasted your time...

config/application.rb
megaslurpie: Changing the goodjob changed the initialization order so that libs/exception.rb was not loaded due to eager loading. It seems that this was issue before this PR but we just got lucky that it loaded correctly

config/application.rb
megaslurpie: nope, it is already in master, my claude must have not rebased this pr correctly

app/jobs/parse_report_job.rb
megaslurpie: https://redhat.atlassian.net/browse/RHINENG-28932

deploy/clowdapp.yaml
megaslurpie: updated

RoamingNoMaD (approved):
LGTM, thank you <3

romanblanco (approved):

PR #2844, opened 2026-07-02
feat(systems): add tombstoning and cleanup for inventory events

petrblaho:
## Jira Issues
- [RHINENG-23282](https://issues.redhat.com/browse/RHINENG-23282) - Implement soft delete logic for inventory events
- [RHINENG-23280](https://issues.redhat.com/browse/RHINENG-23280) - Add Inventory update and delete events consumer

## Description
Implements soft-delete/tombstoning logic for the local `systems` table when processing Kafka inventory events, preventing late/out-of-order delete/update messages from incorrectly recreating deleted systems. 

Also introduces a dedicated `systems:cleanup` Rake task with three configurable subtasks:
1. **`deleted`**: Purges old soft-deleted tombstones (older than configured retention, e.g., 14 days) and their associated compliance records (test results, policy associations).
2. **`stale`**: Stubs stale system identification (disabled by default, runs only when `STALE_CLEANUP_ENABLED` is `'true'`).
3. **`filter`**: Uses four highly readable distinct SQL SELECT statements to find and purge existing systems in the local database that do not match our current Kafka consumer filter eligibility criteria (missing/invalid `insights_id`, CentOS OS, Edge host type, or BootC systems).

## Key Changes
- **Model:** Added `default_scope { where(deleted_at: nil) }` to `KafkaSystem`.
- **Consumer:** Updated `InventoryEventsConsumer` to run `Kafka::SystemRemover` immediately before `Kafka::DeletedSystemCleaner` for delete events.
- **Service (Tombstone):** Created `Kafka::SystemRemover` to write the soft-deletion timestamp to `deleted_at`.
- **Service (Upsert):** Enhanced `Kafka::SystemImporter#upsert_system` to set `deleted_at: nil` and updated the SQL UPSERT conditional check to support resurrection or ignore older, out-of-order messages.
- **Rake Task:** Added the multi-stage, batched `systems:cleanup` task.
- **Testing:** Comprehensive test suite in `spec/tasks/cleanup_systems_spec.rb`, `spec/services/kafka/system_remover_spec.rb`, and updated specs in `system_importer_spec.rb` and `deleted_system_cleaner_spec.rb`.

## Secure Coding Checklist
- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [x] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

- Added soft-delete/tombstoning for Kafka inventory `systems` to prevent out-of-order delete/update events from recreating deleted rows.
- Updated `KafkaSystem` to exclude soft-deleted records by default (`default_scope { where(deleted_at: nil) }`).
- Updated delete processing in `InventoryEventsConsumer` to run `Kafka::SystemRemover` (sets `deleted_at`) before `Kafka::DeletedSystemCleaner`.
- Enhanced `Kafka::SystemImporter#upsert_system` to clear `deleted_at` on valid upserts and adjust UPSERT conflict rules to support resurrection while ignoring older messages.
- Added `systems:cleanup` Rake task with `deleted`, `stale` (optional), and `filter` subtasks to purge old tombstones, optionally remove stale systems, and hard-delete locally ineligible systems based on current eligibility rules.
- Added/extended specs covering importer resurrection vs stale ignoring, `Kafka::SystemRemover` behavior (including invalid timestamp handling and error logging), and `systems:cleanup` for each subtask.

[RHINENG-23282]: https://redhat.atlassian.net/browse/RHINENG-23282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHINENG-23280]: https://redhat.atlassian.net/browse/RHINENG-23280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/kafka_system.rb
+++ b/app/models/kafka_system.rb
@@ -3,4 +3,6 @@
 # Temporary model for ingesting inventory data directly to compliance db
 class KafkaSystem < ApplicationRecord
   self.table_name = 'systems'
+
+  default_scope { where(deleted_at: nil) }
 end
--- a/app/services/kafka/system_remover.rb
+++ b/app/services/kafka/system_remover.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Kafka
+  # Service for marking a KafkaSystem as soft-deleted (tombstoned) in the systems table
+  class SystemRemover
+    def initialize(message, logger)
+      @message = message
+      @logger = logger
+      @id = @message.dig('id')
+      @org_id = @message.dig('org_id')
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def remove_system
+      computed_timestamp = delete_timestamp
+      # rubocop:disable Rails/SkipsModelValidations
+      result = KafkaSystem.where(id: @id)
+                          .where('updated IS NULL OR updated < ?', computed_timestamp)
+                          .update_all(deleted_at: computed_timestamp)
+      # rubocop:enable Rails/SkipsModelValidations
+
+      if result.zero?
+        @logger.info("[Kafka::SystemRemover] Ignored stale delete event or no active system found for ID #{@id}")
+      else
+        @logger.audit_success("[Kafka::SystemRemover] Soft-deleted system #{@id}")
+      end
+    rescue StandardError => e
+      @logger.audit_fail("[Kafka::SystemRemover] Failed to soft-delete system #{@id}: #{e.message}")
+      raise e
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    private
+
+    # rubocop:disable Metrics/MethodLength
+    def delete_timestamp
+      timestamp_str = @message.dig('updated') || @message.dig('timestamp')
+      begin
+        parsed_time = Time.zone.parse(timestamp_str) if timestamp_str
+        if timestamp_str && parsed_time.nil?
+          @logger.warn("[Kafka::SystemRemover] Failed to parse timestamp '#{timestamp_str}' " \
+                       "for system #{@id}, falling back to current time")
+        end
+        parsed_time || Time.current
+      rescue StandardError => e
+        @logger.warn("[Kafka::SystemRemover] Error parsing timestamp '#{timestamp_str}' " \
+                     "for system #{@id}: #{e.message}")
+        Time.current
+      end
+    end
+    # rubocop:enable Metrics/MethodLength
+  end
+end
--- a/lib/tasks/cleanup_systems.rake
+++ b/lib/tasks/cleanup_systems.rake
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+# Service class for cleaning up stale, deleted, and ineligible systems
+class SystemsCleaner
+  def initialize(batch_size:, logger:)
+    @batch_size = batch_size
+    @logger = logger
+  end
+
+  def run(subtasks)
+    start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
+    if subtasks.include?('deleted')
+      @logger.info("Running subtask 'deleted'...")
+      deleted_ids = deleted_system_ids
+      delete_systems_in_batches(deleted_ids, 'deleted') if deleted_ids.any?
+    end
+
+    elapsed = (Process.clock_gettime(Process::CLOCK_MONOTONIC) - start_time).round(2)
+    @logger.info("systems:cleanup completed in #{elapsed}s.")
+  end
+
+  private
+
+  # Subtask: Deleted (Tombstones)
+  # Hard-deletes systems soft-deleted (deleted_at IS NOT NULL) older than retention days.
+  def deleted_system_ids
+    retention_days = ENV.fetch('DELETED_RETENTION_DAYS', 14).to_i
+    cutoff_time = retention_days.days.ago
+
+    sql = <<~SQL
+      SELECT id FROM systems
+      WHERE deleted_at IS NOT NULL
+        AND deleted_at < '#{cutoff_time.iso8601}'
+    SQL
+    ActiveRecord::Base.connection.select_values(sql)
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  # Batch deletion of identified systems and their related records
+  def delete_systems_in_batches(system_ids, subtask_name)
+    total = system_ids.size
+    @logger.info("Found #{total} systems to clean up for subtask '#{subtask_name}'.")
+
+    system_ids.each_slice(@batch_size) do |batch_ids|
+      ActiveRecord::Base.transaction do
+        # Purge related records
+        results_count = V2::HistoricalTestResult.where(system_id: batch_ids).delete_all
+        policy_systems_count = V2::PolicySystem.where(system_id: batch_ids).delete_all
+
+        # Purge the systems (bypassing default_scope deleted_at: nil)
+        systems_count = KafkaSystem.unscoped.where(id: batch_ids).delete_all
+
+        @logger.info("Deleted batch: #{systems_count} systems, #{results_count} test results, " \
+                     "#{policy_systems_count} policy-system associations.")
+      end
+    end
+    @logger.info("Completed subtask '#{subtask_name}'. Total systems cleaned up: #{total}")
+  end
+  # rubocop:enable Metrics/MethodLength
+end
+
+namespace :systems do
+  desc 'Cleanup stale, soft-deleted, and ineligible systems'
+  task cleanup: :environment do
+    subtasks = ENV.fetch('SUBTASKS', 'deleted').split(',')
+    batch_size = ENV.fetch('BATCH_SIZE', 1000).to_i
+
+    logger = Logger.new($stdout)
+    logger.info("systems:cleanup started. Subtasks: #{subtasks.join(', ')} | BATCH_SIZE: #{batch_size}")
+
+    SystemsCleaner.new(batch_size: batch_size, logger: logger).run(subtasks)
+  end
+end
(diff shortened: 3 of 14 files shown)


megaslurpie (changes requested):
left a few comments, mostly around the default_scope interaction with DeletedSystemCleaner and the backfill

app/models/kafka_system.rb
```diff
@@ -3,4 +3,6 @@
 # Temporary model for ingesting inventory data directly to compliance db
 class KafkaSystem < ApplicationRecord
   self.table_name = 'systems'
+
+  default_scope { where(deleted_at: nil) }
```
megaslurpie, line 7: i think this default_scope might cause `DeletedSystemCleaner#remove_kafka_system` to no-op after `SystemRemover` runs. `SystemRemover` sets `deleted_at`, then `KafkaSystem.where(id: @id).delete_all` i believe gets `AND deleted_at IS NULL` appended from default_scope and would match zero rows. the KafkaSystem row might stay as a tombstone uless the cleanup rake task runs.

lib/tasks/cleanup_systems.rake
```diff
@@ -0,0 +1,237 @@
...
+    subtasks = ENV.fetch('SUBTASKS', 'deleted,stale,filter').split(',')
+    batch_size = Integer(ENV.fetch('BATCH_SIZE', '1000'), 10)
+    raise ArgumentError, 'BATCH_SIZE must be positive' unless batch_size.positive?
+
+    logger = Logger.new($stdout)
+    logger.info("systems:cleanup started. Subtasks: #{subtasks.join(', ')} | BATCH_SIZE: #{batch_size}")
+
+    SystemsCleaner.new(batch_size: batch_size, logger: logger).run(subtasks)
+  end
+end
```
megaslurpie, line 237: i think the backfill could resurrect tombstones. line 90 of systems_backfill.rake inserts `deleted_at = NULL`, line 103 sets it on conflict, and the WHERE on line 104 only checks `systems.updated` not `deleted_at`. SystemImporter's upsert seems to handle this corectly in this PR, backfill might need the same treatment.

app/services/kafka/system_remover.rb
```diff
@@ -0,0 +1,32 @@
...
+      # rubocop:enable Rails/SkipsModelValidations
+      # rubocop:enable Layout/LineLength
+    end
+
+    private
+
+    def delete_timestamp
+      timestamp_str = @message.dig('updated') || @message.dig('timestamp')
+      Time.zone.parse(timestamp_str)
+    rescue StandardError
```
megaslurpie: i think this silently swallows parse errors and falls back to `Time.current` with no log. wrong `deleted_at` could mess up resurrection logic since the upsert compares `deleted_at < EXCLUDED.updated`

app/services/kafka/system_remover.rb
```diff
@@ -0,0 +1,32 @@
...
+  # Service for marking a KafkaSystem as soft-deleted (tombstoned) in the systems table
+  class SystemRemover
+    def initialize(message, logger)
+      @message = message
+      @logger = logger
+      @id = @message.dig('id')
+      @org_id = @message.dig('org_id')
+    end
+
+    def remove_system
```
megaslurpie, line 14: `@logger` is set but never used, no way to tell if a soft delete happend or was skipped. SystemImporter and DeletedSystemCleaner both log their outcomes

app/models/kafka_system.rb
petrblaho: Great catch! In commit `512f8d56`, I've delegated the soft-delete task entirely to the new `SystemRemover` service, keeping `DeletedSystemCleaner` focused exclusively on cleaning up associated/related data (such as historical test results and policy systems) without deleting the system row itself. This cleanly resolves the `default_scope` conflict.

lib/tasks/cleanup_systems.rake
petrblaho: Exactly! I've updated the `systems_backfill.rake` script in commit `27db7cb7` (and further refactored in commit `16bb453f`) to use `COALESCE(systems.deleted_at, systems.updated) < EXCLUDED.updated`. This ensures that backfill updates cannot accidentally resurrect older soft-deleted tombstones.

app/services/kafka/system_remover.rb
petrblaho: Agreed. In commit `937f8699`, I added explicit error logging to warn when parsing fails, so we don't silently swallow parse errors while safely falling back to `Time.current`.

app/services/kafka/system_remover.rb
petrblaho: Good call. In commit `705e2a06`, I added proper audit and info logging to `SystemRemover` (similar to `SystemImporter` and `DeletedSystemCleaner`), making full use of the `@logger` instance.

spec/services/kafka/system_remover_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    it 'logs a warning and falls back to current time' do
+      expect(Karafka.logger).to receive(:warn).with(/Failed to parse timestamp 'invalid-timestamp-string'/)
+      service.remove_system
+      system = KafkaSystem.unscoped.find(system_id)
+      expect(system.deleted_at).not_to be_nil
+    end
+  end
+
+  context 'when existing system has a newer updated timestamp' do
```
RoamingNoMaD, line 62: Just some questions:

Is this a realistic scenario? Would such a succession of messages make sense?
I understood a delete message a final thing, no matter if we receive updates after. 
My logic is that even if the messages get processed in an incorrect order, once we receive a delete, it's deleted from Inventory's perspective and if a customer was to re-register, the new record would have a different ID, so resurrection does not make sense. Am I wrong here?

lib/tasks/cleanup_systems.rake
```diff
@@ -0,0 +1,237 @@
...
+              else
+                query
+              end
+
+      eligible_ids = query.pluck(:id)
+      return 0 if eligible_ids.empty?
+
+      # Purge related records
+      results_count = V2::HistoricalTestResult.where(system_id: eligible_ids).delete_all
+      policy_systems_count = V2::PolicySystem.where(system_id: eligible_ids).delete_all
```
RoamingNoMaD, line 201: DeletedSystemCleaner does this immediately when a deleted message arrives. Wouldn't this make the other service useless?

spec/services/kafka/system_remover_spec.rb
petrblaho: Normally yes, but we have to handle re-registration and Kafka lag:

1. **Re-registration:** When a machine is unregistered and then re-registered (like running `insights-client --register`), it keeps the same UUID. Because the host is active again, the legitimately newer update message will have `excluded.updated > systems.deleted_at` so we resurrect it (set `deleted_at = nil`) to make it visible.
2. **Kafka lag:** If a delayed, stale update message from before the deletion arrives late, we compare timestamps (`excluded.updated < systems.deleted_at`) and safely ignore it.

lib/tasks/cleanup_systems.rake
petrblaho: It looks redundant, but we need it for two reasons:

1. **`stale` and `filter` tasks:** These subtasks hard-delete systems that never received a Kafka delete event (for inactivity or failing consumer filters). Since `DeletedSystemCleaner` is never run for them, the Rake task is the only way to delete their associated test results and policies to prevent orphaned rows.
2. **Fail-safe:** For `deleted` tombstones, this acts as a final garbage collector if the initial delete event consumer crashed or failed to clean related records mid-process.

RoamingNoMaD (approved):

petrblaho:
I need to add recurring job configuration for this, please wait with merge.

edit: added

petrblaho:
rebased and re-commited so the history is "nicer"

lib/tasks/cleanup_systems.rake
```diff
@@ -0,0 +1,237 @@
...
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  # Batch deletion of identified systems and their related records
+  # rubocop:disable Metrics/AbcSize, Metrics/BlockLength
+  # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
+  def delete_systems_in_batches(batch_ids, subtask_name, cutoff_time: nil)
+    ActiveRecord::Base.transaction do
+      # Re-select and lock candidates in transaction to avoid TOCTOU race conditions (e.g. resurrection)
+      query = KafkaSystem.unscoped.where(id: batch_ids).lock('FOR UPDATE')
```
megaslurpie: Should this be FOR UPDATE OF systems ? claude is saying current query will have issues locking inventory.hosts table?

app/services/kafka/system_remover.rb
```diff
@@ -0,0 +1,55 @@
...
+      @message = message
+      @logger = logger
+      @id = @message.dig('id')
+      @org_id = @message.dig('org_id')
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def remove_system
+      computed_timestamp = delete_timestamp
+      # rubocop:disable Layout/LineLength
```
romanblanco: Disabling the rule should not be needed here

```suggestion

```

petrblaho: Removed. Thanks for catching that.

lib/tasks/cleanup_systems.rake
petrblaho: Good catch, updated to lock exclusively on systems (`FOR UPDATE OF systems`). Thank you.

romanblanco (approved):
LGTM 👍

megaslurpie (approved):

PR #2845, opened 2026-07-03
refactor(app): remove V2:: references from application code

RoamingNoMaD:


spec/policies/value_definition_policy_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe RulePolicy do
```
romanblanco:
```suggestion
describe ValueDefinitionPolicy do
```

RoamingNoMaD: Good catch, I'll double check my changes too. Thanks :)

romanblanco:
@RoamingNoMaD the https://github.com/RedHatInsights/compliance-backend/pull/2845/changes/6afde49ebe95d3620d174c0b7f0350710a209920 commit's name does not match what's happening in the commit. Could you update it?

RoamingNoMaD:
@romanblanco It was a mistake in rebasing. It should be fixed now.

db/seeds.dev.rb
```diff
@@ -1,117 +1,54 @@
...
+        t = Tailoring.for_policy(policy, minor)
+        t.save ? t : nil
+      rescue Exceptions::OSMinorVersionNotSupported
+        nil
+      end
+      next unless tailoring
+
+      host_count.times do
+        FactoryBot.create(
+          :test_result,
```
romanblanco, line 39: This seems to cause issues when seeding DB on a fresh setup.

```
PG::ObjectNotInPrerequisiteState: ERROR: cannot insert into view "hosts"                     
DETAIL: Views that do not select from a single table or view are not automatically updatable.
```

RoamingNoMaD: Fixed in new version, thanks for trying it out.

romanblanco (approved):

PR #2849, opened 2026-07-07
fix(DB): make TestResults supported by default in function

RoamingNoMaD:
- Default `TestResults.supported` to `true` when omitted.
- Add a versioned trigger function and migration to apply the updated `test_results_insert` behavior.
- Regenerate the database schema and update GoodJob index definitions.

### DIFF
--- a/db/functions/test_results_insert_v02.sql
+++ b/db/functions/test_results_insert_v02.sql
@@ -1,0 +1,32 @@
+CREATE OR REPLACE FUNCTION test_results_insert() RETURNS trigger LANGUAGE plpgsql AS
+$func$
+DECLARE result_id uuid;
+BEGIN
+    INSERT INTO "historical_test_results" (
+      "tailoring_id",
+      "report_id",
+      "system_id",
+      "start_time",
+      "end_time",
+      "score",
+      "supported",
+      "failed_rule_count",
+      "created_at",
+      "updated_at"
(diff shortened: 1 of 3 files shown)


petrblaho (approved):

db/functions/test_results_insert_v02.sql
```diff
@@ -0,0 +1,32 @@
...
+      "created_at",
+      "updated_at"
+    ) VALUES (
+      NEW."tailoring_id",
+      NEW."report_id",
+      NEW."system_id",
+      NEW."start_time",
+      NEW."end_time",
+      NEW."score",
+      COALESCE(NEW."supported", TRUE),
```
opacut, line 23: If I understand correctly, then this is to prevent the insertion of a NULL. I would recommend writing a test for this, so that if this gets changed in the future for any reason, a test will remind us that we need to do this

opacut (approved):

db/functions/test_results_insert_v02.sql
RoamingNoMaD: We don't really test DB functions. :thinking: 
This is more of a fallback and there's no way for a nil to get in there ATM. 

I am a big against testing every minor thing as our tests run long already. Unit tests cover the Report parser so I feel safe with this merge.

PR #2857, opened 2026-07-13
fix(migration): comment out model references in old migrations

RoamingNoMaD:
- Commented out legacy model/data backfills and cleanup operations in old migrations.
- Prevented historical migrations from deleting `Revision` records.
- Disabled rollback logic that depended on application models or recreated benchmark records.
- Removed the `test_results.failed_rule_count` backfill while retaining the column migration.


petrblaho (approved):
LGTM

PR #2858, opened 2026-07-13
fix(rules): use plaintext banner value definitions on banner rules

LightOfHeaven1994:
Fixes https://redhat.atlassian.net/browse/RHINENG-22924 - users could not edit `Login Banner` value.

Starting with scap-security-guide 0.1.81 [JIRA](https://redhat.atlassian.net/browse/RHEL-118499) & [PR #14371](https://github.com/ComplianceAsCode/content/pull/14371)), SSG introduced plaintext *_banner_contents value definitions for remediation, while OVAL checks still export regex *_banner_text refs. During SSG import, openscap_parser only reads OVAL check-export refs, so banner rules were persisted with value_checks pointing at the regex value definition — which the console cannot meaningfully edit.

This change maps banner rules to the plaintext *_banner_contents value definitions when they exist in the security guide, so the UI will allow patching values for banners.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

- Map regex banner references to editable plaintext banner contents when available.
- Preserve fallback behavior when plaintext banner definitions are unavailable.
- Add coverage for login, remote login, MOTD, and dconf banner mappings.

### DIFF
--- a/app/services/concerns/xccdf/rules.rb
+++ b/app/services/concerns/xccdf/rules.rb
@@ -4,12 +4,13 @@ module Xccdf
   # Methods related to parsing rules
   module Rules
     extend ActiveSupport::Concern
+    include BannerValueMappings
 
     included do
       def rules
         @rules ||= @op_rules.each_with_index.map do |op_rule, idx|
           rule_group = rule_group_for(ref_id: op_rule.parent_id)
-          value_checks = op_rule.values.map { |value_ref_id| value_definition_for(ref_id: value_ref_id).id }
+          value_checks = value_checks_for(op_rule)
 
           ::Rule.from_parser(
             op_rule,
@@ -34,6 +35,19 @@ module Xccdf
       end
 
       private
+
+      def value_checks_for(op_rule)
+        op_rule.values.map do |value_ref_id|
+          value_definition_for(ref_id: remediable_banner_value_ref_id(value_ref_id)).id
+        end.uniq
+      end
+
+      def remediable_banner_value_ref_id(value_ref_id)
+        contents_ref_id = BANNER_TEXT_TO_CONTENTS[value_ref_id]
+        return value_ref_id unless contents_ref_id
+
+        value_definition_for(ref_id: contents_ref_id) ? contents_ref_id : value_ref_id
+      end
 
       def new_rules
         @new_rules ||= rules.select(&:new_record?)
--- a/spec/services/concerns/xccdf/rules_spec.rb
+++ b/spec/services/concerns/xccdf/rules_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe Xccdf::Rules do
 
       # NOTE: taken from `app/services/concerns/xccdf/value_definitions.rb`
       def value_definition_for(ref_id:)
-        @value_lookup.fetch(ref_id)
+        @value_lookup[ref_id]
       end
     end.new(
       security_guide: security_guide,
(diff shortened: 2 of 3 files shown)


spec/services/concerns/xccdf/rules_spec.rb
```diff
@@ -141,5 +141,92 @@ def value_definition_for(ref_id:)
...
+          )
+        ]
+      end
+
+      it 'keeps the regex banner value as a fallback' do
+        rule = service.rules.first
+
+        expect(rule.value_checks).to eq([banner_text_definition.id])
+      end
+    end
```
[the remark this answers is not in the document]
LightOfHeaven1994: Good catch! fixed

app/services/concerns/xccdf/rules.rb
```diff
@@ -35,6 +36,19 @@ def save_rules
...
+          value_definition_for(ref_id: remediable_banner_value_ref_id(value_ref_id)).id
+        end.uniq
+      end
+
+      def remediable_banner_value_ref_id(value_ref_id)
+        contents_ref_id = BANNER_TEXT_TO_CONTENTS[value_ref_id]
+        return value_ref_id unless contents_ref_id
+
+        value_definition_for(ref_id: contents_ref_id)&.ref_id || value_ref_id
+      end
```
RoamingNoMaD, line 50:
```suggestion
      def remediable_banner_value_ref_id(value_ref_id)
         contents_ref_id = BANNER_TEXT_TO_CONTENTS[value_ref_id]
         return value_ref_id unless contents_ref_id
         value_definition_for(ref_id: contents_ref_id) ? contents_ref_id : value_ref_id
     end
```

spec/services/concerns/xccdf/rules_spec.rb
```diff
@@ -141,5 +141,92 @@ def value_definition_for(ref_id:)
       expect(results.length).to eq(2)
       expect(results.map(&:ref_id)).to contain_exactly(*op_rules.map(&:id))
     end
+
+    context 'when a rule references a regex banner value' do
```
RoamingNoMaD, line 145: What if a rule references both a regex and a legacy one? 
I think it's what you're trying to do with the `uniq` call. 
If that is a realistic expectation to have, please write a test for this case.

Also this test seems a little spaghetti. Could you please work with the contexts above and try to stay as DRY as possible?

RoamingNoMaD (commented):
Nothing too serious, if you disagree with my review, we could merge it right away.
Thanks for the contribution :)

app/services/concerns/xccdf/rules.rb
LightOfHeaven1994: @RoamingNoMaD great suggestion! applied

spec/services/concerns/xccdf/rules_spec.rb
LightOfHeaven1994: @RoamingNoMaD added a test. I double checked and seems it can't happen the way it's implemented now. But I guess we can keep this safety check just in case (it doesn't make our code more complicated). Since this task is too specific, I don't wanna hit issues in the future.

also restructured text.

LightOfHeaven1994:
@RoamingNoMaD thanks for the review! I haven't noticed you did that 3 weeks ago 😄 card is in the sprint so noticed it now. Updated based on your comments

romanblanco (approved):

RoamingNoMaD (approved):
Thanks for the reaction, LGTM <3

PR #2863, opened 2026-07-14
perf(parsing): use streaming XML reader to reduce memory usage

megaslurpie:
## Summary

Use a streaming `Nokogiri::XML::Reader` to extract only the data needed for report parsing from XCCDF reports, instead of building a full DOM tree for the entire document.

- **New `XccdfReportExtractor` service:** streams through the XML and extracts the `<Benchmark>` id/version and the `<TestResult>` subtree, then assembles a minimal synthetic document (~12-15% of the original size).
- **Transparent integration:** extraction happens inside `XccdfReportParser#initialize`, so all callers (Kafka validation, ParseReportJob persistence) benefit without changes.
- **No behavioral change:** the existing parser receives a smaller document but produces identical results — benchmark id, version, profile id, score, rule results are all preserved. Groups, profiles, rules, and values return empty arrays, which matches their current effective usage (loaded but never consumed during report parsing).
- **DatastreamImporter unaffected:** SSG imports use a separate class (`DatastreamFile`) and never touch the extractor.

RHINENG-1998

## Out of scope / follow-ups

- Extract once in `Kafka::ReportParser#validate_and_enqueue` and enqueue the minimal XML to reduce GoodJob/queue payload size by ~85%

> Draft: opened for early review and CI validation.

## Secure Coding Checklist
- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

- Replace full-DOM XCCDF parsing with streaming extraction.
- Preserve benchmark metadata, scores, and rule results in a reduced document.
- Preserve namespaces and complete `Benchmark` opening tags.
- Handle extraction errors through `XccdfReportParser`.
- Add extraction, validation, error-handling, and idempotency tests.

### DIFF
--- a/app/services/xccdf_report_extractor.rb
+++ b/app/services/xccdf_report_extractor.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+# Streams an XCCDF report and extracts only the Benchmark metadata and TestResult subtree
+class XccdfReportExtractor
+  class ExtractionError < StandardError; end
+
+  def self.extract(xml_string)
+    new(xml_string).extract
+  end
+
+  def initialize(xml_string)
+    @xml = xml_string
+  end
+
+  def extract
+    benchmark_tag = nil
+    benchmark_name = nil
+    benchmark_depth = nil
+    version_xml = nil
+    test_result_xml = nil
+
+    reader = Nokogiri::XML::Reader(@xml)
+    reader.each do |node|
+      next unless node.node_type == Nokogiri::XML::Reader::TYPE_ELEMENT
+
+      if benchmark_depth.nil? && node.local_name == 'Benchmark'
+        benchmark_depth = node.depth
+        benchmark_name = node.name
+        benchmark_tag = build_benchmark_open_tag(node)
+        next
+      end
+
+      next unless benchmark_depth
+      next unless node.depth == benchmark_depth + 1
+
+      case node.local_name
+      when 'version'
+        version_xml ||= node.outer_xml
+      when 'TestResult'
+        test_result_xml = node.outer_xml
+        break
+      end
+    end
+
+    validate!(benchmark_tag, version_xml, test_result_xml)
+    build_minimal_xml(benchmark_tag, benchmark_name, version_xml, test_result_xml)
+  rescue Nokogiri::XML::SyntaxError => e
+    raise ExtractionError, "Malformed XML: #{e.message}"
+  end
+
+  private
+
+  def build_benchmark_open_tag(node)
+    attr_str = node.attributes&.map { |name, value| %(#{name}="#{value}") }&.join(' ')
+    attr_str ? "<#{node.name} #{attr_str}>" : "<#{node.name}>"
+  end
+
+  def validate!(benchmark_tag, version_xml, test_result_xml)
+    raise ExtractionError, 'No <Benchmark> element found in report' unless benchmark_tag
+    raise ExtractionError, 'No <version> element found in report' unless version_xml
+    raise ExtractionError, 'No <TestResult> element found in report' unless test_result_xml
+  end
+
+  def build_minimal_xml(benchmark_tag, benchmark_name, version_xml, test_result_xml)
+    "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n#{benchmark_tag}\n#{version_xml}\n#{test_result_xml}\n</#{benchmark_name}>"
+  end
+end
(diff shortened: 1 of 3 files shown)


app/services/xccdf_report_extractor.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+# Streams an XCCDF report and extracts only the Benchmark metadata and TestResult subtree
+class XccdfReportExtractor
```
RoamingNoMaD, line 6: Could we just enhance the [OpenscapParser](https://github.com/OpenSCAP/openscap_parser) with this instead of creating a new service?
I think it would fit there well. What do you think?

app/services/xccdf_report_extractor.rb
```diff
@@ -0,0 +1,67 @@
...
+  rescue Nokogiri::XML::SyntaxError => e
+    raise ExtractionError, "Malformed XML: #{e.message}"
+  end
+
+  private
+
+  def build_benchmark_open_tag(node)
+    attr_str = node.attributes&.map { |name, value| %(#{name}="#{value}") }&.join(' ')
+    attr_str ? "<#{node.name} #{attr_str}>" : "<#{node.name}>"
+  end
```
RoamingNoMaD: Nokogiri::XML::Reader#attributes returns decoded attribute values (&amp; → &). Re-interpolating without XML-escaping can produce invalid XML if any attribute contains ", <, >, or &. Practically unlikely for XCCDF Benchmark attributes, but IMO we should take care of potential errors like this.

spec/services/xccdf_report_extractor_spec.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe XccdfReportExtractor do
```
RoamingNoMaD, line 5: If Nokogiri::XML::Reader#attributes doesn't include xmlns declarations (they're technically not "attributes" in the XML Namespaces spec), the reconstructed document loses namespace context. The test against the real fixture should catch this — but I'd add an explicit assertion like:

```ruby
it 'preserves namespace declarations' do
  expect(minimal_xml).to include('xmlns="http://checklists.nist.gov/xccdf/1.2"')
end
```

RoamingNoMaD (changes requested):

app/services/xccdf_report_extractor.rb
RoamingNoMaD: Rethinking my comment: this streaming reader should be a wrapper around our parser. No need to merge the two.

app/services/xccdf_report_extractor.rb
megaslurpie: Addressed in b8f64c90.

We stopped rebuilding the Benchmark open tag from decoded Reader attributes. `Reader#attribute_hash` / `#namespaces` call `xmlTextReaderExpand`, which materializes the entire Benchmark subtree and undoes the streaming win.

The open tag is now copied from the source with a quote-aware regex, so original entity escaping stays intact (`&amp;`, quotes, `>` inside attribute values). There is a fallback through `Reader#attribute` (GetAttribute, no expand) plus `CGI.escapeHTML` if the regex cannot match.

Specs cover special characters in attributes and `title="a > b"`.

spec/services/xccdf_report_extractor_spec.rb
megaslurpie: Addressed — added the xmlns assertion you suggested, plus `xml:lang` (the Reader hash was dropping the `xml:` prefix when we reconstructed the tag).

Copying the source open tag keeps `xmlns`, `xmlns:xsi`, and `xml:lang` as written in the report. Would you mind taking another look?

app/services/xccdf_report_extractor.rb
megaslurpie: ^- by grok 4.6 btw

spec/services/xccdf_report_extractor_spec.rb
megaslurpie: ^- by grok 4.6 btw

RoamingNoMaD (approved):
I think this would do :)

Can you please make a follow-up ticket where we would try to enqueue only the useful parts in ParseRerportJob calls? I think we could still optimize here, but I don't expect you specifically to do it.

Thanks for all the work you've done for us <3

megaslurpie:
@RoamingNoMaD added https://redhat.atlassian.net/browse/RHINENG-30211

petrblaho (approved):

PR #2890, opened 2026-07-22
fix(ci): optimize ephemeral deployment probes and init DB pre-check

petrblaho:
## Summary
Fixes 100% of ephemeral test deployment timeouts in Jenkins caused by CPU quota throttling (`67m` limit) during Rails Puma preloading and database init retries.

## Problem
App-interface recently reconciled a `67m` CPU limit override for `compliance-service` in ephemeral (`insights-pr`) test environments. Under this tight CPU quota:
1. **Puma Rails Preload Delay**: Puma preloading Rails across 3 workers takes ~90 seconds. With the default `initialDelaySeconds: 10` + 90s failure window, Kubernetes repeatedly kills Puma via `SIGTERM` (`Container compliance-service failed liveness probe, will be restarted`) right before port 8000 opens, entering an endless restart loop.
2. **Init Container Retries**: `compliance-service-init` runs `scripts/check_migration_status_and_ssg_synced.sh`. When PostgreSQL or `compliance-ssg-service` takes time to start up, every retry boots full Rails (taking ~50-90s per boot), causing 10 retries to waste ~8-10 minutes before the main application container even starts.

Together, these issues caused Bonfire to hit its 20-minute deployment timeout across all recent Jenkins PR runs.

## Solution
1. **Increase Liveness/Readiness Probe Initial Delays** (`deploy/clowdapp.yaml`):
   - Increased `initialDelaySeconds` from `10` to `180` seconds for `compliance-service`.
   - Gives Puma 3 minutes to safely preload Rails and bind to port 8000 on low CPU without being prematurely terminated by Kubernetes.
2. **Fast Database Socket Pre-check** (`scripts/check_migration_status_and_ssg_synced.sh`):
   - Added a lightweight Ruby database connection socket check (using `ClowderCommonRuby` & stdlib `socket`) before executing `bundle exec rake`.
   - Polling the database socket takes ~0.05s per loop, avoiding heavy 90-second Rails reboots during startup connection retries.

## Verification
- Verified on Jenkins build **`PR-2890 #2`** (Ephemeral namespace: `ephemeral-ulb3j1`).
- All ClowdApps (`compliance` 6/6, `rbac` 7/7, `host-inventory` 12/12, etc.) reached **100% READY with 0 container restarts**.
- Bonfire completed deployment cleanly and launched IQE smoke tests (`compliance-iqe`).

- Increased ClowdApp probe delays to 180 seconds to allow Rails/Puma startup under CPU constraints.
- Added database socket readiness checks to reduce unnecessary Rails boot retries while dependencies initialize.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -473,13 +473,13 @@ objects:
         livenessProbe:
           tcpSocket:
             port: web
-          initialDelaySeconds: 10
+          initialDelaySeconds: 180
           periodSeconds: 30
         readinessProbe:
           httpGet:
             path: /api/compliance/v2/status
             port: web
-          initialDelaySeconds: 10
+          initialDelaySeconds: 180
           periodSeconds: 30
         resources:
           limits:
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -565,10 +565,11 @@ objects:
           httpGet:
             path: /
             port: 3000
-          initialDelaySeconds: 30
+          initialDelaySeconds: 180
           timeoutSeconds: 5
           periodSeconds: 30
         readinessProbe:
+          initialDelaySeconds: 180
           timeoutSeconds: 15
           periodSeconds: 30
           exec:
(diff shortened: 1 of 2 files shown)


megaslurpie (approved):
Approving — the shared `initialDelaySeconds: 180` change means prod/stage pods will take up to 3min before liveness/readiness is checked, but per Slack thread this is a known, accepted tradeoff for now to unblock ephemeral deployments. DB socket pre-check and Dockerfile changes look correct.

PR #2908, opened 2026-07-27
ci(image): target intermediate build stage explicitly on PR builds

petrblaho:
Targets the intermediate `build` stage explicitly during PR builds so Buildah queries Quay for Stage 2 (`bundle install`) cache layers.

- Target intermediate `build` stages in PR image builds.
- Push build-stage layers to Quay to enable remote caching for `bundle install`.
- Fetch `origin master` before calculating reproducible build timestamps.
- Apply 30-day expiration labels to PR build-stage images.

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -67,6 +67,7 @@ fi
 CACHE_REPO="quay.io/cloudservices/compliance-backend"
 
 # Determine merge-base with master to extract a consistent commit timestamp for reproducible layer caching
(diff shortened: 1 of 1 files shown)


petrblaho:
Closing as superseded and resolved by PR #2909. PR #2909 enabled `--cache-to` on PR builds and updated `Dockerfile` to use a global `ARG BASE_IMAGE` across build and runtime stages, achieving 100% remote layer cache hits across all worker nodes without requiring an explicit intermediate `--target build` step.

PR #2915, opened 2026-07-30
ci(security): add remote layer cache to platform security scan workflow

petrblaho:
## Summary
- Passes `--cache-from quay.io/cloudservices/compliance-backend` into the `build_arg` parameter of `platsec-security-scan.yml`.
- Enables remote layer caching during container image builds in the ConsoleDot Platform Security Scan workflow, reducing build and scan times.


petrblaho:
Closing. Platform Security Scan Workfow uses Docker instead of Podman as we use in our other jobs where we generate cache layers to Quay. As Docker / BuildKit and Podman / Buildah cache hashes are not compatible pursuing caching in security scan jobs is not worth more time.

PR #2933, opened 2026-08-06
fix(ClowdApp): add kessel to dependency list

RoamingNoMaD:
- Add `kessel` to the ClowdApp dependency list.
- Remove the multi-component `APP_NAME` override from the smoke-test script.


LightOfHeaven1994 (approved):
LGTM!

RoamingNoMaD:
The dependency list works different than expected...

RoamingNoMaD:
We can't drop APP_NAME because that's how bonfire knows which apps to deploy. ClowdApp deps only handle startup order. It all has to stay in the Jenkinsfile if i am not wrong.

PR #2936, opened 2026-08-07
fix(ClowdApp): add kessel to dependency list

RoamingNoMaD:
- Add `kessel-relations`, `kessel-inventory` and `kessel-inventory-consumer` as optional `ClowdApp` dependencies.
- Preserve the environment-defined `APP_NAME` in the Jenkins smoke-test script.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,7 +59,6 @@ pipeline {
                         curl -s ${CICD_URL}/bootstrap.sh > .cicd_bootstrap.sh
                         source ./.cicd_bootstrap.sh
 
-                        export APP_NAME="host-inventory kessel rbac compliance"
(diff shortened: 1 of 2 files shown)


romanblanco (approved):

petrblaho (approved):

RoamingNoMaD:
Using the dep list makes bonfire only resolve top level dependencies. 
`rbac` only deploys `rbac` and not their debezium instance for example.
I'll try to use the same trick in Konflux that we've used in Jenkins by adding the dependencies to the `EXTRA_DEPLOY_ARGS`, like Ingress does:
https://github.com/RedHatInsights/insights-ingress-go/blob/master/pr_check.sh#L15

PR #2942, opened 2026-08-10
feat(System): query HBI for tag scoping in on-premise mode

romanblanco:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

- Add tag filtering to show and aggregate endpoints.
- Use paginated Host Inventory queries in on-premise mode.
- Preserve unfiltered behavior when no tags are provided.
- Permit and document the `tags` query parameter.
- Set the `systems.tags` default to an empty array.
- Update development host tag handling.
- Guard dotenv loading when `Dotenv::Rails` is unavailable.
- Add controller, integration, and Host Inventory test coverage.

### DIFF
--- a/app/controllers/concerns/parameter_handling.rb
+++ b/app/controllers/concerns/parameter_handling.rb
@@ -56,7 +56,10 @@ module ParameterHandling
       ids_only: ParamType.boolean
     }
 
-    permitted_params_for_action :show, id: ParamType.string
+    permitted_params_for_action :show, {
+      id: ParamType.string,
+      tags: ParamType.array(ParamType.string) | ParamType.string
+    }
 
     # Use the params[:parents] configured by the route to construct a permit
     # hash containing each ID passed from the parents of a nested resource.
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -8,6 +8,11 @@ redis:
   cache_password:
 slack_webhook: 'this is set through a env var in Openshift and not shared for security reasons'
 disable_rbac: false
+# Insights-on-Disconnected mode. When true, the app runs as part of an
+# on-premise (IoP) deployment and resolves tag-based host scoping through
+# Host Inventory directly instead of the locally propagated systems.tags
+# column. Overridden via the IOD_MODE env var in non-Clowder environments.
+iod_mode: false
 async: true
 supported_ssg_ds_config: 'ssg-datastreams.yaml'
 supported_ssg_ansible_tasks_config: 'ssg-ansible-tasks.yaml'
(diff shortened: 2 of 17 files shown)


app/controllers/concerns/parameter_handling.rb
```diff
@@ -56,7 +56,10 @@ def permitted_params_for_action(action, params)
       ids_only: ParamType.boolean
     }
 
-    permitted_params_for_action :show, id: ParamType.string
+    permitted_params_for_action :show, {
+      id: ParamType.string,
+      tags: ParamType.array(ParamType.string) | ParamType.string
+    }
```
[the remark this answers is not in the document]
romanblanco: `collection.rb:69` wraps with `Array(permitted_params[:tags]).map(&:to_s)` before `.any?`

config/settings.yml
```diff
@@ -8,6 +8,11 @@ redis:
   cache_password:
 slack_webhook: 'this is set through a env var in Openshift and not shared for security reasons'
 disable_rbac: false
+# Insights-on-Disconnected mode. When true, the app runs as part of an
```
RoamingNoMaD, line 11: Could we call it `Insights on premise` instead of a custom Compliance only name?

db/migrate/20260713120000_fix_systems_tags_default.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class FixSystemsTagsDefault < ActiveRecord::Migration[7.1]
+  def up
+    change_column_default :systems, :tags, from: {}, to: []
+    execute "UPDATE systems SET tags = '[]'::jsonb WHERE tags = '{}'::jsonb"
```
RoamingNoMaD, line 6: I don't think this will update fields that do not contain an empty hash. 
Why is it not reversible? What about the rest of the tag fields that already have data?

romanblanco: This should unify `{}` and `[]` into only `[]`, and in reverse, we can not restore the per-row types.

Also, we already use `[]` everywhere in our environments, so this is more about correcting the default.

config/settings.yml
romanblanco: Thanks, I thought this came from other repo, but looks like it's made up.

db/migrate/20260713120000_fix_systems_tags_default.rb
RoamingNoMaD: Right, thanks for explaining.

petrblaho (approved):

PR #2951, opened 2026-08-17
feat(parsing): flag SSG version mismatches in TestResults

RoamingNoMaD:
https://redhat.atlassian.net/browse/RHINENG-28324

- Flag SSG version mismatches in serialized test results and the API schema.
- Recompute scores and limit rule results to the current rule set when versions differ.
- Use the tailoring security guide to resolve mismatched rule results.
- Skip unknown-rule validation for mismatched versions.
- Persist `mismatched` in historical test results, views, triggers, and migrations.
- Add regression coverage for mismatch detection, scoring, rule resolution, and API output.

### DIFF
--- a/app/services/concerns/xccdf/rule_results.rb
+++ b/app/services/concerns/xccdf/rule_results.rb
@@ -38,9 +38,11 @@ module Xccdf
     private
 
     def rule_ids
-      @rule_ids ||= ::Rule.where(
-        security_guide_id: security_guide.id, ref_id: selected_op_rule_results.map(&:id)
-      ).pluck(:ref_id, :id).to_h
+      @rule_ids ||= begin
+        guide_id = version_mismatched? ? tailoring.profile.security_guide_id : security_guide.id
+        ::Rule.where(security_guide_id: guide_id, ref_id: selected_op_rule_results.map(&:id))
+              .pluck(:ref_id, :id).to_h
+      end
     end
   end
 end
--- a/app/services/concerns/xccdf/rule_results.rb
+++ b/app/services/concerns/xccdf/rule_results.rb
@@ -38,8 +38,9 @@ module Xccdf
     private
 
     def rule_ids
+      guide_id = version_mismatched? ? tailoring.profile.security_guide_id : security_guide.id
       @rule_ids ||= ::Rule.where(
-        security_guide_id: security_guide.id, ref_id: selected_op_rule_results.map(&:id)
+        security_guide_id: guide_id, ref_id: selected_op_rule_results.map(&:id)
       ).pluck(:ref_id, :id).to_h
     end
   end
--- a/app/services/concerns/xccdf/test_result.rb
+++ b/app/services/concerns/xccdf/test_result.rb
@@ -3,14 +3,16 @@
 module Xccdf
   # Methods related to saving TestResult from openscap_parser
   module TestResult
+    EMPTY_SCORE = 0.0
+
     delegate :score, to: :@op_test_result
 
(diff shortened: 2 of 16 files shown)


app/services/concerns/xccdf/test_result.rb
```diff
@@ -33,6 +33,24 @@ def delete_old_test_results
       ::TestResult.where(id: old_ids).delete_all
     end
 
+    def recomputed_score
+      return score unless version_mismatched?
+
+      countable = countable_op_rule_results
+      return 0.0 if countable.empty?
+
+      passed = countable.count { |rr| rr.result == 'pass' }
```
romanblanco:
```suggestion
      passed = countable.count { |rr| rr.result == RuleResult::PASSED }
```

app/services/concerns/xccdf/test_result.rb
```diff
@@ -33,6 +33,24 @@ def delete_old_test_results
       ::TestResult.where(id: old_ids).delete_all
     end
 
+    def recomputed_score
+      return score unless version_mismatched?
+
+      countable = countable_op_rule_results
+      return 0.0 if countable.empty?
```
romanblanco: I'd suggest not to use magical numbers in code and move it into constant.
Also, wouldn't it be better to use `nil` in this case? 


```suggestion
      return nil if countable.empty?
```

app/services/concerns/xccdf/rule_results.rb
```diff
@@ -38,8 +38,9 @@ def test_result_rules_unknown
     private
 
     def rule_ids
+      guide_id = version_mismatched? ? tailoring.profile.security_guide_id : security_guide.id
```
romanblanco: The `guide_id` should be computed inside the block for the memoization not to be recalculated on each call:

```ruby
def rule_ids
  @rule_ids ||= begin
    guide_id = version_mismatched? ? tailoring.profile.security_guide_id : security_guide.id
    ::Rule.where(security_guide_id: guide_id, ref_id: selected_op_rule_results.map(&:id))
          .pluck(:ref_id, :id).to_h
  end
end
```

app/services/concerns/xccdf/test_result.rb
```diff
@@ -33,6 +33,24 @@ def delete_old_test_results
...
+
+      passed = countable.count { |rr| rr.result == 'pass' }
+      (passed.to_f / countable.size * 100).round(2)
+    end
+
+    def countable_op_rule_results
```
romanblanco, line 48: The effect would probably be minor, but this could also be memoized:

```ruby
def countable_op_rule_results
  @countable_op_rule_results ||= begin
    results = selected_op_rule_results || []
    if version_mismatched?
      matched = rule_ids.keys.to_set
      results.select { |rr| matched.include?(rr.id) }
    else
      results
    end
  end
end
```

romanblanco (changes requested):
Overall looks good to me, only minor comments.

romanblanco (approved):

app/services/concerns/xccdf/test_result.rb
RoamingNoMaD: After sleeping on it, I'll make it into a constant. 
Three reasons:
1. Right now if score is nil and someone requests systems with `compliant=false`, systems with `nil` score wouldn't show up even though they'd be non-compliant.
2. I'm quite sure the notification we'd send about the systems would show `nil` too, which is not nice. 
3. The OpenAPI spec clearly says that score is a decimal number.

Changing it back, using a constant this time.

PR #2960, opened 2026-08-26
fix(CodeCov): set required uploads to 1

RoamingNoMaD:
We need to change the legacy setting (from when we had minitests also) so CodeCov doesn't wait forever to signal it's status.

- Configure Codecov to report status after one build.
- Upload `./coverage/coverage.xml` explicitly in the CI workflow.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,4 +8,4 @@ ignore:
   - "lib/tasks"
 codecov:
   notify:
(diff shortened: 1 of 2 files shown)


RoamingNoMaD:
Fix is successful :sunglasses:

LightOfHeaven1994 (approved):
LGTM, great work!

PR #2970, opened 2026-09-03
feat(systems): dual-write Kafka profile fields

petrblaho:
## Jira
RHINENG-30504

## Summary
- Dual-write Inventory operating-system versions and owner ID into native system columns.
- Preserve the trimmed JSONB profile and stale-message guard.
- Log malformed owner IDs while retaining them in JSONB and storing native NULL.

## Verification
- `podman-compose exec rails bundle exec rspec spec/services/kafka/system_importer_spec.rb` — 26 examples, 0 failures
- `podman-compose exec rails bundle exec rake spec:validate` — blocked by unrelated untracked `get_view.rb` and `test_heredoc.rb` RuboCop offenses

- Added native-column writes for OS versions and owner IDs during Kafka imports.
- Preserved trimmed JSONB profiles and stale-message protection.
- Logged malformed owner IDs and stored native `NULL` values.
- Expanded coverage for updates, omitted values, malformed values, and stale messages.

### DIFF
--- a/app/services/kafka/system_importer.rb
+++ b/app/services/kafka/system_importer.rb
@@ -33,10 +33,14 @@ module Kafka
     end
 
     def extract_system_attrs(id, payload, updated)
+      sp = relevant_system_profile(payload)
+      os = sp['operating_system'].is_a?(Hash) ? sp['operating_system'] : {}
+
       {
         id: id, account: payload.dig('account'), org_id: payload.dig('org_id'),
         display_name: payload.dig('display_name'), groups: payload.dig('groups') || [],
-        tags: payload.dig('tags') || [], system_profile: relevant_system_profile(payload),
+        tags: payload.dig('tags') || [], system_profile: sp,
+        os_major_version: os['major'], os_minor_version: os['minor'], owner_id: sp['owner_id'],
         stale_timestamp: payload.dig('stale_timestamp'), created: payload.dig('created'),
         updated: updated, insights_id: payload.dig('insights_id'),
         deleted_at: nil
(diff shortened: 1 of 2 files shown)


app/services/kafka/system_importer.rb
```diff
@@ -33,20 +35,34 @@ def payload_valid_for_import?(payload)
...
+      {
+        os_major_version: operating_system['major'],
+        os_minor_version: operating_system['minor'],
+        owner_id: native_owner_id(system_profile['owner_id'])
       }
     end
 
+    def native_owner_id(owner_id)
+      return owner_id if owner_id.nil? || (owner_id.is_a?(String) && OWNER_ID_FORMAT.match?(owner_id))
+
```
RoamingNoMaD: Could we use the same `UUID.validate` helper as the rest of the app?

```suggestion
      return owner_id if owner_id.nil? || (owner_id.is_a?(String) && UUID.validate(owner_id))
```

app/services/kafka/system_importer.rb
```diff
@@ -33,20 +35,34 @@ def payload_valid_for_import?(payload)
...
         deleted_at: nil
+      }.merge(native_system_profile_attrs(system_profile))
+    end
+
+    def native_system_profile_attrs(system_profile)
+      operating_system = system_profile['operating_system']
+      operating_system = {} unless operating_system.is_a?(Hash)
+      {
+        os_major_version: operating_system['major'],
+        os_minor_version: operating_system['minor'],
```
RoamingNoMaD: Should we have a type check here? 
They probably won't happen, but the DB will reject incorrect types if any error occurs.

Something like: `System.type_for_attribute(:os_major_version).cast(os_major_version)`

RoamingNoMaD (changes requested):
Good PR overall. Just a few type verification related comments.

RoamingNoMaD (approved):

PR #2981, opened 2026-09-15
fix(kessel): pass oauth2 creds as keyword arg to Kessel secure client builder

coderbydesign:
`build_secure_client` called `oauth2_client_authenticated(auth)` positionally, but kessel-sdk 1.11.0 requires the keyword `oauth2_client_credentials:`. The secure+auth path was dead code until `KESSEL_INSECURE` flipped to false at the HCC gateway repoint, so this only surfaced in stage as "wrong number of arguments (given 1, expected 0; required keyword: oauth2_client_credentials)" → 500s on RBAC-gated endpoints. Also capture the builder return value in case the SDK builder is immutable.

Harden the spec so this can't recur silently: use a verifying instance_double of the real ClientBuilder (a plain double accepted the bad call and kept CI green), assert the keyword arg, and add a case proving .build runs on the authenticated builder rather than the original.

Assisted-by: Claude

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

- Pass OAuth2 credentials with the required `oauth2_client_credentials:` keyword.
- Use the authenticated builder returned by `oauth2_client_authenticated` before calling `.build`.
- Update specs to verify keyword arguments and immutable builder behavior.


megaslurpie (approved):
## Approve — correct root-cause fix, well-tested, minimal scope

Verified the fix against `kessel-sdk 1.11.0` source: `ClientBuilder#oauth2_client_authenticated(oauth2_client_credentials:, channel_credentials: nil)` — both are keyword args, so the old positional `oauth2_client_authenticated(auth)` raised `ArgumentError`, surfacing as 500s on RBAC-gated endpoints once `KESSEL_INSECURE` flipped false in stage. Confirmed on the live `compliance-stage` deploy (image `f419258`): `GET /api/compliance/v2/policies` → `Completed 500` traced to `kessel_builder.rb:36`.

### Strengths
- **Correct fix**, and a single change covers every consumer — both `check_permission` and `list_workspaces_with_permission` build through this one method.
- **The `instance_double` swap is the real value.** A plain `double` accepted the bad positional call and kept CI green; the verifying double now fails on arity/kwarg mismatch at call time. This is what prevents recurrence.
- **`builder = ...` reassignment is sound.** The 1.11.0 builder is actually mutable and returns `self` (so it's a no-op today), but this future-proofs against an immutable builder — accurately framed as "in case".

### Non-blocking suggestions
1. **Test isolation (latent):** `KesselBuilder.auth` (`@auth ||=`) and `KesselRbac.client` (`@client ||=`) are process-global class ivars never reset between examples. Passes today only because ordering is `:defined` and the arg-asserting OAuth example runs before the non-asserting one. Under `--order random` (or if an earlier spec triggers `KesselBuilder.auth`), the `.with(oauth2_client_credentials: mock_auth)` assertion could read a stale `@auth`. A `before { KesselBuilder.instance_variable_set(:@auth, nil) }` would make it robust.
2. **Follow-up (out of scope):** any Kessel build/check failure becomes `ConfigurationError → AuthorizationError → 500` per request with no fallback. Worth a separate decision on fail-open vs. fail-closed (403) for Kessel outages.

Note: reviewed statically — I did not execute the suite locally. LGTM. 🚀

RoamingNoMaD (approved):
Makes sense, LGTM

opacut:
the jenkins pipeline is failing due to an error in the pipeline teardown - but all tests passed. I think we can ignore that check for now

RoamingNoMaD:
I'll steal your PR to rebase it :see_no_evil: 
Merging soon...

RoamingNoMaD:
Unrelated, flaky test failures. Thanks for the contribution <3

